PR #2557, opened 2017-12-24
CI: test that angular example works

Hypnosphi:
~~Later we can add an integration test as well~~
did that


ndelangen (approved):

ndelangen:
Great catch @Hypnosphi !

PR #2558, opened 2017-12-24
Move everything from lodash to lodash-es

igor-dv:
Issue: After latest changes in angular  the entire `lodash` package is bundled.

## What I did
Changed lodash to lodash-es in. This way we can import specific method without installing the lodash.* package (that have issues with @types)


Hypnosphi (approved):

jest.config.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     '<rootDir>/examples/cra-kitchen-sink',
     '<rootDir>/examples/official-storybook',
   ],
+  transformIgnorePatterns: ['/node_modules/(?!lodash-es/.*)'],
```
tmeasday: Is this something that storyshots users will need to do too?

PR #2517, opened 2017-12-19
add support for storyshots to save files according to different patterns

zvictor:
# The issue

In my setup, I always have a `cases.spec.js` that contains different props for my components. This file provides cases scenarios for both unit tests and storybook. With that in place, my components always have a standard `stories.js` file next to it, that looks like that:

```js
import React from 'react';
import { storiesOf } from '@storybook/react';

import Component from './component';
import cases from './cases.spec.js';

const stories = storiesOf('Category Icon Hover Menu', module)

for (const [caseName, caseData] of cases) {
  stories.addCentered(caseName, () => (
    <Component {...caseData} />
  ));
}
```

I like organising the project like this because I get storybook "for free" while writing tests. However, the `multiSnapshotWithOptions` does not work the way I intended with such setup. I don't like that it always create a `__snapshots__` folder with a **single file** on it.

screen shot 2017-11-30 at 13 25 24

# The alternative

I rewrote `getStoryshotFileName` to use the stories `kind` and `name` instead of the original filename:

screen shot 2017-11-30 at 13 31 20

The API is quite simple:

```js
import initStoryshots, { snapshotPerStoryAdded, snapshotPerStoryFile  } from '@storybook/addon-storyshots';

initStoryshots({
  // filesPattern: snapshotPerStoryAdded,
  filesPattern: snapshotPerStoryFile,
});
```

with that I also killed the `multiSnapshotWithOptions`. I hope nobody minds 😄

PR #2413, opened 2017-12-01
Add ability to use image snapshots to addon-storyshots

thomasbertet:
Issue: https://github.com/storybooks/storybook/issues/1781

## What I did
- Ability to generate image snapshots from stories using addon-storyshots.

## How to test
- Start fresh & bootstrap everything
- Go to _cra-kitchen-sink_
- Run `yarn run test-image-snapshots`

Is this testable with jest or storyshots?
- Not sure 

Does this need a new example in the kitchen sink apps?
- Yep done.

Does this need an update to the documentation?
- Maybe we should say something about it in the doc as it may persuade people to use storybook ? But I don't think there is a word about storyshots in the doc yet.

PR #2564, opened 2017-12-26
Angular and Vue storyshots

igor-dv:
Issue: Adding storyshots support for angular

## What I did
1. I have a BDay - this is my cake =) 🍰 
2. I've refactored storyshots to make it easier to integrate new frameworks
3. There are these PRs (#2517, #2413) that should happen before
4. ~~I have concerns about adding every new framework to peer dep... Should people install all these angular/polymer/aurelia packages?~~
5. There is a lot of copy-paste - **TBD in a sperate PR - extract common angular logic to some kind of `@storybook/angular-core`**
6. ~~I don't know how to make props to be visible in snapshots =(~~
7. ~~One storyshot is failing. We need to refactor more things to be able to run framework-specific preparations with beforeEach.~~

## How to test
For running only angular sthoryshots:
```
node node_modules\jest\bin\jest.js angularshots
```
 
For running only vue sthoryshots:
```
node node_modules\jest\bin\jest.js vueshots
```


igor-dv:
@storybooks/team , I think it's ready for review.

CC: @ralzinov @alterx - angular part

ndelangen:
I'd assume this would increase our test coverage more? Is it not reported correctly yet because of failing CI?

igor-dv:
Idk, how test coverage is calculated?

igor-dv:
Any ideas why cli is failing? I know it runs with angular v4, but I checked it locally with v4 and storyshots did work...

Hypnosphi:
Here's the error message, for reference:

```
ERROR in /tmp/storybook/node_modules/@angular/platform-browser-dynamic/src/compiler_factory.d.ts
(9,115): error TS2305: Module '"/tmp/storybook/node_modules/@angular/compiler/compiler"' has no exported member 'CompileReflector'.
```

igor-dv:
Yeah, I've seen it. I but I can't understand what is this dependencies hell again.

igor-dv:
Just moved angular deps in storyshots back to peer and cli tests are green now, FML. 
@Hypnosphi, any suggestions? I feel it's the monorepo thingy, but I'm not familiar with cli tests to be sure.

Hypnosphi:
> moved angular deps in storyshots back to peer

Why did you need them as direct deps?

igor-dv:
They are used in the code

Hypnosphi:
Is there a chance someone uses storyshots without storybook?

Hypnosphi:
See https://github.com/storybooks/storybook/pull/2467#issuecomment-355014111

igor-dv:
Hah, I don't think so. Do you mean to remove them at all? In this case, we will get the extraneous deps warning...

Hypnosphi:
As a react user, I'd be quite surprised if I knew that some developer tool installs angular for me

Hypnosphi:
> In this case, we will get the extraneous deps warning...

I see two solutions here:
— Split storyshots into three different framework-specific packages, each of which has their own peerDeps
— Suppress the warnings

igor-dv:
Yeah. I'll suppress it... Agree with you.

addons/storyshots/package.json
```diff
@@ -20,8 +20,11 @@
     "babel-runtime": "^6.26.0",
     "glob": "^7.1.2",
     "global": "^4.3.2",
+    "jest-preset-angular": "^5.0.0",
     "jest-specific-snapshot": "^0.3.0",
     "prop-types": "^15.6.0",
+    "react": "*",
+    "react-test-renderer": "*",
```
Hypnosphi: Please don't use wildcards for direct dependencies

addons/storyshots/package.json
```diff
@@ -20,8 +20,10 @@
     "babel-runtime": "^6.26.0",
     "glob": "^7.1.2",
     "global": "^4.3.2",
+    "jest-preset-angular": "^5.0.0",
     "jest-specific-snapshot": "^0.3.0",
     "prop-types": "^15.6.0",
+    "react-test-renderer": "^16.2.0",
```
Hypnosphi: Can you please check that it works in a react 15 app?

ndelangen: Would this block the merge? I presume this is the same for master?

Hypnosphi: On master, it's a peer dep

addons/storyshots/src/angular/renderTree.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
Hypnosphi: Please disable it for a particular line only (`// eslint-disable-next-line`)

addons/storyshots/package.json
igor-dv: Maybe I can just limit it to the >= 15?

addons/storyshots/src/angular/renderTree.js
```diff
@@ -1,8 +1,10 @@
-/* eslint-disable import/no-extraneous-dependencies */
 import AngularSnapshotSerializer from 'jest-preset-angular/AngularSnapshotSerializer';
 import HTMLCommentSerializer from 'jest-preset-angular/HTMLCommentSerializer';
+// eslint-disable-next-line
```
Hypnosphi: please disable only the specific rule (`// eslint-disable-next-line import/no-extraneous-dependencies`)

igor-dv:
So I'm waiting for #2413 to be finished and I'll be able to solve conflicts, but will be glad to get a review anyway =)

CC. @alterx , @ralzinov  for the angular part.

addons/storyshots/package.json
Hypnosphi: What do you mean? It's not a peer dep. `>=15` will just install latest 16

igor-dv: Yeah, right. I'll check it with 15 so. 🤦‍♂️

igor-dv:
@Hypnosphi, I think there will be indeed a problem with v15 (didn't check it but have seen other issues about it). So probably I have to return the `react-test-renderer` back to peer. It doesn't make sense for angular and vue users, but removing it at all will break other imports in storyshots (that I can't remove for backward compatibility)...
Ideas?

Hypnosphi:
> Ideas?

Separate packages for different frameworks?

igor-dv:
🤦‍♂️  I am lazy to do this

igor-dv:
@Hypnosphi , what do you think of using this: https://github.com/Wizcorp/codependency ?

Hypnosphi:
Description looks a bit outdated (npm stopped installing peers quite a long time ago)

Hypnosphi:
Probably that's because it was last updated in 2014

igor-dv:
I think I've found a solution. Now `react-test-renderer` is loaded only when used with react storyshots. I can remove it from deps (also jest-preset-angular). Need only update the docs.

addons/storyshots/src/angular/renderTree.js
```diff
@@ -1,4 +1,6 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
Hypnosphi: How are people expected to get `jest-preset-angular` dependency? As contrary to `@angular/core`, it's by no means guaranteed to present in an angular project

addons/storyshots/src/react/renderShallowTree.js
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
Hypnosphi: Same here

igor-dv: People who are already using Storyshots were already needed to install `react-test-renderer`, so for them, there won't be a problem. I need to add a section in Readme about these "optional peers", that people have to install by hand.
BTW, For angular and vue these deps are needed to configure jestconfig (I've already added it to readme)

Hypnosphi: Makes sense to me

addons/storyshots/README.md
```diff
@@ -36,6 +38,43 @@ Usually, you might already have completed this step. If not, here are some resou
 
 > Note: If you use React 16, you'll need to follow [these additional instructions](https://github.com/facebook/react/issues/9102#issuecomment-283873039).
 
+### Configure Jest for React
+StoryShots addon for React is dependent on [react-test-renderer](https://github.com/facebook/react/tree/master/packages/react-test-renderer), but 
+doesn't install it, so you need yo install it separately.
```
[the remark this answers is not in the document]
Hypnosphi: Please add the shell command to copypaste:
```sh
npm install --save-dev react-test-renderer
```

addons/storyshots/README.md
```diff
@@ -40,11 +40,19 @@ Usually, you might already have completed this step. If not, here are some resou
 
 ### Configure Jest for React
 StoryShots addon for React is dependent on [react-test-renderer](https://github.com/facebook/react/tree/master/packages/react-test-renderer), but 
-doesn't install it, so you need yo install it separately.
+doesn't install it, so you need to install it separately (read [here](#deps-issue) why).
```
Hypnosphi: It's a super-nit, but I prefer when text keeps making sense if you replace links with their text, that's why I don't really like links with "here" text.

Can we just make "doesn't" a link?

addons/storyshots/README.md
```diff
@@ -75,6 +87,20 @@ module.exports = {
   moduleFileExtensions: ['vue', 'js', 'jsx', 'json', 'node'],
 };
 ```
+
+### <a name="deps-issue"></a>Why don't we install dependencies of each framework ?
+Storyshots addon is currently supporting React, Angular and Vue. Each framework needs its own packages to be integrated with Jest. We don't want people that use only React will need to bring other dependencies that do not make sense for them. 
+
+`dependancies` - will installed an exact version of the particular dep - Storyshots can work with different versions of the same framework (let's say React v16 and React v15), that have to be compatible with a version of its plugin (react-test-renderer).
```
Hypnosphi: `dependancies` — typo

Hypnosphi (approved):

igor-dv:
I'm merging this one since I'm tired solving conflicts. Still didn't get a feedback on the angular part, but I think it will be easier to test in beta.

PR #2566, opened 2017-12-26
[app:angular] Fixed dependency injection for components

ralzinov:
Issue:
DI on dynamic components doesn't work 

## What I did
Fixed dependency injection on dynamic components 

## How to test
inject dependency into component, check that dependency is defined

no

no

no


shilman (changes requested):
@ralzinov I’m seeing an error on the “All knobs” example in `examples/angular-cli` which is not present on `master`. Any idea what's going on here?

storybook

ralzinov:
@shilman Fixed, but i think we need some extra testing on this

igor-dv:
I will test it with v4 today. @ralzinov  can you please add meanwhile a usecase of the problem to the example ?

igor-dv:
Tested more:
1. works for v5
2. doesn't work for v4 - when changing to use ɵReflectionCapabilities started working for both.
```js
import { ɵReflectionCapabilities } from '@angular/core';
const reflectionCapabilities = new ɵReflectionCapabilities();
/// ...
export function getParameters(component: any) {
  const params = reflectionCapabilities.parameters(component);

  if (!params || !params[0]) {
    return getMeta(component, ['parameters'], []);
  }

  return params;
}
```
3. doesn't work for both versions with knobs.
🤕 🤦‍♂️

igor-dv:
Looks like everything is working now. In my previous check, I dumbly used `@storybook/knobs` instead of the `@storybook/knobs/angular`, so I've misled everyone and myself. 

image

igor-dv (approved):

PR #2576, opened 2017-12-27
Fix @storybook/client-logger import

1y8ws7:
Issue:
`@storybook/client-logger` was imported incorrectly in `@storybook/addon-info`, which results in runtime error `TypeError: Cannot read property 'warn' of undefined`

## What I did
Replaced default import with named import.

## How to test
1. Use `marksyConf` in storybook configuration. 
2. When running storybook in the browser, proper warning message should be shown instead of the runtime error.

### DIFF
--- a/addons/info/src/index.js
+++ b/addons/info/src/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import deprecate from 'util-deprecate';
 import nestedObjectAssign from 'nested-object-assign';
-import logger from '@storybook/client-logger';
+import { logger } from '@storybook/client-logger';
 import Story from './components/Story';
 import PropTable from './components/PropTable';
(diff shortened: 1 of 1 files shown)


Hypnosphi (approved):

Hypnosphi:
Nice catch @eur00t! I wonder why eslint didn’t catch it though

eur00t:
@Hypnosphi thanks! I think it'll be helpful to add `import/default` and `import/named` eslint rules from `eslint-plugin-import` (I can see that currently we use `import/no-unresolved` only).
With `import/default` there can be problems though, because some packages doesn't export ES6 modules, for example `@storybook/client-logger`. It might work when `"jsnext:main"` is defined.

Issue #2575, opened 2017-12-27
Haul support for React Native not working on Windows


FredyC:
### Issue details

I am trying to configure `storybook-react-native` with Haul because I want to use Typescript loader. On my Windows machine I end up with this error.

```bash
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^

SyntaxError: missing ) after argument list
    at createScript (vm.js:80:10)
```

### Steps to reproduce

Run following on the Windows machine
`storybook start -p 7007 --haul webpack.haul.storybook.js`

### Please specify which version of Storybook and optionally any affected addons that you're running

- storybook/react-native: 3.3.1

### Affected platforms 

Windows 10

### Screenshots / Screencast / Code Snippets (Optional)

**webpack.haul.storybook.js**
```js
module.exports = ({ platform }, { module, resolve }) => ({
  entry: `./storybook/index.js`,
  module: {
    ...module,
    rules: [
      {
        test: /\.tsx?$/,
        loader: 'ts-loader',
      },
      ...module.rules,
    ],
  },
  resolve: {
    ...resolve,
    extensions: [
      '.ts',
      '.tsx',
      `.${platform}.ts`,
      '.native.ts',
      `.${platform}.tsx`,
      '.native.tsx',
      ...resolve.extensions,
    ],
  },
})
```

Hypnosphi:
Fix released as [`3.3.4`](https://github.com/storybooks/storybook/releases/tag/v3.3.4)

PR #2577, opened 2017-12-27
Use haul/bin/cli.js for cross-platform support

danielkcz:
Issue: #2575 

## What I did

Changed path to Haul CLI instead of using `.bin` folder.


FredyC:
It's kinda shame that this is not getting more attention as it's a rather annoying bug. Every time I install something with Yarn I have manually change dist files to keep it working. Not sure about failing tests, apparently not related to this change.

Hypnosphi:
Why the old path don't work? I should be a symlink to the new one

FredyC:
Well, Windows and symlinks ... I would say a long story :)

danielduan (approved):
looks good to me if this is just a change from using the simlinked bin file to the actual one.

@Hypnosphi is the chromatic test on CircleCI broken?

ndelangen (approved):

ndelangen:
Merge when ready

Issue #2587, opened 2017-12-28
Hot Module Reload not working properly for `@storybook/react@3.3.1`


tmeasday:
### Issue details

Due to https://github.com/storybooks/storybook/pull/2503

if you change a story's or component's implementation, it will not re-render on HMR.

### Steps to reproduce

Simply change any of the story implementations in kitchen-sink

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/react 3.3.1

tmeasday:
See https://github.com/storybooks/storybook/pull/2588

tmeasday:
Fixed in 3.3.2

tmeasday:
I wonder if vue has this same problem: 

https://github.com/storybooks/storybook/blob/03f04e03dab9c1230be1b90afc60664674bbc59f/app/vue/src/client/preview/render.js#L63-L72

[Not introduced in 3.3 but there anyway?]

I'm not sure who to @ here @ndelangen

ndelangen:
Could @ @kazupon ?

PR #2588, opened 2017-12-28
Use store revisions to ensure that stories re-render on HMR.

tmeasday:
For #2587

Issue:

To avoid unnecessarily re-rendering stories we don't re-render if the story name+kind haven't changed.

However on HMR, this leads to not rerendering / breakage, as in this case the name/kind haven't changed, but the store's implementation itself changed.

## What I did

I added the concept of a revision to the storystore. This implies that the same name+kind can be different as the store itself has changed. I got the idea from styleguidist which I think saw something similar in when I was reading their source ;)

Moving forward, if we want dynamic stores (yes please!) we may need to generalise the concept of revision to meaning more just "incrementing on each HMR". 

## How to test

Try changing stories implementations in the kitchen sink.

**We should ensure this is merged into https://github.com/storybooks/storybook/pull/2241**


Hypnosphi:
Should we do the same for other apps?

tmeasday:
They don’t have this immediate problem. We’ll get this revision code in all
apps when we land the general story store PR

Hypnosphi:
> They don’t have this immediate problem

Why do they need the revisions then?

danielduan (approved):
this looks like a decent enough patch. also curious as to why the other two platforms don't need it.

shilman (approved):
LGTM. Have confirmed the bug and the fix on my machine, but don't understand the broader implications of this change (if any).

tmeasday:
@Hypnosphi 

>> They don’t have this immediate problem
> Why do they need the revisions then?

Good question, perhaps we can discuss elsewhere (I'll make this into a PR against #2241 and we can go into detail there).

I think it is conceptually correct that we keep track in the storystore that the stories have *changed*; ie the output of `Foo:bar` (kind="Foo", name="bar") is different to what it was previously. 

In this case, such a distinction is needed because our React code is trying to avoid unnecessary re-renders when the story *choice* hasn't changed. There was no way (until this patch) for the React code to know if the story *implementation* had changed.

@danielduan 

> also curious as to why the other two platforms don't need it

They don't need it immediately because they are not (AFAIK) making this optimization [I don't really know if they need to either]. This patch is a reaction to a bug introduced when I made this optimization in React in https://github.com/storybooks/storybook/pull/2503.

PR #2589, opened 2017-12-28
Add more `imports` eslint rules

Hypnosphi:
Issue: we don't get eslint errors for wrong named and default imports, which sometimes leads to bugs (e.g. #2576)

For those rules to work when importing our own packages, I had to add `jsnext:main` for all the packages that have a compiled file as their `main`


.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
   },
   settings: {
     'import/core-modules': ['enzyme'],
+    'import/ignore': ['node_modules\\/(?!@storybook)'],
```
igor-dv: What is the usecase? I've tried to lint with this setting and without - nothing happened.

igor-dv (approved):

.eslintrc.js
Hypnosphi: That's because there's no errors =) Try replacing `{ logger }` with `logger` [here](https://github.com/storybooks/storybook/blob/19e2761fef219816ffcf6ac91a76ecf1c164bca4/addons/info/src/index.js#L4)

From [docs](https://github.com/benmosher/eslint-plugin-import#importignore):
> In practice, this means rules other than no-unresolved will not report on any imports with (absolute filesystem) paths matching this pattern.

Issue #2570, opened 2017-12-27
Don't require storyshots users to transform `lodash-es`


tmeasday:
### Issue details

As part of https://github.com/storybooks/storybook/pull/2558 we added a dependency on `lodash-es`, an uncompiled JS dependency.

This means that Jest needs to be configured to transform the file with:

```js
  transformIgnorePatterns: ['/node_modules/(?!lodash-es/.*)'],
```

within each users' storyshots-using project.

### Steps to reproduce

Create a storyshots project with 3.3.

tmeasday:
Can we add to the user's jest config? Feels a little hacky, but then storyshots is pretty hacky ;)

tmeasday:
cc @igor-dv @Hypnosphi

tmeasday:
The only other option I can think of is to use something like `babel-register`

tmeasday:
I don't think `babel-register` works with jest, so I think that is out.

igor-dv:
Hmm.. Shit. I've Forgotten about people need their own jest config. Storyshots already use babel-core internally maybe we can somehow use it for that?

tmeasday:
It also overrides `require`, so maybe.. let's try some stuff.

tmeasday:
This line is the place the exception comes from (it isn't clear from the stack trace):

https://github.com/storybooks/storybook/blob/tmeasday/note-jest-changes/addons/storyshots/src/index.js#L46

Hypnosphi:
> I don't think babel-register works with jest

Why?

igor-dv:
DIdn't work for me,

tmeasday:
>> I don't think babel-register works with jest
> Why?

Jest overrides require itself I guess. It doesn't seem like they are compatible.

tmeasday:
If we want to try and get it working with Jest we should probably follow these for insights into a non-hacky solution: https://github.com/facebook/jest/issues/4842 https://github.com/facebook/jest/issues/2205

eddiemonge:
This is a breaking change and is causing test failures for us

igor-dv:
Will revert it today

tmeasday:
This was fixed and released in 3.3.3

PR #2591, opened 2017-12-28
Revert "Move everything from lodash to lodash-es"

igor-dv:
Reverts storybooks/storybook#2558

Since `lodash-es` requires an extra configuration in the users' jest.config - it makes it a breaking change. Sadly I can't find a quick workaround, so have to revert it. This revert will return the problem of the whole `lodash` package being bundled to the angular app, and will need to be addressed separately.

💔


igor-dv:
CC #2570

Hypnosphi (approved):

Issue #1736, opened 2017-08-25
Ability to refresh (force rerender) of story


victorandree:
It would be neat if addons could tell the Storybook API to refresh a given story. Using [storybookAPI.selectStory()](https://storybook.js.org/addons/api/#storybookapiselectstory) does nothing if the story is already selected.

This would be useful when a story wants to customize hot-reloading to maintain parts of its state, through e.g. a Redux store. This would support stories like this:

    import { storiesOf } from '@storybook/react';
    import reselectStory from 'storybook-hot';
    import * as React from 'react';

    import configureStore from './configureStore';
    import HotStory from './HotStory';

    // Configure your Redux store once. If you make edits to *this* module, it
    // will get hot-reloaded by its parent so that your Redux state gets reset.
    const store = configureStore();

    // Store a mutable reference to the story, which gets replaced on hot reload
    let Story = HotStory;

    storiesOf('Hot story').add('Example', () => <Story store={store} />);

    // Whenever the story changes, get the new module and tell Storybook to
    // re-render us.
    module.hot.accept('./HotStory', () => {
      Story = require('./HotStory').default;

      // Tell our addon to tell the API to reselect our story.
      reselectStory('Hot story', 'Example');
    });

I've already written an addon (unpublished) which does this by selecting another (dummy) story and selecting back to the original story. It works, quite well, but requires a dummy story. This is obviously imperfect, and you still have to do hard reloads from time-to-time, but it comes close to the "regular" app workflow but I can isolate mocking my network stack to my stories, specifically.

The relevant "blocker" is in [init_api.js#L48](https://github.com/storybooks/storybook/blob/master/lib/ui/src/modules/api/configs/init_api.js#L48). I realize you wouldn't usually want to emit an event for the same story, but maybe there could be an escape hatch for an explicit re-render?

ndelangen:
This is really interesting!
We actually do need a good method to push new data to a currentSelectedStory, because Vue and Angular work quite different from react. And we want to wrap the storyFn as little as possible actually.

#1209

dangreenisrael:
I'll take a crack at it when I get a chance

Hypnosphi:
Released as [3.4.0-alpha.0](https://github.com/storybooks/storybook/releases/tag/v3.4.0-alpha.0)

netpoetica:
Is there any doc on how to use this/is this actually available?

dangreenisrael:
[...]

PR #2600, opened 2017-12-28
(#1736) Expose forceReRender over `channel` [WIP]

dangreenisrael:
Issue: #1736 

***This is not currently working***

## What I did

Allowed force render to be called via `addonAPI.forceReRender`

## ToDo: 
- [ ] Manually test
- [ ] Automated tests
- [ ] Update current docs site
- [ ] Update new docs site

## How to test

- possibly

- Yes, for addons api

### DIFF
--- a/app/angular/src/client/preview/index.js
+++ b/app/angular/src/client/preview/index.js
@@ -23,9 +23,10 @@ const storyStore = new StoryStore();
 const reduxStore = createStore(reducer);
 const context = { storyStore, reduxStore };
 
+let channel;
 if (isBrowser) {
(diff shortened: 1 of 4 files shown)


dangreenisrael:
This is being closed due to a problem with accessing `StoryStore` on emitted events. It seems to me that the `react/client/preview` code is run twice. The first time (when the listeners are being set) the `StoryStore` has not been initialized. The second time (when the `StoryStore` gets initialized) the listeners don’t seem to be accessible from `AddonStore`.  Because render can’t be done without a `StoryStore`, the render function returns early.

Issue #2602, opened 2017-12-28
[3.3.3] eslint: Unable to resolve path to module @storybook/react


DimaRGB:
### Issue details

After update to `3.3.3` see eslint error: `Unable to resolve path to module @storybook/react. (import/no-unresolved)`

### Please specify which version of Storybook and optionally any affected addons that you're running

- `@storybook/react 3.3.3`

### Screenshots / Screencast / Code Snippets (Optional)

```js
import * as storybook from '@storybook/react';
```

eslintrc.js:
```js
  ...
  parserOptions: {
    ecmaVersion: 8,
    sourceType: 'module',
    ecmaFeatures: {
      impliedStrict: true,
      jsx: true,
    },
  },
  ...
```

Hypnosphi:
Probably related to adding `jsnext:main` in #2589

DimaRGB:
seems like eslint try to use `jsnext:main` but `src` folder not published.

alansouzati:
hey, great that we have a merged PR for this! :) Any news when we will have it released? also, I believe this issue can be closed?

Hypnosphi:
@shilman how about a patch release?

Hypnosphi:
Fix released as [`3.3.4`](https://github.com/storybooks/storybook/releases/tag/v3.3.4)

hswolff:
Sadly I'm still seeing this issue. 

The PR only removed `src` from the root `.npmignore` and not from the individual package itself: https://github.com/storybooks/storybook/blob/master/app/react/.npmignore#L2

I believe we need to remove that from the individual package to have the `src` directory published as well.

Hypnosphi:
Oh I see

hswolff:
Incoming!

hswolff:
PR: https://github.com/storybooks/storybook/pull/2677

csi-lk:
Hmm if I manually remove `src` from the `.npmignore` in the @storybook/react folder I still get this eslint error, it could be i'm misunderstanding the solution?

hswolff:
You mean in your `node_modules` folder in your application? I'm guessing that it's because the `jsnext:main` value points to a file that doesn't exist. Perhaps try copying over the `src/` folder from this repo into your node_modules folder, which should fix the issue. May be good to do that to verify. That's what my PR will change when a new npm package is published.

csi-lk:
I ended up giving up and reverting, will just wait for this change to come through, cheers

PR #2604, opened 2017-12-28
Publish package sources along with transpiled files

Hypnosphi:
Issue: #2602 

In #2589, I introduced `jsnext:main` fields so that eslint can resolve named and default exports when importing one package from another. They point to files in `src` directories that aren't published currently.


igor-dv:
I think we don't want to publish `src` files...

marcin-mazurek:
@igor-dv If you don't publish the `src` folder, then it doesn't make sense to set the `esnext:main` to `src` (which is needed for tree shaking and so on).

Hypnosphi:
> I think we don't want to publish src files

What's exactly the problem with it? I find it very convenient to be able to view external packages' sources directly in `node_modules` directory

tmeasday:
I don't feel very expert on this but I've seen it done before and it feels OK to me.

tmeasday (approved):
Happy to approve this

ndelangen (approved):
This would allow smart bundlers to tree-shake, correct?

LGTM

Hypnosphi:
> This would allow smart bundlers to tree-shake

Well we can actually to that ourselves, by tweaking our default webpack config =)

PR #2605, opened 2017-12-29
Use store revisions to ensure that stories re-render on HMR.

tmeasday:
Reworking #2588 fro #2241.

I think we should also check the revision number on this line: https://github.com/storybooks/storybook/blob/10ed61dd9e82ad00bfa31872ffc140dfeebe9d42/app/vue/src/client/preview/render.js#L63

But I would prefer a Vue expert to weigh in on that.

Also we should probably rebase/something `shilman/refactor-core`. This should hopefully cleanly apply in any case.

Finally, @Hypnosphi perhaps suggested some objections to this approach. I think this is the place to discuss them.

### DIFF
--- a/app/react/src/client/preview/render.js
+++ b/app/react/src/client/preview/render.js
@@ -14,6 +14,7 @@ const logger = console;
 let rootEl = null;
 let previousKind = '';
 let previousStory = '';
+let previousRevision = -1;
 
 if (isBrowser) {
   rootEl = document.getElementById('root');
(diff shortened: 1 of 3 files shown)


tmeasday:
> But I would prefer a Vue expert to weigh in on that.

I can confirm that HMR is *not* working in vue when you change e.g. a story definition in `vue-kitchen-sink`, but it does change with a very similar patch to the one to `react/render.js` above (to `vue/render.js`).

Hypnosphi:
There wasn't objections, only questions

shilman (approved):
LGTM

PR #2614, opened 2018-01-01
Vue storyshots

igor-dv:
Issue:  Adding storyshots support for vue

## What I did
Just wanted to check what is the complexity of adding storyshots for vue based on what is already done for angular, and looks like it's easier than angular stuff. 

Probably I need to do some cherry-pick of the main refactoring.

## How to test
For running only the vue sthoryshots:
```
node node_modules\jest\bin\jest.js vueshots
```


addons/storyshots/package.json
```diff
@@ -45,6 +45,7 @@
     "@storybook/addons": "^3.3.3",
     "babel-core": "^6.26.0 || ^7.0.0-0",
     "react": "*",
-    "react-test-renderer": "*"
+    "react-test-renderer": "*",
+    "vue": "^2.5.13"
```
Hypnosphi: I think we should remove the framework peer dependencies instead of adding them

igor-dv: Yeah, I think so too.

ndelangen (approved):

igor-dv:
@ndelangen , @Hypnosphi , WDYT, I'll merge it to the angular branch and continue from there?

ndelangen:
Sounds like a plan @igor-dv absolutely EPIC work!

PR #2626, opened 2018-01-03
skip serializing (Synthetic)Events

rhalff:
Revert old behavior of handling SyntheticEvents passed to the action handlers.

Fixes: #2590


tmeasday (approved):
Works well, made some other comment on #2590

Issue #2630, opened 2018-01-03
Storyshot's multiSnapshotWithOptions does not work with TypeScript


jelisejev:
### Issue details

The multiSnapshotWithOptions feature is not working correctly when writing stories in TypeScript:
it initially generates the snapshots correctly after a repeated run, the "Storyshots Integrity › Abandoned Storyshots › encountered a declaration exception" test fails reporting all files as missing.

The problem is because of this peace of code:

```
function getPossibleStoriesFiles(storyshotFile) {
  var _path$parse2 = _path2.default.parse(storyshotFile),
      dir = _path$parse2.dir,
      name = _path$parse2.name;

  return [_path2.default.format({ dir: _path2.default.dirname(dir), name: name, ext: '.js' }), _path2.default.format({ dir: _path2.default.dirname(dir), name: name, ext: '.jsx' })];
}
```

It's looking only for js or jsx files, while our stories have a tsx extension and thinks that the stories are orphaned.

### Steps to reproduce

1. Configure storybook to support typescript;
2. Create a story with a tsx extension;
3. Run jest and generate the snapshots;
4. Run jest again and observe the error.

### Please specify which version of Storybook and optionally any affected addons that you're running

- `@storybook/react` 3.3.3
- `@storybook/addon-storyshots` 3.3.3

### Affected platforms 

Not platform specific.

igor-dv:
@jelisejev , we have a PR where storyshots will be working with angular,
[this](https://github.com/storybooks/storybook/pull/2564/files#diff-a3e89e598b48fc8bfcbb947cc9410349R14)  line is fixing the ts issue.

I can probably cherry pick it to a separate PR for a patch release...

jelisejev:
@igor-dv 

> I can probably cherry pick it to a separate PR for a patch release...
The PR should fix the problem, however, if you can deliver the fix earlier, we would be very happy :) We're actively doing code reviews, and this feature indeed makes it much easier.

igor-dv:
CC: @shilman. Patch candidate

Hypnosphi:
Fix released as [`3.3.4`](https://github.com/storybooks/storybook/releases/tag/v3.3.4)

jelisejev:
@igor-dv thanks for the prompt fix, will test it out!

PR #2639, opened 2018-01-04
Add .ts compatibility to storyshots

igor-dv:
Issue: #2630

## What I did

A quick fix to make ts stories to work well with stoyshots and `multiSnapshotWithOptions`.
#2517 might revisit the ability to customize stoyshots file names.


ndelangen (approved):

ndelangen:
Perhaps we should add an example showing typescript usage? if that's not too cumbersome in a monorepo?

igor-dv:
All angular examples are actually in typescript, Do you think we need something for react ?

ndelangen:
No, doesn't seem necessary to me, but if whoever reading this, disagrees, let us know :)

PR #2646, opened 2018-01-04
Removed lowercase-only restriction for knobs

alterx:
Take in account that now the knob name needs to be the same as the angular component @Input attribute. Otherwise it'll not bind to the correct property.

Issue: https://github.com/storybooks/storybook/issues/2641

## What I did
Removed lowercase-only restriction for knobs

## How to test
```
yarn bootstrap --reset --core
cd examples/angular-cli
yarn storybook
open http://localhost:9009
```


addons/knobs/src/angular/helpers.js
```diff
@@ -73,11 +73,10 @@ const getAnnotatedComponent = ({ componentMeta, component, params, knobStore, ch
     const oldValue = knobOptions.value;
     knobOptions.value = value;
     knobStore.markAllUnused();
-    const lowercasedName = name.toLocaleLowerCase();
-    this[lowercasedName] = value;
+    this[name] = value;
```
igor-dv: So as I understand we can't give a different name to knobs? It has to be a name of the prop, right? Or maybe all knobs are working the same way and I missing something?

igor-dv (approved):

addons/knobs/src/angular/helpers.js
alterx: Exactly. The knobs work a bit differently in the other frameworks because in those you can just send the value in as a prop. With Angular we need to bind to an specific field and check for the changes. That's why, in this specific case, to preserve the same api the name of the knob needs to be equal to the field name.

PR #2650, opened 2018-01-04
Add IE11 compatibility meta tag

mynnx:
Issue:

## What I did
I could get IE11 to render the Storybook, but it would not render the iframe view.  It kept defaulting to older document modes which were too outdated to render our JS.

The only way I could reliably get IE11 to render the iframe view was to add the following meta tag, which is present in the Storybook template:

```<meta content="IE=edge" http-equiv="X-UA-Compatible" />```

I'm not well-versed in how this works; my reading suggests that an HTTP header is usually preferred but we're working with `webpack-dev-server` and don't have a ton of control over that.

If anyone has suggestions about how this should work, I'm open to hearing them.

## How to test
Open the iframe view in IE11 with and without the meta tag; observe that it does not render without.

Not unless we use Browserstack or Saucelabs

## Does this need a new example in the kitchen sink apps?
No

## Does this need an update to the documentation?
No

### DIFF
--- a/app/angular/src/server/iframe.html.ejs
+++ b/app/angular/src/server/iframe.html.ejs
@@ -3,6 +3,7 @@
 <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta content="IE=edge" http-equiv="X-UA-Compatible" />
     <base target="_parent">
     <script>
         if (window.parent !== window) {
--- a/app/react/src/server/iframe.html.ejs
+++ b/app/react/src/server/iframe.html.ejs
@@ -3,6 +3,7 @@
 <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta content="IE=edge" http-equiv="X-UA-Compatible" />
     <base target="_parent">
     <script>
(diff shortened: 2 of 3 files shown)


ndelangen (approved):

ndelangen:
Adding the header on static builds will be super annoying too, so this is perfect actually!

Thank you for caring @mynnx !

ndelangen:
Could I request you also perform this change for the Vue and Angular apps?

mynnx:
@ndelangen Thanks for the quick response!  I tried following the instructions at https://github.com/storybooks/storybook/blob/master/docs/src/pages/configurations/add-custom-head-tags/index.md to add this tag, but found that it was not injected.  Can you verify that this doc page is still accurate and that the functionality still works?

ndelangen:
Epic! Merge when ready

mynnx:
@ndelangen I can't merge without write access and it appears that CI failed.  Anything you'd like me to do?

ndelangen:
I'm looking into it, looks like an outage at circleCI.

I leave the "Merge when ready" remark, because I usually don't wait around for the CI to approve, and browse elsewhere. With that comment other maintainers know it was my intention to merge.

Issue #2537, opened 2017-12-22
Handle a11y for components with delayed rendering


Primigenus:
### Issue details

The a11y addon doesn't work with content that loads late. I have a component that waits until data has arrived before rendering, but the a11y addon runs immediately. There is no way to access the underlying axe object used by the addon in order to re-run the test once the content is available.

The best solution would be to magically handle components with delayed rendering. A shortcut solution might be to just add a button to the UI that allows the user to re-run the tests. A way around this issue is to install the aXe Chrome extension, open the preview iframe in a new tab, and run aXe manually on the rendered view.

### Steps to reproduce

Install the a11y addon, then include the story I added below.
The a11y panel reports no violations were found and no checks were passed.

### Please specify which version of Storybook and optionally any affected addons that you're running

- storybook/react 3.2.18
- storybook/addon-a11y 3.2.18

### Affected platforms 

All platforms.

### Screenshots / Screencast / Code Snippets (Optional)

```js
class TestA11yComponent extends React.Component {
  state = {
    show: false
  }
  constructor(props) {
    super(props)
    setTimeout(() => this.setState({ show: true }), 1000)
  }
  render() {
    return this.state.show && <div>testing</div>
  }
}
storiesOf('test', module).add('default', () => <TestA11yComponent />)
```

danielduan:
We'd love a PR if you have time to look into it. I think adding an additional button to re-evaluate seems like a short and simple approach.

Hypnosphi:
Released as `3.4.0-alpha.5`

PR #2651, opened 2018-01-05
Handle a11y for components with delayed rendering

Primigenus:
Issue: #2537 

## What I did

Implemented support in a11y-addon for delayed rendering of components by adding a "re-run tests" button within the panel.

I also corrected the contrast of the addon's tabs (to 7:1) and added the number of violations and passes to each tab (because if you re-run, the number may change, but you wouldn't know until you navigated to the tab).

## How to test

There's a story called "Accessibility" in cra-kitchen-sink that demonstrates a11y-addon with a regular button and with a button that doesn't render until 1s after the preview loads. In order to verify that it works, click the "re-run tests" button. (Ideally, we would be able to verify that results are displayed in the tab.)


Primigenus:
@Hypnosphi What's necessary to get this merged?

Hypnosphi:
I hope I'll be able to review it in upcoming days

Hypnosphi:
@ndelangen would you like do do the review instead of me?

addons/a11y/src/components/Report/index.js
```diff
@@ -16,52 +16,25 @@ const styles = {
...
-  return (
-    <div>
+    ) : (
       <span style={styles.empty}>{empty}</span>
-      <RerunButton />
-    </div>
-  );
-}
+    )}
+    <RerunButton onClick={onRerunClick}>RE-RUN</RerunButton>
```
Primigenus: This change is a regression from "Re-run tests" to "RE-RUN". "RE-RUN" doesn't describe the functionality of the button in full and is also in all-caps; this should be achieved with CSS instead of written out in the HTML.

addons/a11y/src/components/Report/RerunButton.js
```diff
@@ -0,0 +1,17 @@
...
+  bottom: 0,
+  right: 0,
+  border: 'none',
+  borderTop: 'solid 1px rgba(0, 0, 0, 0.2)',
+  borderLeft: 'solid 1px rgba(0, 0, 0, 0.2)',
+  background: 'rgba(255, 255, 255, 0.5)',
+  padding: '5px 10px',
+  borderRadius: '4px 0 0 0',
+  color: 'rgba(0, 0, 0, 0.5)',
+  outline: 'none',
```
Primigenus: This disables the outline on the button, which causes problems with focus.

Primigenus (commented):
Thanks for the commits! I noticed some regressions of accessibility properties in my changes. Please take a look at them.

addons/a11y/src/components/Report/RerunButton.js
ndelangen: Ah, made it consistent with other places, will remove the `outline` line

ndelangen (approved):
@Hypnosphi yeah looks like a good change to me. I'll merge it.

@Primigenus any objections?

addons/a11y/src/components/Report/RerunButton.js
Hypnosphi: Let's do that in other places as well =) I wonder if it's possible to have an eslint rule for that

Hypnosphi:
> @Primigenus any objections?

Looks like he had one https://github.com/storybooks/storybook/pull/2651#discussion_r161380311

ndelangen:
Will address it!

Primigenus:
@ndelangen Thanks! The only other thing is the button label should be "Re-run tests" instead of "RE-RUN".

addons/a11y/src/components/Report/index.js
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { Fragment } from 'react';
```
Hypnosphi: @ndelangen if we want to use fragments, we need to specify `>=16.2.0` as react peer dependency

Issue #2652, opened 2018-01-05
[Addon Info] - Error when using oneOf propType and a evaluation


abrahamgnz:
### Issue details

What did you expect to happen? 
Tried to display the propTable of a component that has a code evaluation in the propTypes declaration.

What happened instead?
Error `propType.value.map is not a function` was displayed

### Steps to reproduce
Create a component with a evaluation in the propTypes declaration.
Add a story withInfo with an object defining the propTables to display the component.
`withInfo({ propTables: [Button] })(() => (...))`
Run the storybook

### Please specify which version of Storybook and optionally any affected addons that you're running

-  `@storybook/react  3.2.12`
-  `@storybook/addon-info  3.2.12`

### Screenshots / Screencast / Code Snippets (Optional)

```
import {values} from 'lodash'

const KIND = {
  PRIMARY: 'primary',
  SECONDARY: 'secondary'
}

const Button = props => (
 ...
)

Button.propTypes = {
/** Type of button **/
  type: oneOf(() => (['hello', 'again'])()),  // <- code evaluation
/** kind of button **/
  kind: oneOf(values(KIND)),  // <- code evaluation
}

```

thompsongl:
Certainly does not solve the underlying problem, but here's a workaround that renders and allows for correct proptype validation:

```js
const KIND = {
  PRIMARY: 'primary',
  SECONDARY: 'secondary'
}
const kindVals = values(KIND)

Button.propTypes = {
  kind: oneOf([...kindVals])
}
```

lcsrinaldi:
Any updates on this issue?

Hypnosphi:
Released [3.3.14](https://github.com/storybooks/storybook/releases/tag/v3.3.14)

PR #2653, opened 2018-01-05
Add type check to PropType on OneOf

theAbeGonzalez:
Issue: [#2652](https://github.com/storybooks/storybook/issues/2652)

## What I did
Added a type check to PropType prop on OneOf component.
## How to test 
Add a story with at component that has a oneOf prop with an evaluation in the propType declaration. 
See issue for further information.

No

No

No


addons/info/src/components/types/OneOf.js
```diff
@@ -1,7 +1,13 @@
 import React from 'react';
 import { TypeInfo } from './proptypes';
 
-const OneOf = ({ propType }) => <span>{propType.value.map(({ value }) => value).join(' | ')}</span>;
+const isObject = propType => typeof propType.value === 'object';
```
Hypnosphi: Should it rather check that `propType.value` is an array, as we're using an array method?

Hypnosphi:
Please add a prop with `oneOf` and evaluation here: https://github.com/storybooks/storybook/blob/master/examples/official-storybook/components/DocgenButton.js

addons/info/src/components/types/OneOf.js
abrahamgnz: I'm not sure why the `typeof propType.value` returns `object` when array. Better to change to `Array.isArray()`

abrahamgnz:
@Hypnosphi 
Somehing like: 
```
  /**
   * `oneOf` is basically an Enum which is also supported but can be pretty big.
   */
  enm: PropTypes.oneOf(['News', 'Photos']),
  enmEval: PropTypes.oneOf({kind: 'News', media: 'Photos'}.keys()),
```
?

Hypnosphi:
Yes, that would do it

Hypnosphi:
You might need to update snapshots of example stories by running `yarn test --core --update`

Hypnosphi (approved):

abrahamgnz:
@Hypnosphi snapshots updated

Hypnosphi (approved):

examples/official-storybook/stories/__snapshots__/addon-knobs.stories.storyshot
```diff
@@ -19,7 +19,7 @@ exports[`Storyshots Addons|Knobs.withKnobs tweaks static values 1`] = `
     My name is Storyteller, I'm 70 years old, and my favorite fruit is apple.
   </p>
   <p>
-    My birthday is: January 20, 2017
+    My birthday is: January 19, 2017
```
Hypnosphi: @tmeasday looks like your fix don't work in some cases

Hypnosphi: @abrahamgnz just for reference, which timezone are you in?

tmeasday: This is puzzling ;)

abrahamgnz: CST @Hypnosphi

tmeasday: @abrahamgnz it would be interesting to know what the following output in the node REPL for you:

```js
new Date('Jan 20 2017 GMT+0').toLocaleDateString('en-US', dateOptions)
```

tmeasday: I guess looking at the above, it's possible that it would come out as Jan 19. I wonder what time `new Date` picks by default.

abrahamgnz: image
@tmeasday

abrahamgnz: new Date picks :
2018-01-08T05:16:33.338Z

tmeasday: @abrahamgnz -- I guess I mean what does `new Date('Jan 20 2017 GMT+0')` pick? 

For me its `2017-01-20T00:00:00.000Z`, and it evaluates to "1/20/2017".

Also:

```js
> +new Date('Jan 20 2017 GMT+0')
1484870400000
```

tmeasday: I would expect all of the above to be system/TZ independent; unless something is weird on your machine with the `en-US` TZ?

abrahamgnz:
```
> new Date('Jan 20 2017 GMT+0')
2017-01-20T00:00:00.000Z
```

abrahamgnz:
```
+new Date('Jan 20 2017 GMT+0')
1484870400000
```
@tmeasday

Hypnosphi: @tmeasday looks like we have to add `timeZone: 'UTC'`  to `dateOptions`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

tmeasday: I can't believe no-one in the US ran into this yet... not even the CI servers!

I can't really explain why it was printing out the 19th in the first place for me then (given `new Date('Jan 20 2017')` was midnight 20 Jan in my TZ). But let's lock it down.

ndelangen: how?

Hypnosphi: This was probably fixed in #2861 

@abrahamgnz can you please try to update snapshots once again?

addons/info/src/components/types/OneOf.js
Hypnosphi: > why the typeof propType.value returns object when array

Because array is an object =)

PR #2656, opened 2018-01-05
Fix some typos in the contributing documentation

arty-name:
This PR is just some cosmetic documentation fixes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Thanks for your interest in improving Storybook! We are a community-driven proje
 
 Please review this document to help to streamline the process and save everyone's precious time.
 
-This repo uses yarn workspaces, so you should `yarn@1.3.2` or higher as package manager. See [installation guide](https://yarnpkg.com/en/docs/install).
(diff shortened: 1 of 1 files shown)


danielduan (approved):

CONTRIBUTING.md
```diff
@@ -31,7 +31,7 @@ yarn install
 yarn bootstrap --core
 ```
 
-The bootstrap command will ask which sections of the codebase you want to bootstrap. Unless you're going to work with ReactNative or the Documentation, you can keep the default.
+The bootstrap command might ask which sections of the codebase you want to bootstrap. Unless you're going to work with ReactNative or the Documentation, you can keep the default.
```
Hypnosphi: In fact, here it won't do it because the `--core` opttion is already specified. Just `yarn bootstrap` would ask those

arty-name: Well, removing "will" altogether seemed like too big of a change for me to make :)

Hypnosphi: I mean, this section probably needs to be revisited. @ndelangen can you help?

PR #2657, opened 2018-01-05
Renamed angular root node to avoid name collisions

alterx:
Angular CLI also uses `app-root` as the root node name and this could lead to weird bugs.

Issue: https://github.com/storybooks/storybook/issues/2647

## What I did
Renamed the angular root node. This is not directly related to the issue linked (the cause was different) but we should rename the node anyway.

## How to test
```
yarn bootstrap --reset --core
cd examples/angular-cli
yarn storybook
open http://localhost:9009
```


igor-dv (approved):

app/angular/src/server/iframe.html.ejs
```diff
@@ -13,6 +13,6 @@
     <%= htmlWebpackPlugin.options.data.previewHead %>
 </head>
 <body>
-    <app-root></app-root>
+    <storybook-dynamic-app-root></storybook-dynamic-app-roott>
```
ralzinov: Typo

ralzinov (changes requested):

ralzinov (approved):

PR #2660, opened 2018-01-05
Update webpack extend warning.

Alfredo-Delgado:
Documentation update:

Flesh out the repercussions of replacing module rules.


docs/src/pages/configurations/custom-webpack-config/index.md
```diff
@@ -38,7 +38,7 @@ Since this config file stays in the Storybook directory, you need to set the inc
 
 You also need to install the loaders (style, css, and sass) used in above config manually.
 
-> Once you create this `webpack.config.js` file, Storybook won't load the [default Webpack config](/configurations/default-config/) other than loading JS files with the Babel loader.
+> Once you create this `webpack.config.js` file, Storybook won't load the [default Webpack config](/configurations/default-config/) other than loading JS files with the Babel loader. This will disable included functionality like svg loading. Read on to learn how to [retain defaults](#full-control-mode--default).
```
Hypnosphi: Extend mode retains defaults as well

Alfredo-Delgado: I don't think I discovered a regression and the existing documentation contradicts that. This is the existing documentation: _Once you create this `webpack.config.js` file, Storybook won't load the default Webpack config other than loading JS files with the Babel loader. I've only extended it to highlight the kind of functionality that's lost when this happens and link to the better solution down the page.

Hypnosphi: I meant that there's a sense in linking to `#extend-mode` instead, but now I see that this is already in this section. It's quite weird actually, need to investigate why is it so.

Hypnosphi: Oh, nevermind. I actually was wrong and extend mode extends `baseConfig`, not the default one

Hypnosphi (approved):

Issue #1896, opened 2017-09-26
Possibility to group knobs in the panel


Aylata:
Hello!

Is it possible to group knobs when they are shown in the panel? I use "chapters" add-on to enable using different components in one story. Each component has its own knobs. At the moment there is a long-long-long list of all knobs and it's very hard to navigate and find something that you need. So is there any chance to group those knobs in a way I need it?

Thank you!

danielduan:
I don't think that's something on the roadmap, but if you think you have a solution for it, feel free to open a PR.

wildeyes:
Does anybody have any pointers on where to look to get this started? this feature is quite important for me, I'm thinking of implementing it.

danielduan:
@wildeyes the code for knobs is under `addons/knobs`. `CONTRIBUTE.md` should have some tips on how to get your dev environment set up.

PR #2661, opened 2018-01-05
Enable groupId option for knobs to be organized into sub panels.

ASeeto:
Issue: #1896 

## What I did

- Added groupId argument to knob functions
- Created GroupTabs component (replica of AddonPanel in storybook/ui) within Knobs panel to serve as navigation when filtering between groups

## How to test
`groupId` has been added as the last argument to each of the knob functions (see updates to README)
```
text(label, defaultValue, groupId);
boolean(label, defaultValue, groupId);
number(label, defaultValue, options, groupId);
color(label, defaultValue, groupId);
object(label, defaultValue, groupId);
array(label, defaultValue, separator, groupId);
select(label, options, defaultValue, groupId);
date(label, defaultValue, groupId);
button(label, handler, groupId);
```

**NOTE: Knobs with optional arguments MUST have default values passed in for use with groupId.**

This was done to avoid breaking changes to current knob function contracts.

e.g. Notice these README updates:
- number(label, defaultValue, options, groupId);
- array(label, defaultValue, separator, groupId);

---

- Example usage has been provided in official-storybook app.

- README has been updated with new argument.

official-storybook


ndelangen:
@ASeeto Sorry for not replying earlier. I haven't seen anyone mentioned before they ran into the issue that knobs became unwieldy and needed filters. But I can imagine it could be useful in large components.

If you could make 1 adjustment I think we could accept this feature.

The groupId should be optional and if 0 knobs use it, the tabs bar should not appear.
This would make it seamless for current users.

Also I'd like this feature documented in docs & the example would be a bit more clear if instead of splitting the knobs up into so many groups, it would just create 2. That would mean less stuff, but the intent of the feature is clear.

Thanks for your PR!

ASeeto:
Hi @ndelangen, no worries! Thank you for getting back to me, I appreciate you reviewing my PR.

I've updated the branch now to address the concerns you've mentioned.

**A Storybook with no Group IDs on Knobs**
*Appears as normal as to not affect current users*

screen shot 2018-02-15 at 6 17 38 pm

**A Storybook with Group IDs on Knobs**
*Appears with a new tabular component GroupTabs at the top of the knobs panel.*

screen shot 2018-02-15 at 6 18 15 pm

Effectively, the contract of each of the available knobs changes to the following:
```
const value = text(label, defaultValue, groupId);
const value = boolean(label, defaultValue, groupId);
const value = number(label, defaultValue, {}, groupId);
const value = number(label, defaultValue, options, groupId);
const value = color(label, defaultValue, groupId);
const value = object(label, defaultValue, groupId);
const value = array(label, defaultValue, ',', groupId);
const value = select(label, options, defaultValue, groupId);
const value = selectV2(label, options, defaultValue, groupId)
const value = date(label, defaultValue, groupId);
button(label, handler, groupId);
```

I've also been updated this in the README.md to explicitly state whether a third argument is required prior to adding a groupId which is the case for the `number` and `array` knobs

ASeeto:
**Alternative Approach for GroupIds in v4 / Breaking Change**

When building out this feature, I also thought of something else that might be useful to introduce in v4 to allow for further customization on a per knob basis. This idea would introduce breaking changes to all knobs however, which is why I chose the implementation I did in this PR as it keeps v3 stable for current users.

Instead of adding `groupId` as an additional argument as this PR illustrates, it might be nice if every knob had a generic `options` object as its third argument. The properties of the `options` object would be any optional content related to a specific knob.

e.g.

**number knob**
*becomes the following:*
```
const options = {
   range: true,
   min: 60,
   max: 90,
   step: 1,
   groupId: 'TESTING',
}
number(label, defaultValue, options);
```

**array**
*becomes the following:*
```
const options = {
   separator: ',',
   groupId: 'TESTING',
}
array(label, defaultValue, options);
```

**Benefits of Alternative for v4**

This would allow the "no more than 3 arguments per function" eslint rule to be abided by as well as allow for faster future development of features like ability to add `description`s, `className`s, or `data-attribute`s to each knob in the knob panel. A knob's `description`, `className`, `data-attribute`, `groupId` would be simply captured and spread from the `options` object argument.

Each knob would have a consistent contract:
```
knobtype(label, defaultValue, options)
```

where `knobtype` is each available knob.

```
text(label, defaultValue, options);
boolean(label, defaultValue, options);
number(label, defaultValue, options);
color(label, defaultValue, options);
object(label, defaultValue, options);
array(label, defaultValue, options);
select(label, defaultValue, options);
selectV2(label, defaultValue, options)
date(label, defaultValue, options);
button(label, handler, options);
```

---

Just a thought that I figured might be worth sharing, thanks for any feedback!

Hypnosphi (changes requested):

ASeeto:
@Hypnosphi Thank you for your review and feedback! I've updated the documentation to reduce the repetitiveness. Please let me know if any more changes should be made.

Hypnosphi (approved):

Hypnosphi:
You need to run `yarn install` in root directory to update lockfile

ASeeto:
@Hypnosphi I've updated the lock file using `yarn install` now, so it should be all set! I do notice it fails the "repo-dirty-check" for `ci/circleci: build` however

Hypnosphi:
Ok, you probably need to wipe all your `node_modules` directories (`yarn bootstrap --reset` will do that)

ASeeto:
@Hypnosphi Thank you, that looks to have fixed it

examples/official-storybook/stories/addon-knobs.stories.js
```diff
@@ -97,6 +97,77 @@ storiesOf('Addons|Knobs.withKnobs', module)
...
+        <p>Birthday: {new Date(birthday).toLocaleDateString('en-US', dateOptions)}</p>
+        <p>Account Balance: {dollars}</p>
+        <p>Years in NY: {years}</p>
+        <hr />
+        <h1>Favorites</h1>
+        <p>Catchphrase: {salutation}</p>
+        <p>Fruit: {fruit}</p>
+        <p>Second Fruit: {otherFruit}</p>
+        <p>Items:</p>
+        <ul>{items.map(item => <li>{item}</li>)}</ul>
```
Hypnosphi: please add keys for `li`s

Hypnosphi (changes requested):

Hypnosphi:
@igor-dv @alterx any idea what causes the angular build failure?

https://circleci.com/gh/storybooks/storybook/38677

Hypnosphi (approved):

Hypnosphi:
@ASeeto please run `yarn test --core --update` to update test snapshot

igor-dv:
> any idea what causes the angular build failure?

I am not sure, why this is happening 🤔 I remember there was a fix by @Quramy that fixed some similar issue. 
Maybe knobs' signature changes somehow conflict with the typing... Byt the way, existing types in DefinitelyTyped will probably break... That makes me think of these two possibilities:

1. Bring the types into the Monorepo
2. Instead of declaring a new groupId argument - add a new "options" argument that will be easily extended.

Hypnosphi:
@igor-dv I think the reason can be lockfile. @ASeeto Is there a chance that you deleted it before generating at some point?

Hypnosphi:
@tmeasday looks like we're getting off-by-day error again: https://circleci.com/gh/storybooks/storybook/38701?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #2666, opened 2018-01-06
Fix config dir detection

Hypnosphi:
Issue: #2665

The problem is that `getConfigDir` don't read from CLI arguments.

## How to test
Since this PR, here's how official-storybook looks like if both of the `*-head.html` aren't loaded:
screen shot 2018-01-06 at 22 45 27

### DIFF
--- a/examples/official-storybook/head-warning.js
+++ b/examples/official-storybook/head-warning.js
@@ -1,0 +1,10 @@
+import { document } from 'global';
+
+export default function addHeadWarning(bundle) {
+  const warning = document.createElement('h1');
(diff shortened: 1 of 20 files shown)


ndelangen (approved):

examples/official-storybook/head-warning.js
```diff
@@ -0,0 +1,10 @@
+import { document } from 'global';
+
+export default function addHeadWarning(bundle) {
```
igor-dv: That's nice one 👍

PR #2565, opened 2017-12-26
Update CLI snapshots as a postpublish step

Hypnosphi:
Issue: after a new version is published, the CLI test snapshots need to be updated. This PR automates it

PR #2671, opened 2018-01-07
Only update CLI snapsots on postpublish script, skip smoke tests

Hypnosphi:
This is a follow-up to #2565 

Smoke tests are too long to run, but there's no actual need to do them after publishing (they will run on CI anyway).

Labelling it with `patch` as it affects publishing process


shilman (approved):
LGTM! Thanks!!!

Issue #2645, opened 2018-01-04
Angular Feedback


amcdnl:
First off, this is awesome guys! I know first hand what a challenge Angular can be to build this type of stuff for! I gave it a whirl today and here is my general 'end-user' feedback.

- Add SASS and LESS support by default when you scaffold
- Make scaffolded the stories and config TypeScript not JS
- Every example should show importing a NgModule
- Add option for defining a template in the story like you did with Vue
- The ability to specify a custom tsconfig for storybook
- The angular cli can include global css in the `.angular-cli.json` that lots of ppl use, it should be picked up here too
- When scaffolding put `/stories` under `/src` so TypeScript will pick it up by default in a cli project

A few notes on some of those items:
- When you implement the template, you will have to make dynamic components. This will loose the AoT compilation so that should be noted. I worked on a project where I did this, you can see how to do that here: https://github.com/amcdnl/component-workshop/blob/master/lib/scenario.ts#L114
- Since Angular uses TypeScript, you can actually read a lot of interesting metadata from the components at runtime that could facilitate in removing some boilerplate. Here is an example of parsing inputs/outputs from Angular components - https://github.com/amcdnl/component-workshop/blob/master/scripts/parser.js#L59
  
//cc @igor-dv

ndelangen:
//cc @alterx @ralzinov

alterx:
Lovely feedback @amcdnl! 

We do use dynamic components currently so adding the template support *shouldn't* be *very*-
 difficult. These examples are amazing (the inputs/outputs parsing code will help a lot)!

Quramy:
> The angular cli can include global css in the .angular-cli.json that lots of ppl use, it should be picked up here too

I think it's nice !

Now, `@storybook/angular` does not depend on `@angular/cli`. I understand that it makes sense for developers not using `@angular/cli`. However, almost all Angular developers create their project from `ng new`, so they(including me) want that `start-storybook` bundles module configured with their `.angular-cli.json`. 

I think storybook should provide a helper function to configure webpack based on `.angular-cli.json`, something like:

```js
/* .storybook/webpack.config.js */

import { webpackConfigHelperForNgCli } from '@storybook/angular';

module.exports = (env) => {
  return webpackConfigHelperForNgCli(); // configures .ts , HTML, and style Rules 
};
```

alterx:
Hey @amcdnl , I have a couple of questions for you: 

[...]

PR #2672, opened 2018-01-07
Change ng stories dir

Quramy:
Issue:

See #2645 

## What I did
- Change generated `stories` directory location.

## How to test


igor-dv:
@Quramy, can you take a look why are the cli test failing ? Looks like something is misconfigured with tsconfig.

Quramy:
@igor-dv. As you said, this error is caused by `src/tsconfig.json`(I think `src/tsconfig.json` fails to configure `typeRoots` 🤔 ).

And the latest `@angualr/cli` generates not `src/tsconfig.json` but `src/tsconfig.app.json`, and using the latest generated tsconfig solves this error. So I'll update the fixture.

igor-dv (approved):
LGTM. 
@Hypnosphi any insights?

Hypnosphi:
>  So I'll update the fixture.

Does it mean that this change will breake people's apps bootstrapped with previous versions of `@angular/cli`

Quramy:
@Hypnosphi 

> Does it mean that this change will break people's apps that were initially bootstrapped with previous versions of @angular/cli?

The `@angular/cli` change related it is https://github.com/angular/angular-cli/pull/4924, and this PR merged on Feb 23, 2017 and is before `@angular/cli` v1.0.0. So I think we don't have to worry. How about it ?

igor-dv:
So is it good to go?

alterx:
I think we're good to go if the change was made pre v1 and we support v1+

Have we tested this?

Hypnosphi:
I'll change the base branch for a moment, just to check if it fixes issues with #2712

PR #2677, opened 2018-01-08
Remove src from indidivudal .npmignore files

hswolff:
Fixes #2602

cc @Hypnosphi

PR #2678, opened 2018-01-08
Remove `src/` from remaining `.npmignore`s.

tmeasday:
Following on from #2677

### DIFF
--- a/addons/knobs/.npmignore
+++ b/addons/knobs/.npmignore
@@ -1,2 +1,1 @@
-src
 .babelrc
--- a/addons/storyshots/.npmignore
+++ b/addons/storyshots/.npmignore
@@ -1,2 +1,1 @@
-src
 .babelrc
--- a/lib/ui/.npmignore
+++ b/lib/ui/.npmignore
@@ -1,2 +1,1 @@
-src
 .babelrc


shilman (approved):
LGTM

joscha:
This change breaks storybook in 3.3.6 via:

```

info @storybook/react v3.3.6
--
  | info
  | info => Loading custom addons config.
  | info => Loading custom webpack config (extending mode).
  | info Building storybook ...
  | ERR! Failed to build the storybook
  | ERR! ./node_modules/@storybook/addon-knobs/src/react/index.js
  | ERR! Module parse failed: Unexpected token (25:9)
  | ERR! You may need an appropriate loader to handle this file type.
  | ERR! \|   const initialContent = getStory(context);
  | ERR! \|   const props = { context, storyFn: getStory, channel, knobStore, initialContent };
  | ERR! \|   return <WrapStory {...props} />;
  | ERR! \| };
  | ERR! \|
  | ERR!  @ ./node_modules/@storybook/addon-knobs/dist/register.js 3:13-29
  | ERR!  @ ./node_modules/@storybook/addon-knobs/register.js
  | ERR!  @ ./conf/storybook/addons.js
  | ERR!  @ multi ./node_modules/@storybook/react/dist/server/config/polyfills.js ./conf/storybook/addons.js ./node_modules/@storybook/react/dist/client/manager

```

because now ` "jsnext:main": "src/index.js",` is picked up, which expects a webpack env that can load an es6-file (which is not true for all projects, for example ours doesn't have any es6, but only TS)

joscha:
removing `src` fixes it, not so sure about the `jsnext:main` part, yetm continuing to investigate.

joscha:
I am not sure it's related to `jsnext:main`, as we `import '@storybook/addon-knobs/register';` in `addons.js`. I am just not following, how we get from the chain `addons.js -> addon-knobs/register.js -> addon-knobs/dist/register.js` suddenly to `addon-knobs/src/react/index.js`

tmeasday:
@joscha let's take it to the other issue but it really seems like the `require('react')` line (in `addon-knobs/dist/register.js`) is somehow requiring `addon-knobs/src/react/index.js`

joscha:
> @joscha let's take it to the other issue but it really seems like the require('react') line (in addon-knobs/dist/register.js 3:13-29) is somehow requiring addon-knobs/src/react/index.js

yes, I can't make sense of that, though.

Hypnosphi:
> I can't make sense of that

@joscha please share your `.storybook/webpack.config.js` file in #2704 and we'll try to figure it out together

PR #2679, opened 2018-01-08
Add parameters to stories in the story store, and render them in app layers

tmeasday:
Issue: https://github.com/storybooks/storybook/issues/993

## What I did

Added support for `.add(name, fn, { parameters })`, and similarly passing the options parameters as part of the story `context`.

### Notes

~~This PR is incomplete, I have another one with some tests for the storystore that I will augment this with. I am interested in feedback on the feature and the approach.~~

Note that the various app layers no longer call `.getStory()` (now deprecated I think) and instead call the wrapped `.getStoryWithContext()` that handles the (generic) task of setting a context for the story. I think I caught all the places where we do this but let me know if not. ~~Also need to test properly for all 4 app layers.~~

### TODO

- [x] Add tests
- [x] Add API to react-native
- [ ] Add documentation
- [ ] Update addons to make use of it? [separate PR?]


examples/official-storybook/stories/core.stories.js
```diff
@@ -0,0 +1,12 @@
...
+
+storiesOf('core/options', module).add(
+  'passed to story',
+  ({ options: { myOption } }) => (
+    <div>
+      Value of <code>myOption</code> is '<code>{myOption}</code>'
+    </div>
+  ),
+  { myOption: 'value' }
+);
```
tmeasday: See this example for insight into how this feature works. More importantly, decorators and addons get passed the `context` also.

joscha:
This is quite cool, it would remove the need for the special abstraction around stories in `percy-storybook` as well. We currently introduce our own story type, add the options to a separate map and then extract them later again. See https://github.com/percy/percy-storybook/blob/3609b6f013e5a6f94a9b294a008b30bad425b9bc/packages/percy-storybook/src/index.js#L43-L62

app/angular/src/client/preview/angular/helpers.ts
```diff
@@ -3,9 +3,9 @@ import {
   enableProdMode,
   NgModule,
   Component,
-  NgModuleRef
+  NgModuleRef,
```
tmeasday: Not quite sure why prettier is reformatting this file. Any insights @igor-dv?

tmeasday:
Yes! Keen to get insight from such use cases.

shilman:
@alterx @tmeasday @ndelangen @igor-dv Any idea why prettier was not applied to the original `app/angular` source files? Inconvenient to review the changes when a lot of the changes are formatting changes. If there's something wrong with our tooling we should fix it? I was under the impression that we run prettier on precommits.

igor-dv:
Looks like tslint is not configured to run on precommit at all.  And eslint handles only js files. Maybe it's your IDE refactored it ? Will take a look on it soon.

Hypnosphi:
After resolving the conflicts, the noisy diff should be gone

tmeasday:
Ok, angular guff sorted.

igor-dv:
Looks legit. What if we have a few decorators, that need different options? Will it be like this :
```
add('..', 
 () => ..., 
 {
  notes: {},
  info: {},
 ...
})
```
By convention ?

tmeasday:
@igor-dv I suspect it might but I think that is the next question to answer after merging this one. There are some other related things, like special options:

```js
.add(..., () => ..., { decorators: ... });
```

Hypnosphi:
Just to be clear: those are not the same options which are set via addon-options?

tmeasday:
> Just to be clear: those are not the same options which are set via addon-options?

No, but this is a good point, we should have a way to set options per-app and per-chapter.

tmeasday:
Let me know what you think about the global and chapter `addOptions` API.

I also remembered that react-native isn't yet using the `@storybook/core` module, so I'll either need to reimplement the API there or change it to do so before we merge this.

Hypnosphi:
> I'll either need to reimplement the API there or change it to do so before we merge this

The latter sounds better to me =)

tmeasday:
Yep, on it (slowly)

tmeasday:
I rebased this off `master` so it now should work in RN too.

Do you think I should add examples of options usage to all example storybooks?

Hypnosphi:
I'm still afraid of possible confusion with addon-options. Can we pick some other name? `parameters` maybe?

tmeasday:
We could call them `storyOptions`? I'm not sure what we call them will be super-user visible?

If I could I would probably rename `addon-options` to something like `addon-settings` (it is really about setting up the manager window, rather than changing options to the stories), but I guess that's a bit much.

I'd be OK to call them `params` or `parameters` though. Any one else have a vote?

Hypnosphi:
>  I'm not sure what we call them will be super-user visible?

`addOptions` will

tmeasday:
> addOptions will

Agreed, we need to change it to either `addStoryOptions` or `addParameters`, or...? I wonder if anyone else wants to weigh in here?

In the meantime I added option stories to all example apps. Seems to work!

igor-dv:
"Parameters" feels to me a better word. What about "context"? Or maybe "state"?

tmeasday:
> What about "context"? Or maybe "state"?

I think these things will mostly be used to configure addons/decorators attached to the story so I would lean towards parameters over something like context.

It looks like `parameters` is leading the race here. If anyone else has any ideas, now is the time!

tmeasday:
Ok, renamed to `parameters`. I think this is ready to merge.

tmeasday:
Well except for the Angular snapshot changes, which are a bit baffling. Should I just accept them @igor-dv?

igor-dv:
Yeah. I don't know yet how to get rid of those ng-💩 things...

Hypnosphi (approved):

tmeasday:
🎉 going to start working on the next phase of my master plan now ;)

PR #2692, opened 2018-01-08
Add html and markup loaders to angular and vue apps

igor-dv:
Issue: html and markup loaders are used in vue/angular apps but are not listed in package json

## What I did
Added them to package.json
 
CC: @amcdnl


amcdnl (approved):
LGTM

Hypnosphi:
@ndelangen how is node security check configured? It seems to be hung

Issue #2688, opened 2018-01-08
Css loading complexity in angular app


igor-dv:
I would like to discuss here the problem we have with the css loading complexity for the angular app.

As I understand from the recent issues, there are several flows of loading css to the app
(Related issues: #2647, #2629, #2616)

### Global Import
Example: 

```js
import './some.css'
...
storyOf(...)
  .add(...);

```
or even in the `config.js`
```js
import '../bootstrap.css'
...
configure(loadStories, module);
```

Storybook should add it to the style tag in iframe 

### With the "styleUrls"
Example: 
```js
@Component({
  selector: 'storybook-app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css'], // <---
})
export class AppComponent {
  title = 'app';
}

```
In this case, it should be loaded via `raw-loader`, and angular will add it as it wishes:

## The problem 
Our angular app includes two rules for css:

1, in webpack.config.js: 
```js
{
  test: /\.(html|css)$/,
  loader: 'raw-loader',
  exclude: /\.async\.(html|css)$/,
}
```
2, in defaults/webpack.config,js:
```js
{
  test: /\.css$/,
  use: [
    require.resolve('style-loader'),
    loader: require.resolve('css-loader'),
    loader: require.resolve('postcss-loader'),
  ]
}
```
And we are getting some collision between these rules, so people need to extend webpack.config.js in order to get their angular components work in Storybook ([Example](https://github.com/storybooks/storybook/issues/2647#issuecomment-355598718) )

IMO extending webpack.config.js should be for some specific/rare usecases, but not for the common usage.

@storybooks/team  WDYT?

ndelangen:
> IMO extending webpack.config.js should be for some specific/rare usecases, but not for the common usage.

Agree 💯 

We have absolutely no preference in the webpack config ourselves in regards to css, we add things so most users have maximum convenience.

Quramy:
@igor-dv I'm very interested in this issue.

> people need to extend webpack.config.js 

The webpack configuration created by Angular CLI is too complex. So it's not easy to extend `@storybook/angular/server`.

I think Storybook should provide the way to configure webpack Angular CLI based project.

And I've created the helper function for PoC: 
https://github.com/Quramy/storybook-angular-cli-helper#install

igor-dv:
@Quramy , looks promising ! Don't you think it's better to make it as a default behavior? So nobody will need to extend webpack.config?

[...]

PR #2703, opened 2018-01-09
Add scss for components in angular apps by default.

igor-dv:
Issue: #2688

Looks like most of the people (I don't know if it's most, but that what I've seen recently) prefer using scss instead of css. 

## What I did
Changed the raw-loader to use scss

For components that use `styleUrls` - scss will be loaded with a raw-loader
For global imports - there will be still style-loader as before.


amcdnl (approved):
LGTM

Hypnosphi:
Is it a breaking change?

amcdnl:
@Hypnosphi - I believe its just for `styleUrls` if I'm correct.

app/angular/src/server/config/webpack.config.js
```diff
@@ -67,9 +67,13 @@ export default function(configDir) {
           loaders: [require.resolve('ts-loader'), require.resolve('angular2-template-loader')],
         },
         {
-          test: /\.(html|css)$/,
+          test: /\.html$/,
           loader: 'raw-loader',
-          exclude: /\.async\.(html|css)$/,
+          exclude: /\.async\.css$/,
```
Hypnosphi: Why do we need to exclude a css file is it doesn't pass the `test` anymore?

Hypnosphi:
> its just for `styleUrls` if I'm correct

This PR changes `styleUrls` in example app. Does it mean that users will need to change it as well?

app/angular/src/server/config/webpack.config.js
igor-dv: oops, I think I need to exclude html here.

igor-dv:
@Hypnosphi, it feels like a breaking change, but I prefer to look at it as a bug fix. Before, in order to make css work, people needed to extend webpack.config + exclude thigngs from css rules (because css didn't really work by default) + add scss. So now they maybe will need just to remove these extended webpack.config ...

ndelangen (approved):

Hypnosphi:
> css didn't really work by default

OK then

PR #2530, opened 2017-12-22
Switch to UglifyJS that supports ES6 and up

ggarek:
Issue:

## What I did
Use `uglify-es` through `uglifyjs-webpack-plugin` to be able to minify es6+ code.

I stumbled upon same issue, then found [this PR](https://github.com/storybooks/storybook/pull/1733), which was not very active recently. So i submit this PR in continuation to original one.

## How to test

I tested as proposed in the original PR. Here are some screenshots.

Build time (is ~28s)
screen shot 2017-12-22 at 16 04 26

Bundle size 
```
1.8M	storybook-static/static/manager.84cf3a1cca5073d55149.bundle.js
1.2M	storybook-static/static/preview.09c4e2758e41ca532a7c.bundle.js
```

screen shot 2017-12-22 at 16 14 57

**Note**
I have no opportunity to check IE 😭

PR #2705, opened 2018-01-10
Don't mangle function names for production builds

pascalduez:
Issue:

Function names are mangled by UglifyJS during production build.
They are important for React stateless components, especially in a styleguide context such as Storybook. This allows playing nicely with third party libraries such as [`react-element-to-jsx-string`](https://github.com/algolia/react-element-to-jsx-string/blob/333591a19420cb52f93b7c638bcaa2ca10bf2d11/src/parser/parseReactElement.js#L12).
The issue happened after #2530.

## What I did

Passed the [`compress: { keep_fnames: true }`](https://github.com/mishoo/UglifyJS2/tree/harmony#compress-options) option to all production builds UglifyJS plugin.
Did not noticed significant bundle size increase on our internal styleguide, which is quite large.

## How to test

Is this testable with jest or storyshots? Nope

Does this need a new example in the kitchen sink apps? Nope

Does this need an update to the documentation? Nope


ndelangen (approved):

Hypnosphi:
Thanks @pascalduez , totally makes sense

pascalduez:
Not sure if this should be applied to Angular, Vue builds as well.

Hypnosphi (changes requested):
> Not sure if this should be applied to Angular, Vue builds as well.

It should, please do that

pascalduez:
@Hypnosphi Done.

Hypnosphi (approved):

PR #2707, opened 2018-01-10
Revert "Add scss for components in angular apps by default."

Hypnosphi:
Reverts storybooks/storybook#2703

Oops I didn't notice the `do not merge` label. We should probably use “request changes” feature instead


igor-dv:
That's ok actually. I just forgot to remove the label =)

PR #2714, opened 2018-01-10
Reflect the new peer dependencies in docs and CLI templates

Hypnosphi:
Issue: As someone has pointed out, both quick and slow start guides don't work currently, because they're missing the newly added peerDependencies

### DIFF
--- a/docs/src/pages/basics/guide-react/index.md
+++ b/docs/src/pages/basics/guide-react/index.md
@@ -16,7 +16,7 @@ In this guide, we are trying to set up Storybook for your React project.
 ## Table of contents
 
 -   [Add @storybook/react](#add-storybookreact)
--   [Add react and react-dom](#add-react-and-react-dom)
+-   [Add react, react-dom, and babel-core](#add-react-react-dom-and-babel-core)
 -   [Create the config file](#create-the-config-file)
 -   [Write your stories](#write-your-stories)
 -   [Run your Storybook](#run-your-storybook)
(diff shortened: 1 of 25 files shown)


docs/src/pages/basics/guide-react/index.md
```diff
@@ -29,12 +29,13 @@ First of all, you need to add `@storybook/react` to your project. To do that, si
 npm i --save-dev @storybook/react
 ```
 
-## Add react and react-dom
+## Add react, react-dom, and babel-core
 
-Make sure that you have `react` and `react-dom` in your dependencies as well because we list these as a peerDependency:
+Make sure that you have `react`, `react-dom`, and `babel-core` in your dependencies as well because we list these as a peerDependency:
```
joscha: I think this might read better as:
```
Make sure that you have `react`, `react-dom` in your dependencies and and `babel-core` in your devDependencies, because we list these as a peerDependency:
```

Hypnosphi: If one has `babel-core` in their dependencies, it will work as well

joscha: yes, but it reads as if babel-core needs to be an actual dependency, I guess I am too literal with the word `dependencies` 🤷‍♂️

shilman (approved):
LGTM, though the snapshot test is failing due to dep ordering. (I tried updating it it on my local machine, but the snapshot order looks correct to me?!!)

Hypnosphi:
Looks like there's some kind of race condition, will fix

Issue #2716, opened 2018-01-11
build-storybook not exitting when fail


davidkwan95:
### Issue details

Expected behavior: build-storybook should exit when err.
Actual behavior: process is stuck and not exitting

### Steps to reproduce

1. Some fail imports in code
2. build-storybook

### Please specify which version of Storybook and optionally any affected addons that you're running

- @ storybook/react 3.3.6
- @ storybook/addon-actions 3.3.6
- @ storybook/addon-info 3.3.6
- @ storybook/addon-links 3.3.6

### Affected platforms 

- Node v8.9.3

Hypnosphi:
Can't reproduce it on fresh CRA
```
$ npm run build-storybook

> test-app@0.1.0 build-storybook /Users/jetbrains/test-app
> build-storybook -s public

info @storybook/react v3.3.6
info
info => Loading custom addons config.
info => Using default webpack setup based on "Create React App".
info => Copying static files from: public
info Building storybook ...
ERR! Failed to build the storybook
ERR! Module not found: Error: Can't resolve '@stoybook/addon-actions' in '/Users/jetbrains/test-app/src/stories'
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! test-app@0.1.0 build-storybook: `build-storybook -s public`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the test-app@0.1.0 build-storybook script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/jetbrains/.npm/_logs/2018-01-11T10_54_03_281Z-debug.log
```

Hypnosphi:
Can you please create a GitHub repo with minimal reproduction of your issue?

davidkwan95:
I am trying to reproduce it in a new project, but so far I've been unable to reproduce it.

ohelixa:
I have the same issue

shilman:
@ohelixa what version?

ohelixa:
Latest 5.0.5

PR #2717, opened 2018-01-11
using process.exitCode instead of process.exit()

davidkwan95:
This PR should fix issue #2716


ndelangen:
Thank you for this PR, Looks good. Can you teach my why the issue occurred and why this fixes it?

Hypnosphi:
I'm intrigued as well. Docs actually [say](https://nodejs.org/api/process.html#process_process_exit_code) that using `process.exit()` is not a recommended practice, but the reason is that it may force the exit *too early* which is contrary to the referenced issue

davidkwan95:
I'm not too sure, but I noticed that calling `process.exit()` inside the callback function seems to wait for some process to be finished before exitting. I'm assuming it has something to do with webpack or node.
Without calling the `process.exit()`, we will be able to exit the program after the event loop is done, but since we got an error, we need to set the exit code to 1 by using `process.exitCode = 1`

Hypnosphi:
> after the event loop is done

If there's some process running (as you say), event loop won't be sone until the process is finished

davidkwan95:
I've checked that the fix works in my current project.

ndelangen:
So I've actually hit this problem now on an other library.
https://github.com/danielstjules/jsinspect/pull/69

I'm ready to approve this.

ndelangen (approved):

PR #2723, opened 2018-01-11
Add loki to the docs

cshell7:
Issue: Loki is not in the docs, more people should know about it cause its handy.

## What I did
Not too sure where this should be in the docs, but with how great of a tool loki is I'm surprised it's not already.
## How to test

Is this testable with jest or storyshots? NR

Does this need a new example in the kitchen sink apps? ??

Does this need an update to the documentation? Y


ndelangen (approved):

ndelangen:
Deployed

PR #2735, opened 2018-01-12
Serve styles and assets using .angular-cli webpack configuration

Quramy:
Issue: See #2645, #2688 

## What I did

Add ability to configure webpack using `@angular/cli` and user's `.angular-cli.json`.

So `@storybook/anguar` serves according to the .angular-cli.json: 

- Global styles
- Static asset files

And if user don't have .angular-cli.json, the minimum configuration is provided as before.

## How to test

See example/angular-cli manually 🙏 . I added new component and story under `examples/angular-cli/src/stories/component-with-style`. Open this story, you can see the following styled DOM:

- Scoped CSS(with templateURL)
- Scoped SCSS(with templateURL)
- Global CSS

And open `localhost:9008/assets/common.css`, you can confirm that static files listed .angular-cli.json are served.


Quramy:
@igor-dv  @alterx how about this?

alterx:
cc @amcdnl :D

app/angular/src/server/angular-cli_config.js
```diff
@@ -0,0 +1,69 @@
...
+  if (!fs.existsSync(fname)) {
+    return null;
+  }
+  const cliConfig = JSON.parse(fs.readFileSync(fname, 'utf8'));
+  if (!cliConfig.apps || !cliConfig.apps.length) {
+    throw new Error('.angular-cli.json must have apps entry.');
+  }
+  const appConfig = cliConfig.apps[appIndex];
+
+  // FIXME dummy value
```
alterx: Which of these are dummy values?

app/angular/src/server/config.js
```diff
@@ -3,6 +3,10 @@ import fs from 'fs';
 import path from 'path';
 import loadBabelConfig from './babel_config';
 import loadTsConfig from './ts_config';
+import {
+  getAngularCliWebpackConfigOptions,
+  applyAngularCliWebpackConfig,
+} from './angular-cli_config';
 
 // avoid ESLint errors
 const logger = console;
```
ndelangen: This should use `@storybook/node-logger`

app/angular/src/server/angular-cli_config.js
```diff
@@ -0,0 +1,69 @@
...
+
+export function applyAngularCliWebpackConfig(baseConfig, cliWebpackConfigOptions) {
+  if (!cliWebpackConfigOptions) return baseConfig;
+
+  let ngcliConfigFactory;
+  try {
+    // We should not require('@angular/cli') at the top script level because user project might not have `@angular/cli`.
+    // eslint-disable-next-line global-require, import/no-extraneous-dependencies
+    ngcliConfigFactory = require('@angular/cli/models/webpack-configs');
+  } catch (e) {
```
ndelangen: I'd like to see this changed a bit.

I'd like to extract a function that detects the installment of `@angular/cli`.

If it's not installed, we just merge the base, we might log that.
If it's installed we try to get their webpack config.
This might fail, since it's a deep file in a npm module.
We should log in either case.

ndelangen (changes requested):

amcdnl (approved):
This is great!

app/angular/src/server/config/webpack.config.js
```diff
@@ -76,10 +76,6 @@ export default function(configDir) {
           loader: 'raw-loader',
           exclude: /\.async\.css$/,
         },
-        {
-          test: /\.scss$/,
```
igor-dv: Don't we want to configure scss in case there is no angular-cli ?

alterx: I think we should do this:
* `angular-cli` present -> use whatever's their config
* not present? Just fallback to good ol' css. 

Technically, if they're not using the cli we aren't officially supporting it. So they should config everything manually. This feels consistent.

@amcdnl what do you think?

app/angular/src/server/angular-cli_config.js
Quramy: it's `outputPath: 'outPutpath'`. I'll amend comment 🙇

app/angular/src/server/config/webpack.config.js
amcdnl: I think the vast majority of people will be using the cli so its pretty safe bet. There is the 1% case when they eject from the CLI we might need to think about ( https://github.com/angular/angular-cli/wiki/eject ) but that just nets the 'guts' of the CLI in a basic webpack config file as if you configured it manually yourself.

Personally, I'd fall under the thought process of try to handle as much as you can w/o config (seems to be a trend now i.e. parcel) so leaving the scss wouldn't hurt. Just my 2 cents tho.

Quramy: > so leaving the scss wouldn't hurt

Thanks your feedback, I'll revet this change 😄

app/angular/src/server/angular-cli_config.js
```diff
@@ -0,0 +1,87 @@
+const path = require('path');
+const fs = require('fs');
+
+const logger = console;
+
+export function isAngularCliInstalled() {
```
igor-dv: Why exporting it?

app/angular/src/server/angular-cli_config.js
```diff
@@ -0,0 +1,87 @@
...
+
+  // cliStyleConfig.entry adds global style files to the webpack context
+  const entry = {
+    ...baseConfig.entry,
+    ...cliStyleConfig.entry,
+  };
+
+  const mod = {
+    ...baseConfig.module,
+    rules: [...cliStyleConfig.module.rules, ...styleRules],
```
igor-dv: Do we actually need these `styleRules` if they should be filtered out?

Quramy: Sorry, the variable name `styleRules` is so bad 🙇 
This array includes rules for `.jsx`, `.ts`, ..., so we need it.

I'll change the variable's name.

app/angular/src/server/config.js
igor-dv: @Quramy, I think this is the last comment to take care of

igor-dv:
@Quramy , you need to update snapshots to fix tests =)

Quramy:
@igor-dv  storyshots for Angular 🎉  !
I added the shapshot file 😄

igor-dv (approved):

alterx (approved):

ndelangen (approved):

igor-dv:
@Quramy , As I understand, after this fix we can't already import global css into `config.js`, because now I see this in the example app:

image

Having this config:

```js
import { configure } from '@storybook/angular';
import addCssWarning from '../src/cssWarning';
import '../src/assets/common.css' // <--- this

addCssWarning();

function loadStories() {
  // put welcome screen at the top of the list so it's the first one displayed
  require('../src/stories');

  // automatically import all story ts files that end with *.stories.ts
  const req = require.context('../src/stories', true, /\.stories\.ts$/)
  req.keys().forEach((filename) => req(filename))
}

```

If it's ok I'll just put the css from `common.css` to the `style.css`

Quramy:
@igor-dv 

Importing global CSS file manually(`import '../src/assets/common.css'`) is no longer needed. 

For instance, in `example` app `src/style.css` is listed in `.angular-cli.json` as global style files:

```js
  "apps": [
    {
      :
      "styles": [
        "styles.css"
      ],
      :
      }
    }
  ],
```

So it's imported automatically as `ng serve` does . Specifically, [this procedure](https://github.com/storybooks/storybook/blob/master/app/angular/src/server/angular-cli_config.js#L66-L70) adds `src/styles.css` to webpack context.

So, I think we can remove the following lines from example's config:

```
import '../src/assets/common.css' // <--- this

addCssWaring();
```

PR #2747, opened 2018-01-14
fixed propTable values + improve UI (addon-info)

SuperOl3g:
Issue:
In Story example, propTypes in propTable doesn't shown at all.
Besides, it works only if propType be passed as an object (in fact it also could be a string).
Besides, it show nothing with `propType.object`.

## What I did

- fixed al this cases;
- improved the view of `propType.oneOf` to differentiate it from `propType.oneOfType;
- improved view of collapsed `propType.shape` view


Hypnosphi:
It behaves weirdly with empty shape (`PropTypes.shape({})`):
https://deploy-preview-2747--storybooks-official.netlify.com/?selectedKind=Addons%7CInfo.Options.inline&selectedStory=Inlines%20component%20inside%20story&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

addons/info/src/components/types/OneOf.js
```diff
@@ -1,7 +1,20 @@
 import React from 'react';
 import { TypeInfo } from './proptypes';
+import PropVal from '../PropVal';
 
-const OneOf = ({ propType }) => <span>{propType.value.map(({ value }) => value).join(' | ')}</span>;
+const OneOf = ({ propType }) =>
+  Array.isArray(propType.value) && (
```
Hypnosphi: So what do we do if it's not an array?

addons/info/src/components/types/OneOf.js
```diff
@@ -1,7 +1,20 @@
...
+    <span>
+      {propType.value.reduce((arr, { value }, i) => {
+        arr.push(<PropVal val={value} />);
```
Hypnosphi: Looks like it will generate react key warnings in development

Hypnosphi:
> In Story example, propTypes in propTable doesn't shown at all.

Which example do you mean?

Hypnosphi:
> improved the view of `propType.oneOf` to differentiate it from `propType.oneOfType`;

It's cool that it differentiates better, but now it adds extra quotes: `"'News'" | "'Photos'"`

PR #2749, opened 2018-01-15
Refactored story store / client api tests

tmeasday:
Issue: unit tests targetted the client_api, where in some cases it was more appropriate to target the story store.

Also the tests basically re-implemented the story store which seems sort of circular (I just use the real implementation if am not testing the details of what internal APIs are called).

## What I did

Refactored the tests

## How to test

Jest tests


igor-dv (approved):
Very nice!

PR #2789, opened 2018-01-20
Fix css warning in angular-cli example

igor-dv:
Issue: https://github.com/storybooks/storybook/pull/2735#issuecomment-359160195

## What I did
Moved the css-rules-warning class to style.css


Quramy (approved):
@igor-dv 👍

PR #2791, opened 2018-01-20
ADD __dirname support

ndelangen:
Add __dirname support to default webpack config & add it to official example & docs

I had a discussion about generating the storybook name based on the directory structure.
It's possible to get the path of the current file, using __dirname. But without a special webpack config, this variable is mostly useless. https://github.com/webpack/webpack/issues/1599

So I added the config to the default webpack config, added it to the official storybook and added a story using it.

Also added it to the docs.

---
This change is [Reviewable](https://reviewable.io/reviews/storybooks/storybook/2791)


ndelangen:
So it's kind of complicated to do, because storyshots run without webpack and thus the implementation of __dirname is different.

igor-dv:
Will test today 👌

ndelangen:
Thanks I'm unsure how to deal with the problem state above.

igor-dv:
I played a bit with the solution and Storyshots. I think a possible workaround could be providing storyName mapping function to Storyshots api... And users will need to map this full path to the name they want to see in their snapshots. But tbh, from the second glance at the solution, I am unsure about having webpack hooks in more places is a good idea

igor-dv:
Maybe there is an option to do something like this: 
```js
import base from 'current-dir-resolve'
```

So we will be able to mock the `current-dir-resolve` in jest

?

tmeasday:
I wonder if this problem could be solved via a babel plugin? This is the common solution for such "needs to work in both webpack and outside" issues in next.

ndelangen:
Yes @tmeasday Is was considering using [babel-plugin-preval](https://github.com/kentcdodds/babel-plugin-preval)

tmeasday:
This is cool! Can we take it a step further and export a macro that people can just use? Like

```js
import React from 'react';
import dirname from '@storybook/react/dirname-macro';

import BaseButton from '../components/BaseButton';

storiesOf(`Other|${dirname}/Dirname Example`, module)
  .add('story 1', () => <BaseButton label="Story 1" />)
  .add('story 2', () => <BaseButton label="Story 2" />);
```

ndelangen:
@tmeasday I've created a babel-plugin called `babel-plugin-paths`, it allows a user to 'import' quite a few path-sections like filename, extention, base, npmRoot etc.

package.json
```diff
@@ -42,6 +42,8 @@
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-polyfill": "^6.26.0",
     "babel-preset-env": "^1.6.1",
+    "babel-plugin-macros": "^2.0.0",
+    "babel-plugin-paths": "^1.0.0",
```
igor-dv: Why are these plugins needed in `ui`?

app/polymer/package.json
```diff
@@ -33,6 +33,8 @@
     "autoprefixer": "^7.1.6",
     "babel-core": "^6.26.0",
     "babel-loader": "^7.1.2",
+    "babel-plugin-macros": "^2.0.0",
+    "babel-plugin-paths": "^1.0.0",
```
igor-dv: As I understand `babel-plugin-paths` is needed only in the example app, why should it be added to every `app/*` ?

igor-dv (commented):
image
 
Sadly windows is a party pooper =(

Also, it just added things to storyshots
image

ndelangen:
uhg of course it is.. I'll look for a solution.

ndelangen:
@igor-dv Does the fix work for you?

tmeasday:
@ndelangen so awesome! Now I want to push you to take this even further!

What about something super cool like:

```js
// In .storybook/config.js
import { configure } from 'paths.macro';
// I can think of a lot of other options we could come up with ;)
configure({ srcRoot: 'src/' });

// If the file is `src/components/Foo.stories.js`
import { kind } from 'paths.macro';

storiesOf(kind, module) === storiesOf('components/Foo', module);
```

Also, as a related thing, we could potentially use this same technique to solve the "passing `module`" problem:

```js
import { stories } from 'paths.macro'; // (<- this would probably be @storybook/macro)

stories.add() === storiesOf(kind, module) === storiesOf('components/Foo', module);
```

Or is that getting into the realm of too much magic?

ndelangen:
I personally think that is too much magic, for now.

It's definitely out of scope for the generic paths macro.

I'll give the idea of some storybook specific macros some more thought. I'll discuss it with a few users within my reach.

tmeasday:
@ndelangen I’m ok with whatever you prefer but I will make the above happen somehow if you don’t. (If I can)

I guess what ends up in the docs is what this PR is about

ndelangen:
I changed the babel macro to always convert to POSIX paths

ndelangen:
If you could verify @igor-dv? 🙇

ndelangen:
Should be good to merge I think.

ndelangen:
Reviewed 3 of 7 files at r7, 2 of 8 files at r8, 1 of 5 files at r9, 7 of 8 files at r10.
Review status: 13 of 29 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[package.json, line 46 at r4](https://reviewable.io:443/reviews/storybooks/storybook/2791#-L3mqQmESg6eHb34C-jD-r4-46:-L3n036fkOsUd1rmSLZX:bo93cld) ([raw file](https://github.com/storybooks/storybook/blob/d77e412854948c36e415fbf3034843b18c5cafcc/package.json#L46)):*
<details><summary><i>Previously, igor-dv (Igor) wrote…</i></summary><blockquote>

Why are these plugins needed in `ui`?
</blockquote></details>

Removed them

---

*[app/polymer/package.json, line 37 at r4](https://reviewable.io:443/reviews/storybooks/storybook/2791#-L3mqQmESg6eHb34C-j4-r4-37:-L3n-mM1m15LWb8ON86d:brlf3dc) ([raw file](https://github.com/storybooks/storybook/blob/d77e412854948c36e415fbf3034843b18c5cafcc/app/polymer/package.json#L37)):*
<details><summary><i>Previously, igor-dv (Igor) wrote…</i></summary><blockquote>

As I understand `babel-plugin-paths` is needed only in the example app, why should it be added to every `app/*` ? 
</blockquote></details>

Removed it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/storybooks/storybook/2791)*

ndelangen:
Reviewed 1 of 6 files at r1, 6 of 19 files at r2, 2 of 12 files at r3, 7 of 21 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/storybooks/storybook/2791)*

ndelangen:
Reviewed 1 of 12 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/storybooks/storybook/2791)*

igor-dv:
@ndelangen , what are these experimentations? 😂 btw, do you need more "windows" help to go on?

igor-dv:
never mind 🤦‍♂️

igor-dv (approved):
May the force be with you

Issue #2655, opened 2018-01-05
Scrollbar problems: Visible disabled scrollbars and overlapping of fixed elements


sapkra:
### Issue details
The decorator set's the `overflow` to `scroll`, but it should be `auto` because if the content is not scrollable the scrollbars will be hidden with `auto`.

There is also an overlapping with fixed elements.
Problem: You can't place a fixed element relative to another fixed element but the backgrounds wrapper is a fixed element.

You can see both problems in the screenshot below.

### Please specify which version of Storybook and optionally any affected addons that you're running
- storybook/react 3.3.3
- storybook/addon-backgrounds 3.3.3

### Screenshot
![image](https://user-images.githubusercontent.com/8087568/34604184-330fc66e-f207-11e7-85cf-4a2be9aec1f8.png)

Hypnosphi:
Hi @sapkra, sorry for the delay

Can you please elaborate a bit more what's the expected behaviour for fixed element in your case?

PR #2807, opened 2018-01-22
Addon-backgrounds: set background on iframe instead of adding a wrapper in preview

Hypnosphi:
It fixes second part of #2655

---
This change is [Reviewable](https://reviewable.io/reviews/storybooks/storybook/2807)


ndelangen:
@Hypnosphi We can actually improve this addon a lot.

We can set the iframe background to be transparent, and set the background in the manager.

What do you think?

Hypnosphi:
@ndelangen Sorry, missed your comment.

Yeah, that sounds like a good idea, will try.

Hypnosphi:
@ndelangen wouls your idea mean an API change? Currently the addon is used as decorator, which implies that it adds something to preview not manager

Hypnosphi:
Another problem is, addons can't currently add things anywhere in manager outside of the addons panel. Should we allow them?

ndelangen:
@Hypnosphi the viewport addon seems to be able to just fine?

igor-dv (commented):
LGTM, but I wonder if these changes are considered backward compatible. For example for people that use both, storyshots and background addons.

Hypnosphi:
Do you mean that people will need to update their snapshots? To me personally it seems OK. Plus we're actually removing noise from storyshots: https://github.com/storybooks/storybook/pull/2807/files#diff-261e00eb9848ef08cfb37eb5e7e0b012

igor-dv:
Yeah, that what I meant. Maybe it worth adding some note about it, Just to reduce confusion... I don't know where though

igor-dv (approved):

Hypnosphi:
I'm pretty sure we're introducing changes that can affect existing snapshots all the time

PR #2809, opened 2018-01-22
ADD autolabeler.yml for https://github.com/probot/autolabeler

ndelangen:
This adds https://github.com/probot/autolabeler to this repo.

This bot adds labels to PRs, very useful for our changelog @shilman


.github/autolabeler.yml
```diff
@@ -0,0 +1,31 @@
+'accessibility': []
```
Hypnosphi: If we need to list tags with empty patterns for some reason, you missed `patch`

.github/autolabeler.yml
```diff
@@ -0,0 +1,31 @@
...
+'addon: a11y': ["addons/a11y/**"]
+'addon: actions': ["addons/actions/**"]
+'addon: backgrounds': ["addons/backgrounds/**"]
+'addon: centered': ["addons/centered/**"]
+'addon: events ': ["addons/events/**"]
```
igor-dv: idk if it's critical, but there are few entries with the extra space at the end

.github/autolabeler.yml
ndelangen: no we don't I just extracted the first page worth of tags, I'll clean up the empty ones and the spaces @igor-dv mentioned

.github/autolabeler.yml
Hypnosphi: I wonder if there is some kind of yamllint =D

ndelangen:
Should be good now right?

.github/autolabeler.yml
```diff
@@ -0,0 +1,24 @@
...
+'addon: viewport': ["addons/viewport/**"]
+'app: angular': ["app/angular/**"]
+'app: polymer ': ["app/polymer/**"]
+'app: react-native': ["app/react-native/**"]
+'app: react': ["app/react/**"]
+'app: vue': ["app/vue/**"]
+'babel / webpack': ["webpack", "babel"]
+'cli': ["lib/cli/**"]
+'compatibility with other tools': []
+'documentation': ["docs", "*.md"]
```
igor-dv: Would you consider adding things to any of `example/*` as a documentation change?  Or changing things in `core` / linter files / as a maintenance?

ndelangen: Shall we merge and add details later?

Hypnosphi (approved):

PR #2826, opened 2018-01-24
extending webpack section is no longer needed for the common usage

Quramy:
Issue: #2688 

## What I did

I Removed extending webpack config section from Angular guide because I think this section gets worthless via #2688 and #2735 .


igor-dv (approved):
Great

PR #2846, opened 2018-01-27
Ts distribution

igor-dv:
Issue: Distribute ts files

## What I did
- [x] add dev:ts for compiling ts to dist
- [x] used "concurrently" to watch both `ts` and `js`
- [x] compile ts with `prepare.js`
- [x] do not copy src ts files to dist with babel ( remove --copy-files ? 🤔  )

Notes: 
1. When running `dev` both babel and tsc will watch on files changes. Since babel is running with --copy-files flag it will also copy changed *.ts files to dist, currently it's not possible to ignore ts files (see comment in code). 
2. When running `bootstrap` (that uses `prepare` script), those unneeded .ts files will be removed. I assume that publishing to npm is done after the fresh bootstrap, so fix me if I am wrong (@shilman)

---
This change is [Reviewable](https://reviewable.io/reviews/storybooks/storybook/2846)


igor-dv:
I see Netlify fails on compiling ts but it does work for me on windows. Can someone take a look?

image

alterx (approved):

ndelangen (approved):

PR #2855, opened 2018-01-27
App-angular: upgrade ts-loader to 3.3.1

Hypnosphi:
[Changelog](https://github.com/TypeStrong/ts-loader/releases)

Angular example seems to work just OK

---
This change is [Reviewable](https://reviewable.io/reviews/storybooks/storybook/2855)

### DIFF
--- a/app/angular/package.json
+++ b/app/angular/package.json
@@ -75,7 +75,7 @@
     "rxjs": "^5.4.2",
     "sass-loader": "^6.0.6",
     "style-loader": "^0.20.2",
-    "ts-loader": "^2.2.2",
+    "ts-loader": "^3.3.1",
     "uglifyjs-webpack-plugin": "^1.2.0",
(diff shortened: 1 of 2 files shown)


danielduan (approved):
```
drop configFileName support (replaced by configFile) BREAKING CHANGE
```
might be safer to save this for the next major or minor instead of patch? otherwise, seems okay.

Hypnosphi:
We don't pass it: https://github.com/storybooks/storybook/blob/master/app/angular/src/server/config/webpack.config.js#L73

igor-dv:
We do pass here : https://github.com/storybooks/storybook/blob/master/app/angular/src/server/config.js#L23

Hypnosphi:
Ok, but the only fiels is `configFile` which isn't deprecated https://github.com/storybooks/storybook/blob/master/app/angular/src/server/ts_config.js#L23

PR #2856, opened 2018-01-28
Add option to pass custom styles for angular components

igor-dv:
Issue: https://github.com/storybooks/storybook/issues/2616#issuecomment-361004910

## What I did

Added an option to provide custom styles for angular components. Works only with templates becasue "dynamic component" is generated only for the templates now: 

```js
storiesOf('Custom Style', module).add('Default', () => ({
  moduleMetadata: {
    declarations: [Button],
  },
  props: {
    text: 'Button with custom styles',
    onClick: action('log'),
  },
  styles: [
    `
    storybook-button-component {
      background-color: yellow;
      padding: 25px;
    }
  `,
  ],
}));
```

---
This change is [Reviewable](https://reviewable.io/reviews/storybooks/storybook/2856)


alterx:
Just a thought, are we sure we don't want this as an add-on?
I mean, is there a similar way to achieve this in the other apps (vue, react, polymer?)

igor-dv:
🤷‍♂️  I don't know... Can we achieve the same behavior for all the apps with custom styles?

alterx:
Dunno, I was just thinking out loud here. I guess if, for example, someone is using react and css-in-js then we wouldn't be able to, but if they're using regular css we could.

Anyway, if we want we can reuse most of this code to create an add-on later on. This LGTM :)

igor-dv:
@alterx , would you mind to accept? 😏

alterx (approved):

alterx:
Of course, good sir. There you go 🎉

Issue #2451, opened 2017-12-09
Multiple story hierarchies


Havelaer:
Hi guys,

I've been working on a feature to have multiple hierarchies (or see it as a flattened first level). 
It's a small iteration on Igor's [story-hierarchy-in-storybook](https://medium.com/sears-israel/story-hierarchy-in-storybook-3-2-7506b5a6ceb9). 

A second 'Addons' hierarchy:
![screen shot 2017-12-08 at 21 36 52](https://user-images.githubusercontent.com/817113/33794001-0ecb3774-dcc3-11e7-9170-dfe4e6dea5bf.png)

Eligible for a PR? And should I submit it to master or to another branch? 
https://github.com/Havelaer/storybook/commits/feature/multiple-hierarchies

Hypnosphi:
Looks promising. Please base on `release/3.3` branch

Hypnosphi:
Released as [3.4.0-alpha.0](https://github.com/storybooks/storybook/releases/tag/v3.4.0-alpha.0)

PR #2857, opened 2018-01-28
UI: remove the frustrating zero

Hypnosphi:
Issue:
screen shot 2018-01-28 at 15 34 56

The issue was introduced with #2451, so it's only present on alpha releases

---
This change is [Reviewable](https://reviewable.io/reviews/storybooks/storybook/2857)

### DIFF
--- a/lib/ui/src/modules/ui/components/stories_panel/index.js
+++ b/lib/ui/src/modules/ui/components/stories_panel/index.js
@@ -25,7 +25,7 @@ const storyProps = [
 ];
 
 function hierarchyContainsStories(storiesHierarchy) {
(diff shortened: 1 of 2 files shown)


lib/ui/src/modules/ui/components/stories_panel/index.js
```diff
@@ -25,7 +25,7 @@ const storyProps = [
 ];
 
 function hierarchyContainsStories(storiesHierarchy) {
-  return storiesHierarchy && storiesHierarchy.map.size;
+  return storiesHierarchy && storiesHierarchy.map.size > 0;
```
igor-dv: I think it's a place for easter eggs 😁

igor-dv (approved):

Issue #2859, opened 2018-01-28
master tests not passing


kn0ll:
first i'd like to say... apologies for this issue... all signs would suggest this is very much *my* problem and not storybook's...

### Issue details

- What did you expect to happen?

on a fresh checkout, i expected the tests to pass.

- What happened instead?

the tests did not pass. instead they fail with snapshot errors, seemingly always involving dates. some dates seem to be off by a day (`January 20, 2017` vs `January 19, 2017)`-- i could imagine this has something to do with regenerating snapshots? forgive me, i am unfamiliar with snapshot testing. but other dates are actually correct, only parsed and displayed differently (`2017-1-20` vs `1/20/2017`). which is a little more suspicious, no? you can find the full output of my tests here: https://gist.github.com/kn0ll/49cbca55faffe74fe1aee40229135521

### Steps to reproduce

```
git clone git@github.com:storybooks/storybook.git
cd storybook
yarn install
yarn bootstrap --all
yarn test --core
```

### Please specify which version of Storybook and optionally any affected addons that you're running

`commit f8bd011e7374ff65f442aa1a8e6963f2d8d2b85d (HEAD -> master, origin/master, origin/HEAD)`

### Affected platforms 

- macOS 10.13.2 (17C205)
- node v9.4.0
- yarn 1.3.2
- [node_modules versions](https://gist.github.com/kn0ll/f3615c98dd56828e716fde750b16b72d)

Hypnosphi:
@tobilen this looks related to your change in date displaying

tmeasday:
@Hypnosphi - remind me, did we end up fixing that on the other issue?

Also weird with the locale printing the date differently. I would propose we just change this test to print out the date as a timestamp or something; we really don't need to test JS date formatting in our tests ;)

Hypnosphi:
> did we end up fixing that on the other issue

Not really

I think we should just set both locale and timeyone explicitly everywhere we display a date

tmeasday:
Hopefully this helps: https://github.com/storybooks/storybook/pull/2861

tmeasday:
@kn0ll can you try latest `master` please?

tvooo:
Had the same issue last week. Can confirm the tests now pass with a fresh checkout of master. Thanks!

kn0ll:
seeing the same issue. will check again this evening to make sure it's not loading old linked module or something.

PR #2860, opened 2018-01-28
files knob

kn0ll:
Issue: my component can accept file input but i can't create a knob for it.

## What I did

created a knob for file input.

## How to test

i added the knob to the `all knobs` story. let me know if this is insufficient.

---

notes: some of these changes break from the pattern of other knobs... with reason...

1.  i added `browser: true` to `.eslintrc.js`. this is so it lints `FileReader`, a global object of window. not trying to impose anything here though. if you'd prefer i lint this another way, like disable linting for those lines, i can do that...

2. files are passed as Data URLs. i ran into security and serialization issues passing input files between iframes and the only in-scope solution i could come up with... was only parsing files as Data URLs. if the component requires other types, it would have to manage that itself (which kind of sucks-- but at least keeps the knob robust). [here is an example wrapper component that could manage this type of deserialization](https://gist.github.com/kn0ll/81df13a7306dcfa15fd69727c7686da2)

3. i set serialize and deserialize to undefined. this is because:
A. certain files, when serialized to the URL, break on reload (404 or something).
B. you cannot actually set the value of input type file, and so it serves no purpose anyway.

3. i was unable to pass tests locally because of #2859... and i am unsure how to clean up the failing checks for this PR... sorry...

thank you all...


Hypnosphi:
> i added browser: true to .eslintrc.js. this is so it lints FileReader, a global object of window

Please use `import {FileReader} from 'global'` instead

addons/knobs/example/stories/index.js
```diff
@@ -20,6 +20,8 @@ stories.add('with all knobs', () => {
 
   const passions = array('Passions', ['Fishing', 'Skiing']);
 
+  const images = files('Happy Picture', ['data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAQAAAC1+jfqAAAABGdBTUEAALGPC/xhBQAAACBjSFJNAAB6JgAAgIQAAPoAAACA6AAAdTAAAOpgAAA6mAAAF3CculE8AAAAAmJLR0QA/4ePzL8AAAAHdElNRQfiARwMCyEWcOFPAAAAP0lEQVQoz8WQMQoAIAwDL/7/z3GwghSp4KDZyiUpBMCYUgd8rehtH16/l3XewgU2KAzapjXBbNFaPS6lDMlKB6OiDv3iAH1OAAAAJXRFWHRkYXRlOmNyZWF0ZQAyMDE4LTAxLTI4VDEyOjExOjMzLTA3OjAwlAHQBgAAACV0RVh0ZGF0ZTptb2RpZnkAMjAxOC0wMS0yOFQxMjoxMTozMy0wNzowMOVcaLoAAAAASUVORK5CYII=']);
```
Hypnosphi: Please add this example to `examples/offisial-storybook` instead. `addons/*/example`s are outdated and probably will be removed. Sorry for the confusion

.eslintrc.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
     sourceType: 'module',
   },
   env: {
+    browser: true,
```
Hypnosphi: See https://github.com/storybooks/storybook/pull/2860#issuecomment-361094736

addons/knobs/src/components/types/Files.js
```diff
@@ -0,0 +1,36 @@
...
+    const fileReader = new FileReader();
+    fileReader.onload = e => resolve(e.currentTarget.result);
+    fileReader.readAsDataURL(file);
+  });
+}
+
+const FilesType = ({ knob, onChange }) => (
+  <input
+    id={knob.name}
+    type="file"
```
Hypnosphi: It would be handy to be able to pass `accept` atribute here
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#Limiting_accepted_file_types

addons/knobs/src/vue/index.js
```diff
@@ -11,10 +11,11 @@ import {
   date,
   select,
   button,
+  files,
   manager,
 } from '../base';
 
-export { knob, text, boolean, number, color, object, array, date, select, button };
+export { knob, text, boolean, number, color, object, array, date, select, button, files };
```
Hypnosphi: Can it be supported in angular and polymer as well?

Hypnosphi (changes requested):

Hypnosphi:
@kn0ll are you still interested in finishing that? It's OK if you aren't, I'll just pick it up then

ndelangen:
@kn0ll This is really useful, let's get this finished and merged!

kn0ll:
sorry i fell out of the loop all. other priorities came up... @Hypnosphi if you've not started i will take it over the finish line today.

Hypnosphi:
No I haven't

Hypnosphi:
Please resolve conflicts

Hypnosphi (approved):

danielduan (approved):

Hypnosphi:
Released as [`4.0.0-alpha.0`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.0)

PR #2861, opened 2018-01-29
Use UTC timezone in formatting too for knobs test

tmeasday:
Issue: https://github.com/storybooks/storybook/issues/2859

Previously I fixed the TZ of the date itself but not the timezone we are formatting relative to.

## What I did

Fixed both.

## How to test

Run storyshots tests


Hypnosphi:
Please apply locale and tz to vue-kitchen-sink example as well

Hypnosphi (approved):

tmeasday:
It looks like there is a date in the Angular stories (although it uses the built in date formatting), and the [polymer tests](https://github.com/storybooks/storybook/blob/1ada6692ea64143c4091d580c37701cf6e298535/examples/polymer-cli/src/stories/advanced.stories.js)

tmeasday:
LMK what you think we should do in the angular case @Hypnosphi

Hypnosphi:
Let's keep it as is until we hit any issues with angularshots

PR #2867, opened 2018-01-29
Added "key" prop to list items inside test failure message.

tniezg:
Issue:
React complains about missing "key" prop in lists displaying a failed Jest test.

## What I did
Added "key" prop equal the list element's index on the list.

## How to test
Create a failing Jest test, let jest output results to a JSON file and have that file load be loaded by Storybook's Jest addon.

Jest's output JSON file containing information about a failed test is required.

No

No

~If your answer is yes to any of these, please make sure to include it in your PR.~


addons/jest/src/components/Result.js
```diff
@@ -35,7 +35,7 @@ const StackTrace = glamorous(({ trace, className }) => (
       .join('')
       .trim()
       .split(/\n/)
-      .map(i => <div>{i.trim()}</div>)}
+      .map(traceLine, traceLineIndex => <div key={traceLineIndex}>{traceLine.trim()}</div>)}
```
Hypnosphi: Is there a chance we can use something other than index as a key? Otherwise, I think you can suppress the corresponding eslint rule for this line

tniezg: I'd like to use something else for the key prop but can't find any reasonable alternatives. I will add the ESLint rule.

Hypnosphi (approved):

Issue #2868, opened 2018-01-29
Browser history broken


chrisjaure:
### Issue details

Clicking back after selecting something in the sidebar does nothing. Expect it to go back to previous page.

### Steps to reproduce

1. Go to http://airbnb.io/react-dates/
2. Click another link in the sidebar
3. Click back

Notice you need to **click back twice** before it will go to the previous page, and also after you do that, the **forward button is disabled**.

### Please specify which version of Storybook and optionally any affected addons that you're running

- storybook/react 3.3.9
- storybook/react 3.4.0-alpha.6

### Affected platforms 

Noticed in Chrome 64 for Mac.

Hypnosphi:
Weird, #2261 was supposed to fix that. Will investigate

Hypnosphi:
Released as `3.3.12`

PR #2870, opened 2018-01-29
Fix browser navigation again

Hypnosphi:
Issue: #2868 

Looks like using `pushState` in iframe creates new history entries in parent window


tmeasday (approved):
Can confirm this works.

Good job.

Issue #2838, opened 2018-01-25
Storyshots doesn't seem to load Storybook config


denkristoffer:
### Issue details

I'm trying to set up Storyshots with Jest, but when the test runs it fails with `ENOENT: no such file or directory, scandir '/site/.storybook/components'`

However, my Storybook config specifies that my stories are placed together with my components with filenames like `Component.stories.js`:

```js
import React from 'react'
import { configure, addDecorator } from '@storybook/react'
import { withKnobs } from '@storybook/addon-knobs'
import { withInfo } from '@storybook/addon-info'

const req = require.context('components', true, /.stories.js$/)

addDecorator((story, context) => withInfo()(story)(context))
addDecorator(withKnobs)

function loadStories() {
  req.keys().forEach(filename => req(filename))
}

configure(loadStories, module)

```

Looks to me like the config file isn't being read or is set up wrong?

### Please specify which version of Storybook and optionally any affected addons that you're running

- storybook/react: 3.3.10
- storybook/addon-storyshots: 3.3.10

igor-dv:
Can you please share your storyshots test file / jest config? and maybe the console output from the test

denkristoffer:
I'm using Create React App, so most of the Jest setup is handled there. I do have a `setupTests.js` file with the following content:

```js
import Enzyme from 'enzyme'
import Adapter from 'enzyme-adapter-react-16'

Enzyme.configure({ adapter: new Adapter() })

// Mock matchMedia
global.window.matchMedia = jest.fn(() => ({
  matches: false,
  addListener: jest.fn(),
  removeListener: jest.fn(),
}))

jest.mock('scroll-to-element', () => 'scroll-to-element')
```

The test file is just
```js
import initStoryshots from '@storybook/addon-storyshots'

initStoryshots()
```

Console output is
```
Summary of all failing tests
 FAIL  src/storyshots.test.js
  ● Test suite failed to run

    ENOENT: no such file or directory, scandir '/site/.storybook/components'

      at Object.fs.readdirSync (fs.js:904:18)
      at requireModules (../../node_modules/@storybook/addon-storyshots/dist/require_context.js:32:28)
      at Function.newRequire.context (../../node_modules/@storybook/addon-storyshots/dist/require_context.js:90:5)
      at evalmachine.<anonymous>:27:19
      at runWithRequireContext (../../node_modules/@storybook/addon-storyshots/dist/require_context.js:102:3)
[...]

PR #2873, opened 2018-01-30
env.NODE_PATH support in Storyshots

igor-dv:
Issue: #2838

## What I did
I've added NODE_PATH support in the mocked "require.context"

## How to test
I have no idea how to add a working example into the monorepo... I've tested it with the repo from the issue.


tmeasday:
@igor-dv would it not make more sense to not call `path.resolve()` if the require path (e.g. `"components"` in the case of #2838) isn't relative?

While we are modifying this file, I would suggest we also fix the glaring bug that it calls `require` [immediately](https://github.com/storybooks/storybook/blob/2ca03e4f03c5b5b813ba4c25a85085ef5f50efa6/addons/storyshots/src/require_context.js#L72), rather than later when `req` is [called](https://github.com/storybooks/storybook/blob/2ca03e4f03c5b5b813ba4c25a85085ef5f50efa6/addons/storyshots/src/require_context.js#L75). This is completely inconsistent with what webpack does.

igor-dv:
> would it not make more sense to not call path.resolve() if the root require path (e.g. "components" in the case of #2838) isn't relative? So essentially doing what this code does in getFullPath?

Probably you are right... Will try to test it with the repo from the issue. 

> While we are modifying this file, I would suggest we also fix the glaring bug that it calls require immediately, rather than later when req is called. This is completely inconsistent with what webpack does.

Why is it a bug? it's a mock of what webpack does, so you can decide how to take care of it =)

tmeasday:
> Why is it a bug?

Because the require happens at a different time -- in this code (direct from the docs):

```js
// In storyshots, the require happens here:
const req = require.context('../src/components', true, /\.stories\.js$/)

function loadStories() {
  // in webpack it happens here:
  req.keys().forEach((filename) => req(filename))
}

```

This has caused bugs in the past, for instance if you add a decorator between the two..

tmeasday:
[We don't have to fix this issue in this PR]

igor-dv:
Is there any open bug on it?

tmeasday:
Only in my head for some reason I think..

igor-dv:
@tmeasday, just checked with the absolute path, and `path.resolve` didn't do anything - left the absolute path as is and everything worked. So I think it probably checks internally for the relativity.

tmeasday:
@igor-dv I created an issue here: https://github.com/storybooks/storybook/issues/2894. Note that the original issue (#2838) will also have this problem ;)

> @tmeasday, just checked with an absolute path, and path.resolve didn't do anything - left the absolute path as is and everything worked. So I think it probably checks internally for the relativity.

Yes, but not so much with "components", it treats this equivalently to "./components". i.e.:

```js
> path.resolve('/a', '/b')
'/b'
> path.resolve('/a', 'b')
'/a/b'
> path.resolve('/a', './b')
'/a/b'
```

igor-dv:
Aha ! Now I understand what you are talking about 😈 I'll do it

addons/storyshots/src/require_context.js
```diff
@@ -40,6 +41,14 @@ function isRelativeRequest(request) {
   );
 }
 
+function getFullPath(dirname, request) {
+  if (isRelativeRequest(request) || !process.env.NODE_PATH) {
```
tmeasday: I don't think it needs to be this complex; `process.env.NODE_PATH` will automatically apply if you require something non-relative, won't it?

So it can just be:

```
return isRelativeRequest(request) ? path.resolve(dirname, request) : dirname;
```

tmeasday (changes requested):

addons/storyshots/src/require_context.js
igor-dv: Checked it now, and it's not applied. It's just resolved according to the cwd. 🤷‍♂️

tmeasday: Oooh, because we need to a `fs.readdirSync` in the directory. Sorry I am bit slow 😊 

Do we not need to worry about other require rules in this case? For instance if I do `require.context('foo')` and I have `node_modules/foo` or `../node_modules/foo` or .. etc

I guess this is better than nothing though!

tmeasday (approved):

tmeasday:
that’s fair..

PR #2885, opened 2018-01-31
Self documented Storybook (A.K.A Addon Stories)

igor-dv:
So, I've been a bit annoyed about too many questions regarding documentation, like:
- "You need to improve docs",
- "How to do this or that"

One of the problems is that though we do have a lot of examples, they are not easily discoverable. So I thought what if we just show the code of the story as an addon. Then we can publish all the examples and link to them from docs. 🤔 Or even put them in the docs with Iframe.

## What I did
So, I've added an experimental addon-stories, which probably won't have any sense for external usages.:

1. ~~It works with a custom loader that should be configured with a custom webpack.config ( maybe there is a better way 🤷‍♂️ ). This loader injects decorator with the source of the file. For the POC purposes, I've used some naive "string.indexOf" implementation, and probably should be replaced with some AST processing.~~
2. ~~Currently, the addon panel just shows the source as a plain text.~~

TBD:
- [x] Use syntax highlighting
- [x] Use AST parsing to find stories and inject decorator
- [x] Add option to highlight specific "add()" because we have multiple "adds" in one story file
- [x] Extract kind and use kind#story as a key in highlight locations map (only when kind is "Literal")
- [x] Rename package to something that makes sense


shilman:
Awesome!!! Suggestion: addon-source? Addon-sourcecode?

Hypnosphi:
`addon-storysource`?

igor-dv:
"addon-storysource" makes me happy =)

igor-dv:
It looks much better with the syntax highlighting =)

image

Hypnosphi:
I have a crazy idea: what if we add CodeMirror?

igor-dv:
Oh it's indeed crazy. So CodeMirror will allow code editing, but what's next 🙀

ndelangen:
Needs syntax highlighting tbh

igor-dv:
image

ndelangen:
What's needed to get this merged?

igor-dv:
I think of a way we can publish every example to some static place and link it to the docs...

Hypnosphi:
> a way we can publish every example to some static place

What's wrong with netlify deploys?

igor-dv:
Can it be automated?

Hypnosphi:
What do you mean? We currently deploy each commit

igor-dv:
Is there any way to link netlify builds to docs?

Hypnosphi:
There are branch deploys with permalinks

For example, this is a link to deploy of this branch that gets updated authomatically with each commit

https://addon-stories--storybooks-official.netlify.com/

On documentation site, we should use links to `release/3.3` branch deploys (because `master` can contain usages of alpha features):
https://release-3-3--storybooks-official.netlify.com/

Hypnosphi (changes requested):
Please add a `README.md` for the addon

examples/angular-cli/.storybook/webpack.config.js
```diff
@@ -0,0 +1,23 @@
...
+  if (originalTsRule) {
+    originalTsRule.exclude = [/\.stories\.tsx?$/];
+  }
+
+  baseConfig.module.rules.push({
+    test: [
+      /\.stories\.tsx?$/,
+      /index\.ts$/
+    ],
+    loaders: [ ...originalTsRule.loaders, require.resolve('@storybook/addon-storysource/loader') ],
```
Hypnosphi: We can add this loader with `enforce: 'pre'` instead of patching the original rule:
https://webpack.js.org/configuration/module/#rule-enforce

Hypnosphi (approved):

igor-dv:
I think I will add Netlify to docs in a separate PR.

Hypnosphi:
Can we add scroll to the highlighted story (maybe in separate PR)?

igor-dv:
> Can we add scroll to the highlighted story (in separate PR)?

Sure

igor-dv:
Can we make it as a patch to 3.3 ?

Hypnosphi:
@shilman technically, it's a feature, but releasing it as patch could improve our documentation right now. WDYT?

shilman:
I think we should just get 3.4 out sooner rather than back-patching a large new feature

Issue #1088, opened 2017-05-21
Improve error reporting with `react-error-overlay`


ndelangen:
https://www.npmjs.com/package/react-error-overlay

https://twitter.com/dan_abramov/status/863482034865483776

MrBoolean:
Mine. PR soon.

ndelangen:
That would seriously be
![awesome](https://media.giphy.com/media/d2Z9QYzA2aidiWn6/giphy.gif)

ndelangen:
Hey @MrBoolean Did we miss your PR for this? or can someone else pick this up?

ndelangen:
@MrBoolean Are you still interested in adding this to storybook?

ijajmulani:
Hi, I would like to solve this. Could you guide me in this? I have cloned the repo.

thongdong7:
I think here is how it work:

In development webpack config, an entry to handle error is added before the main entry

https://github.com/facebookincubator/create-react-app/blob/206f9e84e62434e78d8ad7cf6f2c84476f611d12/packages/react-scripts/config/webpack.config.dev.js#L48-L58

The react-error-overlay handles the error when occurs

https://github.com/facebookincubator/create-react-app/blob/206f9e84e62434e78d8ad7cf6f2c84476f611d12/packages/react-dev-utils/webpackHotDevClient.js#L163

ijajmulani:
Thanks @thongdong7. I get a fair idea about will let you know once it done.

ndelangen:
@ijajmulani YES definitely I can help you, I can tell you about the codebase and help you find what you need.

Can you join our slack?
https://now-examples-slackin-nqnzoygycp.now.sh/
That way we can chat a bit more easily.

ijajmulani:
@ndelangen thanks for sending slack request. I have joined storybook on slack.

ritch:
Any update on this? Is this difficult to add manually?

Spy-Seth:
You could start to use it by creating a global decorator: https://gist.github.com/Spy-Seth/25bdaf2a6aa8c37228d47db79cfd0977

ndelangen:
@Spy-Seth Looks like you might be able to make a PR, adding this for everyone? 🙇

Spy-Seth:
Did you mean add the gist as an official decorator inside storybook mono-repo?

ndelangen:
Activate it by default seems good to me, do you agree?

Spy-Seth:
[...]

PR #2900, opened 2018-02-01
added react Error overlay on cli template by default

swyxio:
closes issue #1088

Issue:

## What I did
formalized the steps outlined in https://github.com/storybooks/storybook/issues/2890 adding it to the middleware and in the react CLI

now it looks like this when you throw an error in the react storybook woo

## How to test
just startup a new storybook project

Is this testable with jest or storyshots? not sure

Does this need a new example in the kitchen sink apps? not sure

Does this need an update to the documentation? probably not


Hypnosphi:
I'm not sure CLI template is the proper place for this. Maybe we should add this directly to [dev webpack config](https://github.com/storybooks/storybook/blob/master/app/react/src/server/config/webpack.config.js) instead as preview entries, so that it gets enabled for all users, not only the new ones?

sw-yx:
@Hypnosphi like this: https://github.com/storybooks/storybook/pull/2900/commits/9dbfd170a74f04a461abb2876632f2ff54f22568 ?

Hypnosphi:
I wonder if we can add a story that throws an error to `examples/storybook-official`, for demo purposes

Hypnosphi:
We also have some logic for displaying errors here (see `ErrorDisplay` usages): https://github.com/storybooks/storybook/blob/master/app/react/src/client/preview/render.js

We still need to find a way to cover all the cases. Just throw an error maybe, so that error overlay catches it?

app/react/src/server/config/webpack.config.js
```diff
@@ -20,6 +20,8 @@ export default function(configDir) {
       preview: [
         require.resolve('./polyfills'),
         require.resolve('./globals'),
+        require.resolve('react-error-overlay'),
```
Hypnosphi: We need to add an explicit dependency on that to `app/react/package.json`

Hypnosphi (changes requested):

Hypnosphi:
For some reason, the overlay doesn't hide when I fix an error

app/react/src/server/config/webpack.config.js
sw-yx: done

sw-yx:
> We still need to find a way to cover all the cases. Just throw an error maybe, so that error overlay catches it?

I am not at all sure what you mean. throw an error where? in `examples/official-storybook`? which story would it be at? would it just be a button that throws an error on click?

>For some reason, the overlay doesn't hide when I fix an error

I am using the overlay work and it does go away when I fix an error. its supposed to reload just like everything else in storybook. why doesn't it work for you?

Hypnosphi:
> I am not at all sure what you mean.

I mean, there's a number of situations when we show red screen, see https://github.com/storybooks/storybook/blob/master/app/react/src/client/preview/render.js
We need to remove red screen (that is, remove the code that renders it), and show error overlay instead. The easiest (and probably the only) way to trigger error overlay is to throw an error

> throw an error where? in examples/official-storybook? which story would it be at? would it just be a button that throws an error on click?

This would make sense as well. It can be a separate story in "Other" namespace. It can just throw inside render

Hypnosphi:
> I am using the overlay work and it does go away when I fix an error. its supposed to reload just like everything else in storybook. why doesn't it work for you?

IDK, I just tried to throw an error in some of the official-storybook components and then remove it, and overlay didn't go away

sw-yx:
Before I delete a bunch of code I wanted to check that I understand what you are trying to do because I think there are two different error cases here.

I added 2 stories to the Other section in the official-storybook example:
```js
storiesOf(`Other|${base}/Error Overlay Example`, module)
  .add('Storybook has a React errorOverlay when a component throws an error', () => <button onClick={() => {throw new Error('example error')}}>Test Error Overlay</button>)
  .add('Storybook has an error', () => {throw new Error('example error')})
```

The first one looks like this (after I click the button): 
err1

the second one looks like this (on render):

err2

I think both are useful and would prefer to keep both?

Hypnosphi:
Oh, so overlay doesn't catch arrors inside story functions? Ok, can we at least remove red box in cases handled by overlay? Now it looks weird when you see red box for a couple of secs and then overlay comes

examples/official-storybook/stories/other-dirname.stories.js
```diff
@@ -8,3 +8,17 @@ import BaseButton from '../components/BaseButton';
 storiesOf(`Other|${base}/Dirname Example`, module)
   .add('story 1', () => <BaseButton label="Story 1" />)
   .add('story 2', () => <BaseButton label="Story 2" />);
+
+storiesOf(`Other|${base}/Error Overlay Example`, module)
```
Hypnosphi: `${base}` not really needed here

sw-yx:
I would guess that it is slower because the ErrorOverlay is done through inserting an iframe (you can see this if you open up the console and watch the bottom) which takes some time to load.

I'll look into resolving the overlap cases but I'm not 100% confident how to do it as the code is a bit convoluted.

sw-yx:
Can you show me where you get the red box + overlay? I don't see the overlap case in my sample story in official-storybook.. see here:

erroroverlay

Hypnosphi:
Try this:

```
function ThrowingComponent() {
  throw new Error()
}

storiesOf(...).add(() => <ThrowingComponent/>)
```

sw-yx:
ok I can replicate it. tracking it down. I am jumping back and forth between `app/react/src/client/preview/render` and `examples/official-storybook` and every time I edit `render` I have to rebuild `yarn bootstrap` or I will not be able to see the changes in `official-storybook`. is there a better way to do this? it makes the iteration cycle very slow.

sw-yx:
this removes the error overlay overlap issues :) but there is still some delay before it shows up. I don't think I can do much about that. probably its just mounting the iframe on the dom.

sw-yx:
hmm. it has inconsistent behavior on refresh. I'm not sure how to fix it :( 

- it works as intended when I switch from one story to the `Storybook tries to render something that throws an error` story, but when I refresh directly to the url it shows the red screen
- it does not work as intended when I switch to `Storybook itself throws an error`, but when I refresh directly to the url it shows the red screen.

Hypnosphi:
> is there a better way to do this?

Yes, you can have `yarn dev` running in background

Hypnosphi:
> when I refresh directly to the url it shows the red screen

This can be related to saving errors in redux store

Hypnosphi:
BTW your latest commit contains only a change in example. Is it intended? https://github.com/storybooks/storybook/pull/2900/commits/8394ef28dddf4fbb5513f2842b102aa634ee6da2

sw-yx:
ah sorry the git hook prevented me from adding the file some how. have a look at this to see what I mean, I commented out the try/catch that is causing the double render but it has knockon effects I described above.

Hypnosphi:
I can confirm both of the problems you mentioned. Looks like just stopping rendering redbox on error isn't a solution

Hypnosphi:
And in `Storybook tries to render something that throws an error`, you can also reproduce the issue I mentioned before: if you replace `throw new Error()` with `return null` in `ThrowingComponent`, the overlay doesn't go away

sw-yx:
agreed. I suggest I undo it and just leave it. I can't figure out how else to reduce the duplicate overlay screen right now.

Hypnosphi:
Looks like `middleware.js` was moved to `core` library in #2937 

@igor-dv can we introduce some extension point there so that frameworks could add middleware specific to them?

igor-dv:
Yeah, should be easy. 

Something like this maybe?

In the `builde-dev.js`:
```js
export function buildDev({ packageJson, getBaseConfig, loadConfig, defaultFavIcon, middlewares }) {
   ...
   app.use(storybook(configDir, loadConfig, getBaseConfig, middlewares));
   ...
}
```

in the `middleware.js`

```js
export default function(configDir, loadConfig, getBaseConfig, middlewares) {
  ...
  middlewares.forEach(router.use)
  ...
}
```

I can do it probably later. Or @sw-yx, @Hypnosphi feel free to add it =)

sw-yx:
I tried to update my fork/my branch up to reflect all your moves but your git hook is giving me a ton of scary errors and not letting me merge the branches. I wrestled with it for a couple of hours and didn't get anywhere. pretty discouraged. I think I will just leave this for now as its too much of a mess for me to be helpful. anyway you have the basic idea from this PR.

sw-yx:
lol nice!

Hypnosphi:
Note: it's not actually merged into master, it's replaced by https://github.com/storybooks/storybook/pull/2941

ndelangen:
Thanks for all the hard work @sw-yx, really appreciated!

Issue #2897, opened 2018-02-01
[addon-info] fails with flow union types


GuyLivni:
### Issue details

Expected storybook to load correctly with addon info, instead it crushes with the following error:
```
Cannot read property 'length' of undefined
TypeError: Cannot read property 'length' of undefined
    at OneOfType (http://localhost:9009/static/preview.bundle.js:71199:31)
```

### Steps to reproduce

1. Start storybook with a story that uses info addon.
2. Make sure that you have flow-bin in project, and at least one of the variables typed as a union.

The error seems to come from addon-info/dist/components/types/OneOfType.js
It fails on:
https://github.com/storybooks/storybook/blob/master/addons/info/src/components/types/OneOfType.js#L7-L10

If I change `propType.value.length` to `propType.elements.length` and `propType.value.map` to `propType.elements.map` everything works as expected.

### Please specify which version of Storybook and optionally any affected addons that you're running

- storybook/react 3.3.11
- storybook/addon-info 3.3.11
- flow-bin 0.64.0

### Affected platforms 
The brake happened when I updated addon-info from 3.2.19 --> 3.3.11

### Screenshots / Screencast / Code Snippets (Optional)

```js
type Props = { value?: string | number  }
```
![screen shot 2018-02-01 at 12 18 22 pm](https://user-images.githubusercontent.com/6113581/35673427-3a978b44-074a-11e8-927a-52055161fca2.png)

Hypnosphi:
Probably related: https://github.com/storybooks/storybook/issues/2603

You're welcome to open a PR fixing that

GuyLivni:
Sure, thanks

Hypnosphi:
Released as [v3.3.13](https://github.com/storybooks/storybook/releases/tag/v3.3.13)

PR #2902, opened 2018-02-01
Check for correct props in Flow or prop-types

GuyLivni:
Issue: #2897 

## What I did
Added a `getPropTypes` function to check if the propType is coming from `Flow(propType.elements)` or `prop-types(propType.value)`, and used it in all type components.
Added `literal` type to display flow enum values correctly in the table.
Added more complex types in FlowTypeButton.js to test the changes.

## How to test

Run storybook kitchen sink app (yarn storybook), no "cannot read property length of undefined" error displayed in [Comments from Flow declarations](https://github.com/storybooks/storybook/blob/master/examples/official-storybook/stories/addon-info.stories.js#L18-L23) story if you define union prop type in FlowTypeButton.js.


Hypnosphi:
Most of the other type components use `value` field. Should we apply this change there as well?

GuyLivni:
Missed that, makes sense to apply it there as well, ill try to create a more centralized solution.
Btw what is the `bitHound - Dependencies` that keeps failing? am I missing something? or should I pull this solution into a different branch than master?

Hypnosphi:
You shouldn't wory about it, it's an optional check

GuyLivni:
Awesome, what do you think about my suggested solution? did I miss anything?

examples/official-storybook/components/FlowTypeButton.js
```diff
@@ -2,11 +2,11 @@
 import React from 'react';
 
 type PropsType = {
-  /** The text to be rendered in the button */
-  label: string,
+  /** A multi-type prop to be rendered in the button */
+  label: string | number,
```
Hypnosphi: Can you please add here as much different types as possible? At the very least, we should repeat all the types present in DocgenButton, to ensure that they work with flow. But there's probably a sense in trying to add some more complex types as well (because when you use flow, usually there are more complex types around)

GuyLivni:
I added more types, but since the last merge new test fails `deploy/netlify — Deploy preview failed`

Hypnosphi:
[Live example](https://5a77801a92226a3dc564bc2b--storybooks-official.netlify.com/?selectedKind=Addons%7CInfo.React%20Docgen&selectedStory=Comments%20from%20Flow%20declarations&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

For some reason, it displays `literal | literal` instead of actual values

Hypnosphi:
Plus, `PrettyPropType` doesn't seem to prettify flow shapes, but it's probably a separate issue
screen shot 2018-02-05 at 01 14 29

GuyLivni:
I added `literal` type for flow enums, it displays the enum values correctly now in the table

Hypnosphi (approved):

PR #2922, opened 2018-02-06
Fix "dev" script to be cross-platform

igor-dv:
Issue: After my changes in "ts distribution", the `dev` script works only in windows

## What I did
I've used [cross-env-shell](https://github.com/kentcdodds/cross-env#cross-env-vs-cross-env-shell) to fix this.


Hypnosphi:
Thanks, will test

package.json
```diff
@@ -16,8 +16,8 @@
     "coverage": "codecov",
     "danger": "danger",
     "dev": "concurrently --kill-others \"yarn dev:ts\" \"yarn dev:js\"",
-    "dev:js": "lerna exec --parallel -- node %LERNA_ROOT_PATH%/scripts/watch-js.js",
-    "dev:ts": "lerna exec --parallel -- node %LERNA_ROOT_PATH%/scripts/watch-ts.js",
+    "dev:js": "lerna exec --parallel -- cross-env-shell node $LERNA_ROOT_PATH/scripts/watch-js.js",
```
Hypnosphi: It fails unless I replace `$` with `\\$`

Hypnosphi (approved):

PR #2926, opened 2018-02-06
Update docs on LinkTo in addon-links

simon360:
The existing docs didn't provide correct examples of how to use LinkTo - it should be the default import from `@storybook/addon-links/react`, but was documented as a named import from `@storybook/addon-links`.

Issue: N/A

## What I did

Updated the example code in `adoon-links`, because the reference to `LinkTo` wouldn't work when used.

## How to test

Is this testable with jest or storyshots? *No*

Does this need a new example in the kitchen sink apps? *No*

Does this need an update to the documentation? *It's purely a documentation update*


Hypnosphi:
Thanks @simon360, I totally forgot to update the docs

Hypnosphi (approved):

PR #2937, opened 2018-02-07
Move builds to core

igor-dv:
Issue: #2636

## What I did

1. Moved index.js to core
2. Moved build.js to core
3. Fixed polymer leftovers - (UglifyJsPlugin and getConfigDir)
4. Moved WatchMissingNodeModulesPlugin to core


ndelangen (approved):

Hypnosphi:
`WatchMissingNodeModulesPlugin` is actually a part of `react-dev-utils` package which we already depend on, in all apps (I guess, `react` is in the package name only because it's a part of `create-react-app` project. Let's import it from there instead of replicating on our side

igor-dv:
Makes sense to me

Hypnosphi:
Ok, I'll make a PR

Hypnosphi:
https://github.com/storybooks/storybook/pull/3141

Issue #2943, opened 2018-02-08
3.4.0-alpha.7 on react-native: require.resolve is not a function


macrozone:
using 3.4.0-alpha.7 on react-native leads to this error on the app:

```
require.resolve is not a function. (In 'require.resolve('./dist/client/manager')', 'require.resolve' is undefined)

<unknown>
    client.js:5:31
loadModuleImplementation
    require.js:213:12
<unknown>
    index.js:39:14
loadModuleImplementation
    require.js:213:12
<unknown>
    index.js:16:15
loadModuleImplementation
    require.js:213:12
<unknown>
    storybook.js:5
loadModuleImplementation
    require.js:213:12
<unknown>
    index.js:1
loadModuleImplementation
    require.js:213:12
guardedLoadModule
    require.js:140:45
global code
    <unknown file>:0
```

Hypnosphi:
Related: https://github.com/storybooks/storybook/pull/2788 #2796

igor-dv:
Yeah... probably should be moved to `core/server.js`

Hypnosphi:
Released as [v3.4.0-alpha.8](https://github.com/storybooks/storybook/releases/tag/v3.4.0-alpha.8)

PR #2947, opened 2018-02-08
Fix React-Native by moving managerPath export to `server.js`

Hypnosphi:
Issue: #2943
This is only present on alpha releases

## What I did
Moved managerPath export to `server.js`

## How to test
Launch CRNA example


igor-dv (approved):
LGTM

PR #2958, opened 2018-02-11
Add built storybooks to official storybook

tmeasday:
Issue:

Screenshot tests for example apps were difficult to work with.

## What I did

Added a iframe story to the official storybook for each example app. This results in these tests getting picked up by Chromatic.

## How to test

Check these screenshots are correct: https://www.chromaticqa.com/build?appId=5a375b97f4b14f0020b0cda3&number=984 -- actually I am not sure what is going on with polymer there, perhaps I shouldn't have added it too?

I'm not completely on top of what the previous tests were used for (and I haven't actually removed them in this PR, maybe I should), but I wanted to try this out and run it by you to see if you think it's useful.


examples/official-storybook/built-storybooks/angular-cli
```diff
@@ -0,0 +1 @@
+../../angular-cli/storybook-static
```
Hypnosphi: How does it work?

tmeasday: I just symlinked the built projects into the static directory of this storybook instead of needing to run a HTTP server for those files or importing them with a `file://` (which I guess is conceptually what it does right now).

tmeasday:
So @ndelangen @Hypnosphi -- do we want to merge this PR? If it is not useful (I'm not sure I totally understand) we can also just close it, I was just feeling it out.

PS although the `polymer-cli` snapshot was denied by @ndelangen it is actually what the output of `build-storybook` in the `polymer-cli` looks like so... ;)

Hypnosphi:
Let's add this as failing test. Someone should fix polymer in another PR then

Hypnosphi (approved):

tmeasday:
Should we remove the existing screenshot tests?

igor-dv:
If you are talking about integration tests, I think we should

Hypnosphi:
Can we add a delay before taking screenshot?

tmeasday:
@Hypnosphi we don't currently have that option in Chromatic (although there are legit use-cases for it, it is often used as a bandaid, and we like to properly hear about problems like this one). In this case I think we can solve it properly by waiting for any `iframe`s rendered by a story to load properly before taking a screenshot. I'm working on a fix; shouldn't be too far off.

igor-dv:
Maybe you are already doing this, so apologize me if I am just speaking obviously =) There is a technique that is used by Webpagetest to calculate a speed index, that compares screenshots during the loading until there is no difference... And no need for a timeout.

Hypnosphi:
@tmeasday the problem is that stories are loaded asynchronously. So the empty storybook interface that you see on the screenshots is exactly how the iframe looks like when fully loaded

tmeasday:
> the problem is that stories are loaded asynchronously. 

Yes, to be precise the problem is that although we wait for all network requests to complete (with a small timeout) before taking a screenshot, if an iframe loads a large JS file (like the SB manager bundle) it can take a while to parse and execute that JS file (which then renders and kicks off further network requests in this case).

It isn't probably possible to determine if "some JS is in progress that may lead to rendering", handling the special case of "story renders an iframe, we should wait for that iframe's JS to load" is something we can do quite easily.

Hypnosphi:
> like the SB manager bundle

Did you mean the preview bundle? Manager gets rendered OK

tmeasday:
It does but that's because there is some timeout between us stopping waiting and the actual screenshot happening... we are getting far into the weeds now ;)

PR #2972, opened 2018-02-12
Added beforeScreenshot config option to addons/storyshots, to allow testing of components with mounting animations

seberik:
Issue:
There is no way to reliably test components with mounting animations with the storyshots-visual regression option.

## What I did
Added a function `beforeScreenshot` in addition to `getMatchOptions` which will be called after page.goto(url) but before page.screenshot(). This allows for visual regression testing of components with mount animations and additionally simulating click, hover events etc.

## How to test

**Is this testable with jest or storyshots?**
I can provide jest test if required but there are currently no tests for the other options which I'm aware of.

**Does this need a new example in the kitchen sink apps?**
Not sure.

**Does this need an update to the documentation?**
I have updated documentation.


igor-dv (approved):
LGTM

PR #3002, opened 2018-02-16
RN: Add error handler on WS to fix crashing on page reload

Hypnosphi:
Issue: https://github.com/storybooks/storybook/issues/2923

## What I did
Added error handling according to https://github.com/websockets/ws/issues/1256#issuecomment-364988689

## How to test
Launch CRNA example, refresh desktop browser tab (it should be Chrome)


lib/channel-websocket/src/index.js
```diff
@@ -52,6 +52,8 @@ export class WebsocketTransport {
       this._handler(event);
     };
     this._socket.onerror = e => {
+      // Ignore network errors like `ECONNRESET`, `EPIPE`, etc.
+      if (e.errno) return;
```
lpinca: This will also catch connection errors like `ECONNREFUSED`, not sure if wanted.

Hypnosphi: What would be the proper filter to only catch crome tab reload?

lpinca: Is `WebSocket` here the browser native `WebSocket` object or `ws` or it depends on the context?

lpinca: I think this is the browser `WebSocket` as per https://github.com/storybooks/storybook/pull/3002/files#diff-432fa23ac718b94ed5be1b2802371d41R3. In this case you don't need this, the fix on the server is sufficient and correct.

danielduan:
saw you denied the two screenshot changes. is that because of the async nature of screenshots or is this PR not ready to go yet?

danielduan (approved):

danielduan:
@tmeasday does the screenshots happen when `document.ready` or is there some sort of timer that gets triggered? I think the screenshots here look like they're taken too early.

There was one I just fixed in my other PR where the early blank screenshot was set as the correct test set. I don't think it's a huge issue but maybe your other customers might mind if the timing is off.

tmeasday:
@danielduan in this case it is due to the stories rendering an iframe, which we then need to wait to load. You can read more detail about what's happening here: https://github.com/storybooks/storybook/pull/2958

nikhedonia:
could you someone publish 3.3.14 to npm? - many thanks

Hypnosphi:
It's actually published:
https://www.npmjs.com/package/@storybook/react-native

> 3.3.14 is the latest of 73 releases

PR #3010, opened 2018-02-17
enable writing api documentation

walkerrandolphsmith:
Issue:

https://github.com/storybooks/storybook/issues/591#issuecomment-361617546

## What I did

I wanted to change what route storybook runs on and attempted to accomplish this with the following "full control" configuration.

```const path = require('path');
const genDefaultConfig = require('@storybook/react/dist/server/config/defaults/webpack.config.js');

module.exports = (baseConfig, env) => {
  const storybookBaseConfig = genDefaultConfig(baseConfig, env);

  storybookBaseConfig.entry.preview = storybookBaseConfig.entry.preview.map(function(entry) {
    if (entry.includes("webpack-hot-middleware")) {
      return `${require.resolve('webpack-hot-middleware/client')}?path=/book/__webpack_hmr&reload=true`;
    }

    return entry;
  });

  storybookBaseConfig.output.publicPath = '/book/';

  return storybookBaseConfig;
};
```

As mentioned in the issue comment I linked, the webpack hot middleware was not receiving the configuration it needed to enable a custom route.

I constructed a minimal configuration for webpack hot middleware including the `path`.

## How to test

Use a "full control" config as described and the default config and ensure HMR works for both storybooks

Not sure

I don't think so its not a common case.

Possibly not since this is an edge case.


Hypnosphi:
Thanks, will test

Hypnosphi:
Actually, modifying `entry` and `output` fields doesn't sound like a good idea to be (actually docs [advise](https://storybook.js.org/configurations/custom-webpack-config/#full-control-mode) not to do that)

I wonder can there be another way to achieve what you need

Hypnosphi:
I think we can make `publicPath` explicitly configurable with a CLI option

walkerrandolphsmith:
I will investigate other solutions! Thanks @Hypnosphi

walkerrandolphsmith:
That suggestion makes sense let me try that path!

Hypnosphi:
By the way, we use `publicPath: ''` for static builds to allow deploying into subpaths. Maybe this can work for devserver as well?

Hypnosphi:
See https://github.com/storybooks/storybook/blob/master/app/react/src/server/config/webpack.config.prod.js#L24-L29

walkerrandolphsmith:
Hey I am in the middle of moving so a bit busy. I see this was added to v4 milestone! Super excited to make it happen.

PR #3019, opened 2018-02-19
Docs live examples

igor-dv:
Issue: Add live examples to docs

## What I did

image

- [x] Added examples to readme
- [x] Added examples to docs


igor-dv:
Have troubles to run docs on windows. Will investigate it later.

README.md
```diff
@@ -91,6 +92,21 @@ For additional help, join us [in our Slack](https://now-examples-slackin-rrirkqo
 
 See [Addon / Framework Support Table](ADDONS_SUPPORT.md)
 
+## Live Examples
+
+### 3.4.alpha
+> Note, this is an Alpha version. Some of the features still might not be released
+
+ - [React Official](https://master--storybooks-official.netlify.com)
```
Hypnosphi: Just `https://storybooks-official.netlify.com`

I'm also not sure about the naming, would like to hear from @tmeasday and @ndelangen

igor-dv: Yeah. I feel some kind of conflict between showing the official storybook UI and showing the examples...

igor-dv:
image

Hypnosphi (approved):

igor-dv:
If nobody rejects the wording, feel free to merge 😉

README.md
tmeasday: Yeah I wouldn't call it that. Don't we have a "React" example too? (`cra-kitchen-sink`).

`official-storybook` is like Storybook's storybook. Not quite sure how best to put that ;) Maybe it doesn't actually need to be on this page?

igor-dv:
@tmeasday, 

> Yeah I wouldn't call it that. Don't we have a "React" example too? (cra-kitchen-sink).

AFAIK, there is almost nothing valuable to user in `cra-kitchen-sink` (documentation-wise), all the examples are located now in the official.

> official-storybook is like Storybook's storybook. Not quite sure how best to put that ;) Maybe it doesn't actually need to be on this page?

I would say, if we want to have a Storybook's storybook, we should put there only the stories of the actual UI. And separate all the other stuff to a different place. I am not sure regarding the naming though. (Official, Storybook's Storybook, StoryStoryBook 😬).

Hypnosphi:
The reason why it's in the same place is that it's much more convenient for development

PR #3023, opened 2018-02-19
Added safety net preventing storybook Jest addon from throwing.

tniezg:
Added safety net preventing storybook Jest addon from throwing on unrecognized test results and forcing those results to be returned no matter their format.

Issue:
Storybook's Jest addon throws an Error when one of failed tests is a Jest snapshot and is not fully recognized by it.

## What I did
Used [enzyme-to-json](https://www.npmjs.com/package/enzyme-to-json) to make better snapshots of React components in Jest when using Enzyme to render them. Failed tests would throw an exception in Storybook's Jest addon, because the addon could not fully recognize the format of the output.

## How to test
1. Use a custom Jest serializer for snapshots, for example enzyme-to-json, and output a failed test by running Jest.
2. Load Jest's output in Storybook's Jest addon.
3. You will see an Error in the web console stating that [`reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) cannot receive an object. This is due to one of reducers inside Storybook's Jest addon being unable to properly parse Jest's output and returning an object instead of an array for the next reducer.

Jest

No

No

~~If your answer is yes to any of these, please make sure to include it in your PR.~~


tniezg:
This is not supposed to be a perfect solution to the issue but a fallback for when everything else fails. Ideally, the addon would recognize failed snapshot tests and format them correctly.

Hypnosphi (approved):

PR #3036, opened 2018-02-20
Use client api in react native

tmeasday:
Issue: RN used the `StoryStore` from `@storybook/core` but not the full `ClientApi`, making things like https://github.com/storybooks/storybook/pull/2679 difficult

## What I did

Refactored a little to use `ClientApi`

## How to test

Test with built-in examples (seems to work). Also try any other apps you can.

## Notes

I didn't refactor the `ConfigApi` out as:
a) it's a pretty small surface area.
b) RN works differently as it doesn't have a redux store or use the same kind of channel logic.

I think we'll be fine leaving that as is.


lib/core/src/client/preview/client_api.js
```diff
@@ -11,14 +11,7 @@ const defaultDecorateStory = (getStory, decorators) =>
...
 
 export default class ClientApi {
-  constructor({
-    channel,
-    storyStore = new StoryStore(),
-    decorateStory = defaultDecorateStory,
-  } = {}) {
-    // channel can be null when running in node
-    // always check whether channel is available
-    this._channel = channel;
```
danielduan: channel here is only used by RN?

danielduan (approved):
makes sense to me.

lib/core/src/client/preview/client_api.js
tmeasday: This code (`client_api`) wasn't previously used by RN.

`_channel` is always there in existing code that uses the client API, but not used by the client API, so given the way I was going to initialize this code in the RN context (without passing the channel as it doesn't exist yet), I figured we get rid of if from this code, so no-one would try and use it in the future.

PR #3040, opened 2018-02-20
Addon Storysource improvements

igor-dv:
## What I did
- Use acorn to calculate locations instead of "line-column" pkg.
- Remove ugly comments from the storysource (currently only eslint things)
- Make "prettier" and "uglyComments" configurable in loader


addons/storysource/src/loader/generate-helpers.js
```diff
@@ -0,0 +1,106 @@
...
+
+    parts.push(start, end);
+    lastIndex = comment.end;
+  });
+
+  return parts.join('');
+}
+
+function prettifyCode(source) {
+  return prettier.format(source, {
```
Hypnosphi: Can we make it optional and configurable?

igor-dv: Only the prettier or the whole comments removal? Also. why?

Hypnosphi: The prettier. Users may have their own preferences about how their code should be formatted

addons/storysource/src/loader/generate-helpers.js
```diff
@@ -0,0 +1,106 @@
...
+    stage3: true,
+    es7: true,
+  },
+};
+
+function generateSourceWithoutUglyComments(source, comments) {
+  let lastIndex = 0;
+  const parts = [source];
+
+  const isUgly = comment => comment.value.trim().indexOf('eslint-') === 0;
```
Hypnosphi: also "global", "flow", "$FlowFixMe", probably others. Let's make it an array of regexp patterns, probably also configurable from outside

addons/storysource/src/loader/generate-helpers.js
igor-dv: Actually, I thought we won't expose it to users 😂 But I can make it configurable.

Hypnosphi: > I thought we won't expose it to users

Why? I would like to use it in my project (and we don't use semicolons =)

igor-dv: 👌

addons/storysource/src/loader/generate-helpers.js
igor-dv: I've added "global" and "eslint" as defaults. Do you think we need to support flow as well? I'm not familiar with its comments syntax.

Hypnosphi: No, it's good defaults

addons/storysource/src/loader/generate-helpers.js
```diff
@@ -20,33 +20,31 @@ const acornConfig = {
   },
 };
 
-function generateSourceWithoutUglyComments(source, comments) {
+function isUglyComment(comment, uglyCommentsRegex) {
+  return uglyCommentsRegex.find(regex => regex.test(comment));
```
Hypnosphi: `.some` not `.find`

addons/storysource/src/loader/generate-helpers.js
```diff
@@ -0,0 +1,104 @@
...
+  locations: true,
+  plugins: {
+    jsx: true,
+    stage3: true,
+    es7: true,
+  },
+};
+
+function isUglyComment(comment, uglyCommentsRegex) {
+  return uglyCommentsRegex.find(regex => regex.test(comment));
```
Hypnosphi: `.some` not `.find`

Hypnosphi (changes requested):
Let's fix array method and update README — and we're good to go

Hypnosphi (approved):

Hypnosphi:
```
#!/bin/bash -eo pipefail
yarn lint
yarn run v1.3.2
$ yarn lint:js . && yarn lint:ts **/*.ts && yarn lint:md .
$ cross-env NODE_ENV=production eslint --cache --cache-location=.cache/eslint --ext .js,.jsx,.json --report-unused-disable-directives .
$ tslint -p . -c tslint.json -t stylish '**/*.ts'
$ remark -q .
addons/storysource/README.md
  1:1  warning  Missing newline character at end of file  final-newline  remark-lint

⚠ 1 warning
Done in 41.52s.
```

Please fix remark warning

PR #3046, opened 2018-02-21
Revisit addon/framework support

igor-dv:
## What I did

Linked every addon to the ADDONS_SUPPORT.md

### DIFF
--- a/addons/a11y/README.md
+++ b/addons/a11y/README.md
@@ -2,7 +2,7 @@
 
 This storybook addon can be helpfull to make your UI components more accessibile.
 
-[Framework Support](../../ADDONS_SUPPORT.md)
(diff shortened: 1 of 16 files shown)


addons/a11y/README.md
```diff
@@ -2,6 +2,8 @@
 
 This storybook addon can be helpfull to make your UI components more accessibile.
 
+[Framework Support](../../ADDONS_SUPPORT.md)
```
Hypnosphi: Please make those links absolute, so that they would work on npm pages

Hypnosphi (changes requested):

addons/a11y/README.md
igor-dv: Like this: https://github.com/storybooks/storybook/blob/master/ADDONS_SUPPORT.md

Hypnosphi (approved):

PR #3058, opened 2018-02-23
Add `__STORYBOOK_CLIENT_API__` for external tools

tmeasday:
So tools like Chromatic and others that interact with the client API can work with storybooks that haven't injected any code.

I would love to refactor this code from preview into core but I will leave to a later PR.

## How to test

Check that `__STORYBOOK_CLIENT_API__` is defined in each example app.


joscha:
cc @timhaines

tmeasday:
Yes, interested to know if this helps Percy and if there is more that would be needed at your end @timhaines

Hypnosphi (approved):

ndelangen (approved):

app/angular/src/client/preview/index.js
```diff
@@ -38,10 +40,10 @@ if (isBrowser) {
 
   // Handle keyboard shortcuts
   window.onkeydown = handleKeyboardShortcuts(channel);
-}
 
-const clientApi = new ClientApi(context);
-export const { storiesOf, setAddon, addDecorator, clearDecorators, getStorybook } = clientApi;
+  // Provide access to external scripts
+  window.__STORYBOOK_CLIENT_API__ = clientApi;
```
Hypnosphi: @tmeasday Is there any particular reason to do that only inside `isBrowser` branch?

tmeasday: I guess `window` is not defined in the other context (i.e. jest). Are you thinking we could add it to `global.__STORYBOOK_CLIENT_API__`?

Hypnosphi: I would like to use it from jsdom, where `window` is actually defined and is not the same as `global`. My usecase is to get stories tree from node.js

tmeasday: This works in JSDOM. Have you tested it? Storybook doesn't know the difference.

tmeasday: I literally tested this yesterday, that's why I feel quite confident about this ;)

Hypnosphi: Did you change default user agent?
https://github.com/storybooks/storybook/blob/c8307fac5bcf61e4f72869c81faa4a3ae05fd56f/app/angular/src/client/preview/index.js#L22

tmeasday:
Would it perhaps also be useful to expose the channel, perhaps on `__STORYBOOK_CHANNEL__`? That way scripts could also fire events to trigger things like stories rendering.

app/angular/src/client/preview/index.js
tmeasday: I guess I did 😊 

When was that line added? That doesn't seem helpful to me.

tmeasday: JSDOM is trying to pretend to be a browser, why would we want to treat it differently?

Hypnosphi: Probably because it lacks api like postmessage that is used in channels

Cc @ndelangen 
https://github.com/storybooks/storybook/pull/2542#discussion_r158586413

tmeasday: Hmm, fair enough.

Well I think it is fine for this particular check just be something `if (typeof window !== 'undefined')` but I suspect there will be other use cases for JSDOM where having `isBrowser === true` might be helpful. If it breaks however I guess we leave it.

tmeasday: Also any thoughts on https://github.com/storybooks/storybook/pull/3058#issuecomment-374431611

Hypnosphi:
Feel free to add as long as you're going to use it =)

tmeasday:
@Hypnosphi => https://github.com/storybooks/storybook/pull/3243

PR #2275, opened 2017-11-09
Refactor addon-action to use glamorous

armandabric:
Hey storybookers,

This PR is related to https://github.com/storybooks/storybook/issues/2263

This commit convert the Action addon to glamourous and extract the `button` as a shared component. 

I don't known how you want to process: 
- Convert one addons at the time to glamourous and after extract what should be shared
- Convert one addons at the time to glamourous and extract what should be shared at the same time
- Convert all addons in the same move and extract what should be shared at the same time

This is my first PR in storybook so I'm probably forgetting somethings ;)

PR #3068, opened 2018-02-24
Addons: avoid mixing manager and preview code together

Hypnosphi:
Issue: #2275 accidentally broke React Native. The problem is that glamorous contains browser-specific code, and this line makes it leak into RN: https://github.com/storybooks/storybook/blob/master/addons/actions/src/index.js#L6

## What I did
Ensured that manager and preview code is separated in all existing addons

### DIFF
--- a/addons/actions/register.js
+++ b/addons/actions/register.js
@@ -1,1 +1,1 @@
-require('./dist').register();
+require('./dist/manager').register();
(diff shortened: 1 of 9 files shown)


igor-dv (approved):
LGTM. 
I hope people don't use "register" from the index file 🤷‍♂️

PR #3070, opened 2018-02-24
Fix links addon examples

Hypnosphi:
Issue: Examples for links addon are broken


igor-dv (approved):
I think other examples are broken too

Hypnosphi:
> I think other examples are broken too

How? Why?

igor-dv:
After I've separated all the examples in the other apps, I didn't check if links are working 😓 So probably they are not.

Hypnosphi:
Can you check it then?

igor-dv:
Yep, tomorrow though.

PR #3073, opened 2018-02-24
Don't use exact versions in peerDependencies

Hypnosphi:
Issue: dependencies.io broke our `peerDependencies` with commits like this, setting them to exact version. This makes those peerDeps quite useless
https://github.com/storybooks/storybook/commit/3bfbc7a7011416c6f374e66103cd48f5f38294b0#diff-4c0bffca46303f18a36ec22ddc6e5755R85

Related issue for dependencies.io: https://github.com/dependencies-io/actor-js-lerna/issues/39


danielduan (approved):

app/react-native/package.json
```diff
@@ -72,6 +72,6 @@
     "babel-core": "^6.26.0 || ^7.0.0-0",
     "babel-runtime": ">=6.0.0",
     "react": "*",
-    "react-native": "0.52.2"
+    "react-native": ">=0.51.0"
```
danielduan: rn is actually pretty tricky to upgrade. it needs matched versions of all the tooling as well for everything to bundle, compile, and work. I still think this is the right move though.

danielduan (approved):

PR #3082, opened 2018-02-25
 Cleanup dependencies

Hypnosphi:
## What I did
— Removed unused dependencies
— Added missing ones
— Removed React-specific stuff from other frameworks apps
— Added `babel-runtime` in CLI templates. We use `transform-runtime` on users' code, that means that they need to have `babel-runtime` package on their side

https://depcheck--storybooks.netlify.com/
https://depcheck--storybooks-official.netlify.com/
https://depcheck--storybooks-angular.netlify.com/
https://depcheck--storybooks-cra.netlify.com/
https://depcheck--storybooks-polymer.netlify.com/
https://depcheck--storybooks-vue.netlify.com/


app/vue/src/server/babel_config.test.js
```diff
@@ -1,110 +0,0 @@
-import loadBabelConfig from './babel_config';
```
igor-dv: Can you explain please the babel changes? Like why don't we need these tests?

examples/cra-kitchen-sink/package.json
```diff
@@ -31,16 +29,14 @@
...
     "@storybook/client-logger": "^3.4.0-alpha.9",
-    "@storybook/components": "^3.4.0-alpha.9",
     "@storybook/react": "^3.4.0-alpha.9",
-    "babel-jest": "^22.4.1",
+    "babel-runtime": "^6.26.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.0",
     "enzyme-to-json": "^3.3.1",
-    "jest": "^22.4.2",
+    "jest": "^20.0.4",
```
igor-dv: Downgrade?

igor-dv (approved):

app/vue/src/server/babel_config.test.js
Hypnosphi: The only thing it tested was the addition of `babel-plugin-react-docgen` which is useless in non-react codebase

examples/cra-kitchen-sink/package.json
Hypnosphi: See https://github.com/facebook/create-react-app/issues/4021#issuecomment-365790095

PR #3102, opened 2018-02-26
fix screenshots tests & add getScreenshotOption to storyshots

thomasbertet:
Issue: none

## What I did
- Image snapshot were not run since [this](https://github.com/storybooks/storybook/commit/85f1f7f64a1af4fd82296f1a3354528720a97d3e#diff-9e6ac008d54aaf3451af361767fb6b82R23) edit.  
- Added a default option to take a screenshot of the full page. It seems reasonnable default to me at least, so if it doesn't for you, let's remove that :question: 
- Added a new method `getScreenshotOptions` to be able to configure _puppeteer_ `screenshot()` behavior.
- Update package.json to be able to run Jest from the root node_modules, since [this commit](https://github.com/storybooks/storybook/commit/def1e04dbb123f3b800f9c6eac18b4dc867d1bb5#diff-93b9bda8380b799931da7a6504b4d8d1L38) removes the Jest deps from official-storybook.  

Also, I wanted to draw your attention that if we specify a threshold higher than 0 we might end-up with some bad visual regression. More info below.
 
## How to test
- At root:  run `yarn run test --image`. 
- At examples/official-storybook:  `yarn run image-snapshots`

- already did.

- nope

- Yep, done.

## Threshold for jest-image-snapshot.
- I ran into the following issue: 

storyshots-image-runner-js-image-snapshots-addons-a-11-y-default-1-diff
This is from the addon-a11-default story.

This diff is ~2.5 % as for jest-image-snapshot. It looks bad to me, so I believe we will not want to have this kind of changes to pass, are we ? Just a question though, if you are fine with it, let it be and I will re-add the threshold to 4%.


addons/storyshots/src/test-body-image-snapshot.js
```diff
@@ -45,14 +55,13 @@ export const imageSnapshot = ({
...
-        })
-      );
+      .then(() => page.screenshot(getScreenshotOptions({ context, url })))
+      .then(image => {
+        expect(image).toMatchImageSnapshot(getMatchOptions({ context, url }));
+      });
   };
 
-  testFn.beforeEach = () =>
+  testFn.beforeAll = () =>
```
thomasbertet: Moving this to `beforeAll` is a perf booster. No need to re-run Chrome for each tests.

Hypnosphi:
> Added a default option to take a screenshot of the full page

Is it a breaking change?

Hypnosphi:
Looks like there are a lot more failures than just one addon-a11y story:
https://circleci.com/gh/storybooks/storybook/40645

BTW, for some reason diffs are not uploaded as artifacts
https://circleci.com/gh/storybooks/storybook/40645#artifacts/containers/0

thomasbertet:
yay @Hypnosphi I believe I broke everything :D The diff were not even generated in the first place :-1: 
I will check to understand why. You have to trust my word for now ahah :)

thomasbertet:
Oh and yes this is breaking. Since the 3.4 is not out yet (still in alpha if I'm correct) I was thinking maybe that's not too bad to throw it now. 
I feel now that it may be not necessary to generate a full page each time ... don't know what you think about this ?

Hypnosphi:
Ok I just forgot it's an alpha feature

thomasbertet:
I re-added the workaround for puppeteer as I suspect this is linked to Chrome headless not willing to run properly. 

It was previously removed in https://github.com/storybooks/storybook/pull/3052, and not noticed it was used by the official-storybook storyshot tests because tests were not run anymore thus nothing broke !

Hypnosphi:
Still lots of errors =(
But now one can see that it's because of different fonts on different OS. That's why I raised the threshold:
https://40938-54173593-gh.circle-artifacts.com/0/official_storybook_image_snapshots/__diff_output__/storyshots-image-runner-js-image-snapshots-addons-a-11-y-disabled-1-diff.png

thomasbertet:
Yep, I see that too.. I was actually trying to determine a good threshold ... not an easy thing to do...
4% is high due to having large image with a lot of white pixels... 
1% might be enough, don't you think ?

Hypnosphi:
> 1% might be enough, don't you think ?

Sure, if you can make current tests pass with it in a way that can be reproduce d by other contributors

thomasbertet:
Yep I get it, can you point me to contributors willing to help test these screenshots for Mac & Windows ? I've only got Linux.

Hypnosphi:
I use Mac, @igor-dv uses Windows

thomasbertet:
Allright, thanks. Could you test this then ? As well as @igor-dv  ? 
Just build the static storybook & run `yarn test --image` :)  Thanks !

Hypnosphi:
Sure, but some of the tests are still failing on CI: https://circleci.com/gh/storybooks/storybook/41017#artifacts/containers/0

thomasbertet:
Yep .. argh not sure what to do, just raise the threshold and let some image still pass even if incorrect ? :(

Hypnosphi:
I think a good solution would involve taking screenshot and verifying them in the same environment (e.g. in docker)

Without that, you'll always have a number of false positives or negatives depending on threshould. To me, personally, false positives (failures without actual error) are much worse, because they block development.

In the case of storybook, we won't miss any visual changes, because Chromatic will catch them for us. But for external storyshots users, with current realization it always will be a tough compromise. That's basically what @tobilen was talking about when you opened the first PR:
https://github.com/storybooks/storybook/pull/2413#issuecomment-348566617

thomasbertet:
> I think a good solution would involve taking screenshot and verifying them in the same environment (e.g. in docker) 

Are we doing this anywhere in this repo ?

Hypnosphi:
No

thomasbertet:
I will just raise the threshold then.

Hypnosphi:
Ok, I'll check how it works on my machine

examples/official-storybook/image-snapshots/storyshots-image.runner.js
```diff
@@ -21,13 +21,22 @@ if (!fs.existsSync(pathToStorybookStatic)) {
     suite: 'Image snapshots',
     framework: 'react',
     configPath: path.join(__dirname, '..'),
-    storyNameRegex: /^((?!tweaks static values with debounce delay|Inlines component inside story).)$/,
```
Hypnosphi: What would be the correct way to ignore those stories?

igor-dv: That's the way to ignore... But adding a proper ignore list to the api will definitely simplify everything...

Hypnosphi: Looks like it didn't work: it ignored all the tests instead

UPD: I just missed `*` in the end, it had to be this:
```
/^((?!tweaks static values with debounce delay|Inlines component inside story).)*$/
```

Hypnosphi: > adding a proper ignore list

#2679 should help a lot here. We could just introduce `skipSnapshots` story parameter (name can be configurable)

igor-dv: Yeah 👍

Hypnosphi:
I have a local failure:

```
● Image snapshots › Addons|Info.Options.inline › Inlines component inside story

    toMatchImageSnapshot(): Received image size must match baseline snapshot size in order to make comparison.

      69 |       return page.screenshot(getScreenshotOptions({ context: context, url: url }));
      70 |     }).then(function (image) {
    > 71 |       expect(image).toMatchImageSnapshot(getMatchOptions({ context: context, url: url }));
      72 |     });
      73 |   };
      74 | 
      
      at diffImageToSnapshot (node_modules/jest-image-snapshot/src/diff-snapshot.js:50:13)
      at Object.toMatchImageSnapshot (node_modules/jest-image-snapshot/src/index.js:50:20)
      at Object.throwingMatcher [as toMatchImageSnapshot] (node_modules/expect/build/index.js:214:24)
      at addons/storyshots/dist/test-body-image-snapshot.js:71:21

  ● Image snapshots › Addons|Info.Options.inline › Inlines component inside story

    expect.assertions(1)
    
    Expected one assertion to be called but received zero assertion calls.
      
      at extractExpectedAssertionsErrors (node_modules/expect/build/extract_expected_assertions_errors.js:46:19)
```

I guess it's because different fonts lead to different line wrapping, which affects page height

thomasbertet:
Thanks @Hypnosphi for helping with this :love_letter: .
Yep I believe your assertion is correct. I was talking about this with the maintainer of jest-image-snapshot [here](https://github.com/americanexpress/jest-image-snapshot/pull/42) and [there](https://github.com/americanexpress/jest-image-snapshot/issues/49)

Hypnosphi:
I think we should just ignore the problematic stories

thomasbertet:
We can do that but we _will_ have the same issues ini a near future when adding more example stories.

Hypnosphi:
Maybe we should restrict storyshot image testing to some limited set of stories. After all, we do it only to ensure that image storyshots work

thomasbertet:
Ah! This is so true... I like this. I believe this is possible to whitelist some stories instead of excluding others, isn't it ? Will do this today.

Hypnosphi:
> I believe this is possible to whitelist some stories instead of excluding others, isn't it ?

It would be even simpler, because ot requires a positive regexp not a negative one. It could look like `/^Addons\|Storyshots/`

thomasbertet:
@Hypnosphi I think we are good to go, WDYT ? 

I added some stories to addon-storyshot to demo the use of the storyshots with image snapshots.
Seems to work fine now on CI with this subset.

examples/official-storybook/image-snapshots/storyshots-image.runner.js
```diff
@@ -19,15 +19,25 @@ if (!fs.existsSync(pathToStorybookStatic)) {
...
     test: imageSnapshot({
       storybookUrl: `file://${pathToStorybookStatic}`,
       getMatchOptions: () => ({
-        failureThreshold: 0.04, // 4% threshold,
+        failureThreshold: 0.02, // 2% threshold,
         failureThresholdType: 'percent',
       }),
+      beforeScreenshot: (page, { context }) => {
+        // Screenshots for this story have to be tweaked.
+        if (context.kind === 'ui/Layout') {
```
Hypnosphi: Looks like it's not needed anymore

Hypnosphi:
Thanks, I hope I'll be able to do the local testing again tomorrow

examples/official-storybook/stories/addon-storyshots.stories.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import BaseButton from '../components/BaseButton';
+
+const text = 'Testing the storyshots addon';
+
+storiesOf('Addons|Storyshots', module)
+  .add('Default', () => <BaseButton label="" />)
```
Hypnosphi: Using an an unstyled button for those stories isn't a really good idea, because it uses appearance of system UI button, which means the differences between OS are huge

E. g., this is how [storyshots story](https://deploy-preview-3102--storybooks-official.netlify.com/?selectedKind=Addons%7CStoryshots&selectedStory=Label&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) looks on mac
screen shot 2018-03-09 at 01 41 03

Hypnosphi:
Please take a look at screenshots before commiting them:
https://github.com/thomasbertet/storybook/blob/b71e110cc6fb5882526b7cdae0547bfc542621b0/examples/official-storybook/image-snapshots/__image_snapshots__/storyshots-image-runner-js-image-snapshots-addons-storyshots-default-1-snap.png

thomasbertet:
My bad, did this a bit too quickly, will be extra careful !

examples/official-storybook/stories/addon-storyshots.stories.js
thomasbertet: Oh so true ! I'll change the test, did a bit too quickly.

Hypnosphi (approved):
Works OK on macOS. @igor-dv please test on Windows

igor-dv:
I am getting this:

image

Not sure if it's a bug or my environment 🤷‍♂️

thomasbertet:
Oh @igor-dv seems a weird issue .... :(
Does it fails every time ?  I mean did you try multiple times ? 
Assuming that it worked before ... not sure what to do, maybe you can try to upgrade puppeteer to check if its better ?

igor-dv:
Yeah, tried it several times, I'll investigate it a bit more

igor-dv (approved):
My bad. Tried it before with an unstable version of Node. Works with the latest stable.

thomasbertet:
Great, thanks @Hypnosphi & @igor-dv !

Issue #1699, opened 2017-08-21
Have stories on each level of hierarchy


igor-dv:
This is very asked improvement from the beginning of the hierarchy feature.

Today we have this:
```js
storiesOf('Cells/Molecules', module)
  .add('with text', () => <Button>Hello Button</Button>)
  .add('with some emoji', () => <Button>😀 😎 👍 💯</Button>);

storiesOf('Cells/Molecules/Atoms', module)
  .add('with text2', () => <Button>Hello Button</Button>)
  .add('with some emoji2', () => <Button>😀 😎 👍 💯</Button>);
```

That will be shown as 
```
- Cells
  - Molecules                
    - with text                
    - with some emoji   
  - Molecules               
    - Atoms                   
      - with text 2
      - with some emoji 2
```

Desired
```
- Cells
  - Molecules                
    - with text                
    - with some emoji   
    - Atoms                   
      - with text 2
      - with some emoji 2
```

In this case we should allow stories and groups to be in the same level.

igor-dv:
I would like us to discuss this. Do we want to implement this ? Do we want to keep it configurable ?

For example the first thing I can think about is to remove the story auto-selection  when clicking the storyKind.

TomiS:
Yes! I ran into this problem in less than 10 minutes with Storybook. I'd also consider this as a bug and not a feature improvement. I think Storybook should merge the nested structure independent of how it is exactly defined.

ndelangen:
I think the desired output IS clearly what everyone expects and wants us to do.

And I agree 100% that's what we want too.
Who's interesting in making a PR for this?

igor-dv:
I'll take it 😉

igor-dv:
What do you think about always putting stories at the beginning ? ()

![image](https://user-images.githubusercontent.com/7867954/29818340-942fd8e2-8cc4-11e7-8971-a3b6b7444f2c.png)

ndelangen:
@igor-dv I'm pretty sure that's what everyone is expecting ✅

ndelangen:
This feature is complete and merged to the 3.3 branch

PR #3105, opened 2018-02-26
Select the first story in tree when opening a tree item

real34:
## What I did

I improved the tree view to automatically select the first story (leaf) when opening a group if possible. It allows the styleguide to open a section more easily so an user could go to the content faster.

## How to test

Is this testable with jest or storyshots? Not sure

### DIFF
--- a/lib/ui/src/modules/ui/components/stories_panel/stories_tree/index.js
+++ b/lib/ui/src/modules/ui/components/stories_panel/stories_tree/index.js
@@ -77,6 +77,11 @@ class Stories extends React.Component {
   onToggle(node, toggled) {
     if (node.story) {
       this.fireOnKindAndStory(node.kind, node.story);
+    } else if (node.children.length) {
+      const defaultStory = node.children[0];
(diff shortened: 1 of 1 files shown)


Hypnosphi:
@igor-dv AFAIR, you intentionally removed this behaviour at some point

igor-dv (changes requested):
Yeah, there was an old discussion here #1699. IMO it doesn't make anything easier. 
Anyway, if this is a wanted behavior at least it should be configurable.

igor-dv:
CC. #2863

real34:
I’d be glad to add a configuration for this.

However I have not yet taken the time to setup the tests locally, and might find time for it only Friday I think. Do you think it could be on time for the `3.4.0` release ?

igor-dv:
I think the release should be somewhere soon. If you won't be ready up to the time, maybe we will be able to sneak it as a patch release after.

CC. @shilman.

creatyvtype:
@real34 did this ever get implemented?

PR #3114, opened 2018-02-28
Update ADDONS_SUPPORT.md

getsetbro:
The viewport plugin works in angular.

Issue:

## How to test


igor-dv (approved):
Thanks !

Issue #2445, opened 2017-12-07
[addon-knob] Display for Number value, min, max when slider is used.


madhaze:
Feature Request:  
It would be nice to have Value, Min and Max displayed when using the number knob range to give the user some context before they grab the slider.

Hypnosphi:
Looks like it may be relatively easy to implement using `<datalist>` element, see "A range control with hash marks and labels" section here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

Feel free to send a PR, but please base it on `release/3.3` branch

cilice:
Unfortunately, even if it's within the spec, no browsers support it fully right now (codepen https://codepen.io/cilice/pen/ypBWKM)

> Currently, no browser fully supports these features. Firefox doesn't support hash marks and labels at all, for example, while Chrome supports the hash marks but doesn't support labels.

Unfortunately, latest Safari supports neither hash marks nor labesls just like Firefox. ☹️

apertureless:
Diplaying the currentValue would be nice. Especially while development when you're tweaking some default values.

mkochendorfer:
I am very interested in the feature request.

Hypnosphi:
@mkochendorfer want to open a PR?

mkochendorfer:
I don't have the time this hit this right now, but possibly in the coming months.

jeffcarey:
I can work on this in the coming days. I'd propose something simple to get started: Minimum on the left, and Current / Maximum on the right. The UI can be enhanced later with tick marks and so on if browsers start supporting that more.

Hypnosphi:
Sorry, 3.4 is feature-frozen

madhaze:
figured.  Thought it was worth asking :) 👍

PR #3128, opened 2018-03-02
#2445 Add min, value, and max labels to range knob

jeffcarey:
Issue: #2445 

## What I did
Added a simple display for the range slider for knobs of type Number: minimum on the left of the slider, and current value and maximum on the right of the slider.

storybook_range_labels

## How to test
Add a knob to a story: of type 'number' with range set to true. View the knobs panel for the story.
Expected: A slider appears with the minimum, current value, and maximum. The current value updates as you move the slider.

No

No (Appears in "official storybook" example).

No


Hypnosphi (approved):

madhaze:
@jeffcarey nice.  this looks great.

PR #3133, opened 2018-03-02
Addon actions: fix slow logging

### LINKED ISSUE #2590: Action addons is slow to log events

rhalff:
All non plain objects will be serialized to a depth of 2.

Events are not treated differently anymore traversing is also cut off at the depth of 2.

All plain objects are still serialized until a depth of 10 (could probably be increased)

Properties starting with double underscore  `__` and the storybook's `STORYBOOK_` vars are not serialized and ignored.

fixes #2590


Hypnosphi:
Can we make it configurable? I think there might be situation where people may need deeper nesting

rhalff:
Yes, it possible to make it configurable.

Should I implement this by listening on the channel for options set by the `Options` addon?

Hypnosphi:
I think exporting a configuration function from actions plugin would be more explicit

rhalff:
Sorry, no code update yet, I was only bringing this branch up to date.

addons/actions/src/lib/retrocycle.js
```diff
@@ -2,7 +2,6 @@ import reviver from './reviver';
 import { muteProperty } from './util';
 import { CYCLIC_KEY } from './';
 
-// eslint-disable-next-line no-control-regex
```
Hypnosphi: Looks like it's still needed

rhalff: Yeah, my ide reported it wasn't, will put it back.

addons/actions/src/index.js
```diff
@@ -1,6 +1,8 @@
+import { action, configureActions, decorateAction } from './preview';
+
 // addons, panels and events get unique names using a prefix
 export const ADDON_ID = 'storybook/actions';
 export const PANEL_ID = `${ADDON_ID}/actions-panel`;
 export const EVENT_ID = `${ADDON_ID}/action-event`;
 
-export * from './preview';
```
Hypnosphi: Yeah, unfortunately this `export *` syntax doesn't play well with `transform-runtime` https://github.com/babel/babel/issues/2877

Hypnosphi:
Can you please add an example with deeply nested object and different `depth` configuration?

rhalff:
It seems the performance gain still relies on setting a fixed `maxDepth` of 2 whenever an object without a name is encountered, I noticed I didn't remove that part yet: https://github.com/storybooks/storybook/pull/3133/files#diff-017c725a54046de74bba2b97b986116aR5

The decision I made there seems a bit arbitrary, but when I remove that part logging something like window will hang again. I'm not sure yet what would be a better way to prevent this.

Perhaps it's also the amount of properties/entries per object and not only the depth which causes the lag.

Hypnosphi (approved):

Hypnosphi:
I assume it's ready

Issue #3044, opened 2018-02-21
Webpack 4 support?


ConneXNL:
Webpack 4 is releasing in a couple days that could potentially speed up the build process. 

Is storybook ready for webpack 4?

I tried to use the beta versions of ts-loader and webpack together with storybook but running into errors that most likely indicate storybook isn't compatible yet with the upcoming webpack 4.

```
I tried to use:
```
    "ts-loader": "^4.0.0-beta.0",
    "webpack": "^4.0.0-beta.2",
```

and I am running into problems:
`Module build failed: TypeError: Cannot set property 'tsLoaderFileVersion' of undefined at successLoader `
```

Hypnosphi:
Storybook uses webpack as direct dependency. So you can't replace its version on your side. Instead, you can fork our repo and try to update corresponding packages to beta version. If things go well, we'll be able to do the upgrade as soon as it becomes stable

walfly:
Anyone doing this yet?

Hypnosphi:
I'll do that, and it will be a breaking change in storybook. The thing is that custom webpack configs may contain plugins, and plugins targeted for webpack 4 are most of the times (if not always) not backwards compatible with webpack 3

walfly:
Great, thanks so much!

patrys:
Currently it's broken as even though storybook explicitly requires `webpack==3.11.0` and gets a private copy of Webpack 3 in its `node_modules` subdirectory, it also requires `dotenv-webpack`.

Now `dotenv-webpack` lists a peer dependency of `webpack==^1 || ^2 || ^3 || ^4` which makes `npm` consider it a good idea to install it in the main `node_modules` next to Webpack 4 instead of next to Webpack 3 under `@storybook/react/node_modules`.

Now as soon as storybook loads `dotenv-webpack` it in turns causes Webpack 4 to be invoked.

patrys:
I believe the above is caused by a bug in `npm`: https://github.com/npm/npm/issues/19944

Kirk-Wang:
@ConneXNL  [ts-loader's README](https://github.com/TypeStrong/ts-loader)

### Compatibility
* TypeScript: 2.4.1+
* webpack: 4.x+ (please use ts-loader 3.x if you need webpack 2 or 3 support)
* node: 6.11.5 minimum (aligned with webpack 4)

Hypnosphi:
Things that block the migration right now

[...]

Issue #2635, opened 2018-01-03
Babelrc resolving to wrong location


alansouzati:
### Issue details

When I have a local babel preset, storybook is not able to locate it.

### Steps to reproduce

- Run `start-storybook` where you have a local babel preset. For example:

```
{
  "presets": [
    "./tools/babel-preset-env",
    "stage-2",
    "react"
  ]
}
```

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/react 3.3.3

### Affected platforms 

- Npm 5.6.0
- Node v8.9.4
- macOS High Sierra

### Screenshots / Screencast / Code Snippets (Optional)

> ERROR in ./storybook/addons.js
Module build failed: Error: Couldn't find preset "./tools/babel-preset-env" relative to directory "/Users/user/workspace/project/storybook"

Hypnosphi:
Looks like it locates the file itself, but fails to resolve relative path from it correctly

Probably related: #2610

alansouzati:
my current work around is to set a `.babelrc` file inside `storybook` folder that does not use the local preset.

I will try to investigate what is going on.

dijs:
Any more word on this? I need to reference a local babel plugin and am running into the exact same issue.

Hypnosphi:
Released as [`4.0.0-alpha.20`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.20)

Issue #2610, opened 2017-12-30
Storybook does not process .babelrc extends correctly


adamchenwei:
### Issue details

_Please provide issue details here - What did you expect to happen? What happened instead?_ 
I expect `npm run storybook` to load correctly without errors show in screenshot below
### Steps to reproduce
get react-static, produce a site, install storybook-cli globally then `getstorybook` in the react-static produced base site.

_Please provide necessary steps for reproduction of this issue. Describe the exact steps a maintainer has to take to make the problem occur. If the problem is non-trivial to reproduce, please link a repository or provide some code snippets._
https://github.com/nozzle/react-static
_(A screencast can be useful for visual bugs, but it is not a substitute for a textual description.)_

### Please specify which version of Storybook and optionally any affected addons that you're running
@storybook/cli@3.3.3
@ "react": "^16.0.0",

### Affected platforms 
Win10

### Screenshots / Screencast / Code Snippets (Optional)
![image](https://user-images.githubusercontent.com/4201698/34457853-57a988d6-ed8c-11e7-9bbf-73890572f8d7.png)

```js
// code here
```

Hypnosphi:
Here's the `.babelrc` this generator produces:
```
{
  "extends": "react-static/.babelrc"
}
```

Looks like we don't process this `extends` properly

Hypnosphi:
Released as [`4.0.0-alpha.20`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.20)

Issue #754, opened 2017-04-10
babel config in package.json


ButuzGOL:

ndelangen:
Please give us more information what is wrong or being requested here.

ButuzGOL:
currently storybook takes config only from .babelrc but if i want to keep it in package.json
https://github.com/storybooks/react-storybook/blob/master/src/server/babel_config.js#L51

https://babeljs.io/docs/usage/babelrc/#use-via-package-json

fczuardi:
@ndelangen I think this is still an issue and we should reopen this bug.

storybook does not honor babel presets and plugins defined inside package.json (the workaround is to copy that babel config to a different file, and name that file .babelrc)

I think it would be nice if storybook is able to use existing custom babel configs when they are present in the package.json, and I can try to write up a patch if this issue gets reopened.

ndelangen:
Sure, I would be happy to get a PR on this!

Here's where the code is:
https://github.com/storybooks/storybook/blob/master/app/vue/src/server/babel_config.js

(there's similar code for `app/vue`)

ndelangen:
@fczuardi Need help help opening a PR?

JulianoRodrigues:
Hey guys, do you still need help with this issue?

Hypnosphi:
Released as [`4.0.0-alpha.20`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.20)

PR #3148, opened 2018-03-04
[BREAKING CHANGE] webpack 4

Hypnosphi:
Issues: #3044, #3083, #3085, #3092, #3115, #3116, #3135

Things that block this right now

- [x] stable `3.4.0` release (we don't want to maintain 3 versions at the same time)
- [x] ~`react-dev-utils`, namely`InterpolateHtmlPlugin` and `WatchMissingNodeModulesPlugin`. See https://github.com/facebook/create-react-app/pull/4077~ (actually they work as is now)
- [x] #3161
- [x] Fix error in angular-cli example https://github.com/storybooks/storybook/pull/3251


Hypnosphi:
Weird, storybook build for polymer works OK locally on my machine, but on CI it's failing:

```
ERROR in /tmp/storybook/addons/knobs/dist/polymer/WrapStory.html
Module not found: Error: Can't resolve 'polymer-webpack-loader/register-html-template' in '/tmp/storybook/addons/knobs/dist/polymer'
 @ /tmp/storybook/addons/knobs/dist/polymer/WrapStory.html 7:27-83
 @ /tmp/storybook/addons/knobs/dist/polymer/index.js
 @ /tmp/storybook/addons/knobs/polymer.js
 @ ./src/stories/addon-knobs.stories.js
 @ ./src/stories sync \.stories\.js$
 @ ./.storybook/config.js
 @ multi /tmp/storybook/app/polymer/dist/server/config/polyfills.js /tmp/storybook/app/polymer/dist/server/config/globals.js /tmp/storybook/node_modules/webpack-hot-middleware/client.js?reload=true ./.storybook/config.js
```

Hypnosphi:
For now, I've just disabled `InterpolateHtmlPlugin` and `WatchMissingNodeModulesPlugin` which means that first 4.0 alpha will probably lack the corresponding features: `.env` support in `*-head.html`, and watching for newly installed packages

igor-dv:
Both plugins are rather simple, maybe it worth copypasting/fixing them into the core until the official support?

Hypnosphi:
They say they're almost ready: https://github.com/facebook/create-react-app/pull/4077#issuecomment-370789259

pkuczynski:
Could you also in this work stream upgrade `babel` to `7.0` (currently in `beta.40`) and `babel-loader` to 8.0.0-beta.2`? That allows using more dynamic `.babelrc.js`...

Hypnosphi:
@pkuczynski sure, definitely makes sense. I think it'll be a separate PR though

pkuczynski:
Cool! I can provide a PR even now, as this is not related to webpack and I am being able to successfully use `babel@7` with `webpack@3` in my project...

And why do you need to wait for https://github.com/babel/babel/pull/7472?

Hypnosphi:
Because it will solve a lot of issues we currently have because of custom babelrc resolution

See e.g. #754 #2635 #2610

pkuczynski:
But then it would require quite some work on your side I guess? While basic upgrade to babel@7 should work out of the box now. So maybe it’s worth to do it in two steps?

Hypnosphi:
Probably. Feel free to give it a try

app/react/src/server/config/webpack.config.prod.js
```diff
@@ -47,21 +46,8 @@ export default function(configDir) {
...
-          ie8: false,
-          mangle: false,
-          warnings: false,
-          compress: {
-            keep_fnames: true,
-          },
-          output: {
-            comments: false,
-          },
-        },
```
Hypnosphi: @ndelangen should we replicate those settings in `optimization.minimizer` option, or are we OK with webpack [defaults](https://github.com/webpack/webpack/blob/4f399322ad1922ae7bafdeea1c4d83bf9c9851a7/lib/WebpackOptionsDefaulter.js#L271-L276)?

Hypnosphi:
We'll be able to add chunks splitting after either https://github.com/jantimon/html-webpack-plugin/issues/882 or https://github.com/webpack/webpack/issues/6717 resolves

app/react/src/server/config/webpack.config.prod.js
ndelangen: Let's try to use their defaults!

TomFoyster:
Is there any timeframe for this? We'd love to use WP4 in our production code, but as the component lib uses Storybook we can't progress.

Hypnosphi:
I think it's ready to merge as is, so we could release `4.0.0-alpha` ASAP
I'm also going to release a temporary fork of `react-dev-utils` under `@storybook` scope

@shilman WDYT

Cryrivers:
How could I test this branch?

app/vue/package.json
```diff
@@ -43,15 +43,14 @@
     "find-cache-dir": "^1.0.0",
     "global": "^4.3.2",
     "html-loader": "^0.5.5",
-    "html-webpack-plugin": "^2.30.1",
+    "html-webpack-plugin": "^3.0.7",
     "json5": "^0.5.1",
     "markdown-loader": "^2.0.2",
     "react": "^16.2.0",
-    "react-dev-utils": "^5.0.0",
+    "react-dev-utils": "zanettin/react-dev-utils",
```
danielduan: is this a repo we're waiting for changes to be merged and published to npm for `react-dev-utils`

app/angular/package.json
```diff
@@ -41,19 +41,18 @@
...
     "html-loader": "^0.5.5",
-    "html-webpack-plugin": "^2.30.1",
+    "html-webpack-plugin": "^3.0.7",
     "json5": "^0.5.1",
     "markdown-loader": "^2.0.2",
     "node-sass": "^4.8.3",
     "raw-loader": "^0.5.1",
     "react": "^16.0.0",
-    "react-dev-utils": "^5.0.0",
+    "react-dev-utils": "zanettin/react-dev-utils",
```
ndelangen: Is this something we could revert back now?

ndelangen (changes requested):

ndelangen:
Besides the forked `react-dev-utils` all looks perfect to me.

If we need to keep the fork, we should publish a forked version, I'd rather not ship with installs from github, that's too unpredictable and users behind a proxy might not be able to install.

app/angular/package.json
Hypnosphi: https://github.com/facebook/create-react-app/pull/4077 is still open

Hypnosphi:
> we should publish a forked version

OK, I'll do that

ndelangen (approved):

danielduan (approved):

DevanB:
Super excited to see this released somewhere. One of our project's Storybook is broken due to Babel@7 and Webpack@4. 

🎉

pkuczynski:
This PR does not include Babel@7...

Cryrivers:
Very excited to see this merged. Could we have a prerelease on npm now?

Hypnosphi:
Cc @shilman

TomFoyster:
What are the chances this'll make it into 3.4.0?

Really keen to see this up on NPM.

Cryrivers:
@TomFoyster I believe it goes into `4.0.0` as it's a breaking change.

TomFoyster:
@Cryrivers You're right, I imagine it will. A shame as we're chomping at the bit to get WP4 into our work, but can't without Storybook's support.

pkuczynski:
And what's the timeline for getting this released? I am blocked with migrating my project to WP4 too :(

danielduan:
@ndelangen @Hypnosphi @shilman let's get an alpha `4.0` published?

If @shilman is unavailable, I think we should have a backup person as well.

Hypnosphi:
Ok, I'll do `3.4.0-rc` and `4.0.0-alpha` releases tomorrow if @shilman doesn't become available

Hypnosphi:
Released as [`v4.0.0-alpha.0`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.0). Please try it and open an issue if something goes wrong

ishantdigitalgenius:
Tried the latest version with the following versions
```
    "@storybook/addon-actions": "^3.3.0",
    "@storybook/addon-info": "^3.3.0",
    "@storybook/addon-knobs": "^3.3.0",
    "@storybook/cli": "^v4.0.0-alpha.0",
    "@storybook/react": "^v4.0.0-alpha.0",
```

When running storybooks, the module building stops at 95% with the following line.

```
95% emitting unnamed compat plugin
```

junish:
> 95% emitting unnamed compat plugin

I am facing with the same error.

I guess that the error is related to the issue of html-webpack-pugin:
https://github.com/jantimon/html-webpack-plugin/issues/870

AndyBan:
It gives you a bit more info when you kill it:

```
95% emitting unnamed compat pluginTypeError: missingDeps.some is not a function
    at compiler.plugin (C:\Javascript\node_modules\@storybook\react\node_modules\react-dev-utils\WatchMissingNodeModulesPlugin.js:25:23)
    at _err1 (eval at create (C:\Javascript\node_modules\tapable\lib\HookCodeFactory.js:24:12), <anonymous>:26:1)
    at Promise.resolve.then.then.then.then.then.then.then.then.catch.then.then.then (C:\Javascript\node_modules\html-webpack-plugin\index.js:222:11)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:160:7)
```

Hypnosphi:
@ishantdigitalgenius @AndyBan can someone of you please create a GitHub repo with minimal reproduction and open an issue with link to it?

SleepWalker:
@Hypnosphi this happens in `react-dev-utils`:

https://github.com/facebook/create-react-app/blob/next/packages/react-dev-utils/WatchMissingNodeModulesPlugin.js#L20-L31

This thing tries to loop over `missingDeps` of type `SortableSet`, which does not have method `some()`:

https://github.com/webpack/webpack/blob/master/lib/util/SortableSet.js

Due to error being thrown, the `callback` won't be called

Hypnosphi:
Any idea why does it work for our [example apps](https://github.com/storybooks/storybook/tree/master/examples)?

SleepWalker:
The next error will be here:

https://github.com/facebook/create-react-app/blob/next/packages/react-dev-utils/InterpolateHtmlPlugin.js#L37

```
ERROR in   TypeError: callback is not a function
  
  - InterpolateHtmlPlugin.js:37 compilation.plugin
    [fc-app]/[react-dev-utils]/InterpolateHtmlPlugin.js:37:11
```

This happens because `html-webpack-plugin` does not provide `callback` argument anymore:

https://github.com/jantimon/html-webpack-plugin/blob/3afd5d228a2775282e751c56eb81c3bf1a2f7b05/index.js#L177-L178
https://github.com/jantimon/html-webpack-plugin/blob/a7d37ca472cb5d9ee9dd58932b169e936416f479/spec/BasicSpec.js#L920-L931

Looks like this is due to new plugin system of webpack4


I have fixed both errors in a very dumb way:

```js
      if (missingDeps.some && missingDeps.some(file => file.indexOf(nodeModulesPath) !== -1)) {

      callback && callback(null, data);
```

After this durty hack the storybook have successfully started

SleepWalker:
Here is more clever candidates for fix:

`InterpolateHtmlPlugin.js`:

```js
...
  apply(compiler) {
    compiler.plugin('compilation', compilation => {
      const handler = (data, callback) => {
        // Run HTML through a series of user-specified string replacements.
        Object.keys(this.replacements).forEach(key => {
          const value = this.replacements[key];
          data.html = data.html.replace(
            new RegExp('%' + escapeStringRegexp(key) + '%', 'g'),
            value
          );
        });
        callback(null, data);
      };

      if (compilation.hooks) {
        compilation.hooks.htmlWebpackPluginBeforeHtmlProcessing
          .tapAsync('InterpolateHtmlPlugin', handler);
      } else {
        compilation.plugin('html-webpack-plugin-before-html-processing', handler);
      }
    });
  }
```

`WatchMissingNodeModulesPlugin.js`:

I've converted Set into an Array, but I'm not sure whether webpack have changed only underlying type or set item type too. I hope that there are still strings as before :)

```js
  apply(compiler) {
    compiler.plugin('emit', (compilation, callback) => {
      var missingDeps = compilation.missingDependencies;
      var nodeModulesPath = this.nodeModulesPath;

      // If any missing files are expected to appear in node_modules...
      if (Array.from(missingDeps).some(file => file.indexOf(nodeModulesPath) !== -1)) {
        // ...tell webpack to watch node_modules recursively until they appear.
        compilation.contextDependencies.push(nodeModulesPath);
      }

      callback();
    });
  }
```

Hypnosphi:
It's actualy already fixed in https://github.com/facebook/create-react-app/pull/4077 and we've released it from our fork as `@storybook/react-dev-tools`. Unfortunately, we have github path [stored] in our lockfile, which made me think that `5.0.0` works again because of some compatibility fix in webpack.

I'll send a hotfix PR ASAP

danielduan:
@ishantdigitalgenius try to use the same version of addons as the rest of the `@storybook` packages. even though there usually aren't any issues, we don't guarantee different versions working together.

Hypnosphi:
Fix released as [`4.0.0-alpha.1`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.1)

peet86:
With 4.0.0-alpha.1 I still have the [jantimon/html-webpack-plugin#870](https://github.com/jantimon/html-webpack-plugin/issues/870) issue.. :(

` 95% emitting HtmlWebpackPlugin/[..]/node_modules/@storybook/react/node_modules/toposort/index.js:29
      throw new Error('Cyclic dependency: '+JSON.stringify(node))
                                                 ^
TypeError: Converting circular structure to JSON`

Hypnosphi:
We probably need to apply workaround from https://github.com/jantimon/html-webpack-plugin/issues/870#comment-370004105

For now, you can try to apply it from custom webpack config (full control mode):

```
export (baseConfig, env, defaultConfig) => ({
  ... defaultConfig,
  plugins: defaultConfig.plugins.map(plugin => {
    if (plugin.constructor.name === 'HtmlWebpackPlugin') {
      return { ...plugin, chunksSortMode: 'none' }
    }
    return plugin
  })
})
```

andrerpena:
Thank you! Storybook is now working with Webpack 4 and custom Webpack configs but, just so you know, I had to add `babel-core` as a `dev-dependency` to my project in order for it to work.

igor-dv:
`babel-core` is a peer dep, so you should add it

Hypnosphi:
We strongly recommend to use the same versions for `@storybook/<framework>` and addons

VRedondo:
Following settings doesn't work for me

```
"devDependencies": {
  "webpack": "^4.12.0",
  "@storybook/angular": "^4.0.0-alpha.18",
  "babel-core": "^6.26.3",
  "babel-loader": "^8.0.0",
```

when running storybook it returns the error 

```

ERROR in ./.storybook/config.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@babel/core'
```

SleepWalker:
@VRedondo you need to use `babel@7` to run the latest alpha, or use the version with babel 6. I've used `alpha14`, but the most recent is `alpha16`

VRedondo:
Thanks @SleepWalker and @Hypnosphi actually both were needed, babel-loader 7 requires babel 7 so it got fixed with
"@babel/core": "^7.0.0",
"babel-loader": "^8.0.0",

Hypnosphi:
> babel-loader 7 requires babel 7 

What made you think so?

VRedondo:
@Hypnosphi the console output says 
```
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Requires Babel "^7.0.0-0", but was loaded with "6.26.3".
``` 
when running with babel-loader 7 in case you don't have babel 7

Hypnosphi:
Yeah, you also need your own babelrc in this case, our default config is for Babel 7

PR #3154, opened 2018-03-06
[STORYSOURCE] Add ability to select stories from inside of the StoryPanel

igor-dv:
## What I did
I've added the navigation from within the addon-storysource panel.

Since it's still in alpha, I'm not sure whether this is a feature request or not.


addons/storysource/src/StoryPanel.js
```diff
@@ -57,29 +102,74 @@ export default class StoryPanel extends Component {
...
+    if (StoryPanel.areLocationsEqual(location, currentLocation)) {
+      return (
+        <div key={storyKey} ref={this.setSelectedStoryRef} style={styles.selectedStory}>
+          {story}
+        </div>
+      );
+    }
+
+    return (
+      <div
```
Hypnosphi: why not just use `button` tag?

igor-dv: I think it's semantically more like an anchor than a button so maybe `a` ?

igor-dv: Just checked and resetting a button to look like a div - is hell...

Hypnosphi: I'd say that making div behave as a button is hell =)

> I think it's semantically more like an anchor than a button so maybe `a`?

Cool idea as soon as you attach a valid `href`. You may want to use `RoutedLink` component from `@storybook/components` to avoid page reload on plain clicks

ndelangen (approved):

PR #3156, opened 2018-03-06
Add 'waitUntil' option to puppeteer of storyshots

markelog:
If some elements use additional xhr requests (like async image loading).
Puppeteer might not imprint the end state of the page. Puppeteer has
built-in method to resolve this -
https://github.com/GoogleChrome/puppeteer/blob/v1.1.1/docs/api.md#pagegotourl-options

This option is fairly general so it should be okay to use for other cases
as well.

However, 'beforeScreenshot' could be used instead, but proposed approach seems
more definite and reliable

As additional further improvement, these `goto` options might be exposed to
consumber as well


Hypnosphi:
Will it work on examples that do some kind of polling?

ndelangen (approved):

ndelangen:
If a user decides to poll every `500ms` in a story...

> networkidle0 - consider navigation to be finished when there are no more than 0 network connections for at least 500 ms.

That is probaly super uncommon... If a story is about a component that changes over time.. My advice is to create a story for a few moment in time, not to do the actual polling in storybook.

What do you think @Hypnosphi ?

Hypnosphi:
Ok, at TeamCity we have some stories that poll data from actual server, but we use a period larger that 500ms, so it's OK

But we may start utilizing websockets for that at some point. Would WS connection count as "network connection"?

markelog:
> Will it work on examples that do some kind of polling?

Good question. 

I think even with old-school comet technics and the new ones like ws it should work in the same way (should be tested, to verify of course). 

In the case, when ws, on user agent side, isn't always receiving new data and if there is no ongoing ping pong frames (https://html.spec.whatwg.org/multipage/web-sockets.html#ping-and-pong-frames) proposed logic should work.

Since of course while test run in in progress, you probably would want, at some point, to stop receiving new data during the screenshot time. And in both ws and comet technics (known to me), you can adjust the heart-beat rate

Hypnosphi (approved):
LGTM

PR #3161, opened 2018-03-07
Move polymer loader to peerDependencies

Hypnosphi:
Issue: `polymer-webpack-loader` [adds](https://github.com/webpack-contrib/polymer-webpack-loader/blob/cab1ebd2bad3243fc3d4cac10694304c95b1d66a/src/index.js#L143) following line to processed files:
```
require('polymer-webpack-loader/register-html-template')
```

It means that the package should always be available on root level. We can't rely on npm hoisting for that, so we need to move it to `peerDependencies`


ndelangen:
@stijnkoopal What do you think?

app/polymer/package.json
```diff
@@ -59,10 +58,12 @@
     "webpack-hot-middleware": "^2.21.2"
   },
   "devDependencies": {
-    "nodemon": "^1.17.1"
+    "nodemon": "^1.17.1",
+    "polymer-webpack-loader": "^2.0.1"
```
stijnkoopal: Not sure that we need this dependency still, but perhaps I'm missing something here.

ndelangen: Thanks! that doesn't sounds like a blocker tbh

ndelangen (approved):

PR #3169, opened 2018-03-07
Move "@types/react" to dev dependencies

MLoughry:
Issue: `@types/react` is listed as an optional dependency in `addon-notes`, which can cause type conflicts to any downstream consumer with a different types version

## What I did
Moved the dependency to devDependencies

## How to test
Build


igor-dv (approved):
Makes sense. I think it was here by mistake.
Thanks.

ndelangen (approved):
Thank you @MLoughry !

danielduan (approved):

PR #3180, opened 2018-03-10
Delaying update of height and width in Layout

### LINKED ISSUE #1990: No throttling on resize event handlers.
**Issue**

When you resize windows fast, everything is laggy, and you feel like:
image

---

My take on this:

**Cause**

The [resize event handler](https://github.com/storybooks/storybook/blob/master/lib/ui/src/modules/ui/components/layout/index.js#L138) is not [throttled](https://lodash.com/docs/4.17.4#throttle).

**Solution**

```jsx
import { throttle } from 'lodash'; // You have it as dependency already.

  {/* ... component stuff */}
   componentDidMount() {
    window.addEventListener('resize', this.onResize);
  }

  componentWillUnmount() {
    window.removeEventListener('resize', this.onResize);
  }

  onResize = throttle(this.handleOnResize, 200, { trailing: true });

  {/* onResize renamed to handleOnResize */}
  {/* ... more component stuff */}
```

**Apart from that**

_Thank you for making this!_

<3
-Dan

wuweiweiwu:
Issue:

## What I did
fixes #1990 
Delayed the `setState` call so we don't rerender a billion times.

Plus removed adding un-needed event handlers to window

## How to test
use React inspector to check state updates

```sh
yarn start
```
resize the windows and see that the `previewPanelDimensions` state isn't updated thus doesn't rerender until after you stopped dragging for a little bit (50ms)

### Is this testable with jest or storyshots?
No

No

No

### DIFF
--- a/lib/ui/src/modules/ui/components/layout/index.js
+++ b/lib/ui/src/modules/ui/components/layout/index.js
@@ -1,4 +1,4 @@
-import { localStorage, window } from 'global';
+import { localStorage } from 'global';
 import PropTypes from 'prop-types';
 import React from 'react';
 import SplitPane from 'react-split-pane';
@@ -140,14 +140,6 @@ class Layout extends React.Component {
     this.onDragEnd = this.onDragEnd.bind(this);
   }
 
-  componentDidMount() {
-    window.addEventListener('resize', this.onResize);
-  }
-
-  componentWillUnmount() {
-    window.removeEventListener('resize', this.onResize);
-  }
-
   onDragStart() {
     this.setState({ isDragging: true });
   }
@@ -157,18 +149,22 @@ class Layout extends React.Component {
   }
 
   onResize(pane, mode) {
+    let resizeTimeout;
+
     return size => {
-      this.layerSizes[pane][mode] = size;
-      saveSizes(this.layerSizes);
+      clearTimeout(resizeTimeout);
+      resizeTimeout = setTimeout(() => {
+        this.layerSizes[pane][mode] = size;
+        saveSizes(this.layerSizes);
 
-      const { clientWidth, clientHeight } = this.previewPanelRef;
-
-      this.setState({
-        previewPanelDimensions: {
-          width: clientWidth,
-          height: clientHeight,
-        },
-      });
+        const { clientWidth, clientHeight } = this.previewPanelRef;
+        this.setState({
+          previewPanelDimensions: {
+            width: clientWidth,
+            height: clientHeight,
+          },
+        });
+      }, 50);
     };
   }
 
--- a/lib/ui/src/modules/ui/components/layout/index.js
+++ b/lib/ui/src/modules/ui/components/layout/index.js
@@ -2,6 +2,7 @@ import { localStorage } from 'global';
 import PropTypes from 'prop-types';
 import React from 'react';
 import SplitPane from 'react-split-pane';
+import _ from 'lodash'; /* eslint-disable-line import/no-extraneous-dependencies */
 
 import USplit from './usplit';
 import Dimensions from './dimensions';
@@ -135,7 +136,7 @@ class Layout extends React.Component {
       isDragging: false,
     };
 
-    this.onResize = this.onResize.bind(this);
+    this.onThrottledResize = _.throttle(this.onResize.bind(this), 200);
     this.onDragStart = this.onDragStart.bind(this);
     this.onDragEnd = this.onDragEnd.bind(this);
   }
@@ -148,24 +149,17 @@ class Layout extends React.Component {
     this.setState({ isDragging: false });
   }
 
-  onResize(pane, mode) {
-    let resizeTimeout;
+  onResize(pane, mode, size) {
+    this.layerSizes[pane][mode] = size;
+    saveSizes(this.layerSizes);
 
-    return size => {
-      clearTimeout(resizeTimeout);
-      resizeTimeout = setTimeout(() => {
-        this.layerSizes[pane][mode] = size;
-        saveSizes(this.layerSizes);
-
-        const { clientWidth, clientHeight } = this.previewPanelRef;
-        this.setState({
-          previewPanelDimensions: {
-            width: clientWidth,
-            height: clientHeight,
-          },
-        });
-      }, 50);
-    };
+    const { clientWidth, clientHeight } = this.previewPanelRef;
+    this.setState({
+      previewPanelDimensions: {
+        width: clientWidth,
+        height: clientHeight,
+      },
+    });
   }
 
   render() {
@@ -212,7 +206,9 @@ class Layout extends React.Component {
           resizerStyle={storiesResizerStyle(showStoriesPanel, storiesPanelOnTop)}
           onDragStarted={this.onDragStart}
           onDragFinished={this.onDragEnd}
-          onChange={this.onResize('storiesPanel', storiesPanelOnTop ? 'top' : 'left')}
+          onChange={size =>
+            this.onThrottledResize('storiesPanel', storiesPanelOnTop ? 'top' : 'left', size)
+          }
         >
           <div style={storiesPanelStyle(showStoriesPanel, storiesPanelOnTop)}>
             <div style={{ flexGrow: 1, height: '100%', width: '100%' }}>{storiesPanel()}</div>
@@ -229,7 +225,9 @@ class Layout extends React.Component {
             resizerStyle={addonResizerStyle(showAddonPanel, addonPanelInRight)}
             onDragStarted={this.onDragStart}
             onDragFinished={this.onDragEnd}
-            onChange={this.onResize('addonPanel', addonPanelInRight ? 'right' : 'down')}
+            onChange={size =>
+              this.onThrottledResize('addonPanel', addonPanelInRight ? 'right' : 'down', size)
+            }
           >
             <div style={contentPanelStyle(addonPanelInRight, storiesPanelOnTop)}>
               {/*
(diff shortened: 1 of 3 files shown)


lib/ui/src/modules/ui/components/layout/index.js
```diff
@@ -140,14 +140,6 @@ class Layout extends React.Component {
     this.onDragEnd = this.onDragEnd.bind(this);
   }
 
-  componentDidMount() {
-    window.addEventListener('resize', this.onResize);
```
danielduan: are these not needed anymore?

I think we still need to handle the case when the browser window gets resized?

wuweiweiwu: I don’t think so. Because `onResize` returns another function the window resize events never actually updates the state. 

I will have to double check but I think the `SplitPane` `onChange` handler fires when browser window is resizes.

lib/ui/src/modules/ui/components/layout/index.js
```diff
@@ -157,18 +149,22 @@ class Layout extends React.Component {
...
+        saveSizes(this.layerSizes);
+
+        const { clientWidth, clientHeight } = this.previewPanelRef;
+        this.setState({
+          previewPanelDimensions: {
+            width: clientWidth,
+            height: clientHeight,
+          },
+        });
+      }, 50);
```
Hypnosphi: Please use `lodash.debounce` package

Hypnosphi (changes requested):

lib/ui/src/modules/ui/components/layout/index.js
Hypnosphi: Yes we need. Try to resize window here https://deploy-preview-3180--storybooks-official.netlify.com

Hypnosphi:
I tried it and I can't say I like it:

https://deploy-preview-3180--storybooks-official.netlify.com/?selectedKind=ui%2FAddonPanel&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel



`Dimensions` panel is used for instant feedback, so picking debouncing over throttling doesn't look like a good choice to me

lib/ui/src/modules/ui/components/layout/index.js
Hypnosphi: Or maybe it should rather be `lodash.throttle`, see https://github.com/storybooks/storybook/pull/3180#issuecomment-372009581

wuweiweiwu: @Hypnosphi Sounds good. I discovered a new issue with using `_.throttle` the way that it is currently implemented. it will create a new throttled function on every state update which is bad. Thus I will be changing how the `onResize` function and how its passed in to the `onChange` handlers

Hypnosphi: Just apply it in constructor

ndelangen:
Can we debounce on requestAnimationFrame?

lib/ui/src/modules/ui/components/layout/index.js
ndelangen: I tend to use that method of writing bound class methods a lot.

ndelangen:
@wuweiweiw thank you for helping, really appreciate the efforts! I think the immediate feedback of the size is really important, if someone is targetting 800px wide for example, it's annoying to not yet fast feedback, and likely 'overshoot' multiple times.. Do you agree?

I think if we can [throttle](https://css-tricks.com/the-difference-between-throttling-and-debouncing/) on requestAnimationFrame everything should be optimal right?

lib/ui/src/modules/ui/components/layout/index.js
danielduan: for readability purposes, it might be better to name it `this.onThrottledResize` to differentiate between the two.

danielduan:
`The window.requestAnimationFrame() method tells the browser that you wish to perform an animation and requests that the browser call a specified function to update an animation before the next repaint.`
https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

From my understanding of these docs, it seems like this should be used when animating or whenever there's a constraint on browser rendering speed. In this case, I believe the `setState` updates are a constraint on CPU cycles so I'm not sure if this is an appropriate use.

I'm okay with either `throttle` or `requestAnimationFrame`. At this point, it seems like a micro optimization.

Hypnosphi:
> Can we debounce on requestAnimationFrame?

It's only about updating numbers on dimension panel, no need for 60 fps animations

wuweiweiwu:
@Hypnosphi @danielduan @ndelangen done!

I rewrote the `onResize` function so it doesn't return a function. And thus a new function wouldn't be created in `render()` every time. The `_.throttle` is set to 200ms but that can be changed :)

lib/ui/src/modules/ui/components/layout/index.js
wuweiweiwu: @Hypnosphi I also doubled checked on `master` resizing the browser window never triggered `onResize`

lib/ui/src/modules/ui/components/layout/index.js
```diff
@@ -1,7 +1,8 @@
-import { localStorage, window } from 'global';
+import { localStorage } from 'global';
 import PropTypes from 'prop-types';
 import React from 'react';
 import SplitPane from 'react-split-pane';
+import _ from 'lodash'; /* eslint-disable-line import/no-extraneous-dependencies */
```
Hypnosphi: We use per-function lodash packages. You need to add `lodash.throttle` as a dependency to `package.json` in `lib/ui`, and import it like that:

```
import throttle from 'lodash.throttle';
```

By the way, that eslint warning you've suppressed was there for a reason. If you use a package that we don't have as a direct or peer dependency, it will break on user side, because they might not have this dependency

Hypnosphi: We use per-function lodash packages. You need to add `lodash.throttle` as a dependency to `package.json` in `lib/ui`, and import it like that:

```
import throttle from 'lodash.throttle';
```

By the way, that eslint warning you've suppressed was there for a reason. If you use a package that we don't have as a direct or peer dependency, it will break on user side, because they might not have this dependency

Hypnosphi (changes requested):

lib/ui/src/modules/ui/components/layout/index.js
Hypnosphi: > I also doubled checked on `master` resizing the browser window never triggered `onResize`

Oh I see. Looks like a bug worth fixing to me

Hypnosphi: We should probably extract the part  that updates `previewPanelDimensions` from `onResize`, because it's the only thing that should be done on window resize

Hypnosphi:
> And thus a new function wouldn't be created in render() every time

Actually, it is created every time, now it's just created directly in `render`

wuweiweiwu:
@Hypnosphi My mistake. But I just got rid of the way `onResize` was returning the function that took in `size` and moved it inside render. I think that makes more sense

lib/ui/src/modules/ui/components/layout/index.js
wuweiweiwu: Fixed! :)

Hypnosphi:
The way it was before is more functional-style, the way you do it is more imperative-style. I have no strong feelings about it though

lib/ui/src/modules/ui/components/layout/index.js
wuweiweiwu: Sounds good!. I'll change that in this PR!

wuweiweiwu:
@Hypnosphi I also added the throttled event handler to window `resize` events.

Is there anything else I should do in this PR?

lib/ui/src/modules/ui/components/layout/index.js
```diff
@@ -156,20 +157,19 @@ class Layout extends React.Component {
     this.setState({ isDragging: false });
   }
 
-  onResize(pane, mode) {
-    return size => {
+  onResize(pane, mode, size) {
+    if (pane && mode && size) {
```
Hypnosphi: Looks a bit hacky to me. Let's extract the part that updates state into a separate function.

In fact, the state updates should be throttled (because they provide visual feedback), and the `saveSizes` part should be debounced (because it only makes sense to store sizes when they are final)

lib/ui/src/modules/ui/components/layout/index.js
Hypnosphi: Is it? I still see this import and `eslint-disable`

lib/ui/src/modules/ui/components/layout/index.js
wuweiweiwu: Sounds good!

lib/ui/src/modules/ui/components/layout/index.js
wuweiweiwu: Oh shoot. I think I accidentally dropped that commit when I was rebasing

lib/ui/src/modules/ui/components/layout/index.js
wuweiweiwu: @Hypnosphi seems that `size` is used to set `defaultSize` and that value is only changed when you save it to local storage. Thus I was getting some weird behavior when I was debouncing. So I opted to throttle both.

wuweiweiwu: I also made saveSize on a smaller interval for smoother user experience

wuweiweiwu:
@Hypnosphi Fixed! I think we can even get rid of the throttling of `saveSize` but I'm not sure how expensive it is to save to localStorage

wuweiweiwu:
And also it seems we can call `getSizes` in `componentDidMount` and save the current `size` in local state. And then we can use `lodash.debounce` to update the `size` in `localStorage` 
Thus we would only access `localStorage` when we actually need to instead of fetching on every render.

But I think that is out of scope for this PR and I can do it in another :)

Hypnosphi (approved):

Hypnosphi:
Cool, works like a charm now!

PR #3209, opened 2018-03-14
react native Typescript transform

cliedeman:
Issue:

## What I did
Added a flag to the react-native storybook cli to allow passing a custom config to the metro bundler

## How to test
Build the project and copied the compiled storybook-start.js, also created a sample project.

No

Yes, added

Docs added


app/react-native/src/bin/storybook-start.js
```diff
@@ -84,6 +85,11 @@ if (!program.skipPackager) {
   }
 
   let cliCommand = 'react-native start';
+
+  if (program.metroConfig) {
+    cliCommand += ` --config ${program.metroConfig}`;
```
Hypnosphi: Is it possible to make it work with haul? This variable is reassigned when using it (see below)

examples/react-native-typesript/.flowconfig
```diff
@@ -0,0 +1,54 @@
+[ignore]
```
Hypnosphi: Why do we need to configure Flow in a Typescript project?

Hypnosphi:
Please add an option for bootstrap script: https://github.com/storybooks/storybook/blob/master/scripts/bootstrap.js

examples/react-native-typesript/.flowconfig
cliedeman: part of the react-native boilerplate. will delete

cliedeman:
Bootstrap script option added

igor-dv (changes requested):
TBH, I don't see why we need to have a full react-native-typescript example as part of the monorepo.

This is just an example of the compatibility between react-native & storybook & typescript, in this manner we may have all the variations of examples with angular/vue/polymer/typescript/flow/(jquery 😓) but it will be very hard to maintain. 

So it can be 
1. A nice blog post
2. A separate repo in the organization
3. A seprate repo in your github account (referenced from the nice blog post).

cliedeman:
Hello, happy to go either route.

"Does this need a new example in the kitchen sink apps?" is a bit open-ended so I just created one anyway. 

If/when this gets merged and released I am happy to do a blog post with a sample repo. It's a lot of fanfare for what is actually just adding a metro bundler flag.

Ciaran

danielduan:
@igor-dv I think this provides us with a test of TS compatibility with RN along with a much more thorough check of the vanilla RN as well. Our current vanilla example doesn't do much at all.

danielduan (changes requested):
`examples/react-native-typesript/.gitignore`
the `typescript` in the folder path is misspelled.

we should also set it up to run in CI so that we can make sure it's working and up to date.

cliedeman:
@danielduan Thanks, renamed

@igor-dv I have added storyshots too now so it should be more useful.

@danielduan Would running the storyshots (npm test) be good enough for CI? Not sure what else we can test via CI

Ciaran

danielduan (approved):

examples/react-native-typescript/tsconfig.json
```diff
@@ -0,0 +1,58 @@
+{
```
igor-dv: please add here

```
"lerna": {
    "disabled": true
  },
```

Otherwise, this project will participate in the ts distribution

igor-dv (commented):
I am still convinced that it's an overkill for this repo.

I would rather create a separate Monorepo "storybook/use-cases" and will add there all the relevant examples. For example `storybook-with-css-modules`, `storybook-with-rn-ts` and so on. 

IMO if something is not vital for the project/brand, it should be in a separate place with its own CI and tests. Think of `webpack-contrib`, `jest-contrib` and others.

ndelangen:
I tend to agree with @igor-dv on the need for a storybooks/storybook-examples repo...

I'd like to discuss this during a roadmap meeting. We can extract it out later, so I do not find this blocking for this PR.

ndelangen (approved):

igor-dv (approved):

PR #3232, opened 2018-03-17
Pin node version to avoid error /bin/bash: yarn: command not found 

dangreenisrael:
## What I did

There is a problem with the latest version of node 8 that is causing circle builds to fail with the error
`/bin/bash: yarn: command not found`  

This pins our version of node in circle to `8.9.4`, which should work.

Issue: https://github.com/nodejs/docker-node/issues/649

## How to test
N/A

No

No

No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: /tmp/storybook
   docker:
-    - image: circleci/node:8
+    - image: circleci/node:8.9.4
 
 version: 2
 dependencies:


Hypnosphi:
What will be the criteria for unpinning it back? Technically, the problem is fixed by now, it just reproduces sometimes because of cached image

dangreenisrael:
@Hypnosphi I imagined just using this as a stopgap so that all the currently approved PRs can be merged. Then changed back once everything gets sorted out.

Just to avoid things getting out of sync

Hypnosphi:
Rerunning builds works for me. I've successfully merged a number of PRs earlier today

dangreenisrael:
Weird, ok thanks.

Hypnosphi:
Please also note that other [approved PRs]() are labeled as `feature request`, so we shouldn't merge it before branching out `release/3.4`, which will probably be done after stable `3.4.0` release on Monday

Issue #3217, opened 2018-03-15
[Angular] Dependency not resolving


MatissJanis:
### Issue details

Component dependencies are not being resolved.

Expected: `store` dependency to be resolved.
Instead, this error is thrown:
```
Can't resolve all parameters for MyComponent: (?)
```

### Steps to reproduce

Snippets below

### Please specify which version of Storybook and optionally any affected addons that you're running

- `@storybook/angular: 3.4.0-alpha.7`

### Affected platforms 

- Angular 5.2.3

### Screenshots / Screencast / Code Snippets (Optional)

```js
// Component
import { Component } from '@angular/core';
import { Store } from '@ngrx/store';

@Component({
    selector: 'my-comp',
    templateUrl: './my-comp.component.html',
})
export class MyComponent {

    constructor(
      store: Store<any>,
    ) {
        console.log(store, store === undefined); // Response: `undefined true`
    }

}
```

```js
// Story
import { storiesOf, moduleMetadata } from '@storybook/angular';
import { StoreModule } from '@ngrx/store';

import { MyComponent } from './my-component.component';

storiesOf('MyComponent', module)
    .addDecorator(moduleMetadata({
        imports: [
            StoreModule.forRoot({}),
        ],
        declarations: [
            MyComponent,
        ],
    }))
    .add('MyComponent', () => {
        return {
            component: MyComponent,
        };
    });
```

igor-dv:
I've added an example usage with the `@ngrx/store`. You can take a look [here](https://deploy-preview-3233--storybooks-angular.netlify.com/?selectedKind=ngrx%7CStore&selectedStory=With%20component&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel), it's injected.

Maybe you can try the latest 3.4.0-rc2 release ?

MatissJanis:
Hey,

Yes, the issue is solved in the newest rc version! Thanks!

Sorry for not replying faster.

PR #3233, opened 2018-03-18
Add example for @ngrx/store

igor-dv:
Issue: Angular people are using https://github.com/ngrx a lot. I think it is nice to have a dedicated section in examples with issues regarding the ngrx. 

## What I did 
Added an example for `@ngrx/store`. CC #3217


Hypnosphi:
@tmeasday it's happening again: screenshots are lost

Hypnosphi (approved):

tmeasday:
@Hypnosphi thanks! I'm pretty sure this is due to a bug that we have fixed and will be releasing soon after a little more testing.

Issue #3196, opened 2018-03-13
Make shortcuts in storybook optionable or move to addon


Kiwka:
### Issue details

On one of the components that we have in our storybook, we have several shortcuts that we want add functionality to. One of them overlaps with storybook default one `⌘ ⇧ C`, and disable us to handle this shortcut properly on the component.

My proposal is to add the possibility to enable or disable shortcuts in storybook through `addon-options`, keeping them as enabled by default.

PunitGr:
Hello @Kiwka @igor-dv 👋 

I'd love to take this up!

Does this spec seem fine to you folks?

```
setOptions({
    ...
    enableShortcuts: false, // true by default
})
```

Hypnosphi:
Sure @PunitGr go for it

Kiwka:
Thanks @PunitGr for looking into this 🙇‍♀️

Hypnosphi:
Released as [`4.0.0-alpha.0`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.0)

PR #3237, opened 2018-03-19
addon-options: Make shortcuts in storybook optional

PunitGr:
Issue: #3196 

## What I did
I've added an option for `enabling/disabling` keyboard shortcuts.

## How to test
One can try this feature by updating the storybook `config.js` as:
```javascript
setOptions({
    ...
    enableShortcuts: false, // true by default
})

This will disable the shortcuts.
```


Hypnosphi:
Should we remove shortcuts help button and popup when shortcuts are disabled?

PunitGr:
Yeah, that makes sense to disable it. Or we can add a text saying that `shortcuts` have been disabled.

PunitGr:
Hey @Hypnosphi, Removed the shortcut help button when `enableShortcuts: false`.
screen shot 2018-03-20 at 6 24 21 pm

lib/ui/src/modules/ui/components/stories_panel/index.js
```diff
@@ -84,6 +97,16 @@ StoriesPanel.propTypes = {
   openShortcutsHelp: PropTypes.func,
   name: PropTypes.string,
   url: PropTypes.string,
+  shortcutOptions: PropTypes.objectof(
```
Hypnosphi: typo: `objectOf`

But it's actually not needed here, it should be just `shape`

Hypnosphi (changes requested):

Hypnosphi (approved):

PR #3251, opened 2018-03-21
Angular: use resolveLoader from cliCommonConfig

Hypnosphi:
Issue: Angular CLI webpack config uses loaders as plain strings (e.g. `use: 'sass-loader'`). By default that means that webpack will try to resolve this loader relative to app root directory. Sometimes it happens to work, sometimes it [doesn't](https://circleci.com/gh/storybooks/storybook/43130)

To deal with it, angular CLI adds `resolveLoader` option which we should use as well


igor-dv (approved):
LGTM

PR #3252, opened 2018-03-21
Fix storyshots renderer and serializer options

theneva:
## What I did

Fixed passthrough of options from `initStoryshots`to `snapshotWithOptions`. In practice, this makes it possible to send `renderer` and `serializer` as options in the intended way (as documented in storyshots' README).

## How to test

I've updated the snapshots from `storyshot.enzyme.test.js` which uses a custom renderer (enzyme's `mount`) and serializer (`enzyme-to-json`).

Edit: CI is red due to some kind of versioning errors. I don't know what causes that, since my branch is not behind master. I'll be happy to fix it if someone points me in the right direction :)

### DIFF
--- a/addons/storyshots/README.md
+++ b/addons/storyshots/README.md
@@ -349,7 +349,7 @@ Run a custom test function for each story, rather than the default (a vanilla sn
 
 ### `renderer`
 
-Pass a custom renderer (such as enzymes `mount`) to record snapshots.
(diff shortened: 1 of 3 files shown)


igor-dv (approved):
Thanks!

theneva:
Thank you! :tada:

PR #3253, opened 2018-03-21
Addon Storysource typescript support

igor-dv:
Issue: [acorn](https://github.com/acornjs/acorn) parser doesn't support typescript out of the box. Until now all the angular stories were parsed successfully because there was no typing in them. But if you want to add a component example into the story (I think it is useful because the user will see a real example and not only a story) the parsing breaks.

## What I did
I've removed the use of acorn from the loader and used `prettier/parser-*` instead (since the addon is already dependant on prettier).

In this case for a regular javascript (which is the default), we will use  `prettier/parser-babylon` (according to the docs babylon is based on acorn)
And for TypeScript we will use `prettier/parser-typescript`.
In order to make it work with typescript, I've added a `parser` option to the loader.
```js
{
  loader: require.resolve('@storybook/addon-storysource/loader'),
  options: {
    parser: 'typescript',
  },
},
```


Hypnosphi (approved):
Cool. AFAIU it will allow parsing Flow as well (at least prettier itself does it with no problem)

igor-dv:
>  AFAIU it will allow parsing Flow as well 

Yeah. they have a `parser-flow` as well =)

https://github.com/prettier/prettier/tree/master/src/language-js

Hypnosphi:
Released as [`4.0.0-alpha.0`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.0)

Hypnosphi:
Looks like prettier doesn't consider those parsers a public API: they changed their exports in a minor release

igor-dv:
Probably we should, they might break it again. Maybe to worth porting the "generic parser" as a separate repo

PR #3259, opened 2018-03-22
Refactor all startup code into `@storybook/core` also

tmeasday:
I'm not sure why we didn't do this already @shilman? It seems an obvious thing to do.

It's a little unfortunate that RN is so different in this respect. Still I am happier making any changes in 2 places (core + RN) than 5+


tmeasday:
PS I suspect the branch name is incorrect here. This is unlikely to be the final refactor.

igor-dv:
Oh, I am ATM doing the same for a config part 😱.

danielduan (approved):

igor-dv (approved):

PR #3272, opened 2018-03-23
Remove onDeviceUI animation to support Detox screenshots

Gongreg:
Issue:

I am implementing Detox and storybook integration to do screenshot tests on React Native. There is a really funny issue with onDeviceUI. Because it is using animation, every single time animation happens the shadows are redrawn in view below the sidebar. Because of that comparing screenshots fails every time since few pixels of shadow look different. To avoid this problem I've added opt in option for animation onDeviceUIAnimated. By default it is set to true.

## What I did

Added option to disable animation (onDeviceUIAnimated=true/false).

## How to test

Could be.

Not really.

I think we need it. Currently I am not sure we have somewhere all the possible options for StorybookUI documented.

### DIFF
--- a/app/react-native/src/preview/components/OnDeviceUI/index.js
+++ b/app/react-native/src/preview/components/OnDeviceUI/index.js
@@ -2,16 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { ifIphoneX } from 'react-native-iphone-x-helper';
 
-import {
-  Animated,
-  Dimensions,
-  Easing,
-  View,
-  TouchableWithoutFeedback,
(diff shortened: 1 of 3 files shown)


danielduan:
I would actually feel comfortable just removing animations in general. It's something that's nice to have but not necessary.

danielduan (changes requested):
If you do decide to keep the animation, please document this prop somewhere.

Gongreg:
I am all in for removing animations, unnecessary opacity too.

Gongreg:
@danielduan, removed the animation.

danielduan (approved):

Issue #3271, opened 2018-03-23
when using onDeviceUI it still tries to create websocket server


Gongreg:
Even when you are using onDeviceUI storybook still tries to create websocket server.

Temporary workaround is to write these lines before rendering storybookUI:

const addons = require('@storybook/addons');
addons.setChannel({on: () => {}, emit: () => {}, removeListener: () => {}});

Proper solution would be to add option to skip websocket channel creation (maybe some people want to use both websockets and onDeviceUI at the same time, so we can't disable this functionality, we need to make it opt in).

This shouldn't be difficult to implement.

Gongreg:
@danielduan, yes. And then we will have all the options to also think how to evolve static build more, by including addons and so on :)

Gongreg:
We should also think whether this should be default option or not. I don't know many RN Storybook users so I can't comment which one should be default behaviour (crashing or skipping websockets when using onDeviceUI). Logic suggests me second option. :D

Gongreg:
@danielduan, lets bring the conversation to the PR.

Gongreg:
This PR was merged into alpha: https://github.com/storybooks/storybook/pull/3686 .

PR #3278, opened 2018-03-24
RN Disables websockets when using onDeviceUI

Gongreg:
This change will by default disable websockets when using onDeviceUI.
I've also added option enable websockets back.

Issue is #3271

There are few questions that I have:

1. Do we need the option in the first place. Is there anybody using onDeviceUI with websockets. Can we brake it for them? Should we brake it?

2. If we keep the option where it should be documented.

3. There is an increasing need to add tests for RN part. 

4. Where is resetStorybook option used at all? What does it bring. Why would we want to reset channel if it already exists?


danielduan:
Breaking this would basically disable all addons by default right? Since our addons don’t run inside the device UI?

I’m definitely in support of just removing it permanently instead of keeping it as an option. It will make maintaining a lot easier. Wondering if we can get some usage data.

Gongreg:
You are correct, if we implement is and user is using onDeviceUI the addons would be disabled (unless he specifically says useWebsockets in options.

Nonetheless the current implementation I've done in our app is to simply check whether storybook is running with simple fetch.

Something like

```
  fetch('http://localhost:7007')
	.then(() => this.setState({UI: getStorybook()}))
    .catch(() => this.setState({UI: getStorybook(..., {onDeviceUI: true})}));
```

So if server is running onDeviceUI is disabled and addons work, if server is down, it simply displays stories without any addons.

danielduan (approved):

danielduan:
Maybe we can put that fetch test as a default in the getstorybook cli template?

Gongreg:
It could be added I guess, just need to think what about port/host? Should we do separate PR for it?

danielduan:
That’s probably a better idea

shilman (changes requested):
This change makes sense to me, and I like that it's still possible to use plugins if you depend on them.

Can we document the `useWebsockets` option as part of this PR, and also document the fact that after this PR, `onDeviceUI` will make the native storybook standalone unless you specify `useWebsockets`? Since it's a big breaking change, it needs a documentation update, both in the normal docs and also in the migration guide (assuming this is a 4.0 feature)

f0rr0:
I was able to get a standalone build working without changing anything in the storybook source. I'm assuming not having a server running throws errors but they are silently consumed?

danielduan:
@f0rr0 do you have a public repo somewhere where you are running Storybook RN without the web UI?

rmevans9:
@f0rr0 It depends on platform too. I think on iOS if the websocket connection fails it just silently eats that error but Android makes a lot of noise (red boxes) if it fails.

Gongreg:
Still relevant.

Issue #3215, opened 2018-03-15
Extending webpack config fails without user-provided `module` property (storybook/vue)


mattixittam:
### Issue details

This issue is exactly as described in https://github.com/storybooks/storybook/issues/1467, only now with `@storybook/vue`. I would have tried the fix from that story, but the structure of the part where the webpack config is extended is a little more ancient in storybook-vue it seems and I couldn't quite make it work.

### Please specify which version of Storybook and optionally any affected addons that you're running

- storybook/vue 3.3.15

igor-dv:
What if you just add empty rules?

```js
{
   module: {
      rules: [],
   },
   plugins: [
     /// ,,,,
   ]
}
```

mattixittam:
Yeah, that's the workaround, but that doesn't seem like something you want a user to have to think of.

Hypnosphi:
Looks like we just need to apply https://github.com/storybooks/storybook/pull/1468 fix to other frameworks

Hypnosphi:
Fix for vue released as [`3.4.0-rc.4`](https://github.com/storybooks/storybook/releases/tag/v3.4.0-rc.4). Please do the same for angular and polymer

igor-dv:
It should've been fixed with moving server/config.js to the core

Hypnosphi:
Yes, but not for 3.4

igor-dv:
I think it can be closed already.

PR #3261, opened 2018-03-22
Move server/config to core

igor-dv:
## What I did

- Moved the server/config to core
- Deleted few deprecated things (default webpack configs, addons from app/react, head.html fallback)

TBD: RN

PR #3279, opened 2018-03-25
Fix issue when extending webpack config

benbakhar:
Issue: #3215 

## What I did

Fix bug when property lookup is performed on undefined object

## How to test

Run storybook after extending the webpack config with the following (Assumes jquery is installed):

webpack.config.js:

```
const webpack = require('webpack');

module.exports = {
  plugins: [
    new webpack.ProvidePlugin({
      $: 'jquery',
    }),
  ],
};
```

### DIFF
--- a/app/vue/src/server/config.js
+++ b/app/vue/src/server/config.js
@@ -71,7 +71,10 @@ export default function(configType, baseConfig, configDir) {
       ...config.module,
       // We need to use our and custom rules.
       ...customConfig.module,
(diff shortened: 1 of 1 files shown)


igor-dv (approved):
LGTM. Though #3261 will fix it in common

Hypnosphi:
Do angular and polymer apps need the same fix for 3.4.0?

igor-dv:
Probably they do

Issue #3223, opened 2018-03-16
Feature request: Set user agent from viewport plugin


SimenB:
### Issue details

We set a `class` on the `body` of the `iframe` so that we can style differently depending on whether we think it's a touch device or not (bigger buttons, less focus state etc). Currently this is just in an interval polling `navigator.userAgent` and passing it though bowser. This is updated if one uses the mobile emulator in Chrome.

I'd like to use the viewport plugin instead of the one built into chrome, but as the user agent remains unchanged the correct `class` is not set on `body`.

### Steps to reproduce

With the viewport plugin installed, select a mobile viewport and call `navigator.userAgent` from within the `iframe`. Observe that it's still a desktop browser.

### Please specify which version of Storybook and optionally any affected addons that you're running

- `@storybook/react 3.3.15`
- `@storybook/addon-viewport 3.3.15`

### Affected platforms 

N/A

### Screenshots / Screencast / Code Snippets (Optional)

I think having a callback every time the viewport changes would be enough for my use case.

Current implementation:

```js
function updateProbablyTouch() {
    const bowserDetect = bowser.detect(navigator.userAgent);

    const probablyTouch = Boolean(
        bowserDetect.mobile || bowserDetect.tablet
    );
    document.body.classList.toggle('probably-touch', probablyTouch);
    document.body.classList.toggle('probably-not-touch', !probablyTouch);
}
```

I don't need the actual `userAgent` to be changed, it's enough if my callback is invoked with the user agent of the viewport chosen

Hypnosphi:
Looks like a valuable feature, please open a PR

mshaaban088:
@SimenB @Hypnosphi As we have an open PR #3099 to make the viewport addon configurable, we can add that callback function to the options, let say
`onViewportChange(prevViewport, newViewport)` if you want to make comparison between the previous and new one, or just
`onViewportChange(viewport)`

and we should also include a property in every device to identify its type (e.g. `desktop/mobile/tablet`)

What do you think?

Hypnosphi:
Looks cool. I'm going to merge #3099 since we've branched off `release/3.4`

mshaaban088:
@Hypnosphi thanks :)

SimenB:
@mshaaban088 that sounds perfect to me! 🙂

mshaaban088:
Let's agree on a signature and I'm willing to work on it.

My proposal is to provide two callbacks

1. Whenever viewport changes
```javascript
/*
[...]

PR #3283, opened 2018-03-26
Viewport-addon Allow setting callback to be called whenever viewport changes

mshaaban088:
Issue: #3223 
@SimenB Please also review

## What I did
Add the ability to pass `onViewportChange` callback to `withViewport` decorator, which will be called whenever viewport changes.

### Signature
It has the following signature
```javascript
function onViewportChange( { viewport } ) {
  // React upon changes
}
```

### How to use
Either with global decorator (`recommended`), or per story

```javascript
import { withViewport } from '@storybook/addon-viewport';

addDecorator(withViewport({
  onViewportChange({ viewport }) {
    console.log('Viewport changed:', viewport);
  }
}));
```

... and below is a screenshot from the storybook

screen shot 2018-03-28 at 00 57 44

## How to test
`yarn test`

Yes

Yes and added to `Viewports`-addon stories

Yes and added

### DIFF
--- a/addons/viewport/src/shared/index.js
+++ b/addons/viewport/src/shared/index.js
@@ -3,6 +3,7 @@ export const PANEL_ID = `${ADDON_ID}/addon-panel`;
 export const UPDATE_VIEWPORT_EVENT_ID = 'addon:viewport:update';
 export const CONFIGURE_VIEWPORT_EVENT_ID = 'addon:viewport:configure';
 export const SET_STORY_DEFAULT_VIEWPORT_EVENT_ID = 'addon:viewport:setStoryDefaultViewport';
+export const VIEWPORT_CHANGED_EVENT_ID = 'addon:viewport:viewportChanged';
 export const INITIAL_VIEWPORTS = {
   responsive: {
     name: 'Responsive',
--- a/addons/viewport/src/shared/index.js
+++ b/addons/viewport/src/shared/index.js
@@ -3,7 +3,6 @@ export const PANEL_ID = `${ADDON_ID}/addon-panel`;
 export const UPDATE_VIEWPORT_EVENT_ID = 'addon:viewport:update';
 export const CONFIGURE_VIEWPORT_EVENT_ID = 'addon:viewport:configure';
 export const SET_STORY_DEFAULT_VIEWPORT_EVENT_ID = 'addon:viewport:setStoryDefaultViewport';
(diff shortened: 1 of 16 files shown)


Hypnosphi (approved):

Hypnosphi:
Please fix unit tests

mshaaban088:
@Hypnosphi thanks for reviewing!

I'm going to fix/add old/new tests probably today

mshaaban088:
@Hypnosphi I've fixed old tests, added new ones, added stories, and updated documentation.

Please take a look.

P.S there is something weird with `danger`, could you also please check it?

Hypnosphi:
It says "PR is marked with "in progress" label." =)

mshaaban088:
facepalm

addons/viewport/src/shared/index.js
```diff
@@ -15,55 +15,63 @@ export const INITIAL_VIEWPORTS = {
       margin: '0',
       boxShadow: 'none',
     },
+    type: 'desktop',
```
Hypnosphi: Is this property documented anywhere?

examples/official-storybook/stories/addon-viewport.stories.js
```diff
@@ -18,10 +20,26 @@ storiesOf('Addons|Viewport.Custom Default (Kindle Fire 2)', module)
       I've inherited <b>Kindle Fire 2</b> viewport from my parent.
     </Panel>
   ))
-  .add('Overridden', () => (
-    <Viewport name="iphone6">
```
Hypnosphi: Does `Viewport` still work? If so, let's keep a story using it

Hypnosphi (changes requested):

addons/viewport/src/shared/index.js
mshaaban088: AFAIK `Viewport` definition is not documented, so it's a good idea to start doing so

examples/official-storybook/stories/addon-viewport.stories.js
mshaaban088: Yeah, it makes sense

addons/viewport/src/shared/index.js
mshaaban088: Done

examples/official-storybook/stories/addon-viewport.stories.js
mshaaban088: Done

Hypnosphi (approved):

PR #3295, opened 2018-03-26
remove confusing drop shadow from info body sections

danielduan:
Issue: https://github.com/storybooks/storybook/issues/1877

Drop shadow creates strange look in Info inline mode

## What I did

Removed it

## How to test

Check Netlify

Y, screenshots may need to be updated
N
N


Hypnosphi:
AFAIU the main problem is the splitting not the shadow

danielduan:
Yes, this only makes the problem a little less bad.

I think the inline info mode needs to be redesigned in general, but that's more of a roadmap thing.

Hypnosphi (approved):

Hypnosphi:
Please update storyshots

PR #3306, opened 2018-03-28
Update react-native symlink resolving and add support for flow

meros:
Issue:

The symlink resolving for react-native does not work all the time. react-native tools moved over to a new way of resolving symlinks last year: https://github.com/facebook/react-native/commit/39f83c4a12b8350e331c3a6c575a2cab64950d7a#diff-d939fe299a50129184bf6b4208f6d497

## What I did
I moved over to the new symlink method according to the react-native PR

## How to test
Run storybook on a project depending on external projects in a workspace monorepo. Make sure roots are all there (just like for react-native start) and storybook is the first root.

No

No

No


danielduan (approved):
looks reasonable. we're adding flow support as a default here too but that's fine by me.

@meros please fix the lint errors, I think you can ignore the global requires with a `// eslint-disable-line global-require`

danielduan (changes requested):
looks reasonable. we're adding flow support as a default here too but that's fine by me.

@meros please fix the lint errors, I think you can ignore the global requires with a // eslint-disable-line global-require

meros:
Done

danielduan (approved):

PR #3316, opened 2018-03-29
Split vendor and runtime chunk in static builds

Hypnosphi:
See https://gist.github.com/sokra/1522d586b8e5c0f5072d7565c2bee693


danielduan (approved):
looks like you'll need to add it for Mathril too.

Hypnosphi:
@danielduan Already done =)

Issue #3317, opened 2018-03-29
Remove usages of deprecated React lifecycle methods


Hypnosphi:
### Work summary

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html
We need to add [`react-lifecycles-compat`](https://github.com/reactjs/react-lifecycles-compat) dependency to use new [`static getDerivedStateToProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) method

### Acceptance criteria

`cWM`, `cWRP`, and `cWU` shouldn't be used in our codebase.
Unfortunately, `eslint-plugin-react` doesn't have a rule for disabling particular methods (see [discussion](https://github.com/yannickcr/eslint-plugin-react/issues/1147#issuecomment-373961468)). So I'll need to check it manually. I'll also investigate how we could utilize [`StrictMode`](https://reactjs.org/docs/strict-mode.html) to detect violations.

### Who to contact

Me

PR #3327, opened 2018-03-31
Remove usages of deprecated React lifecycle methods

Hypnosphi:
Issue: #3317 

## What I did
Added [`StrictMode`](https://reactjs.org/docs/strict-mode.html) to all our example apps, fixed violations on our side, and opened issues for third-party libraries:
https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+author%3AHypnosphi+archived%3Afalse+lifecycle


app/react/package.json
```diff
@@ -67,7 +67,7 @@
   "peerDependencies": {
     "babel-core": "^6.26.0 || ^7.0.0-0",
     "babel-runtime": ">=6.0.0",
-    "react": ">=15.0.0 || ^16.0.0",
-    "react-dom": ">=15.0.0 || ^16.0.0"
+    "react": ">=15.0.0",
+    "react-dom": ">=15.0.0"
```
tmeasday: Should this be `<18.0.0` as well or something? Do we just not set an upper limit on the version range so we work for as long as possible?

app/react/src/client/preview/render.js
```diff
@@ -19,12 +19,19 @@ if (isBrowser) {
   rootEl = document.getElementById('root');
 }
 
+function render(node, el) {
+  ReactDOM.render(
+    process.env.STORYBOOK_EXAMPLE_APP ? <React.StrictMode>{node}</React.StrictMode> : node,
```
tmeasday: Maybe we should add this as a setting so users can also enforce strictness in their storybooks? Either `yarn storybook --strict` or `import { setStrict } from '@storybook/react'; setStrict(true)`.

I'd probably lean toward the second as it is react only.

tmeasday (commented):
A couple of comments/suggestions

app/react/src/client/preview/render.js
Hypnosphi: It's quite trivial to add on user side:
```
addDecorator(story => <StrictMode>{story()}<StrictMode>)
```

app/react/package.json
Hypnosphi: The whole point of this PR is to make storybook future-compatible. So no, I don't think, we should add new restrictions in it (note that `>=15.0.0` part was there before, I just removed the redundant part)

app/react/src/client/preview/render.js
tmeasday: Oh, fair enough!

app/react/package.json
tmeasday: It’s tricky though, right? I mean sort of by definition react v18 may very well have some incompatibility with what we are doing. 

So we just leave it to the user (oh SB breaks with react 18, maybe I should try updating it) rather than the package system?

Hypnosphi: Peer dependencies should be as wide as possible. We shouldn't say we're incompatible with something unless we know it for sure. And React tries hard to make major releases breaking only for those projects that had some unfixed deprecation warnings at the moment of release

Hypnosphi: Actually you can read the lack of upper limit here as our contract to support next major React releases before they become stable. For 16, it was more than half a year between first alpha and stable release, so I think we can afford such a contract

tmeasday: > contract to support next major React releases before they become stable

Right, but this *particular* version of Storybook is unlikely (maybe) to support react@18. No doubt we will try to be on top of it and get a version of SB out well before react@18 that does support it, but it won't be this version.

tmeasday: So when we do ship the (theoretical) changes that support react@18 in storybook 4.x.y we won't actually change the `package.json` to indicate that and users won't be able to tell from the output of `npm`/`yarn` that they should update to sb@4.x.y if they want to use react@18.

tmeasday: Anyway, I don't know if it is a big deal as it's not like peer dependencies even really work properly and it all requires users to actually read changelogs/interpret esoteric errors anyway as a result, so I am not saying we have to do it. But it feels more correct.

Hypnosphi: Ok I think I got the idea, but I think it's out of scope of this PR. Feel free to open another one, and note that the next major version is 17

tmeasday (approved):

Issue #2383, opened 2017-11-27
Background does not persist when changing between story scenarios


clementallen:
When using a custom background, it resets to the default when changing to a different story scenario.  I'm not sure if this is deliberate or not but it seems like it would be more convenient to persist the background when changing.

Demo:
![storybook-background-bug](https://user-images.githubusercontent.com/6305750/33271665-d030eb68-d37f-11e7-9cda-b8c863bd95f7.gif)

wuweiweiwu:
@Hypnosphi I can pick this up :)

Hypnosphi:
Sure, go for it

Hypnosphi:
Released as [`4.0.0-alpha.2`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.2)

PR #3331, opened 2018-04-01
Persist background for @addon/background

wuweiweiwu:
Issue: #2383 

## What I did
don't unset background url when changing stories.

## How to test

yes
no
no

### DIFF
--- a/addons/background/src/BackgroundPanel.js
+++ b/addons/background/src/BackgroundPanel.js
@@ -72,11 +72,11 @@ export default class BackgroundPanel extends Component {
       this.channel = addons.getChannel();
     }
 
-    this.state = { backgrounds: [], currentBackground: null };
+    this.state = { backgrounds: [] };
 
     this.channel.on('background-set', backgrounds => {
       this.setState({ backgrounds });
-      const currentBackground = api.getQueryParam('background') || this.state.currentBackground;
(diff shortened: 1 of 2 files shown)


addons/background/src/BackgroundPanel.js
```diff
@@ -72,11 +72,11 @@ export default class BackgroundPanel extends Component {
       this.channel = addons.getChannel();
     }
 
-    this.state = { backgrounds: [] };
+    this.state = { backgrounds: [], currentBackground: null };
 
     this.channel.on('background-set', backgrounds => {
       this.setState({ backgrounds });
-      const currentBackground = api.getQueryParam('background');
+      const currentBackground = api.getQueryParam('background') || this.state.currentBackground;
```
Hypnosphi: I'd prefer URL to be the single source of truth there. We already persist background changes in URL, let's just stop resetting that on story change:
https://github.com/wuweiweiwu/storybook/blob/12c5af9e513f989e03db87475811b86a4cdd718a/addons/background/src/BackgroundPanel.js#L91

wuweiweiwu: I see! Will update :)

wuweiweiwu:
@Hypnosphi Done! :)

Hypnosphi (approved):

PR #3337, opened 2018-04-01
Mobile device view: toggling stories panel with ☰ button 

purrden:
Issue:

## What I did
* adapted Header component  to have ☰ button when on mobile device
* make stories-panel toggled by clicking ☰ button
* added helper to detect whether the storybook is open on mobile device (this is done through userAgent of window)
* adjusted styles 

Now on mobile devices the layout looks something like this when storybook is rendered:
image

## How to test
Open storybook on mobile device

yes (I added some new tests)
no
not sure


lib/ui/src/modules/ui/libs/is_mobile_device.js
```diff
@@ -0,0 +1,7 @@
+export default userAgent => {
+  if (/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(userAgent)) {
```
Hypnosphi: Do we really need javascript for this? Can we just show/hide things using media queries instead?

Hypnosphi (changes requested):

lib/ui/src/modules/ui/libs/is_mobile_device.js
Hypnosphi: Right now, I can't test it using emulator from devTools

denzenin: I did it to avoid changes to layout in general.
Depending on whether it is mobile device, the header is rendered outside of stories-panel as a standalone component and visibility of stories-panel is just toggled with the existing mechanism. 
Plus for stories-panel as well. To hide it initially for mobile devices (in initial state) :)

Hypnosphi: OK makes sense

lib/ui/src/modules/ui/components/header.js
```diff
@@ -33,44 +33,58 @@ const shortcutIconStyle = {
...
+    {isMobileDevice && (
+      <button style={burgerIconStyle} onClick={onBurgerButtonClick}>
+        ☰
+      </button>
+    )}
+    <a style={linkStyle(isMobileDevice)} href={url} target="_blank" rel="noopener noreferrer">
       <h3 style={headingStyle}>{name}</h3>
     </a>
-    <button style={shortcutIconStyle} onClick={openShortcutsHelp}>
+    <button style={iconStyle(isMobileDevice)} onClick={openShortcutsHelp}>
```
Hypnosphi: Looks like this button is useless on mobiles, it shows keyboard shortcuts

lib/ui/src/modules/ui/components/layout/index.js
```diff
@@ -98,7 +98,11 @@ const defaultSizes = {
   },
 };
 
-const saveSizes = sizes => {
+const saveSizes = (sizes, isMobileDevice) => {
+  if (isMobileDevice) {
```
Hypnosphi: Please add some inline comment explaining why do we need to skip saving here

lib/ui/src/modules/ui/components/layout/index.js
```diff
@@ -177,13 +185,15 @@ class Layout extends React.Component {
...
       showAddonPanel,
-      addonPanelInRight,
       addonPanel,
       storiesPanel,
       preview,
+      isMobileDevice,
     } = this.props;
     const { previewPanelDimensions } = this.state;
 
+    const addonPanelInRight = isMobileDevice ? 0 : this.props.addonPanelInRight;
```
Hypnosphi: Why number not boolean?

Hypnosphi:
It breaks in a weird manner when I try to resize stories panel on mobile: https://deploy-preview-3337--storybooks-official.netlify.com/?selectedKind=ui%2FAddonPanel&selectedStory=default&full=0&addons=1&stories=0&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

lib/ui/src/modules/ui/components/header.js
ndelangen: We can actually make those things clickable and perform the action.. but once the left panel is gone (either by toggling it, or by going full-screen, you can't go back)

ndelangen:
@denzenin Thank you for this amazing first PR, thank you!

lib/ui/src/modules/ui/components/header.js
Primigenus: Note that some people use USB keyboards with mobile devices, so I wouldn't necessarily just remove it because of mobile.

lib/ui/src/modules/ui/components/header.js
```diff
@@ -33,44 +33,58 @@ const shortcutIconStyle = {
...
-};
+});
 
-const Header = ({ openShortcutsHelp, name, url }) => (
-  <div style={wrapperStyle}>
-    <a style={linkStyle} href={url} target="_blank" rel="noopener noreferrer">
+const Header = ({ openShortcutsHelp, onBurgerButtonClick, name, url, isMobileDevice }) => (
+  <div style={wrapperStyle(isMobileDevice)}>
+    {isMobileDevice && (
+      <button style={burgerIconStyle} onClick={onBurgerButtonClick}>
```
igor-dv: Why don't just show this button always? I don't see any cons to it in a web mode, and having it in both modes will reduce complexity.

lib/ui/src/modules/ui/components/layout/index.js
igor-dv: I see two ways of refactoring this part:

1. 

Extract the `saveSizes` and `getSavedSizes` to a separate file and do not change their functionality.
Change only the Layout component to be coupled to the `isMobileDevice`. 

2. 

As an alternative, you can store mobile sizes in a sperate entry of the storage.

denzenin: @igor-dv I like the second option

denzenin:
@Hypnosphi regarding breaking on resize. yes, it is weird indeed. It seems that resizing of SplitPane works far from smooth on mobile. Tried it on master branch. Even when I had succeeded to resize story-panel I could resize it again (after first resize) only once in a bunch of tries.

lib/ui/src/modules/ui/components/layout/index.js
denzenin: missed it, thanks :)

lib/ui/src/modules/ui/components/header.js
denzenin: for me, when I see a ☰ button on mobile websites, I expect it to "always be there for me" :) 
in this particular case, if we leave it on Desktop version and have the same behaviour (toggling story-panel, it will be hidden once story-panel is hidden)

but I have no issue with keeping the button, so if you advise to get rid of the check above – I'll remove it

denzenin:
@ndelangen thank you for the kind words :)

lib/ui/src/modules/ui/components/layout/index.js
Hypnosphi: Sorry, I don't really get the point of different entry names. Different devices cannot share the same `localStorage` AFAIK

denzenin: @Hypnosphi  initially I avoided saving dimensions on mobile as the resizing worked not perfectly when I firstly tried it (I did not face issues that you have found later, but still resizing seemed to be a hard task to perform. So I wanted to avoid situations when you somehow resized the pane and did not like the results but the new unwanted sizes are saved in localStorage, so you cannot reset the layout by refreshing the page.

Maybe I misunderstood Igor's comment above about separation of saved dimensions, if it is not needed, but we still want to save sizes on mobile – let's just drop the changes and leave this part as it is in master right now?

igor-dv: I talked more about the browser emulators. I think in this case the storage is shared.

lib/ui/src/modules/ui/components/header.js
igor-dv: Maybe we can have something like this when it's closed (in both modes):

image

The floating button (with opacity). It will open the stories panel with the header in it. This way we can reuse most of the UI between web and mweb.

denzenin: I assume there is a disadvantage for mobile view with floating button. In case of this floating element you will have to always keep it in mind when writing stories of the components. We would need to add some additional containers/spacing in stories to test the component without visual/touch interference with floating element. In my opinion header is OK and quite simple solution for this.

igor-dv: 👌

ndelangen: @igor-dv I prefer this:
https://user-images.githubusercontent.com/3757759/38176267-ac776684-35f4-11e8-9bd0-2c486937f02e.png

both visually and UX, you should not overlay anything over the preview.

Hypnosphi:
So should we maybe disable resizing on mobile at all?

lib/ui/src/modules/ui/libs/is_mobile_device.js
danielduan: I'm in the media query camp as well. If on desktop, a user scales the window smaller so they can code and render in the browser side by side, they should be able to use the mobile view as well.

denzenin: @danielduan with current version of Storybook there is nothing stopping users from hiding stories-panel on desktop (with shortcuts) and then resize the browser window as it is needed for your particular case. I want to emphasize that this pull request solves the problem with testing components on physical mobile devices :) 
And I tried to affect as little of desktop version as possible on purpose. Cause this is a whole separate task related to layouts in general.

danielduan: I understand that you don't want to expand the scope of this change so this sounds fine with me.

It's still a broader issue that we'll need to address at some point.

ndelangen: Let's take it 1 step at a time? I don't fancy userAgent sniffing either, but it does solve an issue and down-scopes this PR well.

Let's improve the manager head for desktop after.

ndelangen (approved):

denzenin:
Good day to all, could you please advise what changes should I make to my pull request to move it forward? Maybe one step at a time but still I need to know those steps :)

Hypnosphi:
What do you think about this?

> should we maybe disable resizing on mobile at all?

You also need to resolve conflicts, feel free to ask for help if you need one

Hypnosphi:
It would be more explicit if we would just disable handlers

Hypnosphi (approved):

lib/ui/src/modules/ui/components/layout/index.js
```diff
@@ -130,7 +130,7 @@ class Layout extends React.Component {
   constructor(props) {
     super(props);
 
-    this.layerSizes = getSavedSizes(defaultSizes);
+    this.layerSizes = getSavedSizes(defaultSizes, props.isMobileDevice);
```
Hypnosphi: Looks like this second argument isn't used anymore

ndelangen:
Awesome work @denzenin ! 💪

lib/ui/src/modules/ui/components/layout/index.js
denzenin: yeah, you are right ) please advise how could I clean it up?

denzenin:
yay! thank you all for your help and support :)

Hypnosphi:
@Aarbel try latest alpha, `4.0.0-alpha.20`

Issue #2729, opened 2018-01-12
LinkTo saves scroll position


nemanjacvejic:
### Issue details

After clicking on LinkTo component it takes me to right story, but content is scrolled. The page saves scroll position from previous story.
### Version
 @storybook/react 3.3.7
 @storybook/addon-links 3.3.7

Hypnosphi:
Fix released as [`4.0.0-alpha.2`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.2)

PR #3338, opened 2018-04-01
Scroll to top of the page when changing story

Hypnosphi:
Issue: #2729

### DIFF
--- a/lib/core/src/client/preview/start.js
+++ b/lib/core/src/client/preview/start.js
@@ -1,6 +1,6 @@
 import { createStore } from 'redux';
 import addons from '@storybook/addons';
-import { navigator, window } from 'global';
+import { navigator, window, document } from 'global';
 import createChannel from '@storybook/channel-postmessage';
 import { handleKeyboardShortcuts } from '@storybook/ui/dist/libs/key_events';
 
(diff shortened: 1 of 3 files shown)


lib/core/src/client/preview/start.js
```diff
@@ -52,10 +52,20 @@ export default function start(render, { decorateStory } = {}) {
...
 
+  let prevKind;
+  let prevStory;
+
   // initialize the UI
   const renderUI = () => {
     if (isBrowser) {
       render(context);
+      const { selectedKind, selectedStory } = reduxStore.getState();
+      if (selectedKind !== prevKind || selectedStory !== prevStory) {
```
tmeasday: Perhaps we shouldn't call into `render()` at all it if they are the same. I'm pretty sure all frameworks end up making this same check (although Angular ends up re-rendering, it is not clear why)

tmeasday (approved):
Happy to approve this, what do you think about the suggestion? Could be a separate PR

lib/core/src/client/preview/start.js
Hypnosphi: Makes sense
@igor-dv @alterx WDYT?

Hypnosphi: There can be also `state.error`

tmeasday: Yes, good point. This is something I've wanted to refactor out anyway, I will get it in a separate PR.

Issue #3348, opened 2018-04-03
HMR with RN Story of "..." named "..." already exists


duro:
If you are reporting a bug or requesting support, start here:
### Bug or support request summary

I have reviewed issue #1581 but do not seem to be having the same problem mentioned there. I am certain that I have passed `module` to my `storiesOf` call.

When HMR is enabled, and I save the file, I get an error:

```
Story of "..." named "..." already exists
```

### Steps to reproduce

1. Setup a simple react-native storybook
2. Launch app
3. Save file

### Please specify which version of Storybook and optionally any affected addons that you're running

- storybook/react-native: ^3.4.0"
- storybook/addon-actions: "^3.4.0"
- storybook/addon-links: "^3.4.0"
- storybook/addons: "^3.4.0"

### Affected platforms 

- React Native v25.0.0

### Screenshots / Screencast / Code Snippets

```js
import React from 'react'
import { Text } from 'react-native'

import { storiesOf } from '@storybook/react-native'
import { action } from '@storybook/addon-actions'
import { linkTo } from '@storybook/addon-links'

storiesOf('ExploreCard', module).add('default', () => <Text>Hi!</Text>)

```

### Additional Info

One thing I noticed is that when I console.log out the `m` var here: https://github.com/storybooks/storybook/blob/master/lib/core/src/client/preview/client_api.js#L52

The object looks like this:

```
Object {
  "exports": Object {},
  "hot": Object {
    "accept": [Function accept],
    "acceptCallback": null,
  },
}
```

So it does not have a `.dispose()` method. Is this a change in the HMR spec?

duro:
I noticed this very old issue in the react-native repo whis is the only other mention I have seen of `hot.dispose` being missing:

https://github.com/facebook/react-native/issues/8558

duro:
I was able to find a work around for this by modding like so:

```
if (m && m.hot) {
      if (m.hot.dispose) {
        m.hot.dispose(() => {
          this._storyStore.removeStoryKind(kind);
          this._storyStore.incrementRevision();
        });
      } else {
        this._storyStore.removeStoryKind(kind);
        this._storyStore.incrementRevision();
      }
    }
```

https://github.com/storybooks/storybook/blob/master/lib/core/src/client/preview/client_api.js#L52

I'm going to try and work this up as a PR

duro:
[...]

PR #3351, opened 2018-04-03
Handle the fact that RN does not expose a dispose method on hot module

duro:
Issue: #3348 

## What I did

I added a check for if the HMR module has a `.dispose` method and if not, just calls `.removeStoryKind` and `.incrementRevision` directly

## How to test

> It is, but I could use some help here. I tried simply implementing a new mock of `MockModule` and then looping over existing spec with new mock, as well as old mock, but got some errors:

```
Expected value to equal:
      []
    Received:
      [{"fileName": undefined, "kind": "kind", "stories": [{"name": "story", "render": [Function anonymous]}]}]

    Difference:

    - Expected
    + Received

    - Array []
    + Array [
    +   Object {
    +     "fileName": undefined,
    +     "kind": "kind",
    +     "stories": Array [
    +       Object {
    +         "name": "story",
    +         "render": [Function anonymous],
    +       },
    +     ],
    +   },
    + ]
```

I'm not familiar enough with the core code to dig deeper into this testing error.

### DIFF
--- a/lib/core/src/client/preview/client_api.js
+++ b/lib/core/src/client/preview/client_api.js
@@ -49,11 +49,16 @@ export default class ClientApi {
       );
     }
 
-    if (m && m.hot && m.hot.dispose) {
-      m.hot.dispose(() => {
+    if (m && m.hot) {
+      if (m.hot.dispose) {
+        m.hot.dispose(() => {
+          this._storyStore.removeStoryKind(kind);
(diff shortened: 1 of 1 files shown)


duro:
This can be closed. See this comment for reason:

https://github.com/storybooks/storybook/issues/3348#issuecomment-378441828

Issue #3329, opened 2018-04-01
Importing Meteor packages (VulcanJS integration)


SachaG:
A lot of [VulcanJS](http://http://vulcanjs.org/) would love to be able to use Storybook, but the problem is that Vulcan relies a lot on Meteor packages, which if I'm not mistaken cannot currently be imported inside Storybook. 

And even if we could get rid of most exports in Vulcan component files and end up with "pure" React files, there is still the issue that Vulcan uses a global component system, where instead of doing `import MyComponent`, you can just write `<Components.MyComponent/>` (this is in order to enable theming among other things). 

I know there's probably no easy workaround currently, but I would love to start a discussion about how to get Storybook working with Vulcan, even if it requires us changing some thing internally?

SachaG:
See related discussion here: https://github.com/VulcanJS/Vulcan/issues/1939

Hypnosphi:
FWIW, working with globals shouldn't be a problem

Hypnosphi:
>  cannot currently be imported inside Storybook

What's exactly the problem with that?

tmeasday:
The problem is that Meteor projects are full of things like `import X from 'meteor/X'` where `meteor/X` resolves (via Meteor's build system, not webpack) to an export from a Meteor package. 

There are some projects that allow you to use Meteor in webpack. I'm not sure how well they work. That would probably be the place to start @SachaG.

tmeasday:
On the globals thing, how does `Components.MyComponent` get registered? Is it auto imported by Meteor (i.e. from outside the `imports/` folder)? It might not be possible to write a webpack `resolve.context` that simulates what Meteor does here? I'm not sure

Hypnosphi:
So is it possible to build some assets with Meteor and import them in `.storybook/config.js` maybe?

tmeasday:
Well as long as webpack can interpret the imports, the global variables won't be a problem @SachaG

eric-burel:
My idea was to replace Node by Meteor when running tools as Storybook (but also jest and whatever 3rd party tool we use), I don't know if it is doable? That's how we work usually with Electron, this way we can support native modules compiled for Electron for example.

jefflau:
So it sounds like there are two realistic options from what I can see:

1. Replace Meteor to another Node Framework within Vulcan so we are totally npm based and webpack compatible.
2. Extend the webpack build system for storybook so it can resolve Meteor imports.

[...]

PR #3360, opened 2018-04-05
Embed the preview context in the user's app

tmeasday:
Issue: #3329

This is proof-of-concept of an idea to move the preview iframe inside the user's app.

The benefits of this are:
  1. The user does not need to configure webpack a second time for storybook.
  2. If the user's app doesn't run inside webpack, they can use storybook (e.g. Meteor apps, #3329)
  3. If they are running storybook alongside their app anyway, this avoids compiling changed components twice for every change.

The downsides I can see are:
 1. You need to change the code of your app to render the stories (And this might be a little tricky to get right so it doesn't affect your production bundle).
 2. You need to run both storybook and your app (which most do anyway I imagine).
 3. We lose the ability to control the way the stories are bundled, potentially hurting future plans to build small bundles for each storybook entrypoint (? cc @ndelangen)

I'm not necessarily suggesting we would *only* allow this way to do it, but just that we might offer it as an option for Meteor (& maybe other bundler) users, and potentially because it's an easier onboarding path for new storybook users.

## What I did

 - Changed the `@storybook/core` webpack config to not build the preview endpoint
 - Changed the `@storybook/core` provider to render http://localhost:3000/__storybook_preview__
 - Changed `@storybook/react` to (hackily) render inside an exported `StorybookPreview` component.

 - Changed the `cra-kitchen-sink` app to render the `StorybookPreview` if the url path is set to `__storybook_preview__`

 - Added a `meteor-react` example app that does the same.

## How to test

`yarn && yarn bootstrap --core` as usual.

### To test `storybook-cra`

Run `yarn storybook` and `yarn start`, browse to the storybook as usual. Notice the iframe points at the actual app.

### To test `meteor-react`

As meteor [doesn't work](https://github.com/meteor/meteor/issues/9000) with yarn workspaces, you'll need to `yarn link` both `@storybook/core` and `@storybook/react` into the app, then `yarn storybook` and run `meteor`, then browse to the storybook.

You can do:

```
meteor npm install --global yarn
meteor yarn
# after yarn linking in the relevant dirs
meteor yarn link @storybook/core
meteor yarn link @storybook/react

meteor yarn storybook
# and
meteor
```


ndelangen:
@tmeasday I think the most effective way to explain the concept is to have a this set as a topic on the next roadmap session. So you can explain the reasoning, strategy and implementation.

ndelangen:
I get this now, and will work on something that's related when I start the docs-view work.

Do you want to keep this PR open @tmeasday ?

PR #3373, opened 2018-04-06
Use per-story parameters in Notes addon

tmeasday:
Now we have per-story parameters, we can change the addon APIs to be more ergonomic via the use of parameters.

This PR is a first pass at changing the notes addon to use a parameter named `notes`. I want to put it out there to try and flush out any problems with the approach and get feedback before following up with changing the other addons.

### New API:

Instead of:

```js
storiesOf(..)
  .add(
    'using decorator arguments, withNotes',
    withNotes('Notes into withNotes')(() => <Story/>)
  );
```

We now do

```js
storiesOf(..)
  .add('withNotes', () => <Story/>, {
    notes:
      'This is the notes for a button. This is helpful for adding details about a story in a separate panel.',
  })
```

### To test

See the `official-storybook`.


Hypnosphi:
Wow, so cool! Can we do the same for `info`?

addons/notes/README.md
```diff
@@ -29,16 +29,24 @@ Add following content to it:
 import '@storybook/addon-notes/register';
 ```
 
-Then write your stories like this:
+Then add the `withNotes` decorator to all stories in your `config.js`:
 
 ```js
-import { storiesOf } from '@storybook/react';
+import { configure, addDecorator } from '@storybook/react`; // <- whichever storybook version you use
```
Hypnosphi: The "version" word is a bit confusing here =)

Hypnosphi (approved):

addons/notes/src/index.js
```diff
@@ -3,25 +3,49 @@ import addons from '@storybook/addons';
...
+
+const decorator = options => {
   const channel = addons.getChannel();
-  return getStory => context => {
-    marked.setOptions({ ...marked.defaults, options });
-    // send the notes to the channel before the story is rendered
-    channel.emit('storybook/notes/add_notes', marked(text));
+  return (getStory, context) => {
+    const { parameters: { notes } } = context;
+    console.log({ options, getStory, context, notes });
```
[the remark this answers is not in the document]
tmeasday: 😬

tmeasday:
> Can we do the same for info?

Yes, definitely, as well as some others.

I just wanted to get this out there and find out what people thought. For instance, are we OK with addons just claiming parameter namespaces (`notes` in this case)? Obviously there is scope for collision, but I am guessing people can be sensible about it.

Any other objections?

If not I will put a few utilities together that make this easier to do in each case and do the obvious things in the other addons.

tmeasday:
Would like to see some explicit 👍 from some other interested folks before moving forward here.

danielduan (approved):
Interesting. It's not that different from what we used to be doing so I wouldn't call this an API revamp by any means. It is slightly better so I definitely welcome this tweak.

I think for an API v2, we should think about an API that resembles more like the content being rendered. I'm not a huge fan of these multiple function parameters in the `.add()` variants. The chaining of functions is also a bit bizarre `withNotes()()`. The way we declare `storiesOf()` is also a bit strange, and its also easy to miss the `module` parameter at the end as well.

Looks fine to me.

examples/official-storybook/stories/addon-notes.stories.js
```diff
@@ -37,10 +26,20 @@ storiesOf('Addons|Notes', module)
     ))
   )
 ~~~
-    `)(() => (
-      <BaseButton label="Button with notes from inline github-flavored markdown - check the notes panel for details" />
-    ))
-  )
+    `;
+
+storiesOf('Addons|Notes', module)
```
igor-dv: Please add these example to the other frameworks too

igor-dv (approved):
Very cool.

If this PR is targeting the v4, probably, it worth removing unneeded exports.

tmeasday:
> it worth removing unneeded exports.

@igor-dv do you mean the old `WithNotes` component (slated for removal in 4.0), or the support for `withNotes('note')(story)` / `withMarkdownNotes('note')(story)`? As much as 4.0 is a breaking release, I think we should have a period of deprecation..

Either we deprecate in an upcoming 3.4.x release (in which case this all needs to be backported to 3.4, probably not a great plan), or we leave the current API around (marked deprecated) until 5.0. I'm OK with that, really.

examples/official-storybook/stories/addon-notes.stories.js
tmeasday: Yep!

tmeasday:
OK, I'll call that a quorum. Will finish off this PR then follow it up with similar changes to other addons.

Hypnosphi:
> we should think about an API that resembles more like the content being rendered

You should probably take a look at https://github.com/storybooks/SBNext/blob/POC-bundler/examples/react/in/button/button.example.js

tmeasday:
@Hypnosphi / @danielduan  I expect the concept of parameters to come through to the new API however---people will still need a way to annotate their "pure" examples with instructions for SB/documentation. 

This isn't the place to discuss the exact API but one idea might be to add a property to the exported function:

```js
export const example1 = () => <Component>with children</Component>;

example1.storyParameters = {
  notes: 'This is the story note that will appear in the storybook UI'
}
```

tmeasday:
Updated the remaining app layers and addressed the other things on this ticket.

tmeasday:
Also removed `WithNotes` (old currently-deprecated API).

Next up I will factor out a utility to make this easy to generalize and deprecate the current API as part of that.

Hypnosphi (approved):

PR #3378, opened 2018-04-06
Change addon panel keyboard shortcut to shift-meta-z

jcharry:
Issue:
Chrome and Safari both use `SHIFT-CMD-C` to toggle the element inspector in dev-tools. The current hotkey for toggling the Addons Panel clobbers the native browser shortcut. `SHIFT-CMD-Z` appears to be unused in chrome, firefox, and safari, so should be a safe alternative.

Please tag your pull request with at least one of the following:
`["feature request"]`

### DIFF
--- a/lib/ui/src/libs/key_events.js
+++ b/lib/ui/src/libs/key_events.js
@@ -41,7 +41,7 @@ export default function handle(e) {
     case keycode('F'):
       e.preventDefault();
       return features.FULLSCREEN;
-    case keycode('C'):
+    case keycode('Z'):
     // backward-compatibility
     case keycode('D'):
       e.preventDefault();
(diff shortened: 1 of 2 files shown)


danielduan (approved):

Hypnosphi:
I would prefer to switch to single key shortcuts, there's a related discussion in some other issue

danielduan:
@Hypnosphi let's create a new issue for that? I think it makes sense, but for now, we can merge and fix the bug until we refactor?

Hypnosphi:
Ok, I have no solid objections. The [table](https://github.com/storybooks/storybook/issues/403#issuecomment-326461268) I made says that FF uses `cmd + shift + Z` for something but right now I look at their [docs](https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly) and can't find it

PR #3383, opened 2018-04-08
Enforce addons store being a singleton by storing it in global variable

Hypnosphi:
This should be the ultimate fix for #1981, hopefully


tmeasday:
I think this is good.

I want to register the downside of this approach (don't try and reconcile dependencies, just let the "first in" win), which is similar to using a `require.alias` to force a react version: If we ever release a new version of the addons API, we can't use the module install system to tell the user what's going on.  (npm/yarn will happily install a variety of versions of `@storybook/addons` without even knowing this might not be what the user wants.

So in the future if we wanted to release `@storybook/addons@2.0` we'd have to build our own adhoc dependency system: "You seem to have addons@1 and addons@2 installed, this is no good, please go search your node_modules".

Having said that, given that yarn/npm don't actually help developers at all with the "1 only absolutely want one version of this package in my dependency tree" problem (why not in 2018 you might ask?), this solution is probably better than the next best thing which is is instead to keep the peer deps thing going but add an explicit warning/error right now if >1 version of addons is installed. This is what a lot of singleton libraries do.

tmeasday (approved):

PR #3390, opened 2018-04-09
Puppeteer install instructions

strass:
Just a note on how to fix an error I encountered due to not having puppeteer installed

### DIFF
--- a/addons/storyshots/README.md
+++ b/addons/storyshots/README.md
@@ -172,6 +172,8 @@ Internally here are the steps:
 - Browses each stories (calling _<http://localhost:6006/iframe.html?...>_ URL),
 - Take screenshots & save all images under _\_image_snapshots\__ folder.
 
+**Puppeteer is not installed by default** as that would increase install times for all storyshot users. `yarn add puppeteer --dev` will save it to your package.json. This will fix the error `Jest: 'it' and 'test' must return either a Promise or undefined.`
+
 ### Specifying the storybook URL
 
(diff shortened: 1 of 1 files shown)


Hypnosphi:
Weird, `puppeteer` is actually a direct dependency of `@storybook/addon-storyshots`:

https://github.com/storybooks/storybook/blob/master/addons/storyshots/package.json#L27

strass:
@Hypnosphi that is very strange... Do you know if anyone else has experienced the error `Jest: 'it' and 'test' must return either a Promise or undefined.` when trying to run based on the image snapshots alpha instructions

thomasbertet:
Not that I am aware of. Which version of Jest are you using ? 
Did you specify any _before_ hook ?

strass:
@thomasbertet I am using the latest version of CRA, no before hooks (you mean on the script invocation, right?)

ndelangen (approved):

addons/storyshots/README.md
```diff
@@ -172,6 +172,8 @@ Internally here are the steps:
 - Browses each stories (calling _<http://localhost:6006/iframe.html?...>_ URL),
 - Take screenshots & save all images under _\_image_snapshots\__ folder.
 
+**Puppeteer is not installed by default** as that would increase install times for all storyshot users. `yarn add puppeteer --dev` will save it to your package.json. This will fix the error `Jest: 'it' and 'test' must return either a Promise or undefined.`
```
Hypnosphi: > Puppeteer is not installed by default

@ndelangen this is just not true

PR #3403, opened 2018-04-11
upgrade docgen to 2.0rc

danielduan:
Issue:
I fixed some stuff inside our `babel-plugin-react-docgen`

since there are some big changes, I don't feel comfortable publishing it as a new major version. this is just a release candidate for storybook alpha.

Most of the breaking changes are under the hood and not exposed to the users:
* changed default resolver to `findAllExportedComponentDefinitions` so we can search through named exports - users who previously specified this through the resolver in `.babelrc` could not get docgen info passed into the info addon because docgen returns an array of objects instead of an object. This should fix that issue.
* Default to not remove method info and changed `.babelrc` key to `removeMethods` instead of `includeMethods`. If users have `includeMethods` set to `false`, this will not affect them either.

## What I did

https://github.com/storybooks/babel-plugin-react-docgen/pull/49

## How to test

Please tag your pull request with at least one of the following:

### DIFF
--- a/app/react/package.json
+++ b/app/react/package.json
@@ -33,7 +33,7 @@
     "airbnb-js-shims": "^1.4.1",
     "babel-loader": "^7.1.4",
     "babel-plugin-macros": "^2.2.0",
-    "babel-plugin-react-docgen": "^1.9.0",
+    "babel-plugin-react-docgen": "^2.0.0-rc.0",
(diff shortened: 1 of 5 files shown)


Hypnosphi (approved):
It would probably make sense to add some examples with named exports

PR #3415, opened 2018-04-13
Add a link to learnstorybook to get started guide

tmeasday:
I hope this is useful for beginners (!). We could link to it in other places but I don't want to be too aggressive about it ;)


danielduan (approved):
unrelated, but can you remove `Storybook is very easy to use` at the top? it provides no useful information to the user and causes unnecessary frustration if it isn't "easy" to the user

Issue #3427, opened 2018-04-16
Test suite failed to run


CLL80:
This issue introduced to master over the last few days:

```
 Your test suite must contain at least one test.

      at node_modules/jest-cli/build/test_scheduler.js:157:22
```

### Steps to reproduce
Clone storybook

`yarn bootstrap --core && yarn test --core --image`

storybook 4.0.0-alpha.2

Hypnosphi:
You need to run `cd examples/official-storybook && yarn build-storybook` before image tests

https://github.com/storybooks/storybook/blob/master/CONTRIBUTING.md#cra-kitchen-sink---image-snapshots-using-storyshots

CLL80:
Yes, they all pass after this. Thanks.

PR #3428, opened 2018-04-16
Addon info prop options

binomialstew:
Issue:

Previously, `maxPropsIntoLine` had no effect on `object` or `array` props, potentially resulting in very long lines in the story source. Now, `maxPropsIntoLine` affects these in the same way as props, including the correct indentation for each key/value.


CLL80:
Tests/snapshots will pass but were broken with latest merge from master. Due to [this issue](#3427 ).

CLL80:
I'm not sure how to tag the PR as was requested. I don't have permission to add a label.

Hypnosphi:
Can you please add a prop with large object to addon info example in official-storybook

CLL80:
@Hypnosphi, definitely. Good idea.

CLL80:
In looking at this again, I noticed that truncated prop ellipses weren't broken to the next line. The latest push corrects this and adds the test @Hypnosphi suggested.

Hypnosphi (approved):

Hypnosphi:

PR #3433, opened 2018-04-17
Add lit-html support

stijnkoopal:
Adds the ability to render a story based on [lit-html](https://github.com/Polymer/lit-html)

## What I did
Instead of returning either a `string` or a `Node` from a story, now you can also return a `TemplateResult`

## How to test
See polymer stories in the examples dir

### DIFF
--- a/app/polymer/package.json
+++ b/app/polymer/package.json
@@ -53,12 +53,14 @@
     "webpack-hot-middleware": "^2.22.0"
   },
   "devDependencies": {
+    "lit-html": "^0.9.0",
     "nodemon": "^1.17.3",
     "polymer-webpack-loader": "^2.0.2"
   },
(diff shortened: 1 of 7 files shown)


app/polymer/package.json
```diff
@@ -53,12 +53,14 @@
     "webpack-hot-middleware": "^2.22.0"
   },
   "devDependencies": {
+    "lit-html": "^0.9.0",
```
igor-dv: Why do you need the "lit-html" and "polymer-webpack-loader" in devDeps ?

Hypnosphi: For `lit-html`, it's probably because you need a single instance of that package for `instanceof TemplateResult` to pass

Hypnosphi (approved):

PR #3445, opened 2018-04-18
Bring slack logo back to the Footer

mshaaban088:
Reverts storybooks/storybook#2819

### DIFF
--- a/docs/src/pages/_users.yml
+++ b/docs/src/pages/_users.yml
@@ -5,12 +5,12 @@ airbnb:
   source: https://github.com/airbnb/react-dates
   demo: http://airbnb.io/react-dates/
   site: http://airbnb.com
-# slack:
-#   logo: slack.svg
-#   title: Slack Emoji Picker
-#   description: Slack improves performance and maintainability with Storybook!
-#   demo: https://slack.engineering/rebuilding-slacks-emoji-picker-in-react-bfbd8ce6fbfe
-#   site: http://slack.com
+slack:
+  logo: slack.svg
+  title: Slack Emoji Picker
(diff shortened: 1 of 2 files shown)


ndelangen (approved):

docs/src/pages/_users.yml
```diff
@@ -5,12 +5,12 @@ airbnb:
...
-#   logo: slack.svg
-#   title: Slack Emoji Picker
-#   description: Slack improves performance and maintainability with Storybook!
-#   demo: https://slack.engineering/rebuilding-slacks-emoji-picker-in-react-bfbd8ce6fbfe
-#   site: http://slack.com
+slack:
+  logo: slack.svg
+  title: Slack Emoji Picker
+  description: Slack improves performance and maintainability with Storybook!
+  demo: https://slack.engineering/rebuilding-slacks-emoji-picker-in-react-bfbd8ce6fbfe
```
danielduan: wait, I'm confused. this is not a demo.

mshaaban088: This is not a demo per se, but they demonstrated how storybook helped them developing UI development environment. It doesn't qualify as a site either from my understanding, or am I missing something?

danielduan: Demo to me means an actual instance of storybook that a user can play with and see how someone else used it. The rest of the demo links are all usage examples.

The medium article is more of a case study and talks very little about the specifics of their storybook implementation.

I went in a while back and sorted the list to prioritize good demos at the top. This should would probably be at the bottom.

mshaaban088: Fair enough!

We should remove that link then, and bring it back once they share it. WDYT?

PR #3447, opened 2018-04-18
Provide a configuration option to limit the number of actions logged

jcharry:
Issue: The actions logger panel accumulates actions for a story endlessly. After a while, this creates quite a bit of lag. You can hit the clear button to clear out the logger and performance improves again, but it'd be nice to have a configuration hook to set a hard upper bound on the number of actions that should be logged.

## What I did
Added a configuration option called `limit` to the `configureActions` call.

### DIFF
--- a/addons/actions/README.md
+++ b/addons/actions/README.md
@@ -109,3 +109,4 @@ action('my-action', {
 |---|---|---|---|
 |`depth`|Number|Configures the transfered depth of any logged objects.|`10`|
 |`clearOnStoryChange`|Boolean|Flag whether to clear the action logger when switching away from the current story.|`true`|
+|`limit`|Number|Limits the number of items logged in the action logger|`50`|
--- a/addons/actions/README.md
+++ b/addons/actions/README.md
@@ -80,20 +80,26 @@ over the channel.
 This is not very optimal and can cause lag when large objects are being logged, for this reason it is possible 
 to configure a maximum depth.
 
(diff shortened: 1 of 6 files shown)


rhalff:
A default limit could probably be added to [configureActions]( https://github.com/jcharry/storybook/blob/fa075796a83efda1f4c1874a0ff3d0aa4d71c5a9/addons/actions/src/preview/configureActions.js#L2), not sure what the default value should be though.

I've added @Hypnosphi as reviewer, does the PR also need to include an example in the `official-storybook`?

danielduan:
I think we should set a sensible default for all users while we're making this change. This seems like an obvious optimization we can make on behalf of users.

does 50 sound like a good number?

Hypnosphi:
> does the PR also need to include an example in the `official-storybook`?

Yeah, that wouldn't hurt

> I think we should set a sensible default for all users while we're making this change. This seems like an obvious optimization we can make on behalf of all users.
>
> does 50 sound like a good number?

Makes sense to me

danielduan (approved):

danielduan:
@jcharry can you fix the failing test here?

`yarn bootstrap --core` and then `yarn test --core --update`

Hypnosphi:
Please add an example to `example/official-storybooks`

jcharry:
@Hypnosphi Okay, added an example.

jcharry:
Rebased and should be good to go now.

PR #3453, opened 2018-04-19
Improve yarn detection

### LINKED ISSUE #3384: getstorybook shouldn't use yarn if there's no `yarn.lock`
### Bug(?)

Using latest `@storybook/cli` running `getstorybook` installs alpha version of storybook versus latest stable version.

### Steps to reproduce

```bash
npm i -g @storybook/cli
getstorybook
```

The installed version of Storybook dependencies are `alpha` versions. Considering the versions are alpha, it seems unexpected that the most recent version of `getstorybook` would install them rather than the latest stable release.

### Affected platforms 

I'm on OSX (10.13.3)

Keraito:
Issue: Currently, `getstorybook` uses yarn when it's available on the system even in cases where `npm` is used in the current project, causing errors. (closes #3384)

## What I did
I changed the condition for using `yarn` from checking `yarn --version` to looking for a `yarn.lock` file in the current project. 

## How to test
- I ran the e2e tests and all of them passed. Testing this with a (additional) fixture seemed excessive based on the fixtures, but please correct me if I'm wrong.

### DIFF
--- a/lib/cli/lib/has_yarn.js
+++ b/lib/cli/lib/has_yarn.js
@@ -1,10 +1,13 @@
+import { sync as spawnSync } from 'cross-spawn';
 import path from 'path';
 import fs from 'fs';
 
 export default function hasYarn() {
+  const yarnAvailable = spawnSync('yarn', ['--version'], { silent: true });
+  const npmAvailable = spawnSync('npm', ['--version'], { silent: true });
   const yarnLockPath = path.resolve('yarn.lock');
-  if (!fs.existsSync(yarnLockPath)) {
-    return false;
+  if ((fs.existsSync(yarnLockPath) || npmAvailable.status !== 0) && yarnAvailable.status === 0) {
+    return true;
   }
-  return true;
+  return false;
 }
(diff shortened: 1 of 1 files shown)


Keraito:
After pushing I thought of the case where `yarn.lock` exists, but yarn is not installed on the system. The process will stop as no yarn is available, is that a (rare) case that is storybook's responsibility to cover?

Hypnosphi (approved):

danielduan:
I think we should default to using `npm` unless (yarn is installed AND (has yarn lock OR npm not installed))

We're bound to get issues with these edge cases given the number of users we have.

Keraito:
I changed the function to behave as specified in https://github.com/storybooks/storybook/pull/3453#issuecomment-382891993, defaulting to `npm` unless the condition matches.

ndelangen:
This looks mergable, what's wrong with the CI @Hypnosphi ?

Hypnosphi:
Connection timeout, I've restarted

PR #3456, opened 2018-04-20
Adding Meteor import statement support

DylanLyon:
Issue: When importing a component that has a Meteor import (e.g. 'import Random from 'meteor/meteor') inside itself or a child component, Storybook will give the error, 'Module not found: Error: Can't resolve 'meteor/meteor''.

It does this because Meteor does it's own packing independently of webpack and isn't a node module that can be easily imported. I've seen multiple possible suggestions for workarounds that might be worth trying.

- Stubbing Meteor out
- Exporting Meteor to global namespace
- Exporting Meteor as node packages
- Modifying Meteor to use webpack
- (Desperate option) rewrite all components to no longer use Meteor imports. 

I'm unsure if any of these methods would work or even make sense but perhaps someone with more experience than I will have more success with it.


danielduan:
you need to rebase your changes. your changes were made off some old code from a while back.

DylanLyon:
I don't have any changes though, I'm just trying to get help.

PR #3457, opened 2018-04-20
Refactor error and "no preview" views into core

Hypnosphi:
After this change, `preview` part in each app contains only framework-specific things


lib/core/src/client/preview/start.js
```diff
@@ -11,6 +12,50 @@ import reducer from './reducer';
...
+
+function showNopreview() {
+  document.body.classList.remove(classes.MAIN);
+  document.body.classList.remove(classes.ERROR);
+
+  document.body.classList.add(classes.NOPREVIEW);
+}
+
+function showErrorDisplay({ message, stack }) {
+  document.getElementById('error-message').textContent = message;
```
ndelangen: Interesting, we're just using plain JS here to show/hide plain messages.

Makes sense.. I always assumed the error handling and debugging experiences for the different frameworks would be wildly different and thus would warrant having a separate component for each.

But setup is quite nice and simple I guess, description has to be an html-string I guess, and so a framework could submit some rich data down to the error display.

ndelangen (approved):

ndelangen:
Would like approval of some more people

lib/core/src/client/preview/start.js
Hypnosphi: > description has to be an html-string

Well right now it isn't supported (I use `textContent` not `innerHTML`). Error stack can be user-generated, so there's some room for xss here

Hypnosphi: But they were enough so far

app/angular/package.json
```diff
@@ -38,26 +35,19 @@
     "babel-runtime": "^6.23.0",
     "case-sensitive-paths-webpack-plugin": "^2.1.2",
     "core-js": "^2.5.5",
-    "cross-env": "^5.1.4",
```
igor-dv: So this : `"dev": "cross-env DEV_BUILD=1 nodemon -e js,ts --watch ./src --exec \"yarn prepare\"",` is not needed anymore (I think it wasn't in use before anyway) ?

igor-dv (approved):
Very cool!

app/angular/package.json
Hypnosphi: It's executed in root, so no need to have those dependencies on this level

Hypnosphi: Or maybe it's not, then we can remove the script as well

igor-dv: I think it's not (maybe it was, but a long time ago). 
The last time I've been touching there was due to the ts distribution issue, and I've cleaned thing there. 
But maybe I am missing something

igor-dv: Oh, I am late

lib/core/src/client/preview/start.js
tmeasday: What is XSS in this context?

tmeasday: As in who is attacking whom?

tmeasday:
This is great! Thanks for this @Hypnosphi

lib/core/src/client/preview/start.js
Hypnosphi: Somebody giving you some tricky URL when you have public storybook on some route of your main website whicj has auth and cookies

tmeasday: Oooh, that's a good one, touche

PR #3459, opened 2018-04-20
Add option to clear action logger

### LINKED ISSUE #3443: Add option to clear action logger on switching stories
I would like to see an option to clear the action logger (and possibly the console) when switching between stories.

This could simply be `clearActionLogger: true`.

I would be happy to write this feature if someone can help me get started with where to look.

Keraito:
Issue: Currently, the action loggers stacks up all the events without limit and across all stories. There is already a PR (#3447) open for a limiting option. In the same way, this introduces an optional feature that will clear the action logger when switching away from a story. (Fixes #3443)

## What I did
Added a `clearActionLogger` option to the actions addon that clears the action logger when switching away from a story. It uses the `storybook/stories/story-event` event for detecting a switch in story.

## How to test
I updated the `configureAction` unit test and the documentation on action options.

### DIFF
--- a/addons/actions/README.md
+++ b/addons/actions/README.md
@@ -102,4 +102,4 @@ action('my-action', {
 |Name|Type|Description|Default|
 |---|---|---|---|
 |`depth`|Number|Configures the transfered depth of any logged objects.|`10`|
-|`clearActionLogger`|Boolean|Flag whether to clear the action logger when switching away from the current story.|`false`|
+|`clearOnStoryChange`|Boolean|Flag whether to clear the action logger when switching away from the current story.|`true`|
--- a/addons/actions/README.md
+++ b/addons/actions/README.md
@@ -102,3 +102,4 @@ action('my-action', {
 |Name|Type|Description|Default|
 |---|---|---|---|
 |`depth`|Number|Configures the transfered depth of any logged objects.|`10`|
+|`clearActionLogger`|Boolean|Flag whether to clear the action logger when switching away from the current story.|`false`|
--- a/examples/official-storybook/stories/addon-actions.stories.js
+++ b/examples/official-storybook/stories/addon-actions.stories.js
@@ -98,11 +98,11 @@ storiesOf('Addons|Actions', module)
       </Button>
     );
   })
-  .add('Clearing the action logger', () => (
+  .add('Persisting the action logger', () => (
     <div>
-      <p>Moving away from this story will clear the action logger</p>
-      <Button onClick={action('clear-action-logger', { clearActionLogger: true })}>
-        Object (clearActionLogger: true)
+      <p>Moving away from this story will persist the action logger</p>
+      <Button onClick={action('clear-action-logger', { clearOnStoryChange: false })}>
+        Object (configured clearOnStoryChange: false)
       </Button>
     </div>
   ));
--- a/examples/official-storybook/stories/addon-actions.stories.js
+++ b/examples/official-storybook/stories/addon-actions.stories.js
@@ -97,4 +97,12 @@ storiesOf('Addons|Actions', module)
         Object (configured depth: 2)
       </Button>
     );
-  });
+  })
+  .add('Clearing the action logger', () => (
+    <div>
+      <p>Moving away from this story will clear the action logger</p>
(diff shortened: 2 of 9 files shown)


Keraito:
In the corresponding issue @merlinpatt mentioned that he would be willing to do this. I just decided to start working on it and proposed my current progress. If he wants to take over or maybe work together I dont mind.

Hypnosphi:
Please use this option in `examples/storybook-official`

Keraito:
Something like this, @Hypnosphi?

examples/official-storybook/stories/addon-actions.stories.js
```diff
@@ -97,4 +97,12 @@ storiesOf('Addons|Actions', module)
         Object (configured depth: 2)
       </Button>
     );
-  });
+  })
+  .add('Clearing the action logger', () => (
+    <div>
+      <p>Moving away from this story will clear the action logger</p>
```
Hypnosphi: I think it would be more natural if it worked the other way around, and logger would be cleared when opening this story. Can we do this?

Keraito: I at first tried to do it the way @Hypnosphi described it, but it's slightly more difficult that way because the options are part of the actions and not the stories themselves. The way @ndelangen sounds more logical and should be of less complexity, so I will try my hands on that.

merlinpatt:
Sorry I'm late to the conversation but isn't there an overall Storybook configuration that this could go in?

Hypnosphi:
@merlinpatt So far, not really. Each addon exposes its own configuration function when needed

addons/actions/README.md
```diff
@@ -102,3 +102,4 @@ action('my-action', {
 |Name|Type|Description|Default|
 |---|---|---|---|
 |`depth`|Number|Configures the transfered depth of any logged objects.|`10`|
+|`clearActionLogger`|Boolean|Flag whether to clear the action logger when switching away from the current story.|`false`|
```
igor-dv: I would call it "clearOnStoryChange"

Keraito:
> Does the action logger have some sort of top level configuration?

@merlinpatt check out [this](https://github.com/storybooks/storybook/tree/master/addons/actions#configuration) on `configureActions`.

> I also find it likely that users would want to clear it on every story change

Based on the other comments, I will be changing the default behaviour to this

Keraito:
I processed both feedback of https://github.com/storybooks/storybook/pull/3459#discussion_r183218281 and https://github.com/storybooks/storybook/pull/3459#discussion_r183236236. 

Behaviour wise, actions are still only persisted when **switching away** from a story. While changing the default to `true` gets rid of the counter-inituitive behaviour described in https://github.com/storybooks/storybook/pull/3459#discussion_r183210433, I didn't see a way yet to make this happen on entry.

Hypnosphi:
> actions are still only persisted when switching away from a story

That's OK, you're kinda saying "I want logs of this story to be persisted"

Hypnosphi (approved):

rhalff (changes requested):
The event id the story change listener depends on (`storybook/stories/story-event`) seems to be triggered by the storysource addon: https://github.com/storybooks/storybook/blob/master/addons/storysource/src/index.js#L1-L3

Unless the storybook api is changed in the near future anyway, I wouldn't rely on this particular event.

Otherwise at least import the event id as `import { EVENT_ID } from '@storybook/addon-storysource` so it's clear where the event originates from.

Hypnosphi:
I think this one should be emitted by core

rhalff:
Currently there is an `onStory()` method defined on the api which could be used, it is passed in during register of the addon.

Usage of it is a bit awkward though, e.g.: 
```js
export function register() {
    
    addons.addPanel(PANEL_ID, {
      title: 'Action Logger',
      render: () => <ActionLogger channel={channel} api={api} />,
    });
  });
}

// do something with api.onStory in the ActionLogger
api.onStory((kind, story) => ....)
```

Keraito:
Thank you @rhalff, I will look into the `onStory` method and propose changes accordingly

Keraito:
Something like this @rhalff?

Keraito:
Added it, thanks for the info @rhalff.

rhalff (approved):

PR #3467, opened 2018-04-21
Upgrade sass-loader to 7.0.1 & stop bringing node-sass bin to each user

Hypnosphi:
Issue: #3425 

This update allows us to stop making users download a large node-sass binary even if they don't use sass/scss at all.

See also: https://github.com/webpack-contrib/sass-loader/issues/532


ndelangen:
I'm OK with this..

ndelangen (approved):

Issue #3408, opened 2018-04-12
[addon-info] Filter out props by name


mindtricksdev:
I was wondering if there's a way to avoid rendering some props when using `withInfo` from `@storybook/addon-info`.

Right now I'm using the following options to avoid rendering HoC props and it works well:
 ```
{
  propTables: [Icon],
  propTablesExclude: [IconWithStyles],
}
```

But the `Icon` has a `classes: PropTypes.object.isRequired` which gets rendered in the docs and I want to filter this out because it is injected by the HoC. The user does not need to pass this when using the default export `IconWithStyles` component.

Thanks!

andyo729:
I could look into this and try to throw something together.

Hypnosphi:
Released as [`4.0.0-alpha.4`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.4)

PR #3468, opened 2018-04-21
Add excludedPropTypes as an option to info addon

andyo729:
Issue: #3408 

## What I did

I updated the info addon to include an `excludedPropTypes` option to exclude those prop types from the `PropTable`.

## How to test

There are two new tests in `PropTable`, one for the default behavior and another for the new filtering behavior.


addons/info/src/components/PropTable.js
```diff
@@ -22,20 +22,33 @@ export const multiLineText = input => {
       ));
 };
 
+const determineIncludedPropTypes = (propDefinitions, excludedPropTypes) => {
+  if (excludedPropTypes.length === 0) {
+    return propDefinitions;
+  }
+
+  return propDefinitions.filter(
+    propDefinition => !excludedPropTypes.includes(propDefinition.property)
```
Hypnosphi: You need to import a polyfill on top of this file:

```
import 'core-js/fn/array/includes';
```

See https://babeljs.io/docs/plugins/transform-runtime/
> NOTE: Instance methods such as `"foobar".includes("foo")` will not work since that would require modification of existing built-ins

Hypnosphi:
Please add a usage example to `examples/official-storybook`

addons/info/src/components/Story.js
```diff
@@ -331,7 +331,7 @@ class Story extends React.Component {
     const array = Array.from(types.keys());
     array.sort((a, b) => getName(a) > getName(b));
 
-    const { maxPropObjectKeys, maxPropArrayLength, maxPropStringLength } = this.props;
+    const { maxPropObjectKeys, maxPropArrayLength, maxPropStringLength, excludedPropTypes } = this.props;
```
ndelangen: Do we have a place to document this?

andyo729: Added this new option to the README

addons/info/src/components/PropTable.js
andyo729: Pulled the polyfill into the `PropTable`

examples/official-storybook/stories/addon-info.stories.js
```diff
@@ -59,6 +59,14 @@ storiesOf('Addons|Info.Options.inline', module).add(
   })(() => <BaseButton label="Button" />)
 );
 
+storiesOf('Addons|Info.Options.excludedPropTypes').add(
```
Hypnosphi: please don't forget adding second `module` parameter, this is needed for HMR

addons/info/src/components/PropTable.js
```diff
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import 'core-js/fn/array/includes';
```
Hypnosphi: You need to add `core-js` as dependency to `addons/info/package.json`

examples/official-storybook/stories/addon-info.stories.js
andyo729: 👍Added back in

addons/info/src/components/PropTable.js
andyo729: Added

Hypnosphi (approved):

yarn.lock
```diff
@@ -4058,6 +4058,10 @@ copy-webpack-plugin@~4.4.1:
     p-limit "^1.0.0"
     serialize-javascript "^1.4.0"
 
+core-js@2.5.5:
+  version "2.5.5"
+  resolved "http://npm.kroger.com/core-js/-/core-js-2.5.5.tgz#b14dde936c640c0579a6b50cabcc132dd6127e3b"
```
Hypnosphi: Looks like this is your private registry and other people can't install from there. You might need to add an empty `.npmrc` to project root (without commiting it), and run `yarn` again

andyo729: Thanks for pointing that out. This has been updated.

Hypnosphi (changes requested):
It doesn't seem to work: label prop is still shown in the table: https://deploy-preview-3468--storybooks-official.netlify.com/?selectedKind=Addons%7CInfo.Options.excludedPropTypes&selectedStory=Excludes%20propTypes%20that%20are%20in%20the%20excludedPropTypes%20array&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

Hypnosphi (approved):

PR #3470, opened 2018-04-21
Upgrade redux to 4.0.0

Hypnosphi:
Issue: #3436


tmeasday:
LGTM, although I reckon we should just get rid of the redux store already ;)

tmeasday (approved):

Issue #3157, opened 2018-03-06
Can I use Storybook without any JS framework to design my LESS framework?


TotomInc:
I've just discovered storybook which is amazing and I would really like to use it to design a LESS/CSS framework. I wanted to know if it's possible to use storybook without any JS framework such as Vue, Angular, ... but only to display each component without user interaction, kind of a dev-environment. I've found nothing about it in the docs or in the issues, can someone explain if it's possible and how to setup it?

MXTcomunica:
I'm also interested in this (using SASS + twig).

In particular, can storybook generate a styleguide with this component structure? E.g.:

components
|__ component_1
          |__ component_1.scss
          |__ component_1.twig

Hypnosphi:
You can use storybook for vue without actually using any of vue features and just pass plain html as template:

```
import { storiesOf } from '@storybook/vue';

storiesOf('MySnippets', module)
  .add('hello', () => '<div class="hello">Hello World</div>')
```

TotomInc:
@Hypnosphi cannot this be a feature inside storybook? It can be useful to implement a way to use storybook without any JS framework.

Hypnosphi:
What's wrong with my suggestion?

tmeasday:
OTOH it would not be hard to add `@storybook/js` I suppose @Hypnosphi?

Hypnosphi:
I'd rather call it `@stroybook/html`

TotomInc:
@Hypnosphi there is nothing wrong with your suggestion, but it looks like a workaround to make it work without any framework. I know storybook was initially designed to be used with a js framework, but can't we make a version of it (like the polymer version) to use it without any js framework?

TotomInc:
Any plans on this?

igor-dv:
I think it won't happen before the 3.4 release. Let's estimate for the next release. Wana help with it? Actually, it's ok to start working on it before 😉 

> I'd rather call it @storybook/html

Here is my list of names 😄 
- `@storybook/natural`
- `@storybook/naked`
- `@storybook/free`
- `@storybook/vegan`

TotomInc:
`@storybook/natural` looks great, I tried to get into the codebase of storybook to see if I can do something. I'll try again soon and see if I can contribute to this! 😄

igor-dv:
@TotomInc , we've done a bit of refactoring moving common parts to `@storybook/core`, so it's much easier now to create a new framework (or a lack of framework) support.

[...]

PR #3475, opened 2018-04-22
Storybook for HTML snippets

Hypnosphi:
Issue: #3157 
[Live example](https://html--storybooks-html.netlify.com/)

Todo:
  * [x] Docs
  * [x] Addons support
  * [x] Netlify
  * [x] Acceptance story
  * [x] CI


examples/html-kitchen-sink/stories/index.stories.js
```diff
@@ -0,0 +1,26 @@
...
+import './welcome.css';
+import welcome from './welcome.html';
+
+document.addEventListener('click', e => {
+  const { kind, story } = e.target.dataset;
+  if (kind || story) {
+    e.preventDefault();
+    linkTo(kind, story)();
+  }
+});
```
Hypnosphi: We could incorporate something like this into links addon

examples/html-kitchen-sink/stories/index.stories.js
```diff
@@ -0,0 +1,26 @@
...
+
+storiesOf('Welcome', module).add('Welcome', () => welcome);
+
+storiesOf('Demo', module)
+  .add('heading', () => '<h1>Hello World</h1>')
+  .add('button', () => {
+    const button = document.createElement('button');
+    button.innerText = 'Hello Button';
+    button.addEventListener('click', action('Click'));
```
Hypnosphi: I have some thought on adding some decorator with [backbone-style](http://backbonejs.org/#View-delegateEvents) event listeners:

```
.addDecorator(withActions({
  dblclick : 'Double click',
  'click .icon.doc':   'Doc icon click',
  'contextmenu .icon.doc': 'Doc icon right click'
}))
```

examples/html-kitchen-sink/stories/index.stories.js
igor-dv: I can't see how it will work with a real-world example, because probably your components won't be coupled to the "story"/"kind"

igor-dv (approved):
🔥 
Do you need any help with addons support?

examples/html-kitchen-sink/stories/index.stories.js
Hypnosphi: We can use some namespace, like `data-sb-story`

Hypnosphi:
> Do you need any help with addons support?

I'd appretiate any. I'll create a slack channel to coordinate

examples/html-kitchen-sink/stories/index.stories.js
Hypnosphi: The idea is, if you want a link to other story, you just add corresponding data-attribute to the element that you want to be the link

docs/src/pages/basics/guide-html/index.md
```diff
@@ -0,0 +1,112 @@
...
+-   [Create the config file](#create-the-config-file)
+-   [Write your stories](#write-your-stories)
+-   [Run your Storybook](#run-your-storybook)
+
+## Add @storybook/html
+
+First of all, you need to add `@storybook/html` to your project. To do that, simply run:
+
+```sh
+npm i --save-dev @storybook/mithril
```
tmeasday: mithril?

docs/src/pages/basics/guide-html/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+You may have tried to use our quick start guide to setup your project for Storybook. If you want to set up Storybook manually, this is the guide for you.
+
+> This will also help you to understand how Storybook works.
+
+## Starter Guide HTML
+
+Storybook has its own Webpack setup and a dev server.
+
+In this guide, we will set up Storybook for your Mithril project.
```
tmeasday: Mithril?

docs/src/pages/basics/guide-html/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+Storybook can be configured in several different ways. 
+That’s why we need a config directory. We've added a `-c` option to the above NPM script mentioning `.storybook` as the config directory.
+
+For the basic Storybook configuration file, you don't need to do much, but simply tell Storybook where to find stories.
+
+To do that, simply create a file at `.storybook/config.js` with the following content:
+
+```js
+import { configure } from '@storybook/mithril';
```
tmeasday: mithril?

docs/src/pages/basics/guide-html/index.md
```diff
@@ -0,0 +1,112 @@
...
+    const button = document.createElement('button');
+    button.innerText = 'Hello Button';
+    // eslint-disable-next-line no-console
+    button.addEventListener('click', e => console.log(e));
+    return button;
+  });
+
+```
+
+Story is a single HTML snippet or DOM note. In the above case, there are two stories:
```
tmeasday: "The return value of a story should be a HTML snippet or a DOM element" (or do you mean node)

examples/html-kitchen-sink/stories/index.stories.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+  };
+
+  document.addEventListener('click', listener);
+  return () => document.removeEventListener('click', listener);
+};
+
+storiesOf('Welcome', module)
+  .addDecorator(story => {
+    subscriptionsStore.register(linkSubscribtion);
```
tmeasday: I don't really understand the idea of using a subscription+decorator to make this run each time we render a story. Why not just register the global event handler once in `addon-links`?

tmeasday (changes requested):
A couple of missed "mithrils" + a question about `addon-links`

docs/src/pages/basics/guide-html/index.md
Hypnosphi: node, yes

lib/core/src/client/preview/start.js
```diff
@@ -19,6 +20,8 @@ const classes = {
 };
 
 function showMain() {
+  subscriptionsStore.clearUnused();
```
igor-dv: Why is it called here and not right after the render()?

Hypnosphi: Mostly to keep a future possibility of async rendering

addons/centered/src/html.js
```diff
@@ -0,0 +1,39 @@
...
+  position: 'fixed',
+  top: 0,
+  left: 0,
+  bottom: 0,
+  right: 0,
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  overflow: 'auto',
+};
```
Hypnosphi: Can we share this object between frameworks?

addons/centered/src/html.js
```diff
@@ -29,7 +29,6 @@ export default function(storyFn) {
   if (typeof component === 'string') {
     inner.innerHTML = component;
   } else if (component instanceof Node) {
-    inner.innerHTML = '';
     inner.appendChild(component);
```
Hypnosphi: Are you sure it's unneeded? Looks like it will keep the previous children

lib/core/src/client/preview/start.js
igor-dv: But you will be able to `await render()` and then call to the "clearUnused", what am I missing?

addons/centered/src/html.js
```diff
@@ -0,0 +1,39 @@
...
+  justifyContent: 'center',
+  overflow: 'auto',
+};
+
+const innerStyle = {
+  margin: 'auto',
+};
+
+export default function(storyFn) {
+  const inner = document.createElement('div');
```
Hypnosphi: We could use ids and avoid creating wrappers on each render

addons/centered/src/html.js
igor-dv: Yeah, makes sense

addons/centered/src/html.js
igor-dv: Where should this wrapper be then? You mean to place it somewhere in DOM?

addons/storyshots/src/html/renderTree.js
```diff
@@ -0,0 +1,16 @@
...
+  const component = story.render(context);
+
+  if (component instanceof Node) {
+    return component;
+  }
+
+  const section = document.createElement('section');
+  section.innerHTML = component;
+
+  return section.firstChild;
```
Hypnosphi: One should be able to render more than one element. So I think we should use `section.childNodes` instead

lib/core/src/client/preview/start.js
Hypnosphi: OK makes sense

Hypnosphi: My idea was that if `render` didn't call `showMain`, then something went wrong, and it could be the reason for decorator not to be marked as used

addons/centered/src/html.js
Hypnosphi: In the same plase as it is now, just with an id

igor-dv: Ah, it's useful when navigating between stories with the same decorator. Will do.

examples/html-kitchen-sink/stories/index.stories.js
Hypnosphi: > Isn't that what the decorator essentially does though

Nope, it only subscribes the first time, subsequently it only marks subscription as used so that it's not removed after render

> I don't really get what the subscriptionStore has to do with this PR

We need to support addons like knobs which use a lot of subscriptions. In other frameworks we use lifecycle hooks for that, e.g.: https://github.com/storybooks/storybook/blob/master/addons/knobs/src/react/WrapStory.js#L18-L37

The `subscriptionStore` is a framework-agnostic solution for this, so we can:

a) Use such addons in @storybook/html (that's why it's this PR)
b) Remove a lot of existing framework specifics from addons

addons/centered/src/html.js
Hypnosphi: Yes, or when some decorator like knobs triggers a re-render

addons/centered/src/html.js
```diff
@@ -1,34 +1,42 @@
...
+}
+
+function getWrapperDiv() {
+  return getOrCreate(WRAPPER_ID, styles.style);
+}
+
+export default function(storyFn) {
+  const inner = getInnerDiv();
+  const wrapper = getWrapperDiv();
   wrapper.appendChild(inner);
```
Hypnosphi: I think we can assume that if `inner` exists, it's already appendend to `wrapper`

addons/storyshots/src/html/renderTree.js
igor-dv: There is a problem of creating a tree for the array of nodes, so I will return a wrapper in case of a few child nodes

Hypnosphi: can [`DocumentFragment`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) help?

Hypnosphi:
```
const fragment = document.createDocumentFragment()
section.childNodes.forEach(node => fragment.appendChild(node))
```

igor-dv: Nope, didn't help.

addons/knobs/src/html/WrapStory.js
```diff
@@ -0,0 +1,106 @@
...
+    this.setPaneKnobs(this.channel, this.knobStore, false);
+  }
+
+  setPaneKnobs(timestamp = +new Date()) {
+    const { channel, knobStore } = this;
+    channel.emit('addon:knobs:setKnobs', { knobs: knobStore.getAll(), timestamp });
+  }
+
+  render(component) {
+    const wrapper = this.knobsWrapper;
```
Hypnosphi: Is this wrapper needed only for rerendering? Can we use forceRender instead?

examples/html-kitchen-sink/stories/addon-knobs.stories.js
```diff
@@ -0,0 +1,15 @@
...
+
+storiesOf('Addons|Knobs', module)
+  .addDecorator(withKnobs)
+  .add('simple', () => {
+    const title = text('Button title', 'Hello');
+
+    const element = document.createElement('div');
+
+    element.innerHTML = title;
+
```
Hypnosphi: return `<div>${title}<div>` looks simpler to me

addons/knobs/src/html/index.js
```diff
@@ -0,0 +1,33 @@
...
+} from '../base';
+
+export { knob, text, boolean, number, color, object, array, date, select, files };
+
+export function button(name, callback) {
+  return manager.knob(name, { type: 'button', value: Date.now(), callback, hideLabel: true });
+}
+
+function prepareComponent({ getStory, context, channel, knobStore }) {
+  const wrapper = new WrapStory(getStory(context), channel, context, getStory, knobStore);
```
Hypnosphi: You instantiate a new class on each render here, so `connectCallbacks` is a new function each time

igor-dv: Yeah, but since the subscriptions store registers it, it holds the reference of every wrapper class. Then on unregister, it removes the previous reference

Hypnosphi: It's not the same reference. So it subscripes to new one and unsubsribes from previous instead of just keeping the previous

Hypnosphi: Actually, I'm not sure there's a use case for a class here

Hypnosphi: > make it similar to other wrappers

No reason for that — this one should eventually replace them all =)

addons/knobs/src/html/registerKnobs.js
```diff
@@ -0,0 +1,60 @@
+import addons from '@storybook/addons';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { subscriptionsStore } from '@storybook/core/client';
```
Hypnosphi: You can add it as devDependency, `subscriptionsStore` is a global singleton

But maybe a better approach would be to introduce an event for registering subscriptions

addons/knobs/src/html/registerKnobs.js
```diff
@@ -0,0 +1,60 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { forceReRender } from '@storybook/html';
+import { manager } from '../base';
+
+const channel = addons.getChannel();
```
Hypnosphi: channel is set at some point in runtime, so it's unsafe to call it at module top level. You can just inline this call at each usage place, because all that it does is return a field on `addons` object

Hypnosphi: Like this:

```
function disconnectCallbacks() {
  const channel = addons.getChannel()
  channel.removeListener('addon:knobs:knobChange', knobChanged);
  channel.removeListener('addon:knobs:knobClick', knobClicked);
  channel.removeListener('addon:knobs:reset', resetKnobs);
  knobStore.unsubscribe(setPaneKnobs);
}
```

LucaColonnello:
Hey guys, what about using JSX support here?
As in https://jsfiddle.net/tgnksxab/1/

Hypnosphi:
@LucaColonnello looks good but I thing it should be a separate library rather than part of storybook

examples/html-kitchen-sink/stories/index.stories.js
tmeasday: Awesome!

addons/knobs/src/html/registerKnobs.js
Hypnosphi: I'll do the latter

igor-dv:
I think everything is done here? At least for the first big PR.

Hypnosphi:
I was going to work on some more addons

examples/html-kitchen-sink/stories/index.stories.js
Hypnosphi: Did that: https://github.com/storybooks/storybook/pull/3475/commits/100985bfdbcb5b68adeb63840e1e71086bc14db2

Hypnosphi:
Ok I think it's ready for final review
CC @storybooks/team

docs/src/pages/basics/guide-html/index.md
```diff
@@ -0,0 +1,112 @@
...
+```js
+/* global document */
+import { storiesOf } from '@storybook/html';
+
+storiesOf('Demo', module)
+  .add('heading', () => '<h1>Hello World</h1>')
+  .add('button', () => {
+    const button = document.createElement('button');
+    button.innerText = 'Hello Button';
+    // eslint-disable-next-line no-console
```
wuweiweiwu: Should we remove eslint stuff from docs?

wuweiweiwu (approved):
Looks awesome!

docs/src/pages/basics/guide-html/index.md
Hypnosphi: Makes sense

tmeasday (approved):

Issue #3026, opened 2018-02-19
storyshots renderOnlyWithOptions


luetkemj:
### Issue details

I would like to be able to pass createNodeMock options to renderOnly in storyshots. I have refs in my stories that work with snapshotWithOptions but not renderOnly.

### Steps to reproduce

This works:
```
initStoryshots({
  test: snapshotWithOptions({
    createNodeMock: (element) => {
      if (element.type === 'div') {
        return document.createElement('div');
      }
      return null;
    },
  }),
});
```

this does not
```
initStoryshots({
  test: renderOnly({
    createNodeMock: (element) => {
      if (element.type === 'div') {
        return document.createElement('div');
      }
      return null;
    },
  }),
});
```

If there an existing way to handle mocking of refs in renderOnly docs should be improved so the solution is more transparent.

igor-dv:
@luetkemj, do you want to PR it?

luetkemj:
@igor-dv sure, I'll give it a go.

Hypnosphi:
Released as [`4.0.0-alpha.9`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.9)

PR #3479, opened 2018-04-24
feat: add renderWithOptions to @addons/storyshots

wuweiweiwu:
Issue:
fixes: #3026 
## What I did
added a new export from `@addons/storyshots` 

`renderWithOptions` that allows users to pass options to the renderer

ex:
```
initStoryshots({
  test: renderWithOptions({
    createNodeMock: (element) => {
      if (element.type === 'div') {
        return document.createElement('div');
      }
      return null;
    },
  }),
});
```

## How to test
I only passed an empty config. 

Is this testable with Jest or Chromatic screenshots? n/a
Does this need a new example in the kitchen sink apps? n/a
Does this need an update to the documentation? y

### DIFF
--- a/addons/storyshots/src/test-bodies.js
+++ b/addons/storyshots/src/test-bodies.js
@@ -41,8 +41,8 @@ export function shallowSnapshot({ story, context, renderShallowTree, options = {
   expect(result).toMatchSnapshot();
 }
 
-export const renderOnly = options => ({ story, context, renderTree }) => {
-  const result = renderTree(story, context, options);
+export const renderWithOptions = options => ({ story, context, renderTree }) => {
+  const result = renderTree(story, context, { options });
 
   if (typeof result.then === 'function') {
     return result;
@@ -50,5 +50,7 @@ export const renderOnly = options => ({ story, context, renderTree }) => {
 
   return undefined;
 };
+
+export const renderOnly = renderWithOptions({});
 
 export const snapshot = snapshotWithOptions({});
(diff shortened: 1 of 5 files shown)


addons/storyshots/src/test-bodies.js
```diff
@@ -41,14 +41,14 @@ export function shallowSnapshot({ story, context, renderShallowTree, options = {
   expect(result).toMatchSnapshot();
 }
 
-export function renderOnly({ story, context, renderTree }) {
-  const result = renderTree(story, context, {});
+export const renderOnly = options => ({ story, context, renderTree }) => {
+  const result = renderTree(story, context, options);
```
igor-dv: Please make the "options" param to be optional. Passing an empty object is kinda confusing.

wuweiweiwu: I see. I was trying to mimic snapshotWithOptions where options are required.

I will add another export, perhaps named `renderWithOptions` thus people using `renderOnly` will not have their code broken

igor-dv: I am ok with breaking change. The next release will be breaking. Or you wanted to have it in 3.4.x?

wuweiweiwu: I'm ok with either! Whichever you think is best. 

But since I think the use case for `renderOnly` with options is small so adding another export is probably best

wuweiweiwu:
Some extra EOL white space were deleted. Hope that isn't a problem

wuweiweiwu:
This can go out with the next release

Issue #1872, opened 2017-09-20
Switch to suppress verbose build output?


alanhogan:
Our project added storybook's `start-storybook` command to our `Profile` so we can start it and a couple other servers with `nf` (Foreman).

This works great but builds cause an extreme amount of output in this manner:

<img width="722" alt="screen shot 2017-09-20 at 1 18 51 pm" src="https://user-images.githubusercontent.com/6808/30665768-041f1ab6-9e07-11e7-8933-ed5c7fc43a92.png">

Would it be possible for the CLI to take a switch (maybe `--quiet`) that would suppress verbose/progress update messages?

ndelangen:
Hey Alan! We've had this feature request before, I think we should add it, would you like to help us?

It would involve a small CLI change and a webpack config change.
Should be fairly easy to implement, do you want to give it a show?

If so, please use the release/3.3 branch as your base 👍 🙇

alanhogan:
I could take a look. Where in the codebase does `storybook-start` live / examine the flags it gets?

ndelangen:
Here's our guide to contribute:
https://github.com/storybooks/storybook/blob/master/CONTRIBUTING.md

Here's the code points your need:

https://github.com/storybooks/storybook/tree/master/app/react-native/src/bin

https://github.com/storybooks/storybook/blob/master/app/react/src/server/index.js
https://github.com/storybooks/storybook/blob/master/app/react/src/server/build.js

https://github.com/storybooks/storybook/blob/master/app/vue/src/server/index.js
https://github.com/storybooks/storybook/blob/master/app/vue/src/server/build.js

`/app/.../src/server/config/webpack.config.js` should contain the webpack config that has the progress plugin, that you want to disable.

I hope this helps,

Hypnosphi:
@ndelangen I changed all the links to point to `release/3.3` branch

mmmeff:
How could we go about overriding the build output via webpack.config.js? I'm in full control mode and attempting to override it with `config.devServer = { stats: 'errors-only' };` but no dice.

ndelangen:
@mmmeff What you need to do is filter out the ProgressPlugin if you want to remove it that way.

mmmeff:
@ndelangen Since webpack's plugins field is defined as an array, removing that plugin is less than ideal. Here's my working but sloppy fix for disabling the progress output.

```
  config.plugins = config.plugins.filter((plugin) => {
    // Find and omit the plugin that looks like progress plugin.
[...]

PR #3487, opened 2018-04-25
Suppress verbose build output

bmson:
Issue: #1872

## What I did
 This is bit of a stab in the dark to address the output noice in Storybook.
Added `--quiet` to the CLI command list, so we can conditionally load the `webpack.ProgressPlugin`.

## How to test
`yarn test --cli`

### DIFF
--- a/app/angular/src/server/config/webpack.config.js
+++ b/app/angular/src/server/config/webpack.config.js
@@ -17,7 +17,7 @@ import { includePaths, excludePaths, nodeModulesPaths, loadEnv, nodePaths } from
 import babelLoaderConfig from './babel';
 import { version } from '../../../package.json';
 
-export default function(configDir) {
+export default function(configDir, quiet) {
   const config = {
     mode: 'development',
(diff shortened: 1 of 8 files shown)


bmson:
The circleci errors seem unrelated to this PR »`Bad credentials`

Hypnosphi:
I've changed the PR base to `master` because we need to apply this change to mithril app as well. But it’s probably easier to open a new one than to resolve the conflicts

Please also document this option
https://github.com/storybooks/storybook/blob/master/docs/src/pages/configurations/cli-options/index.md

Hypnosphi:
Please update config for app/mithril

Hypnosphi (approved):

benniemosher:
Was this only applied to the `start-storybook` command? The docs don't have the `--quiet` option for `build-storybook` and the CLI is not accepting it. How hard would it be to get it done there too?

ndelangen:
It should work for both in V5 I think! Can you help us test?

PR #3515, opened 2018-04-30
Force re-render event

Hypnosphi:
Issue: #1736 introduced an ability to force rerender story, but you have to depend on a particular framework to use it. I've introduced a `'forceReRender'` event that addons could use in a framework-agnostic way.

This also introduces constants for core events as a separate package (to avoid circular dependency between `core` and `ui`)

### DIFF
--- a/lib/core/src/client/preview/start.js
+++ b/lib/core/src/client/preview/start.js
@@ -171,6 +171,10 @@ export default function start(render, { decorateStory } = {}) {
 
   renderUI();
   reduxStore.subscribe(renderUI);
+  const forceReRender = () => renderUI(true);
+  if (isBrowser) {
+    channel.on(Events.FORCE_RE_RENDER, () => renderUI(true));
+  }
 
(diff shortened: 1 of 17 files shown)


lib/core/src/client/preview/start.js
```diff
@@ -170,6 +171,10 @@ export default function start(render, { decorateStory } = {}) {
 
   renderUI();
   reduxStore.subscribe(renderUI);
+  const forceReRender = () => renderUI(true);
+  if (isBrowser) {
+    channel.on(Events.FORCE_RE_RENDER, () => renderUI(true));
```
tmeasday: Use the `forceRerender` const here?

tmeasday (approved):
Looks great, one minor tweak if you liek

igor-dv (approved):

ndelangen (approved):

lib/core/src/client/preview/start.js
Hypnosphi: Sure, forgot that

PR #3531, opened 2018-05-04
Ignore home package.json no license field

### LINKED ISSUE #3446: [storybook/cli] Fails if package.json at /home/user has no license field.
### Summary
```storybook/cli``` gives error when initiating a repo with ```$ getstorybook```.

#### Here's the error:
```
$ npx -p @storybook/cli getstorybook
  npx: installed 312 in 22.043s

  getstorybook - the simplest way to add a storybook to your project. 
 
  • Detecting project type. ✓
  • Adding storybook support to your "Create React App" based project
      ../../package.json: No license field

      ../../package.json: No license field

      ../../package.json: No license field
```

> #### Seems like ```getstorybook``` fails if ```package.json``` at ```/home/user``` does not have a ```licese``` field.

### Steps to reproduce
1. ```sudo npm i -g @storybook/cli```
2. ```getstorybook```

- ```storybook/cli: 3.4.2```
- Addons: none

Keraito:
Issue: Fixes #3446 

If the `/home/user` folder contained a `package.json` without a `license` field, the installation of `getstorybook` is cancelled. This warning is thrown regardless of anything in the current folder that you're running `getstorybook` in. 

Reproduction would be to [create a package.json without a license field in your home directory](https://github.com/storybooks/storybook/issues/3446#issuecomment-382553884), initialize a project that would normally work with `getstorybook`, like through `create-react-app`, and then run `getstorybook`.

This is caused by the [checking of the version of the generator dependencies](https://github.com/storybooks/storybook/blob/master/lib/cli/lib/latest_version.js#L8) which resulted in the warning being thrown and thus existing the installation at https://github.com/storybooks/storybook/blob/3539625d7f801a0fee1fb8c58f1ff57016c9513f/lib/cli/lib/latest_version.js#L34
(Thanks to @sunilhari for pointing this out).  
In fact, if you for whatever reason have a `package.json` at your home directory without a `license` field (or anywhere along the path towards the current directory), the warning will be thrown at any `yarn` command. The result of actually running the command, however, is then outputted after the warning:
image

## What I did
Using the knowledge from the screenshot above, I basically added an array of warning texts that will not stop the entire installation process.

This fix is entirely based on `yarn` and I did not check yet how this works with `npm` and if there are any differences. But I wanted to open this PR in order to see whether this solution is desired before spending more time on it.

### DIFF
--- a/lib/cli/lib/latest_version.js
+++ b/lib/cli/lib/latest_version.js
@@ -1,5 +1,6 @@
 import { spawn } from 'cross-spawn';
 import hasYarn from './has_yarn';
+import whitelistedWarnings from './whitelisted_warnings';
 
 const packageManager = hasYarn() ? 'yarn' : 'npm';
 
@@ -34,7 +35,11 @@ export default function latestVersion(packageName) {
     command.stderr.on('data', data => {
       // yarn error
       const info = JSON.parse(data);
-      reject(new Error(info.data));
+      whitelistedWarnings.forEach(whitelistedWarning => {
+        if (!info.data.includes(whitelistedWarning)) {
+          reject(new Error(info.data));
+        }
+      });
     });
   });
 }
--- a/lib/cli/lib/latest_version.js
+++ b/lib/cli/lib/latest_version.js
@@ -1,6 +1,5 @@
 import { spawn } from 'cross-spawn';
 import hasYarn from './has_yarn';
-import whitelistedWarnings from './whitelisted_warnings';
 
 const packageManager = hasYarn() ? 'yarn' : 'npm';
 
@@ -35,11 +34,9 @@ export default function latestVersion(packageName) {
     command.stderr.on('data', data => {
       // yarn error
       const info = JSON.parse(data);
-      whitelistedWarnings.forEach(whitelistedWarning => {
-        if (!info.data.includes(whitelistedWarning)) {
-          reject(new Error(info.data));
-        }
-      });
+      if (info.type !== 'warning') {
+        reject(new Error(info.data));
+      }
     });
   });
 }
--- a/lib/cli/lib/latest_version.js
+++ b/lib/cli/lib/latest_version.js
@@ -34,7 +34,7 @@ export default function latestVersion(packageName) {
     command.stderr.on('data', data => {
       // yarn error
       const info = JSON.parse(data);
-      if (info.type !== 'warning') {
+      if (info.type === 'error') {
         reject(new Error(info.data));
       }
     });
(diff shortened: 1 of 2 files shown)


Hypnosphi:
>  I did not check yet how this works with `npm`

Please do that

Hypnosphi:
Should we maybe just ignore `"type": "warning"`?

Keraito:
@Hypnosphi 
> Please do that

Running `npm info babel-runtime version` does not raise a warning for a missing license field in `package.json` in the home directory, so this is an `yarn` only issue as far as I can observe. Nonetheless, I also did a full `getstorybook` installation with `npm`, and all of it works.

While I did this, I noticed that the checking of the package versions is always done with yarn if it's installed (https://github.com/storybooks/storybook/blob/master/lib/cli/lib/latest_version.js#L4), and that the `--use-npm` flag is only pointed at the installation of dependencies. Is this intended? I'm not sure if old versions of yarn will cause issues with it, like what was seen with issues related to https://github.com/storybooks/storybook/pull/3453. 

> Should we maybe just ignore "type": "warning"?

That would be easier yes, but is every warning from `yarn`/`npm`/both ones that can be ignored for the installation of storybook? Asking because I don't have sufficient knowledge of all specifics of both package manager to be able to answer that.

Hypnosphi:
> While I did this, I noticed that the checking of the package versions is always done with yarn if it's installed (https://github.com/storybooks/storybook/blob/master/lib/cli/lib/latest_version.js#L4), and that the --use-npm flag is only pointed at the installation of dependencies.

It would make sense to include `--use-npm` check to `has_yarn` helper (and probably rename it to `use_yarn`
https://github.com/storybooks/storybook/blob/master/lib/cli/lib/has_yarn.js

> is every warning from yarn/npm/both ones that can be ignored for the installation of storybook?

In this particular helper we only want to catch the fact that a package isn't published.

Keraito:
> In this particular helper we only want to catch the fact that a package isn't published. User will still see the warnings when the actual installation will happen.

Makes sense, will change it to ignore yarn warnings in general then.

> It would make sense to include --use-npm check to has_yarn helper (and probably rename it to use_yarn

Does it make sense to put this change into another PR, or should I include it here?

Keraito:
> I'm ok with both, please do whatever you're more comfortable with

I will do it in another PR as at first glans it looks like the change will not be very trivial, isn't necessarily related to the original issue, and I don't want to block this patch cuz of it. 

> `info.type === 'error'` would be more explicit

Changed it

Hypnosphi (approved):

PR #3535, opened 2018-05-06
Use the --use-npm flag also for version checking

Keraito:
Issue: While I was working on #3531, I noticed that the package manager responsible for checking the versions of dependencies in the generators was dependent on `has_yarn`, fully ignoring the `--use-npm` flag https://github.com/storybooks/storybook/pull/3531#issuecomment-386759214.

## What I did
I made the `latestVersion` function to adhere to the `--use-npm` flag. 
How I currently did it is pass the `npmOptions` object all the way through generators in order to reach  `latestVersion`. 

- It works, but I wouldn't call this cleanest way of doing it, as it requires a change in all the generators. Is there a better way of accessing the `--use-npm` flag in the helper?
- It currently only works on a Create React App project, all the other generators will fail due to the destructuring in https://github.com/storybooks/storybook/compare/master...Keraito:use-npm?expand=1#diff-8d17cc91321b260aaa782e8a42cd5634R37, but I wanted to verify whether this is the best way or there is a more cleaner and less impactful way of doing it.


Hypnosphi:
I think it's ok

Keraito:
Tried to make sure that I didn't miss a generator or call to either `latestversion`/`getVersion`. I ran the tests in `lib/cli` and everything went well except for some things at the end, but that could also have been caused by some internet issues on my side. Will test again tomorrow.

Hypnosphi (approved):

Keraito:
Ran the tests again and now they succeeded (`  Done in 714.76s.`), yesterday was very likely to be caused by network issues.

Hypnosphi:
Please update generators for new apps (html, marko) as well

Keraito:
Done, I'm not in a position atm to run the tests again though, I can do that in a few hours.

Hypnosphi:

PR #3536, opened 2018-05-06
chore: add Node.js 10

DanielRuf:
## What I did
Add Node.js for CI builds to test the current stable Node.js version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ defaults: &defaults
   working_directory: /tmp/storybook
   docker:
     - image: circleci/node:8
+    - image: circleci/node:10
 
 version: 2
 dependencies:


DanielRuf:
> failed during stage 'building site': Deploy directory 'netlify-build' does not exist

It seems there is an issue with the Netlify setup in the last few builds.

> PR is not labeled with one of: ["cleanup","BREAKING CHANGE","feature request","bug","documentation","maintenance","dependencies:update","dependencies","other"]

No commitlint style? So should I label it `other: ...`?

DanielRuf:
> No commitlint style? So should I label it other: ...?

It seems no one applies this to the PRs / commits so where is the sense in this atm? Some commitlint / husky solution should be probably used to enforce this (if `--no-verify` is not used).

Hypnosphi:
Actually teamcity builds are running on `node:latest` image

wuweiweiwu:
Might have some unintended consequences. https://github.com/storybooks/storybook/issues/3529

ndelangen:
@wuweiweiwu testing in node 10 should not have any side-effects for users.

If I understand @Hypnosphi correctly we are already testing on Node8 on CircleCI and Node10 on TeamCity.

---

@DanielRuf You may commit with any message you like as long as it's decent. Our release process is mostly manual, but the commit process is beginner friendly.

ndelangen:
We run many test on both CircleCI and TeamCity. TeamCity runs some additional tests on examples.
We do not use TravisCI AFAIK.

> It seems there is an issue with the Netlify setup in the last few builds.

Indeed, I've been meaning to get to that and fix it for the past few days, haven't gotten around to doing that. The TeamCity example tests are mostly a improved version so it hasn't made top priority on my list.

PR #3555, opened 2018-05-07
Generic addon decorators

Hypnosphi:
Based on decorators for [`storybook/html`](https://github.com/storybooks/storybook/pull/3475)


addons/backgrounds/html.js
```diff
@@ -1 +1 @@
-module.exports = require('./dist/html');
+module.exports = require('./dist/deprecated');
```
Hypnosphi: I think we can keep those files for one alpha release

addons/backgrounds/mithril.js
```diff
@@ -1 +1 @@
-module.exports = require('./dist/mithril');
+module.exports = require('./dist/deprecated');
```
Hypnosphi: I think we can keep those files for one alpha release

addons/a11y/src/index.js
```diff
@@ -1,18 +1,35 @@
...
-function checkA11y(storyFn, context) {
+export const configureA11y = (options = {}) => {
+  axeOptions = options;
+};
+
+const runA11yCheck = () => {
   const channel = addons.getChannel();
-  return manager.wrapStory(channel, storyFn, context, axeOptions);
-}
+  const wrapper = document.getElementById('root');
```
tmeasday: Is the pattern we want to encourage? 

I feel like it is more inflexible than we might like, i.e it means stories can only ever be rendered in `#root`, whereas there are other conceivable ways you might want to render a story.

Could we instead do something like:

```js
import { afterStory } from '@storybook/core';

export const checkA11y = story => {
  addons.getChannel().emit(Events.REGISTER_SUBSCRIPTION, a11ySubscription);

  // The idea is that `afterStory` wraps a story in a callback that triggers with the story's 
  // DOM element after the story has rendered to the screen.
  const wrappedStory = afterStory(story, (html) => {
    runA11yCheck(html);
  });
 
 return wrappedStory();
};
```

Not sure if that's the best API, but does the idea make sense?

Hypnosphi: Makes sense, will try something like `STORY_RENDERED` event

Not sure that I get what you mean by other ways to render a story though

tmeasday: I guess what I mean is that at the moment you can call `const story = getStory()` on the story store and then execute that function to get something renderable (how you render that depends on the view layer of course). At the moment that function is also wrapped in the relevant decorators and those decorators will work however the story is rendered.

At the moment the only place a story is rendered is to the `#root` via the `renderMain` function, but other ways you might want to render a story is e.g. inside some `<Story/>` element in a documentation site.

Maybe I'm overthinking it, we would need to think about how decorators/addons make sense in that context a bit anyway, but it's something I think we will want to do in the future.

Hypnosphi: I think it's a pattern we could support in [SBNext](https://github.com/storybooks/SBNext)

CC @ndelangen

Hypnosphi: I've added `STORY_RENDERED` event to avoid `setTimeout(fn, 0)`, but I decided to keep `#root` part for now. Most of the addons, including a11y, have manager part, so they are unusable outside of storybook context. So I think that getting `rootEl` dynamically would bring more complexity than value right now

tmeasday: WFM, thanks for talking through it with me.

ndelangen (approved):

ndelangen (approved):

tmeasday (approved):
LGTM

wuweiweiwu (approved):
addon's are so much cleaner now!

Hypnosphi:

PR #3559, opened 2018-05-08
Refactor transitional decorator from addon-notes

tmeasday:
This should make it easy to transitional other addons to a parameter-based style


lib/addons/src/transitional-decorator.js
```diff
@@ -0,0 +1,34 @@
...
+
+    return wrapper(getStory, context, {
+      options,
+      parameters,
+    });
+  };
+
+  const hoc = deprecate(
+    options => story => context => decorator(options)(story, context),
+    `Passing stories directly into ${name}() is deprecated, instead use addDecorator(${name}) and pass options with the '${parameterName}' parameter`
```
Hypnosphi: Are we sure that we want to deprecate the old way? Honestly, it looks a lot more explicit to me

tmeasday: Oh, I thought we did?

I think the strongest argument in favour is if you want to pass parameters to more than one decorator for the same story:

```js
.add('story', withNotes('foo')(backgrounds(['red', 'white'])(() => <Story />)))

vs

.add('story', () => <Story />, { notes: 'foo', backgrounds: ['red', 'white'] })
```

I was planning on changing all the documentation to not mention the old way at least, so deprecating made sense to me, but I'm open to whatever.

igor-dv: What is a benefit of having the old API supported?

Hypnosphi: You may want to supply the same parameters (e.g. backgrounds) to multiple stories

And even extract it into variable or module, like here: 
https://github.com/storybooks/storybook/tree/master/addons/backgrounds#usage
("Of course it's easy to create a library module")

Hypnosphi:
```
.addDecorator(backgrounds(myBackgrounds))
.add('story', () => <Story />)
.add('other story', () => <OtherStory />)

vs

.add('story', () => <Story />, { backgrounds: myBackgrounds })
.add('otherStory', () => <OtherStory />, { backgrounds: myBackgrounds })
```

Hypnosphi: Or even global `addDecorator(backgrounds(myBackgrounds)` if you want them for all the stories (which is quite often the case)

Hypnosphi: It would be nice if parameters could be used to override common settings, like this:

```
.addDecorator(backgrounds(myCommonBackgrounds))
.add('story', () => <Story />)
.add('other story', () => <OtherStory />)
.add('special story', () => <SpecialStory />, {backgrounds: specialBackgrounds})
```

The implementation could be as simple as
```
const options = {...commonOptions, context.parameters[parameterName]}
```

tmeasday: @Hypnosphi you can!

```js
chapter.addParameters({ backgrounds: myBackgrounds })

// or globally
addParameters({ backgrounds: myBackgrounds });
```

tmeasday: Passing options to the the decorator (like `.addDecorator(backgrounds(myCommonBackgrounds))`) will pass those options as a separate argument to the the `wrapper` function, at which point the decorator can do what it wants with them.

My gut feeling is it confusing to have both:

```js
.addDecorator(backgrounds(myBackgrounds))

// OR, equivalent
.addDecorator(backgrounds)
.addParameters({ backgrounds: myBackgrounds })
```

As two ways to achieve the same thing, so I lean towards the second as it is more flexible (you can add the decorator at any level of the stack and the parameters at any level also).

Hypnosphi:
```
.addDecorator(backgrounds)
.addParameters({ backgrounds: myBackgrounds })
```

This is both more verbose and more magical =(

tmeasday: I mean my idea is that you would just add `addDecorators(backgrounds)` once in your `.storybook/config.js` once (equivalent to adding it to `.storybook/addons.js`), and then you would add the parameters to turn the addon on wherever you needed it.

Hypnosphi: OK this looks like a valid option. In this case we need to make sure that each addon is a noop when it lacks its parameters

tmeasday: Yeah! I was going to do this per-addon (notes already is a no-op if the params are not set), but we could also make this thing do it automatically. I guess it depends on whether the addon makes sense without params.

ndelangen: > I mean my idea is that you would just add addDecorators(backgrounds) once in your .storybook/config.js once (like you need to add it to .storybook/addons.js), and then you would add the parameters to turn the addon on wherever you needed it.

That's perfect I think.

Hypnosphi (approved):

lib/addons/src/transitional-decorator.js
Hypnosphi: @tmeasday can you please do the above in examples-official?

Hypnosphi: @tmeasday I just realised that those are two different usecases

```
// here withSomething is params => (storyFn, context) => output
.addDecorator(withSomething(params))

// here it is params => storyFn => () => output
.add('story', withSomething(params)(storyFn))
```

Actually addons that support both do some checks to figure out which usage is it.

Can we maybe just deprecate the latter but keep the former, because it's quite useful in cases like backgrounds?

Hypnosphi: Oh, now I see that the latter is the thing we're actually deprecating here, good

Hypnosphi: But the former would probably need to be supported when migrating other addons

Hypnosphi:
Looks like tests are failing because of this:

https://github.com/storybooks/storybook/blob/98a99b5c2f4b9c8e6189c2d4d3dd1faeedea58ed/addons/notes/src/__tests__/index.js#L4

tmeasday:
Ok, did another pass. Let me know what you think @Hypnosphi. I suspect we may have to tweak it as we apply the pattern to the various addons. 

Right now, you can use the `decorator` returned by `makeTransitionalDecorator({ ..., wrapper })` in the following ways:

```js
// 1. Pass parameters per-story, ala notes:

// We do this once
addDecorator(withNotes)

// Then, on every story, we do:
add('story', () => <Story />, { notes: 'Note' });

// In this case, wrapper is called with:
wrapper(story, context, { parameters: 'Note' });

// If a story does *not* set the parameters, the wrapper *IS STILL CALLED* but without options:
// It is the responsibility of the wrapper to ignore this story [A]:
wrapper(story, context, {});

// 2. Adding global parameters at "decoration-time", ala backgrounds:

// This is it:
addDecorator(withBackgrounds(['blue', 'red']));

// In this case, wrapper is called (per-story) with the options set [B]:
wrapper(story, context, { options: ['blue', 'red'] });

// 2a. [equivalent, C] Also, you can achieve the above with `addDecorator` + `addParameters`:
addDecorator(withBackgrounds)
addParameters({ backgrounds: ['blue', 'red'] });

// In this case, wrapper is called with parameters set [B]:
wrapper(story, context, { parameters: ['blue', 'red'] }); 

// 3. Per-story via wrapping directly (notes, deprecated):

.add('story', withNotes('Note')(() => <Story />));

// This will call the wrapper with options set:
// *BUT* will trigger deprecation warnings.
wrapper(story, context, { options: 'Note' });
```

Notes from above:
[A]: We could take responsibility to ignore stories instead? I'm guessing some decorators like `addon-info` might want this, and require users to manually option-out (`{info: { skip: true } }` perhaps?).

[B] We could reconcile the two here? (i.e. make it `parameters` in both cases?). Potentially the addon cares about the difference though?

[C] Although 2a. is a bit more verbose it is also more powerful (e.g. you can set background parameters for the whole app, for a specific chapter, *and* for a specific story, and the parameters mechanism handles reconciliation automatically).

lib/addons/src/transitional-decorator.js
```diff
@@ -17,18 +17,24 @@ export const makeTransitionalDecorator = ({ name, parameterName, wrapper }) => {
...
     // Used without options as .addDecorator(decorator)
     if (typeof args[0] === 'function') {
       return decorator()(...args);
     }
 
-    // Input are options, ala .add('name', decorator('note')(() => <Story/>))
-    return hoc(args[0]);
+    return (...innerArgs) => {
+      // Used as [.]addDecorator(decorator(options))
+      if (typeof innerArgs.length > 1) {
```
Hypnosphi: typeof?

lib/addons/src/transitional-decorator.js
```diff
@@ -0,0 +1,40 @@
+import deprecate from 'util-deprecate';
+
+// Create a decorator that can be used both in the (deprecated) old "hoc" style:
+//   .add('story', decorator(options)(() => <Story />));
+//
+// And in the new, "parameterized" style:
+//   .addDecorator(decorator)
+//   .add('story', () => <Story />, { name: { parameters } });
+
+export const makeTransitionalDecorator = ({ name, parameterName, wrapper }) => {
```
Hypnosphi: Should we rename it to something like `makePolymorphicDecorator`?

Hypnosphi:
[A] In viewport addon, parameters are optional, so we should allow addons to handle stories without parameters

lib/addons/src/transitional-decorator.js
```diff
@@ -0,0 +1,40 @@
...
+
+  return (...args) => {
+    // Used without options as .addDecorator(decorator)
+    if (typeof args[0] === 'function') {
+      return decorator()(...args);
+    }
+
+    return (...innerArgs) => {
+      // Used as [.]addDecorator(decorator(options))
+      if (typeof innerArgs.length > 1) {
```
Hypnosphi: typeof?

lib/addons/src/transitional-decorator.js
tmeasday: Sure! Or just `makeDecorator()`?

lib/addons/src/transitional-decorator.js
tmeasday: wow :0

tmeasday:
> [A] In viewport addon, parameters are optional, so we should allow addons to handle stories without parameters

Ok cool. We could also add an argument to the creator `skipIfNoParameters: true` or the like?

> [B] I'm Ok with both

👍 

> [C] I see. I still think that it wouldn't be too hard to support both versions

Oh, yeah, that is the plan. Sorry if I was confusing. I was saying we support both, even though there is some overlap.

lib/addons/src/transitional-decorator.js
ndelangen: I like the makeDecorator name too

lib/addons/src/transitional-decorator.js
tmeasday: I need to add a test for this modality, it just doesn't really make sense for `notes`. I'll do another PR for `backgrounds` and we'll definitely have one.

tmeasday:
Not sure what's happening with the UI change to the angular CLI.

I'll add a unit test to the `makeDecorator` function

tmeasday:
@Hypnosphi do you want to re-review this, I added an extra API and some tests

tmeasday:
(Or just merge away)

Hypnosphi (approved):

Hypnosphi:

PR #3568, opened 2018-05-10
remove nested braces in Story Source block

primetwig:
Issue:
Story Source block has a preview of code used. If preview has objects as properties, then you can not copy-paste it, because it renders redundant braces around nested values.
Preview: http://take.ms/alSVx

## What I did
I've added a prop `braceWrapped` for `PropVal` component, so braces get rendered for a first level entries only. Nested `PropVal` occurrences will no longer wrap themselves into braces.

## How to test
Look at the source of component that has deep object as a property.

Is this testable with Jest or Chromatic screenshots? yes
Does this need an update to the documentation? no

### DIFF
--- a/addons/info/src/__snapshots__/index.test.js.snap
+++ b/addons/info/src/__snapshots__/index.test.js.snap
@@ -700,8 +700,8 @@ exports[`addon Info should render <Info /> and external markdown 1`] = `
                                           <span
                                             style={Object {}}
                                           >
+                                            {
                                             <ThemedComponent
-                                              braceWrapped={true}
                                               maxPropArrayLength={3}
                                               maxPropObjectKeys={3}
                                               maxPropStringLength={50}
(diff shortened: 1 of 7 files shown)


Hypnosphi:
Please add an example with nested object to `examples/official-storybook`

ndelangen:
@primetwig Do you need any help with that?

primetwig:
I'm not sure what to add to that folder.
`examples/official-storybook/components/DocgenButton.js` already has nested props.

Hypnosphi:
You need to pass something to them in `stories/addon-info/stories.js`

Hypnosphi:
Sorry, it's `addon-info.stories.js`

primetwig:
I've removed the prop `braceWrapped` I was trying to add, because I've come with a way to use an existing prop `level`.
I've also added some props to the stories you've asked.

primetwig:
@Hypnosphi @ndelangen, is it all good now?

Hypnosphi (approved):
Had no chance to look through the snapshots diff, I hope it's all about shifting braces to another level

Hypnosphi:

PR #3569, opened 2018-05-10
fix ADDONS_SUPPORT.md view error

klimentru1986:
Issue: [ADDONS_SUPPORT.md](https://github.com/storybooks/storybook/blob/master/ADDONS_SUPPORT.md) now broken

## What I did

fix ADDONS_SUPPORT.md view error


igor-dv (commented):
Didn't know it is broken. But looks like everything should be shifted right.

Hypnosphi:
Still looks broken to me

Hypnosphi:
We probably have lost a column when merging marko and html branches. I'll recover it

PR #3576, opened 2018-05-13
Support webpack4 modules format

theKashey:
Issue: storybook 4.0.0 does not pass `filename` to story context, as long module.filename is not defined.

## What I did

I've changed module.filename to module.name, and added one test, to test `storyOf` function agains real `module` variable. (double checked in browser)

## How to test

As usual. But you might need run storybook to test integration with webpack in real.

Not sure about keeping old way for accessing fileName, but, look like, it's obsolete.

### DIFF
--- a/lib/core/src/client/preview/client_api.js
+++ b/lib/core/src/client/preview/client_api.js
@@ -85,7 +85,7 @@ export default class ClientApi {
       // wrap the first decorator and so on.
       const decorators = [...localDecorators, ...this._globalDecorators];
 
-      const fileName = m ? m.filename : null;
+      const fileName = m ? m.id : null;
 
       // Add the fully decorated getStory function.
       this._storyStore.addStory(kind, storyName, this._decorateStory(getStory, decorators), {
--- a/lib/core/src/client/preview/client_api.test.js
+++ b/lib/core/src/client/preview/client_api.test.js
@@ -214,10 +214,27 @@ describe('preview.client_api', () => {
   describe('reads filename from module', () => {
     const api = new ClientAPI();
     const story = jest.fn();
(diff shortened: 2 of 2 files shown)


igor-dv (commented):
Initially, the `fileName` was added to make it available during this storyshots a.k.a. running in node. 
Does it still happen (I assume that it does, since all the tests are passing)?

theKashey:
@igor-dv - there were no tests against "real" module variable. 
Meanwhile `nodejs` does contain both `filename` and `id`. Webpack actually does not include any of them, only `i` + "adapter", resulting `id` as a getter.

ndelangen:
Thank you @theKashey that's pretty advanced for a first PR 🙇 

What are potential negative side effects of this?

theKashey:
There is only one drawback - I could guarantee that this will work for old node.js versions, but could not say the same for the older webpack versions. But this is not the case.
But I am pretty sure that `.id` is the right way to access file name. I am confident about the change due to my contributions to dependency mocking libraries. And tests. 
As this PR is not changing anything, just fixing the property name, and I need non-emptyfilename to continue with [another PR](https://github.com/tsuyoshiwada/storybook-chrome-screenshot/issues/42).

ndelangen (approved):

ndelangen:
@igor-dv do you agree we can merge this?

igor-dv:
Yeah. Does it make sense to mark it as breaking change?

ndelangen:
Yes, 4.0.0 material for sure.

ndelangen:
Congrats on your first merged PR @theKashey

theKashey:
Will send more :)

Hypnosphi:

PR #3591, opened 2018-05-15
[other] help: allow webpack.config.babel.js

JaKXz:
## Issue:
I want to configure my custom webpack config with the latest syntax and have it transpile through babel, as webpack is fully capable of doing. 

The problem is that to import from my custom config, I already have `webpackBase.config.babel.js` files in my project.

## What I did
This PR is just the result of my digging through the codebase and trying to find where I would enable this manually. I would love if this was allowed though since webpack natively handles `.babel.js` files.

## How to test
Likely testable with Jest. This PR is just a POC / WIP /not mergeable atm but with help could get there.

Thanks!

### DIFF
--- a/lib/core/src/server/config.js
+++ b/lib/core/src/server/config.js
@@ -92,6 +92,7 @@ export default options => (configType, baseConfig, configDir) => {
     wrapInitialConfig = noopWrapper,
     wrapBasicConfig = noopWrapper,
     wrapDefaultConfig = noopWrapper,
+    webpackBabel = false,
   } = options;
 
   const config = wrapInitialConfig(baseConfig, configDir);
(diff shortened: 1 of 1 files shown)


ndelangen:
Hey @JaKXz thank you fo this PR.

I'd also want to support the newest syntax in all config files 👍 

I'm fairly sure the solution in this PR could not work:

https://github.com/storybooks/storybook/pull/3591/files#diff-f9dd41ea5873e4c8869d3f38bd83cc0bR115

that require is executed in Node directly, not via webpack, in fact the resulting object|function is passed to webpack as config.

So in order to this we'd have to use babel-register or similar, or fs-load and parse the webpack.config.js manually instead of require-ing it.

What do you think?

JaKXz:
I agree, I think your solution would work! :)

I don't fully understand what is going on under the hood, but, why can't we just pass the webpack config path straight to webpack and just let webpack handle it?

ndelangen:
We run webpack via the Node api. That needs a config object, not a path.

We extend and pass the config as an object|function to the user to extend before passing it into webpack.

That's why this hasn't been done (yet), because we're editing the object from the webpack config file at runtime, before passing it into webpack.

Hypnosphi:
We probably need something like `babel-register` for that. I wonder what does webpack use internally in this case

ndelangen:
@JaKXz I'd love this feature in, but this PR as it is, won't work.

Would love to help you with a solution that would work, what would you need?

PR #3607, opened 2018-05-18
fix(knobs): cancel debounced onChange on unmounting

kumarharsh:
Issue: The debounced onChange handlers in the knob components are not cancelled when the components are unmounted, which make cause edge cases where if the component is unmounted while the debounced function is still being called, an unhandled error might be thrown.

## What I did
Cancelled the debounced function on component unmount

## How to test
Too difficult to test reliably, but this is how it'd go:
1. In the kitchen-sink, use knobs to create some inputs which can be manipulated (text or number would be best)
2. In the knob/number component, change the debounce duration to be much bigger - say 2000ms (to demonstrate the edge case).
3. Open the storybook app and change the knob value really fast (press up key, for example)
4. Select some other component within 2 seconds - the issue should manifest itself.

### DIFF
--- a/addons/knobs/src/components/types/Array.js
+++ b/addons/knobs/src/components/types/Array.js
@@ -36,6 +36,10 @@ class ArrayType extends React.Component {
     this.onChange = debounce(this.props.onChange, 200);
   }
 
+  componentWillUnmount() {
+    this.onChange.cancel();
+  }
+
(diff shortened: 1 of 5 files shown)


ndelangen (approved):

Hypnosphi:

PR #3610, opened 2018-05-20
addon-viewport: Use the new parameterized way of decorators

mshaaban088:
## What I did
I used the new parameterized way of decorators described in #3559 

## How to test
yarn test

No

No, I just updated the already existing one in `official-storybook` examples

Yes and added

### DIFF
--- a/addons/viewport/src/preview/withViewport.js
+++ b/addons/viewport/src/preview/withViewport.js
@@ -1,4 +1,4 @@
-import addons from '@storybook/addons';
+import addons, { makeDecorator } from '@storybook/addons';
 import CoreEvents from '@storybook/core-events';
 import deprecate from 'util-deprecate';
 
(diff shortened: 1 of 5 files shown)


addons/viewport/src/preview/withViewport.js
```diff
@@ -9,38 +9,41 @@ import {
 } from '../shared';
 
 function noop() {}
-let handler = noop;
 
-const subscription = () => {
+const subscription = handler => () => {
```
mshaaban088: I noticed an issue when you create 2 stories with `subscription`/`onViewportChange`, the 2nd one receives the event twice every time you change the viewport, which should not be

Hypnosphi: Now it creates a new subscription on each rerender (`subscription(handler)`) is a new function each time

Hypnosphi: You may want to store `currentHandler` as module-level variable instead

addons/viewport/src/preview/withViewport.js
```diff
@@ -9,38 +9,41 @@ import {
...
-  const options = typeof nameOrOptions === 'string' ? { name: nameOrOptions } : nameOrOptions;
+const withViewport = makeDecorator({
+  name: 'withViewport',
+  parameterName: 'viewport',
+  wrapper: (getStory, context, { options, parameters }) => {
+    const storyOptions = parameters || options;
+    const viewportOptions =
+      typeof storyOptions === 'string' ? { name: storyOptions } : storyOptions;
+
+    applyViewportOptions(viewportOptions);
```
Hypnosphi: is it needed when there are neither options nor parameters?

addons/viewport/src/preview/withViewport.js
Hypnosphi: Can you please add a second example with `onViewportChange` so that everyoune could reproduce the issue

Hypnosphi: Oh I see, changing of `handler` variable prevents the subscription from removing. This should fix it without affecting performance:

```
const subscription = () => {
  const currentHandler = handler
  channel.on(VIEWPORT_CHANGED_EVENT_ID, currentHandler);
  return () => channel.removeListener(VIEWPORT_CHANGED_EVENT_ID, currentHandler);
};
```

mshaaban088: Yep, this should work.

addons/viewport/src/preview/withViewport.js
mshaaban088: No, not needed in such case

addons/viewport/src/preview/withViewport.js
```diff
@@ -9,38 +9,41 @@ import {
...
-    return context ? decorated() : decorated;
-  };
-}
+export default withViewport;
 
 export const Viewport = deprecate(({ children, ...options }) => {
-  setViewport(options);
+  applyViewportOptions(options);
   return children;
 }, `<Viewport> usage is deprecated, use .addDecorator(withViewport(viewport)) instead`);
```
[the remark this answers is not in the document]
mshaaban088: I updated it.
I wasn't aware of such method to be honest :)

addons/viewport/src/preview/withViewport.js
mshaaban088: @Hypnosphi do you still need the second example?

Hypnosphi: As you wish. Or maybe a testcase would be better, just to make sure it doesn't break in future

addons/viewport/src/preview/withViewport.js
Hypnosphi: Then you can pass `skipIfNoParametersOrOptions: true` to `makeDecorator`

Hypnosphi (approved):

addons/viewport/src/preview/withViewport.js
mshaaban088: I agree, test case is even better. Will add it

Issue #2423, opened 2017-12-04
Change the Theme of storybook


Mujaddadi:
I am interested in customizing the UI of storybook. I am talking about the styles not just layout. At the moment the chances of changing it are zero to none. The UI even don't use classes so we can't overwrite it. Is there is anyway to do it without forking the repo? If it is, I will appreciate.

majapw:
We would also be interested in this feature. We use storybook to show off our internal design language, and it would be nice if we could leverage the features of the design language in the storybook as well.

A high-level styles object with a font stack, default spacing, colors, etc. would be a good start. 

cc @noratarano

Mujaddadi:
I would like to work on it whenever I will get time, which I don't have at the moment.

Mujaddadi:
Not sure what's the point of the bot to close issues even if they are inactive. Doesn't solve the issue or fulfill feature request.

danielduan:
As an open source project, it's hard to accommodate every feature request ourselves. We use the bot to help close out any inactive issues so we spend more time coding and less time managing GitHub issues.

We do welcome any help we can get if anyone is interested in helping to implement support for theming. While there has been discussions about supporting theming, it isn't in our core roadmap in the near future.

uliana-sirotina:
Hi, we are also interested in adjusting the styles of storybook. If there could at least be a className for every part of the story. Like "story-header" etc.

ckeeney:
Is there a way to make stale bot ignore an issue?  An issue has been created for this topic a number of times:
- https://github.com/storybooks/storybook/issues/313
- https://github.com/storybooks/storybook/issues/428
- https://github.com/storybooks/storybook/issues/1107
- https://github.com/storybooks/storybook/issues/1360
- https://github.com/storybooks/storybook/issues/1576

The stale bot repeatedly closing these issues just fragments the conversation.

ndelangen:
+1 on that @ckeeney. This is on the roadmap, I kinda like @uliana-sirotina idea as a intermediate solution.

I could do that, but it's also possibly easy enough for someone new to do it. Anyone here interested in taking it on?

Mujaddadi:
I would like to do it as my first open source project. Although I am quite busy but its time to give back to the community. I will check it on this Sunday hopefully.

Keraito:
If you need some help at some point, @Mujaddadi, I can gladly help, just @ me!

Hypnosphi:
[...]

PR #3628, opened 2018-05-22
Storybook UI theming

ndelangen:
Issue: #2423 #3679

## What I did
- [x] Consolidate UI components (WIP)
- [x] Add theme variables
- [x] implement a theme
- [x] implement a second theme
- [x] add ability to set theme from config.js

## Also
- Remove knobs from URL (still able to load from URL, but they get cleared)
- Remove debounces from knobs input


lib/addons/src/index.js
```diff
@@ -52,7 +53,16 @@ export class AddonStore {
   }
 
   addPanel(name, panel) {
-    this.panels[name] = panel;
+    // supporting legacy addons, which have not migrated to the active-prop
+    const original = panel.render;
```
tmeasday: Does this technique work reliably? What if panel is a stateless functional component?

Perhaps we can check for `propTypes`?

lib/components/src/tabs/tabs.stories.js
```diff
@@ -0,0 +1,91 @@
...
+panels.test3.propTypes = panelProps;
+panels.test4.propTypes = panelProps;
+panels.test5.propTypes = panelProps;
+panels.test6.propTypes = panelProps;
+
+const onPanelSelect = action('onPanelSelect');
+
+storiesOf('Components|Tabs', module)
+  .addDecorator(s => <div style={{ height: 'calc(100vh - 20px)' }}>{s()}</div>)
+  .add('statefull - no initial', () => <TabsState panels={panels} />)
```
tmeasday: stateful

tmeasday (commented):
Am I missing something? I don't see the themes or support for them via `config.js` in here. 

This PR seems more about refactoring the addons component to use the `active` parameter, which seems fine in itself.

lib/addons/src/index.js
ndelangen: the `render` should be just a function as per the current docs. Sure in react a function === component, but neither we nor anyone in the ecosystem has put propTypes on that function.

Class-based components as render here, are not supported, not before of after this refactor.

ndelangen:
@tmeasday Indeed, This PR so far is mainly consolidating, I'll try and consolidate some more components and then start adding theming to them.. It doesn't make a lot of sense to add theming in 60 places and then refactor then after I think.

I'll continue working on this PR, so there's more to come!

tmeasday:
Why don't we whack an "in progress" label on it so others like me don't get confused

ndelangen:
Ow I should have applied the label indeed, I prefixed my PR with "WIP" to indicate.

Still I love early reviews, so thanks, all feedback is appreciated!

addons/a11y/src/components/Panel.js
```diff
@@ -1,65 +1,93 @@
...
-            label: <Passes>{passes.length} Passes</Passes>,
-            panel: <Report passes items={passes} empty="No a11y check passed" />,
-          },
-        ]}
-      />
-    );
+    return active ? (
+      <div>
+        <Tabs
+          tabs={[
```
shilman: Implement this as `children`?

addons/knobs/src/KnobManager.js
```diff
@@ -1,5 +1,5 @@
 /* eslint no-underscore-dangle: 0 */
-import deepEqual from 'deep-equal';
+import equal from 'fast-deep-equal';
```
shilman: `equal` => `deepEqual`?

addons/knobs/src/KnobStore.js
```diff
@@ -12,7 +12,9 @@ export default class KnobStore {
     this.store[key] = value;
     this.store[key].used = true;
     this.store[key].groupId = value.groupId;
-    this.callbacks.forEach(cb => cb());
+
+    clearTimeout(this.timer);
```
shilman: comment this? non-obvious

addons/knobs/src/components/Panel.js
```diff
@@ -146,14 +130,22 @@ export default class Panel extends React.Component {
...
+              onFieldClick={this.handleClick}
+            />
+          </TabWrapper>
+        ),
         title: knobKeyGroupId,
       };
     });
 
-    if (groupIds.length > 0) {
+    if (groupIds.length >= 0) {
```
shilman: what happens in the `== 0` case?

addons/knobs/src/components/types/Date/index.js
```diff
@@ -1,73 +1,50 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import styled from 'react-emotion';
+
 import Datetime from 'react-datetime';
-import insertCss from 'insert-css';
-import debounce from 'lodash.debounce';
-import style from './styles';
+// import { Input } from '@storybook/components';
```
shilman: remove?

addons/a11y/src/components/Panel.js
ndelangen: Well I could I'm unsure it would be an improvement. As each tab gets a `active` prop passed from the `Tabs`|`TabState` component.

Do you have an implementation in mind, that looks better than this?

addons/knobs/src/KnobStore.js
ndelangen: Knobs can now stop rate limiting themselves because it's mostly handled here.

You can notice this has massive positive impact on the responsiveness and performance of knobs overall.

addons/knobs/src/components/Panel.js
ndelangen: Sorry, left a bit of a mess here, the length of an array is always `>= 0`..

What happens is:

addons/a11y/src/components/Panel.js
Hypnosphi: This would look better to me:

```
<Tabs>
  <Tab label={<Violations>{violations.length} Violations</Violations>}>
    <Report passes={false} items={violations} empty="No a11y violations found." />
  <Tab>
  <Tab label={<Passes>{passes.length} Passes</Passes>}>
    <Report passes items={passes} empty="No a11y check passed" />
  </Tab>
</Tabs>
```

To implement it, you can just change [those lines](https://github.com/storybooks/storybook/blob/c49dff998321bdb8447cd6b19d0a30b2fb67eaf0/addons/a11y/src/components/Tabs.js#L65-L74) to

```
{React.Children.map(children (tab, index) => (
  <Item
    // eslint-disable-next-line react/no-array-index-key
    key={index}
    active={active === index ? true : undefined}
    onClick={() => this.onToggle(index)}
  >
    {tab.props.label}
  </Item>
))}
```

Hypnosphi: Ideally, we could also use some tab identifier other than index

addons/knobs/src/KnobStore.js
shilman: Right. I mean it could use a comment in the code, so that a reader can understand why it's done this way.

lib/addons/src/index.js
```diff
@@ -52,7 +53,16 @@ export class AddonStore {
...
+    if (original && original.toString() && !original.toString().match(/active/)) {
+      this.panels[name] = {
+        ...panel,
+        render: ({ active }) => TabWrapper({ active, render: original }),
```
tmeasday: I feel pretty confused about what is going on here. Isn't this throwing away most of the implementation of the original panel? In the docs we talk about `componentDidX` etc -- creating a new component that just contains the `render` function isn't going to work in general.

lib/addons/src/index.js
tmeasday: Can't we just check if `panel` is a function, and use it as `original` if so? Then we'd support SFCs too.

> Sure in react a function === component, but neither we nor anyone in the ecosystem has put propTypes on that function.

Not sure what you mean here.

If the user is using es6 classes (as our docs current use) or a SFC, they may have `panel.propTypes = { active: ... }`. Especially *if we tell them to in the new docs!*.

I would have thought a check of `panel.propTypes && !panel.propTypes.active` is a much better indicator of not using the new API than the fairly unreliable string check of the function body.

app/angular/src/demo/welcome.component.ts
```diff
@@ -46,34 +46,34 @@ import { Component, Output, EventEmitter } from '@angular/core';
...
-        main {
-          margin: 15px;
-          max-width: 600;
-          lineHeight: 1.4;
-          fontFamily: "Helvetica Neue", Helvetica, "Segoe UI", Arial, freesans, sans-serif;
-        }
+      main {
+        margin: 15px;
+        max-width: 600;
+        lineheight: 1.4;
```
tmeasday: I'm guessing this is a prettier change? Is this a bug? (shouldn't it be `line-height`?)

docs/src/pages/addons/writing-addons/index.md
```diff
@@ -107,33 +107,27 @@ Then add the following code to the register.js.
...
-    overflow: 'auto',
-  }
-};
+const NotesPanel = styled('div')({
+  margin: 10,
+  width: '100%',
+  overflow: 'auto',
+});
 
 class Notes extends React.Component {
```
tmeasday: I think we should add `propTypes` to the example ;)

tmeasday: Also, do you want to mention themes in this doc?

docs/src/pages/configurations/theming/index.md
```diff
@@ -0,0 +1,83 @@
...
+  background: 'url("/path/to/logo.svg")',
+}
+```
+
+The styles provided here support everything [emotion](https://emotion.sh/) does. So that included things like nested selectors!
+
+## Adding more theme variables for addons
+
+If addons have a need for specific theme variables, the user has to add them. 
+We advice addons to reuse the variables listed above as much as possible.
```
tmeasday: s/advice/advise/

tmeasday (commented):
Looks great! I reckon the UI looks better with the new normal theme :)

Keraito:
Looks super awesome! Have a few remarks and questions regarding the extent that the theming should go. The new dark theme mode looks super nice! 🎉 

## Extent of theming

So right now, the theming also flows into the different addon tabs/panels, adjusting their style (like fontcolor to white and backgroundcolor to black most noticable) in different ways. The theme does not flow into the preview area, which makes sense in most cases because those are the components of the user and we won't want to mess with those. However, there are a few addons that make use of the preview area (based on the examples in `official-storybook`):

- a11y (Typography)
- Logger
- Info (Both inline and with the button)
- Storyshots
- Viewport

All of these just render some text of some kind, which are not affected by theming and thus create black text on a black theme. Or in the case of `info` for example, still stays entirely white. So the `info` page is basically the same as before, and `inline` is kind of like a zebra 
image

While not all of the addons I mentioned need this, I think that it would be great to also let the theming flow into the different addons so that stuff that gets rendered *by the addon* into the preview area will match the rest of the UI.

There's also the GraphQL addon. That's kind of a different case since it uses the `graphiql`. I seems possible to apply an editor theme to that: https://github.com/graphql/graphiql#applying-an-editor-theme, but I'm not sure if that covers all the UI as well. But whether we want this is up for debate.


## Some remarks

The strip to highlight the selected addon/tab is not that visible, compared to the normal theme.
image
image

The same for these 2 buttons, they basically look as 1 button right now:
image
image

Should the shortcut page also be black with white text?
image


This was what I got from playing and looking around in the `official-storybook`. In general, it looks super awesome 😍 Love it that you can just change the theme instantly with a single change in option.

addons/knobs/src/components/Panel.js
```diff
@@ -169,33 +156,32 @@ export default class Panel extends React.Component {
...
-          <GroupTabs
-            groups={groups}
-            onGroupSelect={this.onGroupSelect}
-            selectedGroup={this.state.groupId}
-          />
-        )}
-        <PanelInner>
+        {groupIds.length > 0 ? (
+          <TabsState panels={groups} />
+        ) : (
```
Keraito: The `PropForm` component went from always being rendered to conditionally being rendered, is that intended?

lib/ui/package.json
```diff
@@ -38,7 +39,7 @@
     "react-icons": "^2.2.7",
     "react-lifecycles-compat": "^3.0.4",
     "react-modal": "^3.4.4",
-    "react-treebeard": "^2.1.0"
+    "@ndelangen/react-treebeard": "^2.1.0"
```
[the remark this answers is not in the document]
ndelangen: I have been given ownership of the original package, and will be updating and publishing it when I find the time.

lib/ui/src/modules/ui/components/search_box.stories.js
```diff
@@ -4,7 +4,7 @@ import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 import { document, navigator } from 'global';
 
-import SearchBox from './search_box';
+import Component from './search_box';
```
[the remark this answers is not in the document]
ndelangen: eslint import rule 😞

app/angular/src/demo/welcome.component.ts
ndelangen: indeed it should

tmeasday: I guess the problem was there before ;)

addons/knobs/src/KnobManager.js
ndelangen: ✅

addons/knobs/src/KnobStore.js
ndelangen: ✅

lib/addons/src/index.js
ndelangen: This works just fine, TabWrapper is called when render is called. TabWrapper calls the render prop.

```js

File: app/angular/src/demo/welcome.component.ts (hunk 9 above)
Comment (ndelangen) [reply app/angular/src/demo/welcome.component.ts]: ✅

File: docs/src/pages/addons/writing-addons/index.md (hunk 10 above)
Comment (ndelangen) [reply docs/src/pages/addons/writing-addons/index.md]: I'll mention styling addon components and `lib/components` in this doc.

Do you want to mention/change the section about this:

File: docs/src/pages/configurations/theming/index.md (hunk 11 above)
Comment (ndelangen) [reply docs/src/pages/configurations/theming/index.md]: ✅

File: lib/addons/src/index.js (hunk 8 above)
Comment (tmeasday) [reply lib/addons/src/index.js]: Yeah but about the other lifecycle methods on `panel`?

File: docs/src/pages/addons/writing-addons/index.md (hunk 10 above)
Comment (tmeasday) [reply docs/src/pages/addons/writing-addons/index.md]: Yeah, let's do this as a separate PR, I'll make a note to do it.

Comment (ndelangen) [reply docs/src/pages/addons/writing-addons/index.md]: 👍

File: lib/ui/package.json (hunk 13 above)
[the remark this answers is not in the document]
Comment (ndelangen) [reply lib/ui/package.json]: I can can give ownership of this package to others too if they want to pick this up 👍

File: addons/knobs/src/components/Panel.js (hunk 12 above)
Comment (ndelangen) [reply addons/knobs/src/components/Panel.js]: Indeed, deliberate change

File: lib/ui/src/modules/ui/components/search_box.js [hunk 15]
```diff
@@ -62,12 +62,8 @@ const suggestionTemplate = (props, state, styles, clickHandler) =>
   ));
 
 class SearchBox extends React.Component {
-  constructor(props) {
-    super(props);
-
-    const { theme = {} } = this.props;
-
-    this.state = {
+  static getDerivedStateFromProps({ theme }) {
```
Hypnosphi: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

Hypnosphi: Note that in React 16.4 `gDSFP` is called as often as `render`, so there's actually no gain in using it like that

addons/knobs/src/KnobStore.js
```diff
@@ -12,7 +15,12 @@ export default class KnobStore {
     this.store[key] = value;
     this.store[key].used = true;
     this.store[key].groupId = value.groupId;
-    this.callbacks.forEach(cb => cb());
+
+    // debounce the execution of the callbacks for 50 milliseconds
+    if (this.timer) {
```
danielduan: please use `lodash.debounce`. i believe it's already a dependency for knobs

danielduan:
@ndelangen thanks for doing this. Is there any way we can break this PR down into multiple smaller PRs?

Smaller PRs tend to solicit better feedback and this one is on the verge of unreviewable.

lib/addons/src/index.js
igor-dv: I can see that all the addons are implemented with something like `active => active ? <X> : null` (even in the docs) . What are the other ways of using the `active` field (What am I missing) ?
I just don't understand why we need to controll this. Why just not to use `TabWrapper` always and hide the "activation" from the users?

examples/official-storybook/config.js
```diff
@@ -15,6 +15,7 @@ addHeadWarning('Dotenv file not loaded', 'dotenv-file-not-loaded');
 setOptions({
   hierarchySeparator: /\/|\./,
   hierarchyRootSeparator: /\|/,
+  theme: themes.dark,
```
igor-dv: Will be this a final api of providing the theme?

ndelangen: Yeah seems like the right choice for now to me.

addons/knobs/src/KnobStore.js
ndelangen: @danielduan not anymore:

igor-dv:
My question here - https://github.com/storybooks/storybook/pull/3628#discussion_r195919953  is a bit hidden, but I would like to understand anyway =)

ndelangen:
@danielduan I completely understand your concern, I will try to do better in the future. There was so much to improve ;)

Do you see any reasons NOT to merge this? @storybooks/team

addons/a11y/src/components/Panel.js
ndelangen: just pushed a refactor to make this component behave more like your proposal. Do you like it?

lib/addons/src/index.js
ndelangen: @igor-dv The usecase is:

ndelangen:
@tmeasday :
> If the user is using es6 classes (as our docs current use) or a SFC, they may have panel.propTypes = { active: ... }. Especially if we tell them to in the new docs!.

We'd be forcing addon developers to adopt propTypes... Not everyone uses proptypes.

ndelangen:
@Hypnosphi thanks, I refactored the Tabs component, do you like it better this way?

Hypnosphi (approved):

igor-dv (approved):

addons/actions/src/manager.js
```diff
@@ -8,7 +8,8 @@ export function register() {
     const channel = addons.getChannel();
     addons.addPanel(PANEL_ID, {
       title: 'Action Logger',
-      render: () => <ActionLogger channel={channel} api={api} />,
+      // eslint-disable-next-line react/prop-types
```
RWOverdijk: Same here.

addons/backgrounds/package.json
```diff
@@ -27,7 +27,6 @@
     "@storybook/addons": "4.0.0-alpha.10",
     "@storybook/core-events": "4.0.0-alpha.10",
     "babel-runtime": "^6.26.0",
-    "emotion": "^9.1.3",
```
RWOverdijk: When you remove emotion, what's left is abstract emptiness. Mah hart. Mah sole...

addons/backgrounds/src/BackgroundPanel.js
```diff
@@ -130,8 +129,11 @@ export default class BackgroundPanel extends Component {
   }
 
   render() {
+    const { active } = this.props;
     const backgrounds = [...this.state.backgrounds];
-
+    if (!active) {
+      return null;
+    }
     if (!backgrounds.length) return <Instructions />;
```
RWOverdijk: Newline between statements, and curly braces? Not sure what the rules here are.

addons/backgrounds/package.json
ndelangen: 💔

addons/actions/src/containers/ActionLogger/index.js
```diff
@@ -56,21 +56,25 @@ export default class ActionLogger extends React.Component {
   }
 
   render() {
+    const { active } = this.props;
     const props = {
       actions: this.state.actions,
       onClear: () => this.clearActions(),
     };
-    return <ActionLoggerComponent {...props} />;
+    return active ? <ActionLoggerComponent {...props} /> : null;
```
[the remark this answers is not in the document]
ndelangen: What would your solution look like?

An if statement above?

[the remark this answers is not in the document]
RWOverdijk: Yeah :) 

```
if (!active) {
  return null;
}

return <ActionLoggerComponent {...props} />;
```

I do this because I can change the two conditions easier if needed and it makes it more readable imho.

addons/a11y/src/register.js
```diff
@@ -5,12 +5,12 @@ import Panel from './components/Panel';
...
 function init() {
-  addons.register(ADDON_ID, () => {
+  addons.register(ADDON_ID, api => {
+    const channel = addons.getChannel();
     addons.addPanel(PANEL_ID, {
       title: 'Accessibility',
-      render() {
-        return <Panel />;
-      },
+      // eslint-disable-next-line react/prop-types
```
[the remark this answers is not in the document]
ndelangen: I'm passing down a function, not a component..

Or is the comment regarding the switched based on `active`?

addons/backgrounds/src/BackgroundPanel.js
Hypnosphi: If eslint passes then it's OK

PR #3635, opened 2018-05-23
React Native Refactored list view

Gongreg:
Moved React native story list view to use sectionList instead of listview.
The old implementation is already deprecated, also it didn't update selected story.


ndelangen (approved):
Code looks good to me

PR #3636, opened 2018-05-23
Using only single channel for react native

Gongreg:
Issue:

Made sure that only one channel is used when working with React Native, so we can reach all events everywhere. If running without storybook server the channel is set to event emitter, otherwise it is websockets. This PR extends PR #3278. So it also solves websockets error.

I am not sure if this is a breaking change or not, I guess it is not common for user to use both websockets and on device ui at the same time. I don't want to mark this as a breaking change.


ndelangen (approved):
LGTM

PR #3639, opened 2018-05-23
Don't scroll story on knob change

Hypnosphi:
Issue: When changing knobs, example is scrolled to top

### DIFF
--- a/examples/official-storybook/config.js
+++ b/examples/official-storybook/config.js
@@ -17,11 +17,7 @@ setOptions({
   hierarchyRootSeparator: /\|/,
 });
 
-addDecorator(Story => (
-  <ThemeProvider theme={themes.normal}>
-    <Story />
-  </ThemeProvider>
-));
+addDecorator(story => <ThemeProvider theme={themes.normal}>{story()}</ThemeProvider>);
 
 configureViewport({
   viewports: {
--- a/lib/core/src/client/preview/start.js
+++ b/lib/core/src/client/preview/start.js
@@ -139,8 +139,10 @@ export default function start(render, { decorateStory } = {}) {
       return;
     }
 
-    // Scroll to top of the page when changing story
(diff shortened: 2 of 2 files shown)


ndelangen (approved):

Hypnosphi:

Primigenus:
I don't know if this change is what caused it, but now I'm seeing the page scroll to the bottom (in fact, all scrollable elements on the page get scrolled to the bottom/to the right) when changing a knob. From your code change it doesn't seem like it would be related, but perhaps the reason the code was there was to offset this behavior?

Hypnosphi:
I can't reproduce it in our [official examples](http://storybooks-official.netlify.com/?knob-Dollars=12.5&knob-Name=Storyteller&knob-Years%20in%20NY=9&knob-background=%23ffff00&knob-Age=15&knob-Items%5B0%5D=Laptop&knob-Items%5B1%5D=Book&knob-Items%5B2%5D=Whiskey&knob-Birthday=1484870400000&knob-Nice=true&knob-Styles=%7B%22border%22%3A%223px%20solid%20%23ff00ff%22%2C%22padding%22%3A%2210px%22%7D&knob-Fruit=apple&selectedKind=Addons%7CKnobs.withKnobs&selectedStory=tweaks%20static%20values&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs), can you?

ndelangen:
I can't either

Primigenus:
Hmm. I did some testing to figure out what's different on my end and it looks like it's the addon-a11y decorator, which is running each re-render and causing everything to scroll. Has something changed about it recently? I've had the addon since I started this storybook and it's only recently started doing this. (There is also a significant performance hit associated with axe running each time you change a knob -- ah, just found #3683)

PR #3650, opened 2018-05-25
Force render on knob button click and update example

Keraito:
Issue:  As discussed on Slack, [the callback of the button knob didn't seem to do something](https://storybooks.slack.com/archives/C55DBHZL0/p1526589520000555). What actually happened was that clicking the button didn't trigger a re-render, so the results of the callback were never visually visible.

There was an example in the official-storybook, but that basically used the knob button inside a react component defined in the stories file and triggered a `setState` call to force a render of the component itself. The example itself described that this was not the best way to do this.

## What I did
1. Force a re-render on button click.
2. Change the example to make the knob button change the values of the injected props to the component.

cc @ndelangen as he was part of the discussion


Hypnosphi (approved):

Hypnosphi:
Feel free to merge after CI passes

Hypnosphi:
Please update storyshots: `yarn test --core --update`

pvinis:
this doesnt seem to work for react native..

PR #3655, opened 2018-05-27
Refactor webpack and babel configs to core

Hypnosphi:
Issue: https://github.com/storybooks/storybook/issues/2240

This also simplifies API for apps, making it a single options object
Looks like this is the last thing to refactor (putting aside RN which is quite specific)


shilman (approved):
LGTM but did not test!!

ndelangen (approved):

Issue #3657, opened 2018-05-28
Support Request: Support storysource to Prettier v1.13.0


isoppp:
### Request Support Prettier v1.13.0

Prittier default parser option changed  babylon to empty and if settting empty, ouput error.

See: https://prettier.io/blog/2018/05/23/1.13.0.html#don-t-default-to-the-javascript-parser-4528-by-duailibe

### Steps to reproduce

New create using storysource project and run.

Out this error:

```
Error: No parser and no file path given, couldn't infer a parser.
```

### platform and version

- Mac 10.12.6
- Node.js v8.9.4

- "@storybook/vue@^3.4.6"
- "@storybook/addon-storysource@^3.4.6"

### Where to start

https://github.com/storybooks/storybook/blob/master/addons/storysource/src/loader/default-options.js

I tried add `parser: 'babylon'` in default option. And it works.
But I do not know if this modification alone is enough.

Add:
prettier changed output from error to warning in 1.13.1

golgistudio:
Referencing the related prettier issue.

https://github.com/prettier/prettier/issues/4567

isoppp:
@igor-dv
I think this need fix. But there's no rush.
Because Prettier changed to warning, but will change again error in the future.

Now this is output:
> No parser and no filepath given, using 'babylon' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.   

But if [this commit](https://github.com/storybooks/storybook/commit/6b3604c72588028f5a6e97f737eedac4a4b7f109) is enough (locke Prettier version forever) , don't need it.

Hypnosphi:
> But if this commit is enough

That was a temporary solution until your PR gets merged

isoppp:
It was merged, I close this.

@Hypnosphi 
Thanks for reviewing many times :bow:

Hypnosphi:
Released as [`4.0.0-alpha.9`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.9)

PR #3660, opened 2018-05-28
storysource: Add default parser option. Support prettier v1.13.0

isoppp:
Issue: #3657 

## What I did

Add default parser option for prettier v1.13.0 in storysource

## How to test

> Is this testable with Jest or Chromatic screenshots?

Probably can't.

> Does this need a new example in the kitchen sink apps?

No.

> Does this need an update to the documentation?

No.

## Note

> I tried add parser: 'babylon' in default option. And it works.
> But I do not know if this modification alone is enough.

I try to make a PR of this comment on the ticket


Hypnosphi:
Looks like it doesn't recognize typescript now:

```
[12:02:47][Step 2/2]   ● inject-decorator › positive - ts › encountered a declaration exception
[12:02:47][Step 2/2] 
[12:02:47][Step 2/2]     SyntaxError: Unexpected token (10:11)
[12:02:47][Step 2/2]        8 | })
[12:02:47][Step 2/2]        9 | class WithStoreComponent {
[12:02:47][Step 2/2]     > 10 |   private store: Store<any>;
[12:02:47][Step 2/2]          |           ^
[12:02:47][Step 2/2]       11 | 
[12:02:47][Step 2/2]       12 |   constructor(store: Store<any>) {
[12:02:47][Step 2/2]       13 |     this.store = store;
[12:02:47][Step 2/2] 
[12:02:47][Step 2/2]       65 | }
[12:02:47][Step 2/2]       66 | 
[12:02:47][Step 2/2]     > 67 | export function generateStorySource({ source, ...options }) {
[12:02:47][Step 2/2]       68 |   let storySource = source;
[12:02:47][Step 2/2]       69 | 
[12:02:47][Step 2/2]       70 |   storySource = generateSourceWithoutUglyComments(storySource, options);
```

CC @igor-dv

addons/storysource/src/loader/default-options.js
```diff
@@ -5,6 +5,7 @@ const defaultOptions = {
     bracketSpacing: true,
     trailingComma: 'es5',
     singleQuote: true,
+    parser: 'babylon',
```
Hypnosphi: We actually [add parser option](https://github.com/storybooks/storybook/blob/master/addons/storysource/src/loader/generate-helpers.js#L33-L38) on call site

So we could just remove `parser !== 'javascript'` check, and replace `'javascript'` string with `'babylon'` in rest of the addon codebase

But I think a more elegant solution would be to pass `filepath` option to prettier so that it could infer the correct parser from extension. You can get file path inside [transform function](https://github.com/storybooks/storybook/blob/master/addons/storysource/src/loader/index.js#L6) as [`this.resourcePath`](https://webpack.js.org/api/loaders/#this-resourcepath), add it to options object, and then pass to prettier instead of parser

Hypnosphi: Both solutions would fix https://github.com/storybooks/storybook/pull/3660#issuecomment-392971907

isoppp: Thank you for information! I will try it.

isoppp:
I fixed one point with additional.
Since Prettie's parser was similarly changed in version 0.13,
When the test was executed, the function called in `parser-js.js` outputted `not function` error.

And I did not do the test locally. I'm sorry.
In this time, we confirmed that errors are not generated by executing the test locally.

However, it seems that errors occur in places which are not touched by CircleCI.
How can we solve this?
Tell me if you have a good idea.

addons/storysource/src/loader/default-options.js
isoppp: I fixed this.
Pass `this.resourcePath` to prettier config.

Thank you for telling me the information.
It was very helpful in fixing it.

addons/storysource/package.json
```diff
@@ -25,7 +25,7 @@
     "babel-runtime": "^6.26.0",
     "estraverse": "^4.2.0",
     "loader-utils": "^1.1.0",
-    "prettier": "~1.12.1",
+    "prettier": "^1.12.1",
```
Hypnosphi: Please update it to latest version: looks like we don't support `1.12` after change in parser imports

addons/storysource/src/loader/generate-helpers.js
```diff
@@ -27,14 +27,21 @@ function generateSourceWithoutUglyComments(source, { comments, uglyCommentsRegex
...
-  if (!config.parser && parser && parser !== 'javascript') {
-    config = {
-      ...prettierConfig,
-      parser,
-    };
+  if (!config.parser) {
+    if (parser) {
+      config = {
+        ...prettierConfig,
+        parser: parser === 'javascript' ? 'babylon' : parser,
```
Hypnosphi: Not needed anymore

isoppp: I thought the same.
But in [this test](https://github.com/storybooks/storybook/blob/master/addons/storysource/src/loader/inject-decorator.test.js) , if the extension is txt and the contents are something else, prettier does not support to txt file, so it gets an error.

error:
```
No parser could be inferred for file: (LOCAL_PATH)/addons/storysource/src/loader/__mocks__/inject-decorator.stories.txt
```

Should i change the test?
I think that probably this is a rare case, so I think that do not have to deal with it unless  need it.

Hypnosphi: After giving it a second thought, we probably should always pass parser to prettier when user specifies it explicitly, which is exactly what we do now. So let's keep it as is

Hypnosphi: But we probably should put `if (filepath)` inside `if (!config.parser)`, to cover cases when parser is not passed

Hypnosphi: Or maybe we should just pass `babylon` when `parser` isn't set explicitly, because this is what we use by default ourselves. In this case, `filepath` will be never needed

addons/storysource/src/loader/generate-helpers.js
```diff
@@ -28,20 +28,25 @@ function generateSourceWithoutUglyComments(source, { comments, uglyCommentsRegex
...
-      };
-    }
+  if (prettierConfig.parser) return prettier.format(source, prettierConfig);
+
+  let config = {};
+
+  if (parser) {
+    config = {
+      ...prettierConfig,
+      parser: parser === 'javascript' ? 'babylon' : parser,
```
Hypnosphi: let's put `...prettierConfig` after this line, so that it wouldn't override explicitly passed `prettierConfig.parser`

isoppp: [I made a slight change](https://github.com/storybooks/storybook/pull/3660/files#diff-4a68ffd7414441778cfa050339c5a114R31) to the beginning of this function.
When `prettierConfig.parser` is explicitly set `return` will be done at this point

> if (prettierConfig.parser) return prettier.format(source, prettierConfig);

After this processing, shoud not set prettierConfig.parser.
So I do not think it is necessary to change it

isoppp: But, if this code is not good, I think it can try changes.

Hypnosphi: I don't feel really good about this double `prettier.format` call in different logical branches. Let's maybe revert it to how it was, with `if (!prettier config.parser)` check?

isoppp: OK, I understood.
Many thanks for having you review it over and over. And, sorry...

I reverted before commit.
After, modified to make parser or filepath always specified
If the parser and file path are empty, set the parser babylon

Hypnosphi (approved):

PR #3686, opened 2018-05-30
using disableWebsockets instead of useWebsockets

Gongreg:
Additional PR for #3636. As discussed with @shilman it would be better to maintain backwards compatibility with using both onDeviceUI and storybook server at the same time. So I swapped useWebsockets with disableWebsockets. So now if somebody will want to use storybook without server in the app he will have to pass in the disableWebsockets option. This will be described in some documentation when we start separating RN storybook into two parts.

### DIFF
--- a/app/react-native/readme.md
+++ b/app/react-native/readme.md
@@ -149,6 +149,28 @@ The following parameters can be passed to the start command:
     Override the root(s) to be used by the packager
 ```
 
+## getStorybookUI Options
+
+You can pass these parameters to getStorybookUI call in your storybook entry point:
+
+```
+{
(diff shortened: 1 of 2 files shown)


shilman (changes requested):
LGTM, can we also document this in the README?

shilman (approved):

Hypnosphi:
@shilman IDK, some problem with cached dependencies maybe. Corresponding TeamCity build passes

Hypnosphi:
@shilman To make it more deterministic, you can add `marko-widgets` as top-level `devDependency`

PR #3690, opened 2018-05-30
Addon-a11y: Run analysis on demand

DerJacques:
Issue: [3683 - a11y accessibility plugin slow on larger components
](https://github.com/storybooks/storybook/issues/3683)

## What I did
I moved around some event emitters, so that the a11y analysis is now run only if the accessibility addon is currently active. The test is also run whenever the addon _becomes_ active.

## How to test
In the UI, everything should work exactly the same.


ndelangen (approved):
LGTM 👍

PR #3701, opened 2018-06-03
An update function for knobs to complete the data binding cycle

aherriot:
Issue: I wish I could interact with the components in the storybook and see the knob values update in response. This PR demonstrates how to do that.
## What I did
See in this excerpt how the onChange function on the TextField can call a new storybook/knobs function to `update` the values in the knob store.

```JavaScript
import React from 'react'

import { storiesOf } from '@storybook/react'
import { action } from '@storybook/addon-actions'
import { linkTo } from '@storybook/addon-links'
import { withKnobs, text, update } from '@storybook/addon-knobs'

import TextField from './'

const stories = storiesOf('TextField', module)
stories.addDecorator(withKnobs)

stories.add('Controlled', () => {
  const label = text('label', 'label')
  const value = text('value', 'default value')
  return (
    <TextField
      label={label}
      value={value}
      onChange={(event) => {
        update('value', event.target.value)
        action('onChange')(event)
      }}
      onFocus={action('onFocus')}
      onBlur={action('onBlur')}
    />
  )
})
```

## How to test
I would need to update all the different types of knobs, add documentation and testing for this to be complete, but before I did, I wanted to gage the community interest. Does any one else feel like this is an idea worth pursuing?

### DIFF
--- a/addons/knobs/src/components/Panel.js
+++ b/addons/knobs/src/components/Panel.js
@@ -43,6 +43,7 @@ export default class Panel extends React.Component {
     this.handleClick = this.handleClick.bind(this);
     this.setKnobs = this.setKnobs.bind(this);
     this.reset = this.reset.bind(this);
+    this.update = this.update.bind(this);
     this.setOptions = this.setOptions.bind(this);
     this.onGroupSelect = this.onGroupSelect.bind(this);
 
@@ -56,6 +57,7 @@ export default class Panel extends React.Component {
   componentDidMount() {
     this.props.channel.on('addon:knobs:setKnobs', this.setKnobs);
     this.props.channel.on('addon:knobs:setOptions', this.setOptions);
+    this.props.channel.on('addon:knobs:update', this.update);
 
     this.stopListeningOnStory = this.props.api.onStory(() => {
       this.setState({ knobs: [], groupId: DEFAULT_GROUP_ID });
@@ -104,6 +106,10 @@ export default class Panel extends React.Component {
 
   reset() {
     this.props.channel.emit('addon:knobs:reset');
+  }
+
+  update() {
+    this.props.channel.emit('addon:knobs:update2');
   }
 
   emitChange(changedKnob) {
--- a/addons/knobs/src/KnobManager.js
+++ b/addons/knobs/src/KnobManager.js
@@ -66,6 +66,11 @@ export default class KnobManager {
     return this.getKnobValue(knobStore.get(name));
   }
 
+  update(name, value) {
+    this.knobStore.update(name, value);
+    this.channel.emit('addon:knobs:update');
+  }
+
   _mayCallChannel() {
     // Re rendering of the story may cause changes to the knobStore. Some new knobs maybe added and
     // Some knobs may go unused. So we need to update the panel accordingly. For example remove the
--- a/addons/knobs/src/KnobStore.js
+++ b/addons/knobs/src/KnobStore.js
@@ -31,6 +31,11 @@ export default class KnobStore {
     this.store = {};
   }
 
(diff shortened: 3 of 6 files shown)


igor-dv:
Did you try to achieve this with `forceReRender`?

aherriot:
@igor-dv currently, there is no way to update the value of a knob from a component. This PR adds an update method to do this.

addons/knobs/src/components/Panel.js
```diff
@@ -106,6 +108,10 @@ export default class Panel extends React.Component {
     this.props.channel.emit('addon:knobs:reset');
   }
 
+  update() {
+    this.props.channel.emit('addon:knobs:update2');
```
Hypnosphi: Not a fan of this kind of naming =(

aherriot: Absolutely, this should be changed. This PR is definitely not complete or even good. All I am asking if the community feels there is value in an `update` function to complete the one-way data flow loop. Often my state lives in a state management library like Redux/Mobx or a form library like react-final-form/redux-form/formik, etc or parent components, so I wanted to use storybook knobs to test the complete cycle. What do you think of this idea?

Hypnosphi: I don't like the fact that it introduces multiple sources of truth (component callbacks and UI knobs)

Maybe some of the [State addons](https://storybook.js.org/addons/addon-gallery/#state) 
or just recompose's [`withState`](https://github.com/acdlite/recompose/blob/master/docs/API.md#withstate) would solve your usecase better?

aherriot: When I am making a re-usable React component, like a simple TextField for example, it seems there are two parts to the interface that I need to consider when I am developing it. 

  1) the props that it receives (and how it updates based on changes) and 
  2) the callbacks that it fires when it is interacted with.

I need to ensure that it updates when a new value is passed in as a prop, and I need to make sure it passes the changed value in a callback when I type in the TextField.

Storybook Knobs does a great job highlighting the first part, and the actions addon is good at highlighting the second, but they remain disconnected. Thank you for the other suggestions. It seems they complete the React data cycle, but without the added visibility of a panel of knobs to also interact with the components.

I don't believe it is introducing any new state, the `update` function simply changes the pre-existing knobStore. The panel of knob controls and the actual component being showcased both receive their values from the knobStore. It is all one source of truth, just two ways to view it and change it.

Anyway, I don't think I have convinced you. Thank you for your efforts with storybook. It is a fantastic project. Maybe I will make my own addon.

danielduan: I think this brings up a good problem and I'm actually quite surprised the community hasn't built a stateful knobs addon yet.

You're probably better off forking this addon and making some architectural changes as to how the data binding is made, rather than adding a patch that kind of does what you need it to do.

Keraito:
Based on the discussion this is likely not gonna get merged, so I'm closing this PR.

Hypnosphi:
See https://github.com/storybooks/storybook/issues/3855

IanVS:
@aherriot let me know if there's anything I can do to help move this across the finish line.

aherriot:
@IanVS Thank you. I will try to get it rebased and updated based on the feedback within next 8 hours or so, unless other obligations come up.

JohnAlbin:
@aherriot I was following the discussion in #3855 and I'd love to see this finished as I've tried (and failed) doing this in my own stories.

ndelangen:
Hey @JohnAlbin Let's meet on discord and discuss?

ndelangen:
I might rebuild knobs from the grounds up, and could use help.

PR #3731, opened 2018-06-07
Remove linebreaks in notes text when it are html elements

### LINKED ISSUE #3724: storybook/notes addon is rendering markdown tables preceeded by numerous <br> tags
If you are reporting a bug or requesting support, start here:
I am using the `storybook/notes` addon. When rendering a table using markdown, I am getting a bunch of `<br>` elements between the proceeding element and the table. This is causing the table to display much farther down the page than intended.

I don't know if this is from something in `Notes` or a bug in `marked`.

### Steps to reproduce
We are using `Notes` to display the `README.md` file associated with a component.

- `@storybook/react` 3.4.3
- `@storybook/addon-notes` 3.4.3

```js
// package.json
"@storybook/addon-console": "1.0.0",
"@storybook/addon-info": "3.4.3",
"@storybook/addon-knobs": "3.4.3",
"@storybook/addon-notes": "3.4.3",
"@storybook/addon-options": "3.4.3",
"@storybook/react": "3.4.3",
```

### Affected platforms 

- Google Chrome 66.0.3359.181 (Official Build) (64-bit), OSX 10.12.6
- Seen elsewhere: FF on OSX (don't know versions)

##### Markdown
storybook_markdown

##### Rendered page
storybook_markdown_rendered

##### DOM
storybook_markdown_dom

Keraito:
Issue: Fixes #3724 

When the user adds a markdown table to the notes addon, a lot of linebreaks will be created. This is caused at https://github.com/storybooks/storybook/blob/master/addons/notes/src/register.js#L48. The table creates a string of HTML elements, with each of them proceeded by a \n. All of them thus get extended with a `<br />`. 

## What I did

Unfortunately, there is no negative lookbehind in js, so instead I added a regex before the existing one to filter out any existing linebreaks if they are immediatly after a HTML element. The detection of HTML elements could be more detailled, if that's prefered. 

## How to test
Added a new story.

### DIFF
--- a/addons/notes/src/register.js
+++ b/addons/notes/src/register.js
@@ -45,7 +45,12 @@ export class Notes extends React.Component {
 
   render() {
     const { text } = this.state;
-    const textAfterFormatted = text ? text.trim().replace(/\n/g, '<br />') : '';
+    const textAfterFormatted = text
+      ? text
+          .trim()
+          .replace(/(<\S+.*>)\n/g, '$1')
+          .replace(/\n/g, '<br />')
+      : '';
 
     return (
       <Panel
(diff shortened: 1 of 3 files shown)


Hypnosphi (approved):

rapheld:
Thanks for the fast turn-around on this!

PR #3744, opened 2018-06-10
Replaced 'dotenv-webpack' with 'universal-dotenv' to support multiple dot env files (like CRA)

swernerx:
Issue: https://github.com/storybooks/storybook/issues/3743

## What I did

Replaced the 'dotenv-webpack' plugin with Webkits native DefinePlugin with a configuration generated by 'universal-dotenv'. This enables usage of multiple `.env` files depending on e.g. NODE_ENV and BUILT_TARGET.

## How to test

Does this need an update to the documentation? Probably, not sure whether DotEnv capabilities are mentioned at all already.


Hypnosphi (changes requested):
Now it only affects react-native. Please apply it to `lib/core` as well

swernerx:
@Hypnosphi Thanks for the feedback. I think it should be ready now.

igor-dv (commented):
You should add `universal-dotenv` to the core's package.json as well. Also, take a look at all these failing things

Hypnosphi:
> Error: The NODE_ENV [development|production] environment variable is required but was not specified.
[08:06:21][Step 2/2]     at Object.<anonymous> (/mnt/agent/work/a1796458d49a7849/node_modules/universal-dotenv/lib/index.cjs.js:21:9)
[08:06:21][Step 2/2]     at Module._compile (internal/modules/cjs/loader.js:678:30)
[08:06:21][Step 2/2]     at Object.Module._extensions..js (internal/modules/cjs/loader.js:689:10)
[08:06:21][Step 2/2]     at Module.load (internal/modules/cjs/loader.js:589:32)
[08:06:21][Step 2/2]     at tryModuleLoad (internal/modules/cjs/loader.js:528:12)
[08:06:21][Step 2/2]     at Function.Module._load (internal/modules/cjs/loader.js:520:3)
[08:06:21][Step 2/2]     at Module.require (internal/modules/cjs/loader.js:626:17)
[08:06:21][Step 2/2]     at require (internal/modules/cjs/helpers.js:20:18)
[08:06:21][Step 2/2]     at Object.<anonymous> (/mnt/agent/work/a1796458d49a7849/lib/core/dist/server/config/env.js:3:1)
[08:06:21][Step 2/2]     at Module._compile (internal/modules/cjs/loader.js:678:30)

swernerx:
I figure dotenv-universal should regard the default NODE_ENV as being "development", right, @Hypnosphi

swernerx:
The reported error about NODE_ENV not being defined should be fixed in the latest version.

swernerx:
I also rebased the changes for alpha.9

Hypnosphi:
Looks like something broke: https://deploy-preview-3744--storybooks-official.netlify.com

swernerx:
I am not sure how this relates to the changes. Wondering whether this is just the overall state of the alpha right now.

Hypnosphi:
Nope http://storybooks-official.netlify.com

swernerx:
Rebased on upstream

swernerx:
Is there any way of reproducing these errors locally? I have now tested with "npm run test", "npm run bootstrap" and "npm start". All of them green and looking good.

That said, I studied the code more and figure there might be areas where the previous plugin might have replaced `process.env.` variables where the new plugin does not. In combination they should work fine though. Essentially the idea is that `universal-dotenv` reads in the multiple dotenv files like CRA does and then the Webpack DotEnv plugin replaces all remaining environment access fields like before.

Hopefully this fixes the issues.

Hypnosphi:
I wasn't able to reproduce it locally, but it reproduces on all the CIs. This is probably a platform-specific issue. Anyway, I've added yarn resolution as suggested in the warning, this should help

Hypnosphi (approved):

swernerx:
Thanks for sharing how to reproduce locally.

The issue seems to be related to the not very maintained graphiql (last release December 2017) which relies on an old version of graphql. I played an hour with different versions. Unfortunately still without any success.

I am just relatively sure, that this is just some side-effect as the package introduced by this PR "dotenv-universal" has quite a small dependency list not anyhow related to graphql at all:

```
universal-dotenv@^1.5.1:
  version "1.5.1"
  resolved "https://registry.yarnpkg.com/universal-dotenv/-/universal-dotenv-1.5.1.tgz#fe256777b5757908b24e01ad25007cc16994e78a"
  dependencies:
    app-root-dir "^1.0.2"
    dotenv "^6.0.0"
    dotenv-expand "^4.2.0"
```

from yarn.lock.

swernerx:
Okay, it seems like my previous assumption was wrong. I got the hint by this thread on the GraphQL issue tracker: https://github.com/graphql/graphql-js/issues/1182#issuecomment-388563851

It's more related to the fact the NODE_ENV is not correctly applied as `production` when building storybooks.

In the docs it's mentioned that "When building storybook, we set NODE_ENV to ‘production’". https://storybook.js.org/configurations/env-vars/#node_env-env-variable - I am just not sure where this actually happens as when the CLI imports "universal-dotenv" it still seems to be unset.

I figure it's related to the fact that Storybooks CLI is setting the value during runtime, while universal-dotenv expects it to be set before that already e.g. via `export`/`set` or on the command line. This approach is required to let `universal-dotenv` correctly load environment specific configuration e.g. `.env.development`.

swernerx:
I think I found a solution. Hopefully CI runs through successfully :)

app/angular/src/server/build.js
```diff
@@ -1,5 +1,9 @@
-import { buildStatic } from '@storybook/core/server';
+// Building for distribution defaults to production target.
+process.env.NODE_ENV = process.env.NODE_ENV || 'production';
```
Hypnosphi: Can we do it inside `buildStatic` function, to avoid duplication?

Hypnosphi: Can we move it to `@storybook/core` to avoid duplication?

swernerx:
Unfortunately not. I don't like repeated code as well. Order is very important here. And we have to make sure that NODE_ENV is correctly set before importing the other stuff. Doing it here keeps CommonJS (which is required for these order requirements) at its minimum.

Hypnosphi:
OK, then I think `bin/build.js` are a more appropriate place for that. They already use commonjs, so there's no need for comments

swernerx:
That looks like a good place... let be try it out...

swernerx:
Wouldn't hurt to get this finally merged. @Hypnosphi can you have another look? This is the far cleaner approach!

Hypnosphi (approved):

pksunkara:
@swernerx I am looking at this part of code feel like the `.env` files should be loaded when the plugin is actually used instead of doing it early. Currently, `loadEnv` from [this line](https://github.com/storybooks/storybook/blob/master/lib/core/src/server/config/webpack.config.js#L69) takes all the system variables and filters the only `STORYBOOK_` ones because the others might be system sensitive. The `.env` files are supposed to be safe and should be completely allowed. But the way your plugin is currently working, you are loading the files before `loadEnv` is called, which means only `STORYBOOK_` variables from the `.env` files are being passed to webpack. All the variables from the `.env` files should be passing to the webpack.

swernerx:
@pksunkara I figure this was my intention. I think of environment variables being mostly static during one run. At lot of other libraries seem to have the assumption as well. They e.g. enable specific code branches for debugging with NODE_ENV being development, but this branching happens at declaration level, not during runtime of the function. That said, for the multi .env situation as supported by the library it would be quite tricky to suggest that it is loadable and initializable at any level because that is not true. Once values have been applied to `process.env` they are never overrided again so code like this:

```js
process.env.NODE_ENV="development";
universalDotEnv.init();
```

will not work correctly when the init call happened before already.

pksunkara:
Yeah, but your package doesn't have an init call. Just doing `require('universal-dotenv')` is loading the `.env` files. I would prefer if your package has an `init()` thing which I can use after filtering the `STORYBOOK_` env variables. The usecase I mentioned in my previous comment is definitely something that would need loading at a different level of code, because the environment variables is not static here. They are being deleted by some parts of the code.

PR #3753, opened 2018-06-11
Update Extend Mode webpack.config docu for Vuejs

meirotstein:
After long struggle with adding storybook on my Vuejs project with sass support I have found that additional resolve section must be added to the webpack.config.js file - I think that this can spare some time to others.

Issue:

## How to test

### DIFF
--- a/docs/src/pages/configurations/custom-webpack-config/index.md
+++ b/docs/src/pages/configurations/custom-webpack-config/index.md
@@ -30,6 +30,12 @@ module.exports = {
         include: path.resolve(__dirname, "../")
       }
     ]
+  },
+  // This relevant only for Vuejs projects:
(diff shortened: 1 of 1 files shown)


ndelangen:
Sorry for the late reply @meirotstein, This should not be necessary for Vue, I think.. 

Maybe with sass support it's different, I'm unsure.

meirotstein:
@ndelangen thanks for your response.

This does necessary if you use custom webpack config with full control mode (without defaults) as in the story you'll probably use templates and the _runtime only_ build is will fail in that case - if you take the storybook defaults it already does that for you.

See here:
https://vuejs.org/v2/guide/installation.html#Runtime-Compiler-vs-Runtime-only

Issue #3406, opened 2018-04-12
require.context in nested modules


goodmind:
### Bug or support request summary

`require.context` only polyfilled in top config, but not in other required file in `@storybook/addon-storyshots`

quinn:
I think I am having this same issue:

```
$ node_modules/.bin/jest
 => Loading custom .babelrc
 FAIL  stories/test.js
  ● Test suite failed to run

    TypeError: require.context is not a function

      1 | import { configure } from '@storybook/react';
      2 |
    > 3 | const context = require.context('../components', true, /stories.js$/)
        |                         ^
      4 |
      5 | configure(() => context.keys().forEach(context), module)
      6 |

      at Object.<anonymous> (stories/index.js:3:25)
      at newRequire (node_modules/@storybook/addon-storyshots/dist/require_context.js:83:14)
      at loadStories (evalmachine.<anonymous>:6:2)
      at ConfigApi.configure (node_modules/@storybook/core/dist/client/preview/config_api.js:63:9)
      at evalmachine.<anonymous>:9:22
      at runWithRequireContext (node_modules/@storybook/addon-storyshots/dist/require_context.js:113:3)
      at Object.load (node_modules/@storybook/addon-storyshots/dist/react/loader.js:33:33)
      at loadFramework (node_modules/@storybook/addon-storyshots/dist/frameworkLoader.js:36:17)
      at testStorySnapshots (node_modules/@storybook/addon-storyshots/dist/index.js:72:54)
      at Object.<anonymous> (stories/test.js:3:1)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        2.854s
Ran all test suites.
```

Hypnosphi:
Released as [`4.0.0-alpha.10`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.10)

PR #3757, opened 2018-06-12
Storyshots - Replace require_context.js with babel-plugin-require-context-hook

igor-dv:
Issue: #2894, #3406

## What I did

Moved webpack's `require.context()` polyfilling out of the storyshots implementation to the babel. It removes the complexity of manual transpyling `config.js`, but just **requiring** it.


MIGRATION.md
```diff
@@ -46,6 +45,7 @@ The `@storybook/react-native` had built-in addons (`addon-actions` and `addon-li
 
 1. `imageSnapshot` test function was extracted from `addon-storyshots` and moved to a new package - `addon-storyshots-puppeteer` that now will be dependant on puppeteer
 2. `getSnapshotFileName` export was replaced with the `Stories2SnapsConverter` class that now can be overridden for a custom implementation of the snapshot-name generation
+3. Storybook that was configured with Webpack's `require.context()` feature will need to add a babel plugin to polyfill this functionality. A possible plugin might be [babel-plugin-require-context-hook](https://github.com/smrq/babel-plugin-require-context-hook)
```
tmeasday: Is this advice for migrating users? Perhaps something more definite would be good here?

addons/storyshots/storyshots-core/README.md
```diff
@@ -38,6 +38,53 @@ If you aren't familiar with Jest, here are some resources:
...
+  "plugins": ["..."],
+  "env": {
+    "test": {
+      "plugins": ["require-context-hook"]
+    }
+  }
+}
+```
+
+Make sure that it is added under the needed environment,
```
tmeasday: "Make sure not to include this babel plugin in the config environment that applies to webpack" -- might be a bit clearer?

tmeasday (commented):
Looking forward to getting rid of that crazy code!

A couple small README comments

MIGRATION.md
igor-dv: 🤔 I can link to the README explanation about this. WDYT?

addons/storyshots/storyshots-core/src/frameworks/config-loader.js
```diff
@@ -1,43 +0,0 @@
-import fs from 'fs';
-import path from 'path';
-import { getBabelConfig } from '@storybook/core/server';
```
Hypnosphi: This export can probably be removed from `@storybook/core` now, as well as `@storybook/<framework>/options.js` endpoints

igor-dv: Will remove

Hypnosphi (approved):

PR #3768, opened 2018-06-15
Load Angular library projects

rhalff:
Issue:

## What I did

Angular 6 has the ability to generate new library projects using the `ng generate library my-lib` command.

This will by default create projects in the the `projects/` folder.

Currently the storybook angular cli does not consult these locations.

This pull request adds support for these libraries.

It does so by:
 - loading angular.json.
 - collecting all projects which are defined as being a library.
-  loading the ng-package.json file of the library to determine the module's entry point (normally src/public_api.ts)
- loading the package.json of the library to determine module name.
- creating webpack aliases for the libraries so the modules can be referenced from anywhere.

In order to also glob the stories in `projects/` the user has to manually add this to `./storybook/config.js`

Hot reload works out of the box and didn't need any code changes.

## How to test

Checkout this branch and run storybook, the `test-lib` library appears under Projects.
If you modify any of the project files the story should reload and reflect the changes.

N.B. The test-lib project was generated with the cli, I didn't make any code changes on it, beside naming the selector to 'storybook-test-lib' while this was required by the lint rules.


rhalff:
closed, same can be accomplished configuring additonal paths in the tsconfig.

rhalff:
Well, it is *a* solution. For `ng serve` I've switched to adding the paths to `src/tsconfig.app.json`.

e.g.
```
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    ....,
    "paths": {
      "@chix-ui/action-buttons": [
        "../projects/chix-ui-action-buttons/src/public_api.ts"
      ],
      "@chix-ui/bottombar": [
        "../projects/chix-ui-bottombar/src/public_api.ts"
      ],
      ...
```
This works, but it seems that doesn't work out of the box with storybook, not sure why, perhaps because the storybook angular-cli does not execute all the steps the angular-cli is doing.

For now I'm solving it with extra code in my own `.storybook/webpack.config.js`

You could still try it out and determine whether this pull request is useful addition or not.

Terrafire123:
Hey, @rhalff !

Can you perhaps add additional details to what you put into your .storybook/webpack.config.js ?

Thanks!

PR #3775, opened 2018-06-19
[addon-knobs] Allow number knob to be empty

paradoxxxzero:
Issue: When emptying a knob number field, currently the value 0 is returned, due to this [line](https://github.com/storybooks/storybook/blob/master/addons/knobs/src/components/types/Number.js#L59) because:
`Number('') === 0`
which is quite problematic if you want your knob to return nothing if nothing has been input.

## What I did

I checked for the empty string in `handleChange` and returned null in this case as it were a NaN value.
I also handled the case where the value is null or undefined, replacing it in the field with an empty string to avoid the classic _Warning: Input is changing an uncontrolled input of type text to be controlled_ when giving the knob a null or undefined default value.

### DIFF
--- a/addons/knobs/src/components/types/Number.js
+++ b/addons/knobs/src/components/types/Number.js
@@ -39,10 +39,11 @@ const RangeWrapper = styled('div')({
 class NumberType extends React.Component {
   constructor(props) {
     super(props);
-
-    this.state = {
-      value: props.knob.value,
-    };
+    let { value } = props.knob;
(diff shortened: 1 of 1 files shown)


danielduan (approved):

wuweiweiwu (approved):
Nice!

Hypnosphi:
This doesn't seem applicable to `release/3.4`, so I removed the `patch` label

Issue #3780, opened 2018-06-21
Support webpack.config.ts


Ailrun:
### Support request summary
Webpack supports something like `webpack.config.ts` with `https://github.com/js-cli/js-interpret`, as they mentioned in https://webpack.js.org/configuration/configuration-languages/.
However, in storybook, it detects fixed `webpack.config.js` in https://github.com/storybooks/storybook/blob/8757938b4104e96e4d33c292329222cdd3a68e98/lib/core/src/server/config.js#L89
This makes it hard to centralize webpack config, when I use `webpack.config.ts`.

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/core 3.4.7

### Affected platforms 

- Webpack
- TypeScript

Ailrun:
Resolved in #3785.

swernerx:
Seems that there is some issue when dealing with "normal" "webpack.config.js". After updating I get this error:

```
WARN => Custom configuration file /Users/swerner/Workspace/init/display/node_modules/edge-storybook/lib/webpack.config.js is detected
WARN    but impossible to import loader for .js
INFO Using default webpack setup based on "create-react-app".
```

I overlooked this first as this does not throws (which it probably should do). Also I am wondering what it should do for configs which do not require transpilation.

Ailrun:
@swernerx Why do you think it should throw?

swernerx:
I think it should throw because there might be unwanted side-effects if my very own config is not loaded. This has to be very explicit and visible to the user. Ignoring the issue somewhat by only raising a warning is IMHO not enough. Just falling back to some default is clearly not what the user expects when defining a custom webpack config.

But this is really another discussion. The bug right now is that actually two folded:

1. Should correctly handle .js configs like before without showing a warning.
2. Should output a warning or even throw when requiring the custom webpack config fails.

The second issue is about this code:

```
      try {
        return requireConfig(customConfigPath);
      } catch (e) {
        // do nothing
      }
```

We really shouldn't hide e.g. parse errors.

Ailrun:
For 1., handling `.js` or `.json` is not that hard (since `interpret` package have `null` interpreter for them), and for 2., ye I think you're right. Could you make a PR? Nowdays I'm little busy because of work...

FYI, you can use permalink to link specific code like following (which gives a hyperlink to the file)
[...]

PR #3785, opened 2018-06-22
Support other type of webpack configs

Ailrun:
Issue: #3780.

## What I did
Add `interpret` package to support other type (`.ts`, `.babel.js`, ...) of webpack config files.

## How to test
Install `ts-node` and `typescript` in `official-storybook` in example directory, and change `webpack.config.js` to `webpack.config.ts`, and finally change `require`s to `import`s.

Does this need an update to the documentation? I'm not sure.


Ailrun:
@Hypnosphi It looks like the check fails because of yarn registry. Could you rerun it?

Hypnosphi (approved):
I'm not very comfortable with copy-pasting the code from webpack sources, but it's probably the best we can do now.

Can you please open an issue for https://github.com/webpack/webpack-cli and ask them to make config resolution logic available for external usage? Babel does a similar thing in v7: https://github.com/babel/babel/pull/7472

Ailrun:
@Hypnosphi the problem is that storybook can accept different type of configurations (even the file names are same)

lib/core/src/server/config.js
```diff
@@ -58,6 +59,87 @@ function informAboutCustomConfig(defaultConfigName) {
   logger.info(`=> Using default webpack setup based on "${defaultConfigName}".`);
 }
 
+// Copied and modified from
```
igor-dv: Please extract all these to a separate file.

igor-dv (changes requested):

Hypnosphi:
Yes, we only need to resolve the filename

lib/core/src/server/config.js
Ailrun: OK, I will extract them into `webpack_config`. Is it what you expect?

Ailrun:
@igor-dv Could you check this again?

igor-dv (approved):
Let's rename `webpack_config` to the `loadCustomWebpackConfig` and we are OK.

Ailrun:
@igor-dv Is there any reason for snake_case file name and camelCase file name? I mean, `babel_config` use snake_case, where `loadCustomWebpackConfig` use camelCase.

igor-dv:
I am not surer about the convention, but it's more about maintenance/readability. I would also change the `babel_config` to `loadBabelConfig` because it confuses.

nickmccurdy:
I think I found a regression:
```
info @storybook/react v4.0.0-alpha.21
info
info => Loading presets
WARN => File /Users/nick/Repos/ui/.storybook/webpack.config.ts is detected
WARN    but impossible to import loader for .ts

```

87Hz:
> I think I found a regression:
> 
> ```
> info @storybook/react v4.0.0-alpha.21
> info
> info => Loading presets
> WARN => File /Users/nick/Repos/ui/.storybook/webpack.config.ts is detected
> WARN    but impossible to import loader for .ts
> ```

Do not forget to install `@babel/register`.

kristojorg:
@87Hz What do you mean? Why am I installing that and where would I use it?

87Hz:
> @87Hz What do you mean? Why am I installing that and where would I use it?
> 
> I'm getting the same error as @nickmccurdy but with `.storybook/main.ts`

That is to ensure `.ts` files are properly transcompiled by Babel. You do not need to use it anywhere, it is handled by Babel.

PR #3832, opened 2018-07-04
fix: add missing `moment` dependency

alexlafroscia:
Issue: I noticed that `action-knobs@4.0.0-alpha.12` depended on `react-datetime` which assumes a `moment` peer dependency, but did not provide it. This caused an error when starting Storybook.

## What I did

Add the missing dependency

## How to test

Try starting a Storybook with the `knobs` addon installed before/after this change.

### DIFF
--- a/addons/knobs/package.json
+++ b/addons/knobs/package.json
@@ -21,6 +21,7 @@
     "escape-html": "^1.0.3",
     "fast-deep-equal": "^2.0.1",
     "global": "^4.3.2",
+    "moment": "^2.22.2",
     "prop-types": "^15.6.1",
     "qs": "^6.5.2",
     "react-color": "^2.14.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -7862,7 +7862,13 @@ graphql-request@^1.4.0:
   dependencies:
     cross-fetch "1.1.1"
 
-graphql@^0.12.3, graphql@^0.13.2:
+graphql@^0.12.3:
+  version "0.12.3"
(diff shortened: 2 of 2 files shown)


Hypnosphi (dismissed):

Hypnosphi (changes requested):
This should be a peer dependency, so that user installs it on their side.

See https://github.com/storybooks/storybook/issues/3335#issuecomment-396078461

Alternatively, we could use `resolve.alias` in our webpack configs to enforce the correct version. In this case, the dependency should be added in `lib/core`. This has a downside that it will bring moment to everyone, even if they don't use knobs

alexlafroscia:
I agree that it's a `peerDependency`, but it's a peer dependency of `react-datetime`, not `addon-knbos`. It doesn't make sense that a user of `addon-knobs` would care at all about `moment` or the version installed.

Hypnosphi:
Here's the problem:

— We depend on `A@2` which has a peer dependency on `B@3`
— We add a direct dependency on `B@3` to satisfy that
— User app has a direct dependency on `B@2` and doesn't have any dependency on `A`
— `npm` hoists `A@2` to root `node_modules`, but leaves `B@3` under `@storybook/*/node_modules` to avoid clashing with `B@2`
— inside `A@2`, `import 'B'` resolves to `B@2` instead of the expected `B@3`, which leads to an error

In this case, `A` is `react-datetime`, and `B` is `moment`. If user has some incompatible version of `moment`, the addon will break

PR #3842, opened 2018-07-05
RFC/Propposal: Rendering with enzyme to allow e2e tests

DianaSuvorova:
## Issue:
We want to use storybook to run our e2e/(selenium) tests. We don't want to introduce any data attributes in our DOM for test-only purposes (we  also don't use traditional css classes in our React codebase so getting elements by xpath or css is not feasible either). Enzyme provide [elements query API](http://airbnb.io/enzyme/docs/api/ReactWrapper/findWhere.html) where one can look up elements by Component name for instance. Which makes looking up and testing real DOM elements really easy.
I have written in details about `enzyme` for e2e idea for `create-react-app` [here](https://github.com/DianaSuvorova/e2e). Same idea applies to storybook.

## What I did

Instead of rendering stories with `ReactDOM.render` I did it with `Enzyme.Mount` which in turn will call `ReactDOM.render` after wrapping it in its API. I also created a global `storyEl` variable selenium would be able to read. So example of selenium script is something like below.

```js
const enzymeEl = await driver.executeScript('return window.storyEl;');
const componentEl = enzymeEl.findWhere(c => c.name() === 'MyComponent');
componentEl. getDOMNode().click()
expect(...)
```

Alternatively there can be additional `e2e-storybook` build where it is rendered with enzyme while default one is unchanged and rendered with ReactDOM.

Let me know your thoughts

Yes
> Does this need a new example in the kitchen sink apps? 

Yes
> Does this need an update to the documentation? 

Yes

> If your answer is yes to any of these, please make sure to include it in your PR.

I have hacked it around to make sure it works as expected. I am going to create comprehensive PR once I get an initial feedback on the proposal above.

### DIFF
--- a/app/react/package.json
+++ b/app/react/package.json
@@ -31,7 +31,9 @@
     "global": "^4.3.2",
     "lodash.flattendeep": "^4.4.0",
     "prop-types": "^15.6.1",
-    "react-emotion": "^9.1.3"
+    "react-emotion": "^9.1.3",
+    "enzyme": "^3.3.0",
+    "enzyme-adapter-react-16": "npm:enzyme-react-adapter-future"
   },
   "peerDependencies": {
     "react": ">=15.0.0",
--- a/app/react/src/client/preview/render.js
+++ b/app/react/src/client/preview/render.js
@@ -1,15 +1,19 @@
-import { document } from 'global';
+import Enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { document, window } from 'global';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { stripIndents } from 'common-tags';
 import isReactRenderable from './element_check';
 
+Enzyme.configure({ adapter: new Adapter() });
+
 const rootEl = document.getElementById('root');
 
 function render(node, el) {
-  ReactDOM.render(
+  window.storyEl = Enzyme.mount(
     process.env.STORYBOOK_EXAMPLE_APP ? <React.StrictMode>{node}</React.StrictMode> : node,
-    el
+    { attachTo: el }
   );
 }
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -5921,6 +5921,18 @@ enzyme-adapter-react-16@^1.1.0, enzyme-adapter-react-16@^1.1.1:
     react-reconciler "^0.7.0"
     react-test-renderer "^16.0.0-0"
 
+"enzyme-adapter-react-16@npm:enzyme-react-adapter-future":
+  version "1.1.3"
+  resolved "https://unpm.uberinternal.com/enzyme-react-adapter-future/-/enzyme-react-adapter-future-1.1.3.tgz#f0c102f098086a0ad0270bbdaf9da5113297dc05"
+  dependencies:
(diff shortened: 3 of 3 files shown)


Hypnosphi:
Hello @DianaSuvorova. You may want to try using [storyshots](https://github.com/storybooks/storybook/tree/master/addons/storyshots/storyshots-core) for that. You can render the stories with Enzyme using `renderer` option and write your custom tests logic in `test` option

DianaSuvorova:
@Hypnosphi , thank you! Haven't heard about it.  I will check it out.

DianaSuvorova:
@Hypnosphi  it looks like it uses Jest meaning jsdom. I want to use real DOM. The whole idea is to render with enzyme into browser. And test it with real DOM. Do you think storyshots can support it?

Hypnosphi:
I'm not sure, @igor-dv may know better

tmeasday:
@DianaSuvorova -- I'm not 100% clear on what you are trying to achieve, so bear with me if I get it wrong. I have thought a little bit about re-using stories as the basis for other types of testing. 

Assuming that you do not use any addons/decorators that actually alter the stories, could it be as simple as a small refactor of how you write your stories?

```js
// Say this is a file `Component.examples.js`:
import Component from './Component';

export exampleOne = () => <Component props={some} values={here} />
export exampleTwo = () => ...;

// Now, use those examples to define stories in storybook (`Component.stories.js`):
import { exampleOne, exampleTwo } from './Component.examples';

  .add('example number one', exampleOne)
  .add('example number two', exampleTwo);

// Additionally, use those examples as the basis of your tests (`Components.tests.js`):
import { exampleOne, exampleTwo } from './Component.examples';

// This would probable be in some library
function mountWithEnyzme(example) {
  // Do whatever you need to do
  return Enzyme.mount(example, '#root');
}

describe('Component', () => {
  it('example ones responds to clicks', () => {
   const componentEl = mountWithEnzyme(exampleOne);
   componentEl. getDOMNode().click()
   expect(...)
  });
});
```

Would that achieve what you wanted? Alternatively, if you already have stories written, you could grab them out of the storybook in your test file with something like:

```js
import { getStorybook } from '@storybook/react';
const exampleOne = 
  getStorybook()
    .find(({ kind }) => kind === 'Component').stories
    .find(({name}) => name === 'example one').render;

// same as above
```

If the second method is useful, we should probably add an API like `getStory('Component', 'example one')`

DianaSuvorova:
@tmeasday, Thank you for taking your time to respond! 

I think the first approach might actually be the appropriate solution to what we am are trying to do. We would just need to refactor our code in way where examples are defined separately from storybook. And then we have an option to use as part of storybook or render it with enzyme for e2e testing. 

We do use few addons and decorators in our stories, so examples would need some additional set up to work for both storybook and e2e purposes. 

Does the second method you mentioned with `getStorybook()` returns final version of a story with all the addons/decorators applied?

tmeasday:
@DianaSuvorova 

> Does the second method you mentioned with getStorybook() returns final version of a story with all the addons/decorators applied?

That is correct. If it is useful to you, maybe make an issue/feature request to add a `getStory` API as I mentioned above.

Good luck!

PR #3846, opened 2018-07-06
Group deprecated stories

tmeasday:
Issue: Deprecated stories were muddled up with other stories.

Also: most of the info stories were still using the deprecated APIs (cc @Keraito)

## What I did

Put all deprecated stories into their own sections. (Was this what you had in mind @shilman?)

## How to test

Hopefully name changes are self explanatory and snapshots will match up.


shilman (commented):
Great! But I think snapshots need to be updated?

ndelangen (approved):

Keraito (approved):
Looks good, thanks for covering up what I should've done 😅🙇

tmeasday:
This is weird, it seems like the snapshots are up to date (no changes and tests pass when I run `yarn test --core --update` on the branch), yet the builds still fail. Anyone have any idea?

ndelangen:
I dunno @tmeasday but I ran it, and it found 2 snapshots to be in need of some updating.

PR #3853, opened 2018-07-08
Pass Jest done callback to testMethod

hisapy:
Issue: NO ISSUE

## What I did

I added the [Jest done callback](https://jestjs.io/docs/en/asynchronous) to be passed to the testMethod in StoryShot addon. Also updated the addon's README.

## How to test

**I don't think so**

**I don't think so**

I updated the StoryShot addon README with usage example.

### DIFF
--- a/addons/storyshots/storyshots-core/README.md
+++ b/addons/storyshots/storyshots-core/README.md
@@ -185,6 +185,122 @@ initStoryshots({
 })
 ```
 
+
+### StoryShots for async rendered components
+
+You can make use of [Jest done callback](https://jestjs.io/docs/en/asynchronous) to test components that render asynchronously. This callback is passed as param to test method passed to `initStoryshots(...)` when the `asyncJest` option is given as true.
+
+#### Example
+
+The following example shows how we can use the **done callback** to take StoryShots of a [Relay](http://facebook.github.io/relay/) component. Each kind of story is written into its own snapshot file with the use of `getSnapshotFileName`.
+
+Add _stories of UserForm_ in the file: UserForm.story.jsx
+
+```jsx
+/* global module */
+import React from "react";
+import { QueryRenderer } from "react-relay";
+import { storiesOf } from "@storybook/react";
+
+// Use the same queries used in YOUR app routes
+import { newUserFormQuery, editUserFormQuery } from "app/routes";
+import UserFormContainer from "app/users/UserForm";
+
+// YOUR function to generate a Relay Environment mock.
+// See https://github.com/1stdibs/relay-mock-network-layer for more info
+import getEnvironment from "test/support/relay-environment-mock";
+
+// User test data YOU generated for your tests
+import { user } from "test/support/data/index";
+
+// Use this function to return a new Environment for each story
+const Environment = () =>
+  getEnvironment({
+    mocks: {
+      Node: () => ({ __typename: "User" }),
+      User: () => user
+    }
+  });
+
+/**
+  
+  NOTICE that the QueryRenderer render its children via its render props.
+  
+  If we don't take the StoryShot async then we will only see the QueryRenderer in the StoryShot. 
+  
+  The following QueryRenderer returns null in the first render (it can be a loading indicator instead in real file) and then when it gets the data to respond to query, it renders again with props containing the data for the Component
+ */
+const renderStory = (query, environment, variables = {}) => (
+  <QueryRenderer
+    environment={environment}
+    query={query}
+    variables={variables}
+    render={({ props, error }) => {
+      if (error) {
+        console.error(error);
+      } else if (props) {
+        return <UserFormContainer {...props} />;
+      }
+      return null;
+    }}
+  />
+);
+
+storiesOf("users/UserForm", module)
+  .add("New User", () => {
+    const environment = new Environment();
+    return renderStory(newUserFormQuery, environment);
+  })
+  .add("Editing User", () => {
+    const environment = new Environment();
+    return renderStory(editUserFormQuery, environment, { id: user.id });
+  })
+```
+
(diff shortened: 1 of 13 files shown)


addons/storyshots/storyshots-core/src/api/snapshotsTestsTemplate.js
```diff
@@ -3,12 +3,13 @@ import { describe, it } from 'global';
 function snapshotTest({ story, kind, fileName, framework, testMethod, testMethodParams }) {
   const { name } = story;
 
-  it(name, () => {
+  it(name, (done) => {
```
igor-dv: As far as I understand, if `done` is passed to the test function, jest recognizes it as an async test, and waits for the `done` to be invoked. As you can see all the tests failed because `done` was not invoked...

igor-dv (changes requested):

addons/storyshots/storyshots-core/src/api/snapshotsTestsTemplate.js
hisapy: thanks for your review @igor-dv  ... I'll fix this implementation ASAP

ndelangen: @hisapy are you able to finish this PR?

hisapy: Sorry, I've been a little bit busy in some backend stuff ... I'll work on this next week

hisapy:
Hello again ... I finally made some time to update and finish this PR. Added tests, updated to use the new API of the `getSnapshotFilename`.

hisapy:
So far the only check failing is the one at **Better Code Hub** ... I tried to access it but it says only organization members can access the code ...

hisapy:
I'm going to rename the option to avoid conflicts with `async` keyword

hisapy:
@igor-dv can you take a look at this again please ?

ndelangen:
Don't worry about Better Code Hub

ndelangen (approved):

ndelangen:
Let's get @igor-dv's approval too

addons/storyshots/storyshots-core/README.md
```diff
@@ -185,6 +185,121 @@ initStoryshots({
...
+    const environment = new Environment();
+    return renderStory(newUserFormQuery, environment);
+  })
+  .add("Editing User", () => {
+    const environment = new Environment();
+    return renderStory(editUserFormQuery, environment, { id: user.id });
+  })
+  
+
+// file: StoryShots.test.js
```
igor-dv: please separate each example file to a its own block

igor-dv (commented):
Can we call it just `async`?

addons/storyshots/storyshots-core/README.md
```diff
@@ -185,6 +185,121 @@ initStoryshots({
 })
 ```
 
+
+### StoryShots for async rendered components
+
+You can make use of [Jest done callback](https://jestjs.io/docs/en/asynchronous) to test components that render asynchronously. This callback is passed as param to test method passed to `initStoryshots(...)` when the `asyncJest` option is given as true.
```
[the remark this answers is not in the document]
hisapy: I've just updated this PR with changes you requested @igor-dv .

By the way, I've used just `async` first but then I've opted for `asyncJest` to avoid issues with syntax _highlightings_ that were assuming it was the `async` ES keyword

igor-dv (approved):
Thanks. Can you please merge from master, to pass CI?

hisapy:
Thx for accepting this PR ladies and gentlemen

SimenB:
Note that `done` is mostly for legacy reason in Jest, we recommend to return a promise instead. Mostly because if an assertion fails, `done` is never called unless you're careful with try-catches etc. Using `new Promise(resolveCb => {})` is less of a footgun.

Not saying that you shouldn't expose a `done` - there are definitely valid use cases, but the docs makes it seem as the callback is the only way if you need to do async work

hisapy:
Thx and wow! I haven't heard of `done` being legacy until now ... I'll try to investigate about the promise approach and create a new PR to address async components with it ASAP.

SimenB:
"legacy" might be too strong. It's probably never going away because of old code bases, and it's a nice way to programatically fail a test. But most of the time doing `return Promise.reject(new Error('fail test'))` is better, and `async-await` makes working with promises quite nice. So we definitely wanna encourage promises over a done callback.

If you work with a callback API I would still use promises, maybe like this:

```js
test('some api with cb pattern', async () => {
  const result = await new Promise((resolve, reject) => {
    callApiWithCallback((err, res) => {
      if (err) {
        reject(err);
      } else {
        resolve(res);
      }
    });
  });

  expect(result).toBeSomething();
});
```

Or use something like https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original.

So yeah, `done` is perfectly valid, but we should try to encourage people to use promises instead. At least mention it in the docs section added in this PR

hisapy:
I'd like to submit a PR to mention about using promise instead of `done()` but I need to do my homework about how to use promise to test async rendered components... currently the ones I've been testing don't have a render or afterRender callback... I just wait 1 ms so the component is updated and then start the assertions based on `enzyme`'s mount and jest. Perhaps it should be as easy as rendering the component inside the _awaiting_ Promise but I'm not so sure ... 

Anyway, for people that might be worry, I can tell you that you still get a full test run with failing assertions and `done()`.

SimenB:
You can wait a millisecond using ` await new Promise(resolve => setTimeout(resolve, 1));`. As long as a promise is returned (which using an async function does implicitly), Jest will wait for it

PR #3866, opened 2018-07-10
Fix knobs for React < 16.3

IanVS:
Issue:

The `getDerivedStateFromProps` lifecycle method was added in React 16.3,
and older versions of react won’t execute it.  Many of the knob components
were relying on this function to initialize component state, and so in
older versions of react state was undefined, and grabbing `value` would
blow up.

## What I did

Most of the usages in these components were to memoize props, but that
can be handled more cleanly through the use of `PureComponent`, as 
mentioned in:
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

I wasn't sure quite how to avoid derived state in the `object` knob, so instead I followed the same approach as https://github.com/storybooks/storybook/pull/3829 and polyfilled it.

## How to test

The only thing I can think of is to test with a version of react prior to the introduction of `getDerivedStateFromProps`.  I'm unfamiliar with your testing setup though, so I don't know how to accomplish this in an automated method.  Do you already have any smoke tests for older react versions?  If so, perhaps it's just a matter of a simple UI test of these components to ensure they don't go up in flames?

Is this testable with Jest or Chromatic screenshots? I don't know.  :(

### DIFF
--- a/addons/knobs/src/components/types/Array.js
+++ b/addons/knobs/src/components/types/Array.js
@@ -10,28 +10,18 @@ function formatArray(value, separator) {
   return value.split(separator);
 }
 
-class ArrayType extends React.Component {
-  static getDerivedStateFromProps(props, state) {
-    if (!state || props.knob.value !== state.value) {
-      return { value: props.knob.value };
-    }
-    return null;
-  }
-
+class ArrayType extends React.PureComponent {
   handleChange = e => {
     const { knob } = this.props;
     const { value } = e.target;
     const newVal = formatArray(value, knob.separator);
-
-    this.setState({ value });
     this.props.onChange(newVal);
   };
 
   render() {
     const { knob } = this.props;
-    const { value } = this.state;
 
-    return <Textarea id={knob.name} value={value} onChange={this.handleChange} size="flex" />;
+    return <Textarea id={knob.name} value={knob.value} onChange={this.handleChange} size="flex" />;
   }
 }
 
--- a/addons/knobs/src/components/types/Array.js
+++ b/addons/knobs/src/components/types/Array.js
@@ -10,7 +10,11 @@ function formatArray(value, separator) {
   return value.split(separator);
 }
 
-class ArrayType extends React.PureComponent {
+class ArrayType extends React.Component {
+  shouldComponentUpdate(nextProps) {
+    return nextProps.knob.value !== this.props.knob.value;
+  }
+
   handleChange = e => {
     const { knob } = this.props;
     const { value } = e.target;
--- a/addons/knobs/package.json
+++ b/addons/knobs/package.json
@@ -26,6 +26,7 @@
     "react-color": "^2.14.1",
     "react-datetime": "^2.14.0",
     "react-emotion": "^9.1.3",
+    "react-lifecycles-compat": "^3.0.4",
     "util-deprecate": "^1.0.2"
   },
   "devDependencies": {
--- a/addons/knobs/src/components/types/Color.js
+++ b/addons/knobs/src/components/types/Color.js
@@ -19,17 +19,10 @@ const Popover = styled('div')({
   zIndex: '2',
 });
 
-class ColorType extends React.Component {
+class ColorType extends React.PureComponent {
   state = {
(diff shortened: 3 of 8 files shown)


IanVS:
I also pushed a commit in an attempt to fix the error I got from CircleCI about a dirty repo.  I don't see how this could be related to any of my changes, but maybe???

addons/knobs/src/components/types/Array.js
```diff
@@ -10,28 +10,18 @@ function formatArray(value, separator) {
   return value.split(separator);
 }
 
-class ArrayType extends React.Component {
```
wuweiweiwu: I'm not sure how useful `PureComponent` is here. 

Since `knobs` is a object, shallow comparisons done by `PureComponent` are always going to return `false`. and maybe will even have a slight performance hit since we're always doing a shallow comparison of prop values

wuweiweiwu: I mean that using `PureComponent` is unnecessary since the shallow comparison of props (in this case `this.props.knob`) is always going to return `false`

Thus they should remain `Component`

There is no need for an implementation of `shouldComponentUpdate`

wuweiweiwu:
After minor comment is addressed LGTM

IanVS:
As @wuweiweiwu pointed out, it's not possible to rely on `PureComponent` in these components.  It's because the parent is re-creating it for each render: https://github.com/storybooks/storybook/blob/cdca2e892c762ccd89b2fc08eaefc6407ff7c2c4/addons/knobs/src/components/Panel.js#L127.

I did a little experimenting, and sure enough the components were re-rendering when a different knob was changed.  So I added a `shouldComponentUpdate` to check only whether the `knob.value` was changed.  I don't _think_ that `knob.name` is generally changed between renders, right?  Same goes for array's `knob.separator` and number's `knob.range` and friends.  If that's a bad assumption, let me know and I can update the methods.

wuweiweiwu:
@IanVS I think you _shouldn't_ implement `shouldComponentUpdate` at all in knobs.

Since each component is being recreated every single time a value changed. `shouldComponentUpdate` would be redundant.

IanVS:
When I tested it, the `shouldComponentUpdate` was indeed able to prevent unnecessary re-renders.  

@ndelangen has done a lot of performance work on these components, perhaps he has some insight here?  

I'm happy to take them out if that is the consensus, but I would hate for this PR to have the end effect of harming performance for all users (since it will be removing the memoizing effect of `getDerivedStateFromProps`).

wuweiweiwu:
@IanVS I'm not sure what you mean by the memoizing effect of `gDSFP`. It doesn't have any performance impacts since it only calculates state based on prop on every time `render` is called. And the only way to make sure some props have not changed is storing it in state since `this` is not available in a static function. There was actually an [RFC](https://github.com/reactjs/rfcs/pull/40) to include `prevProps` but it was abandoned.

Unless you're implementing `cSU` - which can reduce extra renders. However, in this case, we can't rely on assumptions that `knob.name` or `knob.range` won't commonly change because they can and we will need to re-render if they do.

You can see [here](https://github.com/IanVS/storybook/blob/1652d64635b5df8c9e382957faa67de17dbea81e/addons/knobs/src/components/PropForm.js#L35) that the only prop we would have to compare is `knobs` and since we care about all the values in `this.props.knobs` it doesn't make sense to implement `cSU`.

I hope my point is clear.  I'm happy to answer any questions.

ndelangen (approved):
Performance seems really good in the official example so LGTM @IanVS

IanVS:
I'm a bit concerned that the changes here will prevent renders if the name of knobs is being dynamically changed (which seems like a weird thing to do, but maybe there's a use case?)  I can submit another PR to address that requirement if desired.

Barring that, is there a chance to get a new alpha release?  Our stories are currently broken until this can go out (I can't figure out how to point our app at my own fork, since it's a lerna project).

Hypnosphi:
Released as (`4.0.0-alpha.15`)[https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.15]

PR #3877, opened 2018-07-12
storybook with Node HTTP Server

rwieruch:
Many people may be more familiar with NPM and node packages, thus it may be valuable to add the  Node HTTP Server to your documentation too.


igor-dv (approved):
Thanks!

Hypnosphi:
I'd vode for removing the python part at all

ndelangen:
I'd say we should advocate the usage of `npx` over installing packages globally:

`npx serve .out` will be favorable IMHO.

PR #3882, opened 2018-07-13
Fix addons panel when using preact

sanpoChew:
Issue: 

Addons such as knobs were no longer rendering in the addon panel after installing the latest alpha, stemming from the changes in this merge request: https://github.com/storybooks/storybook/pull/3628. This is due to children in preact always being an Array.

## What I did

use Array.isArray to to check if content is an array, pull out the first value if it is.

## How to test

Nope.

Nope.

Nope.

### DIFF
--- a/lib/components/src/tabs/tabs.js
+++ b/lib/components/src/tabs/tabs.js
@@ -148,7 +148,8 @@ export const Tabs = ({ children, selected, onSelect, absolute, bordered }) => {
       title,
       id,
       render:
-        typeof React.Children.only(content) === 'function' ? React.Children.only(content)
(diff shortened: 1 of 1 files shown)


ndelangen:
Useful thanks!

Do you think you could resolve the minor lint issue?

sanpoChew:
@ndelangen done 👍

danielduan (approved):

ndelangen:
Great work @sanpoChew! Thank you so much!

Hypnosphi:
@ndelangen please only add `patch` label to bugs reproduced in stable version

PR #3884, opened 2018-07-14
MOVE ui into it's own group

ndelangen:
improved the official storybook a bit so the ui section has a proper own heading


danielduan (approved):

igor-dv (approved):
Maybe you can also remove this:

image

😳

ndelangen:
@igor-dv sure, I'll try and get that in a better shape as well soon!

PR #3885, opened 2018-07-14
Stylize webpack references

langri-sha:
It is officially written as __webpack__.

See https://twitter.com/wsokra/status/855800490713649152.

## What I did

Replaced all __Webpack__ occurrences with __webpack__, except when these appeared in titles.

### DIFF
--- a/docs/src/pages/configurations/custom-webpack-config/index.md
+++ b/docs/src/pages/configurations/custom-webpack-config/index.md
@@ -5,9 +5,8 @@ title: 'Custom Webpack Config'
 
 ## Default mode
 
-The default Webpack config of Storybook is balanced for a medium-size project (specially created with [Create React App](https://github.com/facebookincubator/create-react-app)) or a library. But if you already have your own Webpack setup, that's not useable.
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/configurations/custom-webpack-config/index.md
```diff
@@ -5,9 +5,8 @@ title: 'Custom Webpack Config'
 
 ## Default mode
 
-The default Webpack config of Storybook is balanced for a medium-size project (specially created with [Create React App](https://github.com/facebookincubator/create-react-app)) or a library. But if you already have your own Webpack setup, that's not useable.
```
danielduan: I know the existing doesn't make much sense anymore but we should still keep some of the same ideas presented here.

Something about the default being a sensible starting point for most projects and frameworks.

ndelangen: agreed

ndelangen: @langri-sha could you re-add something along these lines, do you think?

ndelangen (approved):

ndelangen:
@danielduan I'll visit this line again later, I have no great replacement, but maybe in the future we'll find a good line.

Meanwhile I'm going to merge this PR, to move it along; it's been open way too long already.

PR #3910, opened 2018-07-20
doc(test): image test is official-storybook, not cra-kitchen-sink

JalilArfaoui:
Issue:

The documentation about `yarn test --image` says it depends on examples/cra-kitchen-sink whereas called code and my experience indicate that is runs `examples/official-storybook`, and there for that it's inside `examples/official-storybook` that we should `yarn build-storybook`. 

## What I did

I updated the documentation according to my understanding

## How to test

Verify that what I've written is correct.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,8 +71,15 @@ Before these tests are ran, the project must be bootstrapped with the React Nati
 
 `yarn test --image`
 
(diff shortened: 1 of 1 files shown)


ndelangen (approved):
Yes checks out, must have been missed when moving the stories to the official example, thanks!

PR #3914, opened 2018-07-23
REFACTOR layout and REMOVE usplit component

ndelangen:
Issue: the way the split drag handles worked was quite flacky

## What I did
I removed the additional component for the handle, and made them part of the react-split Resizer component css class.

slight regression in theming (handles are no longer themable) might find a fix soon.


ndelangen:
Please let me know if you think the non-theme-ability of the 2 grey bars in the handle is an breaking-issue for you.

Personally I think I can fix it a bit later in another PR.

lib/components/src/layout/desktop.js
```diff
@@ -42,21 +91,18 @@ const AddonPanelWrapper = styled('div')(
   })
 );
 
-const resizerCursor = isVert => (isVert ? 'col-resize' : 'row-resize');
+// const resizerStyles = (isShown, isVert) => ({
```
igor-dv: Do you need all these comments later?

lib/components/src/layout/desktop.js
```diff
@@ -2,15 +2,64 @@ import React, { Component } from 'react';
...
+import { injectGlobal } from 'emotion';
 import styled from 'react-emotion';
 import throttle from 'lodash.throttle';
 
 import SplitPane from 'react-split-pane';
 
-import USplit from './usplit';
 import Dimensions from './dimensions';
 
+// eslint-disable-next-line no-unused-expressions
```
igor-dv: 🤦‍♂️

igor-dv (commented):
IMO, the theming could be done in a separate PR indeed

danielduan (approved):

lib/components/src/layout/desktop.js
ndelangen: Well it makes sense, it's a deliberate side-effect..

lib/components/src/layout/desktop.js
ndelangen: No I will remove them

ndelangen:
Would love your approval @igor-dv 🙇

igor-dv (approved):
✌

PR #3923, opened 2018-07-26
addons-jest : bug with the jest parameter

libetl:
The jest param is not working on my machine. Array.from({0: 'zero', 1: 'one'}) does not work on my nodejs.

Instead, Array.prototype.slice works pretty well.

Issue:

## What I did
1°) Made a storybook with the jest addon included, version 4.0.0.alpha-14
2°) tried to invoke the jest addon decorator : 
   2°a) in the config.js : withTests({results: require('../.jest-test-results.json'),  filesExt: '(\\-specs)?(\\.js)?$'}))
   2°b) in the story file :  storiesOf('One card preview', module).addParameters({ jest: ['creditCardPreview'] })

## How to test
Is this testable with Jest or Chromatic screenshots? I don't know
Does this need a new example in the kitchen sink apps? I am not sure
Does this need an update to the documentation? I think not, it looks like a regression.

### DIFF
--- a/addons/jest/src/index.js
+++ b/addons/jest/src/index.js
@@ -2,7 +2,7 @@ import addons from '@storybook/addons';
 import deprecate from 'util-deprecate';
 
 const findTestResults = (testFiles, jestTestResults, jestTestFilesExt) =>
-  Array.slice({...testFiles, length: Object.keys(testFiles).length}).map(name => {
+  Array.prototype.slice.call({...testFiles, length: Object.keys(testFiles).length}).map(name => {
     if (jestTestResults && jestTestResults.testResults) {
       return {
         name,
--- a/addons/jest/src/index.js
+++ b/addons/jest/src/index.js
@@ -2,7 +2,7 @@ import addons from '@storybook/addons';
 import deprecate from 'util-deprecate';
 
 const findTestResults = (testFiles, jestTestResults, jestTestFilesExt) =>
-  Array.prototype.slice.call({...testFiles, length: Object.keys(testFiles).length}).map(name => {
+  Object.values(testFiles).map(name => {
     if (jestTestResults && jestTestResults.testResults) {
       return {
         name,
--- a/addons/jest/src/index.js
+++ b/addons/jest/src/index.js
@@ -2,7 +2,7 @@ import addons from '@storybook/addons';
 import deprecate from 'util-deprecate';
 
 const findTestResults = (testFiles, jestTestResults, jestTestFilesExt) =>
-  Array.from(testFiles).map(name => {
+  Array.slice({...testFiles, length: Object.keys(testFiles).length}).map(name => {
     if (jestTestResults && jestTestResults.testResults) {
       return {
         name,


libetl:
I don't know (yet) where to add a jest test in the jest addons.

libetl:
https://jsfiddle.net/djvrcqh3/23/

libetl:
Sorry I missed putting a tag in the PR. Can you do that for me ? I think it should be 'bug'

addons/jest/src/index.js
```diff
@@ -2,7 +2,7 @@ import addons from '@storybook/addons';
 import deprecate from 'util-deprecate';
 
 const findTestResults = (testFiles, jestTestResults, jestTestFilesExt) =>
-  Array.from(testFiles).map(name => {
+  Array.slice({...testFiles, length: Object.keys(testFiles).length}).map(name => {
```
Hypnosphi: Are you sure it works? `Array.slice` is `undefined` for me

libetl: I prefered the Array.slice shortcut rather than Array.prototype.slice.call
MDN says there is a full support for Array.slice.
But I think I can replace that with Array.prototype.slice.call

Hypnosphi:
Which Node version is it? [ES6 compatibility table](http://kangax.github.io/compat-table/es6/#node4) says that `Array.from` is supported even in Node 4.

libetl:
image

addons/jest/src/index.js
Hypnosphi: Wow, never heard of it. Do you have a link to MDN?

libetl:
Can you try the jsfiddle again, I have updated it https://jsfiddle.net/djvrcqh3/26/

Hypnosphi:
```
Uncaught TypeError: Array.slice is not a function
```

Hypnosphi:
Looks like `Array.from({0: 'zero', 1: 'one', length: 2})` works as well

libetl:
Yes you are right @Hypnosphi 
But I have replaced by Array.prototype.slice.call already

Hypnosphi:
I mean your jsfiddle doesn't work for me

libetl:
Sorry
https://jsfiddle.net/djvrcqh3/33/

Hypnosphi:
Maybe we could just use `Object.values(testFiles)`?

libetl:
@Hypnosphi that is even better

libetl:
Thanks for pointing that out

Hypnosphi (approved):

libetl:
I am not an expert in JS. Are the values going to be sorted anyway ?
We are going to lose the keys

ndelangen:
@Hypnosphi is this a for a patch release?

ndelangen:
Thank you @libetl so great to get this fixed!

addons/jest/src/index.js
wuweiweiwu: here it is! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

wuweiweiwu:
@libetl keys in Objects in JS are not guaranteed to be sorted.

wuweiweiwu (approved):
LGTM 👍

PR #3939, opened 2018-08-01
Clean READMEs

igor-dv:
Issue: There is a few copy-pasted content from place to place, that is being hard to maintain because there is no single source of truth 

## What I did
1. Removed badges from the inner packages (there was even broken greenkeeper badge somewhere)
2. Commented out the BrowserStack since it is not implemented.
3. Removed unused ancient images
4. Moved all the copy-pasted demo.gifs to one place. If any framework want its own (like in marko), then it might be changed. I would even prefer some abstract Storybook drawings, because UI changes frequently


ndelangen (approved):

ndelangen:
Nice cleanup!

PR #3941, opened 2018-08-01
Improve BettercodeHub 

igor-dv:
Issue: BettercodeHub is not impressed by our codebase (6/10), and I am trying to improve something.

## What I did
1. Excluded `.stories.js` from the analysis - most of the stories are for the example apps = documentation. And you don't really test story, it's more like story ~= test. 
2. Added `.storyshot` to tests, since we have tons of them and looks like they are not counted


danielduan (approved):

igor-dv:
Looks like it analyzes stories anyway 🤔 

image

PR #3950, opened 2018-08-02
Inline emotion css calls that require theme to avoid using state

### LINKED ISSUE #3840: Cannot read property 'mainBorder' of undefined
After upgrading from 4.0.0-alpha.9 to 4.0.0-alpha.12 i got this error ` Cannot read property 'mainBorder' of undefined`.

Anyone know what cause this error ? 
It looks like it comes from the theme.

I set the theme in `setOptions`.
```
import { themes } from '@storybook/components';
setOptions({ 
 ...
  theme: themes.normal
});
```

tkh44:
Issue:

`getDerivedStateFromProps` can be called before the prop `theme` is set by react-emotion which causes errors.

## What I did

Inlined the `css` calls instead of placing them on the component's state. `withTheme` will rerender this component when the `theme` changes so there is no need to use `getDerivedStateFromProps`.

## How to test

fixes #3840

### DIFF
--- a/lib/cli/test/snapshots/meteor/package.json
+++ b/lib/cli/test/snapshots/meteor/package.json
@@ -10,8 +10,8 @@
   "dependencies": {
     "babel-runtime": "^6.20.0",
     "meteor-node-stubs": "~0.2.4",
-    "react": "^16.4.1",
-    "react-dom": "^16.4.1"
+    "react": "^16.4.2",
+    "react-dom": "^16.4.2"
   },
   "devDependencies": {
     "babel-core": "^6.26.3",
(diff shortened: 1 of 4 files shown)


tkh44:
I've tried to get all the checks but I am unable to get it right. If someone could give me some pointers I'd be grateful.

ndelangen (approved):

ndelangen:
Thank you @tkh44 !

tkh44:
@ndelangen I should have been more clear that this is a bugfix and the bug this issue causes keeps the storybook from rendering at all in the latest alpha releases in some situations. I couldn't pin down the exact cause of the bug but this PR fixes it.

image

ndelangen:
@shilman as @tkh44 mentions, let's do an alpha release rather quick?

PR #3958, opened 2018-08-04
Make addon-options work with story parameters

nm123github:
## What I did
- use makeDectorator to upgrade withOptions
- update documentation

## How to test
I've added usage to some stories in the marko-cli

addon-options-eg

### DIFF
--- a/addons/options/src/preview/index.js
+++ b/addons/options/src/preview/index.js
@@ -1,4 +1,4 @@
-import addons from '@storybook/addons';
+import addons, { makeDecorator } from '@storybook/addons';
 import { EVENT_ID } from '../shared';
 
 // init function will be executed once when the storybook loads for the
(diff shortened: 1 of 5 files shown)


nm123github:
Attempting to make addon-options use parameters. Few things,

- I have not deprecated the global setOptions. I think its probably a good idea to let it be there but would like to know what others think.
- Right now when you set options (using params) for a story, they get set globally. Ideally i would probably like it to be reset or something (maybe to the options we set using setOptions?) if we switch to a story where we don't set options. Is it possible to setOptions specifically for a story?

ndelangen (approved):
SUPER AWESOME!

ndelangen:
setOptions being persistent is expected behavior I think?

I don't see this as a big problem.

addons/options/src/preview/index.js
```diff
@@ -41,3 +41,32 @@ export function setOptions(newOptions) {
 
   channel.emit(EVENT_ID, { options });
 }
+
+export const withOptions = makeDecorator({
+  name: 'withOptions',
+  parameterName: 'options',
+  skipIfNoParametersOrOptions: false,
+  allowDeprecatedUsage: true,
```
tmeasday: @nm123github -- was there a particular reason why you decided to allow deprecated usage for this decorator?

nm123github: Just noticed bunch of other add-ons that had set it to true, so presumed this was the pattern to follow and maybe there was a plan to set them to false at a later date.

tmeasday: Gotcha. That option is to allow `add('story', withOptions(() => <Story />))` style usage (which is the old way addons did it pre-4.0). Given this API never existed for `withOptions` I don't think we need it.

PR #3959, opened 2018-08-05
[BUG FIX] Use fixed version of react-dev-utils

igor-dv:
Issue: #3953 #3956

## What I did
Somehow, having a caret in `"react-dev-utils": "^6.0.0-next.3e165448"` is resolved to an older version of `react-dev-utils` - `react-dev-utils@6.0.0-next.b2fd8db8`, which is not compatible with webpack 4.

It worked in our examples because it locked with the right version in the yarn.lock, but a fresh install might change this.

### DIFF
--- a/app/react-native/package.json
+++ b/app/react-native/package.json
@@ -54,7 +54,7 @@
     "json5": "^1.0.1",
     "prop-types": "^15.6.2",
     "raw-loader": "^0.5.1",
-    "react-dev-utils": "^6.0.0-next.3e165448",
+    "react-dev-utils": "6.0.0-next.3e165448",
     "react-native-compat": "^1.0.0",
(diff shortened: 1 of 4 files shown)


pksunkara (approved):

pksunkara:
Yup, a fresh install in vue-kitchen-sink is encountering the same issue. 4.0.0-alpha.15 is completely non-working because of this. I would recommend merging this and publish a new alpha.

igor-dv:
CC. @shilman

dartharaknus:
And I would also recommend the examples folder to be not part of yarn workspaces so that they can be tested as examples rather than as monorepo components. Atleast in the CI.

ndelangen:
@shilman please release this asap if you can.

PR #3961, opened 2018-08-05
Improve typescript config docs

web-devel:
## What I did

Updated typescript configuration docs so that stories made with typescript picked up by default.

### DIFF
--- a/docs/src/pages/configurations/typescript-config/index.md
+++ b/docs/src/pages/configurations/typescript-config/index.md
@@ -69,6 +69,15 @@ The above example shows a working config with the TSDocgen plugin also integrate
 
 This is for the default configuration where `/stories` is a peer of `src`. If you have them all in just `src` you may wish to replace `"rootDirs": ["src", "stories"]` above with `"rootDir": "src",`.
 
+## Import tsx stories
+
+Change `config.js` inside the Storybook config directory (by default, it’s `.storybook`) to import stories made with Typescript:
+
+```js
+// automatically import all files ending in *.stories.js
+const req = require.context('../stories', true, /.stories.tsx$/);
+```
+
 ## Using Typescript with the TSDocgen addon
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/configurations/typescript-config/index.md
```diff
@@ -69,6 +69,15 @@ The above example shows a working config with the TSDocgen plugin also integrate
 
 This is for the default configuration where `/stories` is a peer of `src`. If you have them all in just `src` you may wish to replace `"rootDirs": ["src", "stories"]` above with `"rootDir": "src",`.
 
+## Import tsx stories
+
+Change `config.js` inside the Storybook config directory (by default, it’s `.storybook`) to import stories made with Typescript:
```
igor-dv: BTW, it can be also `config.ts`

danielduan (approved):

ndelangen (approved):

ndelangen (approved):

web-devel (commented):
I don't think renaming `config.js` to `config.ts` is a good idea since `.storybook` directory is not included in ts compiler (so it won't work out of the box without more changes in this guide). 

Also it seems `@storybook/react` looks exactly for `config.js` :
> Error: => Create a storybook config file in "./.storybook/config.js".

I suggest leaving `config.js` as is.

igor-dv:
If you configured Storybook to work with TS, it will also be able to load `config.ts` and `addons.ts`.

web-devel:
> If you configured Storybook to work with TS, it will also be able to load config.ts and addons.ts.

If you follow this particular guide it won't. At lest we also have to change `tsconfig.json` to include files in `.storybook`.

igor-dv:
Can you elaborate how did you change it? Also, what is your version of Storybook?

web-devel:
@igor-dv I used `latest` which is `v3.4.10`, which has `config.js` lookup hardcoded: https://github.com/storybooks/storybook/blob/release/3.4/app/react/src/server/config.js#L26
It seems things changed in master though

ndelangen:
What's the next course of action then? Do we add a remark about typescript support to the docs?

web-devel:
@igor-dv @ndelangen I was able to test it with latest alpha and it works (config.ts/addons.ts).

ndelangen:
I'll turn this into a non-patch, since this now documents 4.0.0 behavior

PR #3963, opened 2018-08-05
sorting by kind and nested stories

dioxmio:
Issue:

## What I did
sorting nested stories when activating: `sortStoriesByKind: false`

## How to test

```import { setOptions } from '@storybook/addon-options';
setOptions({sortStoriesByKind: false});
```

make sure that now both the root list of stories and the nested ones are ordered. At the moment just the root stories where sorted.

["bug", "maintenance"]
#3963

### DIFF
--- a/lib/ui/src/modules/ui/libs/filters.test.js
+++ b/lib/ui/src/modules/ui/libs/filters.test.js
@@ -74,7 +74,7 @@ describe('manager.ui.libs.filters', () => {
       expect(res).toEqual([stories[1], stories[2], stories[0]]);
     });
 
-    test('should sort nested stories by kind', () => {
+    test('should sort nested stories', () => {
       const unsorted = ['bb', 'aa'];
       const stories = [
         { kind: 'ss', namespaces: ['ss'], stories: unsorted, rootName: '' },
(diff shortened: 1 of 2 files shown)


ndelangen:
We should add some tests for this function

ndelangen:
@dioxmio There's lots of information in https://github.com/storybooks/storybook/blob/master/CONTRIBUTING.md

In short: you need to run `yarn` before running the unittests

dioxmio:
@ndelangen awesome thanks. I have added the test however the build on Danger is failing. Could you check? All the rest is good and green.

cheers

ndelangen (approved):

lib/ui/src/modules/ui/libs/filters.test.js
```diff
@@ -74,6 +74,20 @@ describe('manager.ui.libs.filters', () => {
       expect(res).toEqual([stories[1], stories[2], stories[0]]);
     });
 
+    test('should sort nested stories by kind', () => {
```
igor-dv: The name of a test is a bit misleading, it should be something like "should sort nested stories", because terminologically it's not "by kind".

ndelangen: ✅

igor-dv: Looks like you changed a wrong test 🤣

ndelangen:
Sorry @igor-dv that should do it 😄

igor-dv (approved):

PR #3965, opened 2018-08-06
Add params-based API to options addon

tmeasday:
Issue:

Options API wasn’t using params and couldn’t really be used per-story

## What I did

 - Added a params-based API: `.add(name, story, { options: {… } })`
 - Added stories to official storybook
 - Updated `.setOptions` internally to *reset* each time, rather than remembering the previous state (and only overriding new keys). This didn’t really make sense and was broken when you switched between stories that set options, unless they all set all options, which users aren’t going to do.

### STILL OUTSTANDING

When you change panels via options it does not re-calculate sizing. So if you hide the addons panel, it will leave a bunch of empty space. Can someone who knows the internals give me some advice on how to re-calculate sizes when options are set?

There are some jest test that touch on it, but the UI is not currently captured by our screenshots for these stories.

Yes, done.

Yes, <strike>done</strike> (TODO).


ndelangen:
@tmeasday I saw this PR first, looked great to me, but maybe you have thoughts on it?

https://github.com/storybooks/storybook/pull/3958

You seem to be be of the opinion options should reset. Perhaps you can rebase this PR and implement that, if you so desire?

tmeasday:
Wow I was totally gazumped by @nm123github -- nice ;)

From playing around with it a bit, I think the options need to be reset, otherwise you get super weird behaviour as the options drift as you change stories.

I'll rebase this PR against current `master`. New questions based on that PR:

1. I'm not sure we should have a story about this in the marko example app? I think these kind SB-wide stories should just be in official storybook, right?

2. We should ensure that the global `setOptions` function actually calls `setParameters` so the "default" options are the user's. (As pointed out by @nm123github)

3. We need to deal with panel resizing still, when you toggle between "addon panel open" and "addon panel closed" at any rate.

nm123github:
@tmeasday had no idea you were working on this too 🙂 , thanks for the review..

**- I'm not sure we should have a story about this in the marko example app? I think these kind SB-wide stories should just be in official storybook, right?**
Agreed, i will remove those and add something like that in official storybook

**- We need to deal with panel resizing still, when you toggle between "addon panel open" and "addon panel closed" at any rate.**
Can you please tell me little more about this? Maybe a gify or some screenshots might help. Also, feel free to PM on slack.

**From playing around with it a bit, I think the options need to be reset, otherwise you get super weird behaviour as the options drift as you change stories.**
^ you know, i would kinda agree with that, but @ndelangen mentioned "setOptions being persistent is expected behavior I think?". Also, i am not sure if there is current functionality that allows to setOptions specifically for a story, or if that will need to be built.

ndelangen:
Let me know when it's ready for review?

tmeasday:
I've done 1. in this PR also. So from my perspective the remaining issue is just resizing the panels when they are hidden. Let me make a couple of gifs.

tmeasday:
> "setOptions being persistent is expected behavior I think?". Also, i am not sure if there is current functionality that allows to setOptions specifically for a story, or if that will need to be built.

I think maybe I have confused things a little -- this PR retains top-level `setOptions` being persisted, what it does not persist is story-level options being persisted. For instance in current master we have this behaviour:

ezgif com-video-to-gif

Notice how the Actions story's addons panel location depends on the previous story you looked at. It's even more of a problem when a story is e.g. hiding the panel.

If the user sets options on *every* story it is not a problem, but this isn't what users will expect i don't think. In any case I think this PR has the best of both worlds (story-level options are not remembered, global options set via `setOptions` are).

nm123github:
Thanks for taking care of this @tmeasday

tmeasday:
No worries @nm123github -- just standing on your shoulders

tmeasday:
On the subject on panel resizing I cannot reproduce the problem on `master` so hopefully I can figure out what's different on this branch and solve it. Will look later today.

igor-dv:
@tmeasday, probably I am missing something, but playing with [preview](https://deploy-preview-3965--storybooks-official.netlify.com) of your PR (didn't review the code yet), I do see options are persistent between stories. For example, the "Custom Storybook" name is remaining when navigating out of the story that changed it. 

BTW I would also limit what people **can** change. For example, not allow changing hierarchy options.

tmeasday:
@igor-dv 

>  probably I am missing something, but playing with preview of your PR (didn't review the code yet), I do see options are persistent between stories. For example, the "Custom Storybook" name is remaining when navigating out of the story that changed it.

Nope, you are not missing something, I thought I had fixed the problem but it turns out only partially.

Any ideas on how we could fix it properly? Would using a subscription help? I'm aware we don't want options to thrash as you go between stories. Ideally you would have the `withOptions` decorator applied at all times, but I don't think we can guarantee that.

> BTW I would also limit what people can change. For example, not allow changing hierarchy options.

Yeah I think maybe we should just do the `shortcutOptions` and not the `uiOptions`

tmeasday:
Here is a video of the issue with panel resizing. For some reason it doesn't happen for the deployed storybook. Anyone have any idea why?

ezgif com-video-to-gif-3

tmeasday:
Ok, I investigated the panel resizing. I think the reason it doesn't happen on the deployed version is that it only happens in react dev mode. More details here: https://github.com/tomkp/react-split-pane/issues/309

I'm not sure if we should ship this feature until this bug is fixed? It seems like storybook is buggy to have a story hide the addon panel and the layout gets broken.

Can anyone think of an easy workaround?

nm123github:
I could repro this locally. Does seem strange.

tmeasday:
I could also fork / PR react-split-pane to try and get the bug fixed quicker

wuweiweiwu:
@tmeasday If you could PR a fix that'll be amazing. I'm also a collaborator on `react-split-pane` so i can review it. I just don't have the bandwidth right now to write a fix :(

igor-dv:
> Any ideas on how we could fix it properly? 

I will try to play with this PR a bit =)

tmeasday:
@igor-dv I hope you saw my branch attached to the issue at
https://github.com/tomkp/react-split-pane/issues/309

igor-dv:
I've checked it out (global options thing). Apparently, the problem is that we don't know that story is "navigated out" to apply the `globalOptions` again. 

Maybe we should have `globalOptions` set through the global decorator? 
This worked (a bit ugly but we can use it as a decorator): 
```js
addDecorator(story => {
  setOptions({
    hierarchySeparator: /\/|\./,
    hierarchyRootSeparator: /\|/,
    theme: themes.dark,
  });
  return story();
});
```

Also, we can deep-equal between the "new options" and the "last options" and do not trigger `channel.emit` if nothing really changed.

shilman:
@tmeasday @igor-dv is there anything blocking this PR?

tmeasday:
Apologies,

There are two outstanding problems on this PR:

1. Resetting options when you navigate away from the story. @igor-dv's solution is probably fine for now, people can either add a global options decorator (and not use `setOptions` directly) if they are using per-story options, or just call `setOptions` once (the old API) if not.

2. Resolving the issue with split pane. I'll have another look at this.

tmeasday:
Ok, I think resolved the issue with `StrictMode` + `SplitPane`, hopefully with @wuweiweiwu's help we can get that out reasonably promptly. I will now look at @igor-dv's suggestion

Hypnosphi (approved):

wuweiweiwu:
Ill open an issue to update react-split-pane to 0.1.83, which would solve the `StrictMode` issues. Thanks again @tmeasday

Hypnosphi:
I don't think it can be merged to 3.4

wuweiweiwu:
On further investigation, @tmeasday I think I'll have to make additional changes, updating to 0.1.83 breaks the drag action in layouts somehow, it drags extremely slow.

tmeasday:
@igor-dv, yeah this v4 only as params aren’t in v3

@wuweiweiwu any idea why? Afaict the function I changed is only called from `componentDidMount` and `getDerivedStateFromProps`, it’s explicitly not called from the drag handler. 

Let me know if I can assist!

igor-dv:
We can just add a deprecation warning to v3.4. Just so sad to pollute a beauty of the v4 with deprecations 😢

wuweiweiwu:
@tmeasday not sure yet. I think `getDerivedStaetFromProps` being called multiple times per render in dev mode breaks stuff since our size calculation is not pure. I'll dig into it tonight.

PR #3967, opened 2018-08-06
Emit messages when stories fail to render

tmeasday:
Issue:

There is no way for external tools or the manager context to know when a story fails to render

## What I did

Emitted two events when the story errors or throws.

## How to test

I added stories for the two cases, however there are a couple of problems/questions:

1. The current use of emotion's `ThemeProvider` in official storybook means that my "error" story ends up throwing rather than triggering the react layer's error code. Any ideas on how to work around this @ndelangen? -- I don't want this decorator to apply for this story -- Perhaps we need to do something hacky like `addDecorator(story, ({kind})) => kind === 'Errors' ? story() : <ThemeProvider>...)`

2. I'd like to also throw the error if the story *throws sometime after the initial render*. Is this out of scope? It would be useful for Chromatic to know that a component has errored even if it does it async. This would require work from all the view layers so perhaps is a separate PR. Could be out of scope for storybook in any case.

3. Is there a way to test the event is emitted correctly?

### DIFF
--- a/examples/official-storybook/config.js
+++ b/examples/official-storybook/config.js
@@ -18,7 +18,14 @@ setOptions({
   theme: themes.dark,
 });
 
-addDecorator(story => <ThemeProvider theme={themes.normal}>{story()}</ThemeProvider>);
+addDecorator(
+  (story, { kind }) =>
+    kind === 'Core|Errors' ? (
+      story()
+    ) : (
+      <ThemeProvider theme={themes.normal}>{story()}</ThemeProvider>
+    )
+);
 
 configureViewport({
   viewports: {
--- a/examples/official-storybook/stories/core.stories.js
+++ b/examples/official-storybook/stories/core.stories.js
@@ -3,6 +3,7 @@ import { storiesOf, addParameters } from '@storybook/react';
 import addons from '@storybook/addons';
 import Events from '@storybook/core-events';
 import { Button } from '@storybook/components';
+import { navigator } from 'global';
 
 const globalParameter = 'globalParameter';
 const chapterParameter = 'chapterParameter';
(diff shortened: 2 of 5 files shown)


ndelangen:
Please explain why the `<ThemeProvider>` is involved here at all?

Would a try-catch decorator work here?

tmeasday:
> Please explain why the <ThemeProvider> is involved here at all?

Sorry, that was a bit low on detail.

So the React integration has some code that calls `setError` if a story returns something that is not renderable by React. However, the use of the `<ThemeProvider>` decorator means that when a story wrapped by it returns something not renderable, the `ThemeProvider` internally breaks and an exception is thrown. Ultimately this leads to `setException` being called (as you can see from this PR, there is a subtle difference in the meaning).

I think we need a story that results in `setError` being called. AFAICT, the only way we can make this happen is to not apply the decorator to this story. Can you think of a better way than my ternary above?

tmeasday:
What do you think @ndelangen? Also any thoughts on how to test it?

tmeasday:
I added a test and sorted out the theming thing (let me know if you are happy with that @ndelangen)

igor-dv:
Yeah, it's a nice API. What if you would like to exclude stories from one test but include them into another one?

tmeasday:
> What if you would like to exclude stories from one test but include them into another one?

Do you mean in a different "suite"? We could also add something like `{ storyshots: { suite: 'X' } } ` -- would that be good?

tmeasday:
This is ready to merge I think, we can add the params API to storyshots later https://github.com/storybooks/storybook/issues/4010

ndelangen (approved):

thedamon:
I'm confused... is this now a feature of storyshots? 
If so.. a way to distinguish between storyshots and imageSnapshot from storyshots-puppeteer in these story options would be a, maze, ing.

Issue #3981, opened 2018-08-08
Bogus warning: impossible to import loader for .js


zsgalusz:
If you are reporting a bug or requesting support, start here:
### Bug or support request summary

This may already be an older issue, but I just noticed it. When building storybook it shows a nonsensical warning:

```
info @storybook/react v4.0.0-alpha.16
info
info => Loading custom addons config.
WARN => Custom configuration file redacted/.storybook/webpack.config.js is detected
WARN    but impossible to import loader for .js
info => Loading custom webpack config (full-control mode).
```

From a quick investigation of the [webpack config loader](https://github.com/storybooks/storybook/blob/master/lib/core/src/server/loadCustomWebpackConfig.js) this is because  `interpret` returns null for `.js` extensions (which is fine, since no loader is needed for them), which on the other hand storybook's webpack config loader misinterprets and warns that it can't load the loader (which it doesn't need). Regardless of the warning then it proceeds to require it so this is just a minor cosmetical issue.

### Steps to reproduce

Build storybook with a custom webpack config in a js file. At the start of the build, it will show the above warning.

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/react v4.0.0-alpha.16

### Affected platforms 

- I assume any, so long as a custom webpack.config.js file is present

igor-dv:
Thanks, @zsgalusz, It's not an old issue. This ability was introduced a few alphas ago.

Hypnosphi:
Released as [4.0.0-alpha.18](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.18)

PR #4009, opened 2018-08-12
Fix custom webpack config warnings

igor-dv:
Issue: Fix the #3981 + a bit more refactoring

## What I did
I've refactored `loadCustomWebpackConfig.js` and now we have a `serverRequire.js` that will be able to load client's scripts based on their preferred compiler on **server** (Currently the only usecase is webpack.config, but I assume will be more, for example: presets, single config file, etc')


ndelangen (approved):

emmenko:
Any ETA on when this fix will be released? 😇

PR #4060, opened 2018-08-23
#3780: Do not display wrong warning when loading ".js" or ".json"

BR0kEN-:
Issue: https://github.com/storybooks/storybook/issues/3780
Description: https://github.com/storybooks/storybook/issues/3780#issuecomment-402491084

### DIFF
--- a/lib/core/src/server/serverRequire.js
+++ b/lib/core/src/server/serverRequire.js
@@ -84,7 +84,7 @@ export default function serverRequire(filePath) {
 
   const moduleDescriptor = interpret.extensions[candidateExt];
 
-  if (registerCompiler(moduleDescriptor) === 0) {
+  if (moduleDescriptor !== null && registerCompiler(moduleDescriptor) === 0) {
     logger.warn(`=> File ${candidatePath} is detected`);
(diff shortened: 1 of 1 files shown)


lib/core/src/server/serverRequire.js
```diff
@@ -84,7 +84,7 @@ export default function serverRequire(filePath) {
 
   const moduleDescriptor = interpret.extensions[candidateExt];
 
-  if (registerCompiler(moduleDescriptor) === 0) {
+  if (moduleDescriptor !== null && registerCompiler(moduleDescriptor) === 0) {
```
igor-dv: So, is the problem only with a `.json`? because there is already a check ☝️  for `.js`.

BR0kEN-: Yeah, you are right. My proposition to not show the warning when `moduleDescriptor` is `undefined` (no key in `interpret.extensions`) or `null` (as for `.js` and `.json`).

igor-dv (approved):
Thanks. we are having some troubles with CI, once it fixed I will merge.

igor-dv:
Can you please merge from master?

BR0kEN-:
Done. Let's wait for the builds.

Issue #4056, opened 2018-08-22
HMR for react-native < 0.55 "Story of ... named ... already exists"


michaelduminy:
### Summary
When hot-reloading occurs I get the "Story of ... named ... already exists" error which breaks the flow of development and removes a lot of the reason I'd like to use storybook.

When I remove the thrown error hot-reloading works just fine.

I don't fully understand how HMR works so I can't offer a solution in that regard but why does storybook throw an error when you attempt to add a story that already exists? Is it really necessary? 

### Solution Ideas
- Move message to a `console.warn`
- Keep Error but don't throw when an `ignoreXError` property is set

I am happy to create a PR with one of these changes, but I'd like to understand the reasoning behind the error first.

jqn:
A quick workaround I found was to follow the examples here https://github.com/infinitered/ignite-ir-boilerplate-andross/blob/master/boilerplate/storybook/storybook.ejs which keeps only on storiesOf per story

PR #4061, opened 2018-08-23
Allow replacing of stories with a warning (rather than an error)

mikeduminy:
Issue: #4056

## What I did
Replaced the error thrown when adding an already existing story with a console.warn. Created a test that covers this case. Note: console.warn is mocked as the jest config throws an error when this is called.


michaelduminy:
I originally made the changes on a release branch so I'm just running through the tests on master to make sure nothing has broken.

michaelduminy:
I can fix the tests for master, but that will break the tests for 3.4 when and if the commit is cherry picked.

lib/core/src/client/preview/client_api.js
```diff
@@ -95,7 +95,7 @@ export default class ClientApi {
       }
 
       if (this._storyStore.hasStory(kind, storyName)) {
-        throw new Error(`Story of "${kind}" named "${storyName}" already exists`);
+        logger.warn(`Story of "${kind}" named "${storyName}" already exists`);
```
igor-dv: Basically, I agree with this change, though I wonder whether we want to replace the existing story with this `kind`
@tmeasday you've worked a lot in this area. WDYT ?

igor-dv:
@storybooks/team, does anyone want to give a feedback?

wuweiweiwu (commented):
LGTM. Nice tests!

Hypnosphi (approved):

michaelduminy:
Thanks for the approval. If there's anything I can do to get this into the v3 stream please let me know.

ndelangen (approved):

ndelangen:
I think we need to merge in master, I can do it today.

Hypnosphi:
@michaelduminy I will cherry-pick it to `release/3.4` branch right before the next patch release, you don't need to do anything

PR #4076, opened 2018-08-25
Add missing `@babel/plugin-proposal-class-properties` dependency

Hypnosphi:
Issue: #4071 

This also bumps all `@babel/*` dependencies to `rc3`


shilman (approved):
LGTM

Issue #2771, opened 2018-01-18
Cannot find relative presets defined in `.babelrc`


vjpr:
### Issue details

```
ERROR in ./.storybook/addons.js
Module build failed: Error: Couldn't find preset "./packages/babel-preset-foo" relative to directory "/Users/Vaughan/dev-live/foo/.storybook"
```

Project tree:

```
- .babelrc
- packages
  - babel-preset-foo
```

_.babelrc_

```
{"presets": ["./packages/babel-preset-foo"]}
```

### Please specify which version of Storybook and optionally any affected addons that you're running

```
    "@storybook/react": "^3.3.9",
    "@storybook/addon-actions": "^3.3.9",
    "@storybook/addon-links": "^3.3.9",
    "@storybook/addons": "^3.3.9"
```

vjpr:
Simple fix was to add a `.babelrc` file to `.storybook` and correct the relative preset.

vjpr:
This didn't actually solve my problem. It appears that babel tries to read relative presets from the dir of the file it is processing. So the preset path would need to be absolute, but this is not possible because `.babelrc` is static...

vjpr:
Seems related to https://github.com/storybooks/storybook/issues/172

vjpr:
The extends functionality should also be added for presets and plugins perhaps:

https://github.com/storybooks/storybook/blob/83d55c23fa124fa87cd061bdb5b5ce9453b393c1/app/vue/src/server/babel_config.js#L58-L66

Hypnosphi:
Also related: #2610 #2582

Hypnosphi:
Released as [`4.0.0-alpha.20`](https://github.com/storybooks/storybook/releases/tag/v4.0.0-alpha.20)

PR #4077, opened 2018-08-25
Use `babelrc: false` only for custom config in `.storybook` directory

### LINKED ISSUE #2582: Support for `.babelrc.js`
### Issue details

Currently `@storybook/react` only looks for Babel configuration in `.babelrc` file:

https://github.com/storybooks/storybook/blob/d63bf65afa75f2f6f4c0f63b0fedb990a8ea98d0/app/react/src/server/babel_config.js#L48

In `@babel/core` they support more configuration sources (`.babelrc.js`, `.babelrc`, `package.json`):

https://github.com/babel/babel/blob/7d798952d2a305cf77d24d1707f723f0a162aadd/packages/babel-core/src/config/files/configuration.js#L29

Looks like the functionality of reading Babel configuration is common enough to be extracted into its own package. I guess it's not good enough just to reimplement the same logic in Storybook repo. Or we can just use Babel's internals – though it will be not easy to maintain down the road.

Any thoughts?

### Steps to reproduce

1. Have Babel configuration in `.babelrc.js` file.
2. Run `@storybook/react` with something like `start-storybook -p 9001 -c .storybook`.
3. `.babelrc.js` is ignored

Expected behaviour: Storybook reads Babel configuration from `.babelrc.js` file.

- @storybook/react 3.3.1

### Affected platforms 

babel-loader 7.1.2
babel-core ^7.0.0-0
@babel/core 7.0.0-beta.34

Hypnosphi:
Fixes  #2582 #2771 #2610 #754 #2635 

This also removes the need to duplicate our presets (react, env etc) in user's app babel config


lib/core/src/server/loadCustomBabelConfig.js
```diff
@@ -43,24 +45,25 @@ function loadFromPath(babelConfigPath) {
...
-      babelConfig.extends = inConfigDir
-        ? path.resolve(configDir, babelConfig.extends)
-        : path.resolve(babelConfig.extends);
+      babelConfig.extends = path.resolve(configDir, babelConfig.extends);
     }
+    return babelConfig;
   }
 
-  return babelConfig || defaultConfig;
+  return isBabelLoader8() ? defaultConfig : {};
```
igor-dv: So, our default babel config is for babel8. After removing `babelrc: false`, can't we just return the default?

igor-dv:
BTW, by

> This also removes the need to duplicate our presets (react, env etc) in user's app babel config

Can we get rid of `wrapDefaultBabelConfig` ?

lib/core/src/server/loadCustomBabelConfig.js
igor-dv: Ah, just understood my mistake 🤦‍♂️

examples/marko-cli/package.json
```diff
@@ -19,8 +19,7 @@
   },
   "dependencies": {
     "marko": "^4.12.4",
-    "marko-starter": "^2.0.4",
-    "marko-widgets": "^7.0.1"
```
ndelangen: why delete the widgets?

examples/marko-cli/src/stories/addon-knobs.stories.js
```diff
@@ -17,27 +15,3 @@ storiesOf('Addons|Knobs/Hello', module)
       age,
     });
   });
-
-storiesOf('Addons|Knobs/Hello (Marko Widget)', module)
```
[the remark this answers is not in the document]
Hypnosphi: Looks like they decided to drop this functionality

examples/marko-cli/package.json
Hypnosphi: They are deprecated and don't support marko 4

Hypnosphi:
> You mean we can get rid of wrapDefaultBabelConfig ?

No, it means that our default config is now being merged with user's

examples/marko-cli/package.json
ndelangen: okay, clear!

ndelangen (approved):
This is great, this makes so much sense!

Issue #4039, opened 2018-08-17
@storybook/addon-storysource doesn't work when storiesOf is imported as alias


Everettss:
If you don't use default `storiesOf` name `@storybook/addon-storysource` doesn't find story source

```js
import { storiesOf as storiesOfAlias } from '@storybook/react';

const stories = storiesOfAlias('Example', module);
stories.add('default', () => <div>Story source</div>);
```

This problem is related to searching of fixed name `const STORIES_OF = 'storiesOf';` in [file](https://github.com/storybooks/storybook/blob/master/addons/storysource/src/loader/parse-helpers.js#L1).

igor-dv:
Yeah, I knew this moment would come... Currently, It supports a common use-case... For a more robust solution, it will require some more AST manipulation, like finding aliases. Also, the `stories.add` will probably break the navigation from within the storysource panel, because it searches for the pattern of `storiesOf().add().add()`

Everettss:
@igor-dv I totally agree that this problem is more general than I initially described. It's easy to solve only this problem of `import { storiesOf as storiesOfAlias } from '@storybook/react';`. But new usecases can occur like:

```js
import { storiesOf } from '@storybook/react';
const storiesOfAlias = storiesOf;

const stories = storiesOfAlias('Example', module);
stories.add('default', () => <div>Story source</div>);
```
Question is open: **is it worth to work on it?**

igor-dv:
I think any improvement is welcomed =) Do you want to work on it?

Everettss:
I've got some implementation only for this problem `import { storiesOf as storiesOfAlias } from '@storybook/react';`. I will take a look at some more general solutions.

PR #4012, opened 2018-08-13
feature: noStoriesOf loader option support for custome story

466023746:
Issue:#3993

PR #4079, opened 2018-08-26
fix(addon-storysource): use global CLIENT_API instead of injecting .addDecorator

Everettss:
Add `window.__STORYBOOK_CLIENT_API__` at the top of the file instead of adding .addDecorator with estraverse.

Originally this PR is intended to fix issue #4039. But I decided to use global [`__STORYBOOK_CLIENT_API__`](https://github.com/storybooks/storybook/pull/3058) and it fixes also other problems with renaming `storiesOf` function.


ndelangen:
@igor-dv could you take a look at this?

igor-dv:
Sure. Will check this today. 
Also, there is this #4012 PR that adds some functionality to storysource.

igor-dv:
@Everettss, sadly but `__STORYBOOK_CLIENT_API__`, is not a good solution, because it adds decorators to the `globalDecorators` list. 
That means every new story that is declared, adds its storysource to the list of `globalDecorators`, then this list is copied to the list of the story's decorators:

image

image

image

The first story that is imported, will have `x` decorators, then the story number `100`, will have `x + 100` decorators, and it's definitely not a good thing.

I have tested this with the `@storybook/html`  [preview](https://deploy-preview-4079--storybooks-html.netlify.com) from this PR, and here is the visualization of the problem:

storysource-bug

ndelangen:
@igor-dv how do we proceed with this?

Everettss:
@ndelangen give me 1 or 2 weeks (I'm on vacation now) an I will search for some replacement or simply admit that this is only worth burying.

ndelangen:
@Everettss

I'm absolutely not trying to pressure you, please don't feel as if you're required to do anything. You're not.

I'll leave it open, the bot will come along in a few weeks and remind us, and then close it automatically.

If you need help, please do reach out to us 🙇

ndelangen:
debatable whether it's a bug or feature, doesn't really matter IMHO

ndelangen:
@Everettss Let's have a chat on slack or discord about this, I have some ideas!

ndelangen:
Storysource should add it's data about stories via addParameters, I think this will pretty much solve all problems.

igor-dv:
I am going to close this. We will adjust storysource anyway.

PR #4084, opened 2018-08-27
Support `.storybook/babel.config.js`

Hypnosphi:
Issue: https://github.com/storybooks/storybook/issues/2758#issuecomment-416367489

### DIFF
--- a/lib/core/src/server/loadCustomBabelConfig.js
+++ b/lib/core/src/server/loadCustomBabelConfig.js
@@ -12,11 +12,12 @@ function removeReactHmre(presets) {
   }
 }
 
-// Tries to load a .babelrc and returns the parsed object if successful
-function loadFromPath(babelConfigPath) {
+// Tries to load a .babelrc or babel.config.js and returns the parsed object if successful
+function loadFromPath(babelConfigDir, defaultConfig) {
   let config;
-  if (fs.existsSync(babelConfigPath)) {
-    const content = fs.readFileSync(babelConfigPath, 'utf-8');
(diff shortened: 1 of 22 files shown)


lib/core/src/server/loadCustomBabelConfig.js
```diff
@@ -25,6 +26,17 @@ function loadFromPath(babelConfigPath) {
       logger.error(`=> Error parsing .babelrc file: ${e.message}`);
       throw e;
     }
+  } else {
+    try {
+      // eslint-disable-next-line import/no-dynamic-require,global-require
+      const jsConfig = require(path.resolve(babelConfigDir, 'babel.config.js'));
```
igor-dv: I think it's better to use `serverRequire` to support also `babel.config.ts`

igor-dv (commented):
I've started to work on presets and created these initial PRs #4027 #4043 (and also [presets-merge-babel-configs](https://github.com/storybooks/storybook/tree/core/presets-merge-babel-configs) branch). Considering this (and if everybody agrees with the API), maybe we can wait with this PR? If it will be released in v4, we won't be able to deprecate it in the next release

lib/core/src/server/loadCustomBabelConfig.js
Hypnosphi: Sure, I think I'll just close this one in favor of yours

igor-dv: Cool =) Would love to hear your thoughts there

PR #4092, opened 2018-08-29
Do not stale "good first issue"

igor-dv:
Issue: ☝️

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -10,6 +10,7 @@ exemptLabels:
   - 'do not merge'
   - 'needs review'
   - 'high priority'
+  - 'good first issue'
   - dependencies:update
 
 # Label to use when marking an issue as stale


Keraito (approved):

Keraito:
The message on the issues page also mentions the "help wanted" label for first time contributors. Would it make sense to also not stale that one.

danielduan:
We can also establish a protocol where if one of us verified and reproduced the bug, we apply the bug label and prevent it from automatically closing either.

danielduan:
I disagree that they are the same. Help wanted means it is an issue that needs attention, while good first issue means it is well documented and easy for a beginner to get started.

Keraito:
The message that is shown on top of the issues page:
image

I meant it more as that the reason we added "good first issue" to be excluded from stale is to keep it around for new contributors. Based on the above message, the "help wanted" label shares the same purpose towards new contributors, hence I suggested to also exclude it from stale. I didn't mean they were the same, sorry if that was unclear

PR #4095, opened 2018-08-29
fix #3933: Duplicated keys in PropTypes table 

### LINKED ISSUE #3933: Multiple empty lines in JSDoc causes key error in React when using Info Addon.
### Bug summary

Creating a JSDoc comment for a prop type with multiple empty lines causes an error in the browser development console.

```
warning.js:33 Warning: Encountered two children with the same key, ``. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in td (created by Td)
    in Td (created by PropTable)
    in tr (created by PropTable)
    in tbody (created by PropTable)
    in table (created by Table)
    in Table (created by PropTable)
    in PropTable
    in Unknown (created by Story)
    in div (created by Story)
    in div (created by Story)
    in div (created by Story)
    in div (created by Story)
    in div (created by Story)
    in Story
```

### Steps to reproduce

1. Create a Story which utilizes the Info Addon:
2. Create a JSDoc comment for a prop type which contains more than one empty line:

Reproduction: https://github.com/strothj/storybook-info-addon-docgen-key-error

- @storybook/react 4.0.0-alpha.14
- @storybook/addon-something 4.0.0-alpha.14

### Affected platforms 

- _If UI related, please indicate browser, OS, and version_
- _If dependency related, please include relevant version numbers_
- _If developer tooling related, please include the platform information_

### Screenshots / Screencast / Code Snippets (Optional)

image

1. Create a Story which utilizes the Info Addon:
```
import React from "react";
import { storiesOf } from "@storybook/react";
import SomeComponent from "../SomeComponent";

storiesOf("SomeComponent", module)
  .addParameters({ info: { inline: true } })
  .add("error", () => <SomeComponent />);
```

2. Create a JSDoc comment for a prop type which contains more than one empty line:
```
class SomeComponent extends React.Component {
  static propTypes = {
    /**
     * This is a required prop.
     *
     * With a multiple line description.
     *
     * After second line, causes error.
     */
    color: PropTypes.string
  };

  render() {
    return <div>Some Component</div>;
  }
}

export default SomeComponent;
```

gombek:
Issue: #3933 

## What I did
I've added check for duplicated lines, if line of text occur multiple times, index value will be added (as postfix) to the key.

## How to test

Appropriate test is included in this PR.

### DIFF
--- a/addons/info/src/components/PropTable.js
+++ b/addons/info/src/components/PropTable.js
@@ -6,9 +6,6 @@ import { Table, Td, Th } from '@storybook/components';
 import PropVal from './PropVal';
 import PrettyPropType from './types/PrettyPropType';
 
-const willItOccurAgain = (collection, item, itemIndex = 0) =>
-  collection.indexOf(item, itemIndex + 1) !== -1;
-
 export const multiLineText = input => {
   if (!input) {
     return input;
@@ -18,13 +15,9 @@ export const multiLineText = input => {
   const isSingleLine = arrayOfText.length < 2;
   return isSingleLine
     ? text
-    : arrayOfText.map((
-        lineOfText,
-        i // note: lineOfText is the closest we will get to a unique key
-      ) => (
-        <span
-          key={willItOccurAgain(arrayOfText, lineOfText, i) ? `${lineOfText}.${i}` : lineOfText}
-        >
+    : arrayOfText.map((lineOfText, i) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <span key={`${lineOfText}.${i}`}>
           {i > 0 && <br />} {lineOfText}
         </span>
       ));
--- a/addons/info/src/components/PropTable.js
+++ b/addons/info/src/components/PropTable.js
@@ -6,6 +6,9 @@ import { Table, Td, Th } from '@storybook/components';
 import PropVal from './PropVal';
 import PrettyPropType from './types/PrettyPropType';
 
+const willItOccurAgain = (collection, item, itemIndex = 0) =>
+  collection.indexOf(item, itemIndex + 1) !== -1;
+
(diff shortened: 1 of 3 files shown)


addons/info/src/components/PropTable.js
```diff
@@ -19,7 +22,9 @@ export const multiLineText = input => {
         lineOfText,
         i // note: lineOfText is the closest we will get to a unique key
```
wuweiweiwu: Comment is misleading. Please remove

addons/info/src/components/PropTable.js
```diff
@@ -6,6 +6,9 @@ import { Table, Td, Th } from '@storybook/components';
 import PropVal from './PropVal';
 import PrettyPropType from './types/PrettyPropType';
 
+const willItOccurAgain = (collection, item, itemIndex = 0) =>
```
wuweiweiwu: Any particular reason that there is a default? Seems like its always specified

addons/info/src/components/PropTable.js
```diff
@@ -6,6 +6,9 @@ import { Table, Td, Th } from '@storybook/components';
...
+  collection.indexOf(item, itemIndex + 1) !== -1;
```
wuweiweiwu: Also a bit hesitant to use this to generate a key, seems a bit unperformant and a bit of an overkill

ndelangen (approved):

addons/info/src/components/PropTable.js
wuweiweiwu: I think that would be good! It'll simplify the logic which is good

Keraito (approved):

Issue #4063, opened 2018-08-23
React-Native: "Can't find variable: document"


sregg:
### Bug

Getting "Can't find variable: document" error when launching React-Native app with storybook.

![image](https://user-images.githubusercontent.com/1479283/44545483-30202480-a6e3-11e8-92fe-e301b74a8840.png)

### Steps to reproduce

1. Create React_Native 0.56 app: `react-native init TestStoryBookReactNative`
2: Install storybook CLI 4.0.0-alpha.16: `npm i -g @storybook/cli@4.0.0-alpha.16`
3. Install storybook in project: `cd TestStoryBookReactNative` and `getstorybook`
4. Run storybook: `npm run storybook`
5. See this error: https://github.com/storybooks/storybook/issues/3897
6 Install Babel 7: `yarn remove babel-core` and `yarn add -D @babel/core@7.0.0-beta.47 babel-core@7.0.0-bridge.0`
7. Run storybook again: `npm run storybook` no more error
8. Check that http://localhost:7007/ works
9. Launch iOS app: `react-native run-ios`
10. Get "Can't find variable: document" error

### Please specify which version of Storybook and optionally any affected addons that you're running

my `package.json`:

```
{
  "name": "TestStoryBookReactNative",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start",
    "test": "jest",
    "storybook": "storybook start -p 7007"
  },
  "dependencies": {
    "react": "16.4.1",
    "react-native": "0.56.0"
  },
  "devDependencies": {
    "@babel/core": "7.0.0-beta.47",
    "@storybook/addon-actions": "^4.0.0-alpha.16",
    "@storybook/addon-links": "^4.0.0-alpha.16",
    "@storybook/addons": "^4.0.0-alpha.16",
    "@storybook/react-native": "^4.0.0-alpha.16",
    "babel-core": "7.0.0-bridge.0",
    "babel-jest": "23.4.2",
    "babel-preset-react-native": "^5",
    "babel-runtime": "^6.26.0",
    "jest": "23.5.0",
    "prop-types": "^15.6.2",
    "react-dom": "16.4.1",
    "react-test-renderer": "16.4.1"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

ritz078:
did you find any solution for this?

sregg:
@ritz078 nope, I gave up...

[...]

PR #4097, opened 2018-08-30
Extract channels store

igor-dv:
Issue: #4063

## What I did
I've extracted ChannelsStore out of the AddonsStore

### DIFF
--- a/app/react-native/src/preview/index.js
+++ b/app/react-native/src/preview/index.js
@@ -3,12 +3,11 @@
 import React from 'react';
 import { NativeModules } from 'react-native';
 import parse from 'url-parse';
-import addons from '@storybook/addons';
-
+import channels from '@storybook/addons/channels';
 import Events from '@storybook/core-events';
 import createChannel from '@storybook/channel-websocket';
-import { EventEmitter } from 'events';
 import { StoryStore, ClientApi } from '@storybook/core/client';
+import { EventEmitter } from 'events';
 import OnDeviceUI from './components/OnDeviceUI';
 import StoryView from './components/StoryView';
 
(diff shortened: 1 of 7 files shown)


Gongreg:
I am afraid this will not help when we start using addons inside react native.
I would like to reuse same addon logic, that means I need to import @storybook/addons inside react-native code.

Is it really necessary to have UI code inside addons? Couldn't it be moved somewhere to keep addons without UI code?

igor-dv:
CC @ndelangen 

There is a `TabWrapper` that was added for performance optimization. Maybe we can refactor it to have a RnTabWrapper for rn usage...

igor-dv:
BTW, we need to fix this to v4 anyway

Gongreg:
@igor-dv I will be talking with @ndelangen soon. I will also try to talk about this issue.

Gongreg:
Hey @igor-dv,
What we've decided to do is to remove that check for legacy addons from this module.

It will be moved somewhere into /lib/ui.

So for now only these lines should be removed.
https://github.com/storybooks/storybook/blob/master/lib/addons/src/index.js#L56-L62

This will solve the blocker and reimplementing the check in ui can be a separate task.

Will you update the PR or should I create a new one?

igor-dv:
> Will you update the PR or should I create a new one? 

I can finish it later, or if you want you can create a new one. I think we need to fix it asap. at least for the next alpha. So if you have time, it will be great.

Gongreg:
@igor-dv Do we still want to extract the channels?

igor-dv:
Nope =) at least not now.

igor-dv:
If our twitter could to make it public that'd be great 😆

PR #4100, opened 2018-08-31
Disabled import for TabWrapper

Gongreg:
Temporary disabled import for TabWrapper to fix RN version.

#4063 #4097


ndelangen (approved):
I will bring back this feature in lib/ui somewhere

PR #4107, opened 2018-09-02
[WIP] presets - merge default babel configs

igor-dv:
Issue: This PR is based on the #4043 (for the reviewability)

## What I did
I've removed `wrapDefaultBabelConfig` in apps, and added the ability to merge babel configs. This will ease the app creation process and presets usage


lib/core/src/server/config.js
```diff
@@ -2,16 +2,12 @@ import path from 'path';
...
-
 function customPreset({ configDir }) {
   return serverRequire(path.resolve(configDir, 'presets')) || [];
 }
 
 function getWebpackConfig(options, presets) {
-  const { configDir, wrapDefaultBabelConfig = noopWrapper } = options;
-
-  const babelOptions = presets.extendBabel({}, { configDir, wrapDefaultBabelConfig });
+  const babelOptions = presets.extendBabel({}, options);
```
Hypnosphi: This breaks Babel 6 support, because it adds our defaults (which are Babel 7 only) to all configs

Hypnosphi: We really need separate `extendDefaultBabel` and `extendBabel`

Hypnosphi: Actually, react-docgen is probably the only usecase for the latter

igor-dv: You mean `react-docgen` is the only working usecase?

Hypnosphi: It's the only plugin that we need to add to any configuration because it typically isn't used in user app

PR #4110, opened 2018-09-03
fixed wrong start script in docs

pavel1992:
Issue:

## What I did

Fixed issue, that i reported - https://github.com/storybooks/storybook/issues/4015


Gongreg (approved):
If this helps there is no reason not to add it.

Keep in mind we are going to remove packager from storybook in the future.

ndelangen (approved):

PR #4131, opened 2018-09-06
Fix incorrect use of withNotes on using-addons page

### LINKED ISSUE #3442: addon-notes causes story to not render and shows error instead
If you are reporting a bug or requesting support, start here:

Using `addon-notes` causes story to not render. Instead it renders an error "Expecting a valid React element from the story...Seems like you are not returning a correct React element from the story.
Could you double check that?"

image

### Steps to reproduce

Steps to reproduce:

Here is a link to a repository where I have recreated the problem: https://github.com/terrencewwong/storybook-addon-notes-example

To summarize, this is what I did:

- install react, react-dom, storybook, and addon-notes
- follow the setup instructions for addon-notes
- write a simple story that uses the notes addon

This is the code for the story that is in the repo I linked:

```js
import React from 'react';

const Component = () => <div>Hello</div>

    .add('default', () => <Component />, { notes: 'A very simple component' });
```

- `@storybook/react v3.4.2`
- `@storybook/addon-notes v3.4.2`

jneuendorf:
This PR fixes #3442.


jneuendorf:
Sorry, this PR accidentally includes unwanted commits.

PR #4133, opened 2018-09-06
Upgrade to babel-plugin-react-docgen@2

danielduan:
Issue:
https://github.com/storybooks/babel-plugin-react-docgen/issues/60

## What I did
I published a final release version and tested with Babel 7

## How to test


ndelangen (approved):

ndelangen:
@danielduan 
```
error Couldn't find any versions for "babel-plugin-react-docgen" that matches "^2.0.0"
```
is the error in the CI

Did you spell the version correctly in the `package.json` or during publishing?

Hypnosphi (changes requested):
CLI test fails on Babel 6 fixture.

Looks like we should add different versions of this plugin depending on whether the user has Babel 6 or 7. You can use something like this for detection: https://github.com/storybooks/storybook/blob/master/lib/core/src/server/loadCustomBabelConfig.js#L48

pksunkara:
@danielduan Also, we need the release of react-docgen after this PR is merged, https://github.com/reactjs/react-docgen/pull/294. And then a new release of babel-plugin-react-docgen.

Hypnosphi:
Right now `babel-plugin-react-docgen@1` seems to work ok with Babel 7. So maybe we shouldn't upgrade unless we deprecate Babel 6.

ndelangen:
@danielduan please re-open if you think we should do this, or open a new PR

PR #4134, opened 2018-09-07
CHANGE index.html.ejs to use files over chunks && UPGRADE generate-page-webpack-plugin

ndelangen:
Issue: generate-page-webpack-plugin was using object spread & it now exposes a more reliable entrypoint.getFiles() as `files`.

This PR upgrades to the latest version & changes the html.ejs template to use the `files` property.

This implements compatibility with https://github.com/webpack-contrib/mini-css-extract-plugin, possibly other css extraction plugins as well.

### DIFF
--- a/lib/core/src/server/templates/index.html.ejs
+++ b/lib/core/src/server/templates/index.html.ejs
@@ -7,17 +7,23 @@
     <title>Storybook</title>
     <link rel="shortcut icon" href="favicon.ico?v=1" />
 
+    <% if (options.headHtmlSnippet) { %>
+      <%- options.headHtmlSnippet %>
+    <% } %>
+      
     <% if (options.links) { %>
       <% for (item of options.links) { %>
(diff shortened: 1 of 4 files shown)


lib/core/src/server/templates/index.html.ejs
```diff
@@ -7,17 +7,23 @@
     <title>Storybook</title>
     <link rel="shortcut icon" href="favicon.ico?v=1" />
 
+    <% if (options.headHtmlSnippet) { %>
+      <%- options.headHtmlSnippet %>
+    <% } %>
+      
     <% if (options.links) { %>
```
diagramatics: Is there still a point on having `options.links` when we don't (well, can't) use them?

Hypnosphi (approved):

lib/core/src/server/templates/index.html.ejs
ndelangen: Well not really no. we do not expose a way to set this.

Issue #4144, opened 2018-09-08
Stories Panel is quite slow


theKashey:
### Bug or support request summary

One could not change stories quickly, due to performance issues. Stories pannel update itself requires about half second to redraw.

### Steps to reproduce

Create 100 stories.

### Screenshots / Screencast / Code Snippets (Optional)

Perfomance graph
<img width="281" alt="screen shot 2018-09-08 at 5 29 11 pm" src="https://user-images.githubusercontent.com/582410/45251865-b51c6680-b390-11e8-94f9-4cf2fe9bc423.png">

300ms to change one story in storybook's own storybook
<img width="216" alt="screen shot 2018-09-08 at 5 28 17 pm" src="https://user-images.githubusercontent.com/582410/45251866-b51c6680-b390-11e8-89c3-fa117a0f1dd9.png">

theKashey:
For everyone, interested in speeding up storybook right now, by __removing__ slow stories pannel
```js
import Stories from '@storybook/ui/dist/modules/ui/components/stories_panel/';
import { remock } from 'react-remock';
remock.mock(Stories);
```
^ it will just delete this pannel in runtime, without any codechanges from storybook.

PR #4145, opened 2018-09-08
Remove Stories and Addon panels when invisible

theKashey:
Issue: #4144

## What I did

I was trying to speedup Stories pannel rendering, but it's not so easy, as long problem lays inside `@ndelangen/react-treebeard`.
For now I am doing the simplest thing possible - not rendering Stories and Addon pannel when they are not visible. This will make tools like screenshot generation much faster.

## How to test

There are no visual or behavior changes.


ndelangen (approved):

ndelangen:
Code looks fine to me, can you update the tests?

theKashey:
Ah, sorry. Forgot to run the tests. My bad.
The root cause is hidden inside storybook/lib/ui/src/modules/ui/components/stories_panel/stories_tree/index.js : mapStoriesHierarchy - for every render it produces a "deeply" new object, bypassing all possible optimizations.

As I could see this file does not exists in your PR, and also there is no logic around generation some complex object for `Treeview.js`, so look like you already handled this.

ndelangen:
Indeed my refactor has completely changed the way stories are stored and retrieved.

It should be much faster, though I have not benchmarked it, It definitely feels faster.

theKashey:
> Error: Couldn't find package "remap-istanbul@^0.6.4" required by "karma-remap-istanbul@^0.2.1" on the "npm" registry.

🙀🤔

ndelangen:
I think removing the addons panels shouldn't be done, their state will be lost, and they'll never be able to re-acquire some stories data.

Storysource breaks for example. 🤔 

The nav can be removed I think, it's state is managed by the manager store. Though I do not know if removing the component will bring big performance gains, as the calculations will probably still happen in the store.

theKashey:
Let's wait for your PR - if it will solve performance issues - then ok. If no - then lets return to more radical solutions, like removing panels :)
Closing this for now, and wishing you luck with a state management.

ndelangen:
Haha thanks @theKashey, it's going well actually!

Thank you for your time and energy, I'm hoping you'll find something else in storybook you want to fix/change! 🙇

PR #4148, opened 2018-09-09
Update README.md

minaseem:
Correcting heading

Issue:

## How to test

### DIFF
--- a/addons/storyshots/storyshots-core/README.md
+++ b/addons/storyshots/storyshots-core/README.md
@@ -130,7 +130,7 @@ module.exports = {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


addons/storyshots/storyshots-core/README.md
```diff
@@ -130,7 +130,7 @@ module.exports = {
 };
 ```
 
-### <a name="deps-issue"></a>Why don't we install dependencies of each framework ?
+### <a name="deps-issue"></a>Why do we install dependencies of each framework ?
```
Hypnosphi: We actually don't

PR #4149, opened 2018-09-09
start-storybook: open browser tab on first compilation

Hypnosphi:
> norbert [11:35 AM]
@hypnos next up the auto-open browser feature?

## How to test

Run `yarn start`

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -88,6 +88,10 @@ Storybook now uses Babel 7. There's a couple of cases when it can break with you
     ```
   * If you're using Babel 6, make sure that you have direct dependencies on `babel-core@6` and `babel-loader@7`.
 
+### start-storybook opens browser automatically
+
+If you're using `start-storybook` on CI, you may need to opt out of this using the new `--ci` flag.
+
 ## From version 3.3.x to 3.4.x
 
 There are no expected breaking changes in the 3.4.x release, but 3.4 contains a major refactor to make it easier to support new frameworks, and we will document any breaking changes here if they arise.
--- a/lib/core/src/server/build-dev.js
+++ b/lib/core/src/server/build-dev.js
@@ -43,6 +43,7 @@ export async function buildDev({ packageJson, ...loadOptions }) {
     .option('--ssl-cert <cert>', 'Provide an SSL certificate. (Required with --https)')
     .option('--ssl-key <key>', 'Provide an SSL key. (Required with --https)')
     .option('--smoke-test', 'Exit after successful start')
+    .option('--ci', "CI mode (skip interactive prompts, don't open browser")
     .option('--quiet', 'Suppress verbose build output')
     .parse(process.argv);
 
@@ -59,7 +60,7 @@ export async function buildDev({ packageJson, ...loadOptions }) {
 
   const port = await getFreePort(program.port);
 
-  if (!program.smokeTest && program.port != null && port !== program.port) {
+  if (!program.ci && !program.smokeTest && program.port != null && port !== program.port) {
     const { shouldChangePort } = await inquirer.prompt({
       type: 'confirm',
       default: true,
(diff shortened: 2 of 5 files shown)


lib/core/src/server/build-dev.js
```diff
@@ -153,6 +155,8 @@ Would you like to run Storybook on port ${port} instead?`,
     logger.info(`Storybook started on => ${chalk.cyan(address)}\n`);
     if (program.smokeTest) {
       process.exit(stats.toJson().warnings.length ? 1 : 0);
+    } else {
+      opn(address);
```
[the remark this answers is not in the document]
Hypnosphi: Done

MIGRATION.md
```diff
@@ -88,6 +88,10 @@ Storybook now uses Babel 7. There's a couple of cases when it can break with you
     ```
   * If you're using Babel 6, make sure that you have direct dependencies on `babel-core@6` and `babel-loader@7`.
 
+### start-storybook opens browser automatically
+
+If you're using `start-storybook` on CI, you may need to opt out of this using the new `--ci` flag.
```
tmeasday: Maybe check for `process.env.CI` too? It's set by most CI systems

igor-dv: I actually think that `ci` is too limiting

package.json
```diff
@@ -53,7 +53,7 @@
     "postpublish": "yarn --cwd lib/cli test -o",
     "publish:alpha": "npm run publish -- --npm-tag=alpha",
     "repo-dirty-check": "node ./scripts/repo-dirty-check",
-    "start": "npm --prefix examples/official-storybook run storybook",
+    "start": "yarn --cwd examples/official-storybook storybook",
```
gabrielcsapo: does this require people to have yarn installed?

libetl: not really, I believe this is only applicable to the storybook project itself

libetl:
Devs restarting locally their already opened storybook will have to provide --ci, which might not be the most meaningful parameter name. WDYT ?

package.json
Hypnosphi: > does this require people to have yarn installed?

You need it anyway when you contribute to storybook, see `CONTRIBUTING.md`

Hypnosphi:
I don't think that opening more than one tab is so much a problem. CRA tries to [solve](https://github.com/facebook/create-react-app/tree/next/packages/react-dev-utils#openbrowserurl-string-boolean) it on OS X this using appleScript.

The problem is, it opens Google Chrome regardless of which browser user has as default. I don't think we should do it unless Google acquires us.

libetl:
I used react-scripts and I can tell that in windows/linux, it opens a new tab for each start command issued, which can be disturbing.
Ok, so I guess the situation cannot be improved today.

libetl (approved):

ndelangen:
I think it's worth it Yes I've had like 10 tabs open with the same dev setup once.. Thats' not great, but overall I think the experience is better then not having anything opened at all.

Thanks @Hypnosphi 👍👍👍

einarlove:
I'm on alpha.23 and it opens a new tab every time i restart storybook, regardless if I already have it open. I've used [react-dev-utils/openBrowser.js](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openBrowser.js) before,  and it uses opn like this pull-request adds, but also check with AppleScript if there is a tab already open.

MIGRATION.md
SleepWalker: yep. it would be much better to have something like `BROWSER=none start-storybook`. This is the same, that is used by cra or docusaurus.

Hypnosphi:
> The problem is, it opens Google Chrome regardless of which browser user has as default. I don't think we should do it unless Google acquires us.

ndelangen:
@einarlove question:
Why do you need to restart storybook so many times?

I mean -I- do restart it all the time, so i know there are use cases. But mine is I maintain storybook's code...

I'd like to know what's yours. Perhaps you don't need to restart storybook all the time?

---
If you have multiple storybooks you can run them on a different port

If storybook breaks down somehow and the only fix is a restart, let's fix it?

einarlove:
I always have it running and switching between branches might break it after a while. Suddenly there's a conflict in package.json and etc, and a restart might be required to get it up to speed again.

That's at least my experience. Maybe I'm overdoing it?

ndelangen:
Hmm, no that sounds like it could be quite common.

I guess our CLI could be smarter about broken code and recovering afterwards.

PR #4152, opened 2018-09-10
Code injector addon to dynamically inject code in the iFrame

nm123github:
Issue: At this point i don't believe there is a way to add remove css resources dynamically in a running storybook instance. I think there might be use-cases where folks want to check if a component works fine in different versions of a css/js library.

## What I did
[GIF demo](https://drive.google.com/file/d/1xjDT8cEKSCSNOFtzHtzDpkKFfOp0sFxt/view)
There is still work to be done on this. **Just wanted to get some preliminary comments/feedback**. Also, it probably wont be trivial to swap in/out javascript versions so mostly this will probably work with just css (at-least to begin with).

## Updated
Updated PR with new code and details below (in comments). Please ignore ^ initial details and demo.


Hypnosphi:
Makes sense to me

Hypnosphi:
Please add some docs on usage

addons/resources/src/ResourcePanel.js
```diff
@@ -0,0 +1,120 @@
...
+      throw new Error('Cannot find Storybook iframe');
+    }
+  }
+
+  componentDidUpdate() {
+    this.loadResources();
+  }
+
+  static getDerivedStateFromProps(props) {
+    return { resources: props.resources };
```
Hypnosphi: it's called on every `setState` in React 16. So `state.resources` will always be the same as `props.resources`

nm123github:
Thanks @Hypnosphi, i'm probably going to change a lot in here. but thanks for the initial feedback.

ndelangen:
Hey Nevilla, I know others are requesting a similar addon.

Think you'd have time to work on this more or share the work with someone else?

nm123github:
Hey @ndelangen , let me try to finish or atleast get it to a point where someone else could take over 👍. Let me know if you have any deadlines as far as releases go..

nm123github:
- Allow users to add/remove scripts and css in the iframe at runtime 🏃
- Allow users to add a default 
```
addDecorator(
  withResources({
    resources: `<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
<link rel="stylesheet" type="text/css" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css"></link>
<link rel="stylesheet" type="text/css" href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"></link>`,
  })
);
```
- Resources added/removed will be sticky (meaning they will persist across stories).
- Resources added/removed via the add-on will not change any of the resources added statically (via preview-head.html).
- Remember, in some cases (like adding event listeners on components) you might need to re-apply resources for the story!

examples/official-storybook/stories/addon-resources.stories.js
```diff
@@ -0,0 +1,23 @@
...
+);
+
+const myIcon = (label, className) => () => (
+  <div>
+    <i className={className} /> {label}
+  </div>
+);
+
+storiesOf('Addons|Resources', module)
+  .add('Primary Large Button', myButton('Primary Large Button', 'btn btn-lg btn-primary'))
```
igor-dv: Why do you use this strange components syntax? I didn't see it in use in the other stories... I think we need to be consistent with the examples.

addons/resources/src/index.js
```diff
@@ -0,0 +1,29 @@
+import addons, { makeDecorator } from '@storybook/addons';
+
+let hasSetInitialResources = false;
+
+export const withResources = makeDecorator({
+  name: 'withResources',
+  parameterName: 'resources',
+  skipIfNoParametersOrOptions: true,
+  allowDeprecatedUsage: true,
```
igor-dv: should we?

addons/resources/src/ResourcePanel.js
```diff
@@ -0,0 +1,161 @@
+import { document, DOMParser } from 'global';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '@storybook/components';
+import styled from '@emotion/styled';
+import AceEditor from 'react-ace';
```
igor-dv: Note for the future: I think we need to stick to a single editor in all these code highlighting components.
- Storysource has now a `react-syntax-highlighter` which is not an editor thought,
- Here it's an `AceEditor`
- IMO @libetl is adding another one

addons/resources/src/ResourcePanel.js
```diff
@@ -0,0 +1,161 @@
...
+            style={{ 'box-shadow': '5px 7px #888888' }}
+            value={resources}
+            name="resourcesdiv"
+            height="300px"
+            width="800px"
+            editorProps={{ $blockScrolling: true }}
+          />
+        </div>
+        <div style={{ float: 'right', 'padding-top': '17px' }}>
+          <Button onClick={this.apply.bind(this)}>APPLY RESOURCES</Button>
```
igor-dv: I think we need to have a "Disable" button ? Looks like it's needed to delete everything from the editor to disable the resources.

addons/resources/src/ResourcePanel.js
```diff
@@ -0,0 +1,161 @@
...
+        newEl.setAttribute(addedElAttr, '');
+      }
+      return newEl;
+    });
+  }
+
+  loadResources() {
+    const { resources = `` } = this.state;
+
+    const addElements = (domElements, i) => {
```
igor-dv: Looks like this function could be extracted, it doesn't use any inner state from the `loadResources` func

igor-dv (commented):
Should be listed in 
- https://github.com/storybooks/storybook#addons
- https://github.com/storybooks/storybook/blob/master/ADDONS_SUPPORT.md

(Also readme is missing)

addons/resources/src/ResourcePanel.js
libetl: We are not yet decided whether to add codemirror or monaco for the moment.
The change will probably happen in the tech/overhaul-ui branch

ADDONS_SUPPORT.md
```diff
@@ -15,6 +15,7 @@
 |[links](addons/links)                       |+|+|+|+|+|+|+| |+|+|+|
 |[notes](addons/notes)                       |+|+*|+|+|+|+|+| |+|+|+|
 |[options](addons/options)                   |+|+|+|+|+|+|+| |+|+|+|
+|[resources](addons/resources)               |+|+|+|+|+|+|+|+|+|+|+|
```
igor-dv: I can assume that RN is not supported

addons/resources/src/ResourcePanel.js
nm123github: added one.

igor-dv:
@nm123github, you need to update the snapshots

igor-dv:
IDK how but looks like `svelteshots` test is also broken =/

nm123github:
Just had a brief chat with @shilman . This will go in 4.1 (so lets not merge for now).

ndelangen:
Yeah that's what I feel is best. also.

Let's release 4.0.0 first.

igor-dv (approved):

nm123github:
Adding css is probably a much more common use case which should be covered in the addon-cssresources. Also, looks like there's a PR being worked on to integrate monaco-editor so in anycase probably best to wait for that.

Maybe will resurrect this sometime in future if there's a need.

wildeyes:
@nm123github what is the PR that integrates monaco-editor?

ndelangen:
@PlayMa256 has been working on a tiny lightweight story editor, but focus has shifted a bit. The storyState was changed for Args & my time was spend on the composition feature.

I'm sure we'll work on a true editor within storybook Q2.

The status is "on hold"

PR #4155, opened 2018-09-10
Presets - add babelDefault extension

igor-dv:
Issue: https://github.com/storybooks/storybook/pull/4107#discussion_r216251083

## What I did 
Added an option to extend "babelDefault" method in presets


Hypnosphi (approved):

pksunkara:
IMO, we should move `babelDefault` to `babel`.

igor-dv:
That was my initial thought. But I can see why it may break... maybe when we will remove the support of babel 6 we will be able to remove `babelDefault` as well 🤷‍♂️

Hypnosphi:
IMO, we should remove `babelDefault` to `babel`, and move `babel` to `webpack` =)

igor-dv:
can you please share an example ?

Hypnosphi:
```
export function webpack(config) {
  return {
    ...config,
    // find babel-loader here and add docgen plugin
  };
}
```

But maybe it's not worth it

igor-dv:
Yeah, I don't like it much. It is less API friendly and more error-prone. I have another idea about what to do with `babelDefault`. Will PR asap

PR #4158, opened 2018-09-11
[addon-storysource] Add `parser` for `prettierConfig`

stramel:
Issue: `No parser and no filepath given, using 'babylon' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.`

## What I did

Added `parser: 'babylon'`

## How to test

Specify a `prettierConfig` without a `parser` property


igor-dv:
So, it's not needed?

Issue #4136, opened 2018-09-07
TypeScript modules not resolved when using "baseUrl".


cwmrowe:
### Support request summary

`@storybook/angular` is not resolving non-relative imports. I believe non-relative imports are a relative new feature of TypeScript that allow you to reference imports from a `baseUrl` rather than always using relative imports to your own code.

I think that support for this feature should be a priority as VS Code will now rename all your modules to use non-relative imports automatically when you move/rename a file so its difficult problem to avoid.

### Steps to reproduce

Create a strorybook project with Angular 6.1.1 and use non-relative imports in your code, e.g.

**src/app/somefile.ts**
```
import { environment } from 'environments/environment';
// instead of import { environment } from '../environments/environment';
```

### Please specify which version of Storybook and optionally any affected addons that you're running

    "@storybook/addon-actions": "^4.0.0-alpha.20",
    "@storybook/addon-centered": "^4.0.0-alpha.20",
    "@storybook/addon-knobs": "^4.0.0-alpha.20",
    "@storybook/addon-links": "^4.0.0-alpha.20",
    "@storybook/addon-notes": "^4.0.0-alpha.20",
    "@storybook/addon-storysource": "^3.4.10",
    "@storybook/addons": "^4.0.0-alpha.20",
    "@storybook/angular": "^4.0.0-alpha.20",

### Affected platforms 

Windows 10

### Workaround

I have managed to work around this using the `TsconfigPathsPlugin` and modifying my `tsconfig.json` file to use `compilerOptions.paths` (note the mappings look slightly ridiculous).

**.storybook/tsconfig.json**
```
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "module": "es2015",
    "paths": {
      "app/*": ["app/*"],
      "pages/*": ["pages/*"],
      "shared/*": ["shared/*"],
      "environments/*": ["environments/*"]
    }
  },
  "exclude": [
    "../src/test.ts",
    "../src/**/*.spec.ts",
    "../src/**/*.stories.ts"
  ],
  "include": ["../src/**/*"]
}
```
**NB:** `compilerOptions.baseUrl` has a value of "./src" in my root `tsconfig.json`

**.storybook/webpack.config.js**
```
const path = require('path');
const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');

module.exports = (baseConfig, env, config) => {
  config.resolve.plugins = config.resolve.plugins || [];
  config.resolve.plugins.push(
[...]

Issue #2718, opened 2018-01-11
[Angular] Storybook doesn't work with non-relative imports


user753:
Non-relative imports like this
```
import {TestComponent} from "app/test/test.component"
```
doesn't work in stories file
```
stories/index.stories.ts
```
```
Module not found: Error: Can't resolve 'app/test/test.component' in 'stories'
```
I can fix it by using relative path
```
import {TestComponent} from "../src/app/test/test.component"
```
but if `test.component` has non-relative import it would have the same error

storybook/angular 3.4.0-alpha.1

johnslay:
Same thing is happening with create-react-app.

igor-dv:
By "non-relative imports", you mean paths mapping in tsconfig.json?

tigredonorte:
I have the same problem: "Module not found: Error: Can't resolve '@gorila/core/utils'"

in tsconfig.json:
`{
    ...
    "paths": {
      "@env/*": ["src/environments/*"],
     }
    ...
}
`

igor-dv:
A possible workaround could be this - https://github.com/storybooks/storybook/issues/2616#issuecomment-360973839
I've seen some other solutions, for example adding plugin - https://github.com/s-panferov/awesome-typescript-loader#advanced-path-resolution-in-typescript-20

I think we will need to add something like this by default to storybook.

morninng:
I have set following code in webpack.config.js
```
const path = require('path');
const genDefaultConfig = require('@storybook/angular/dist/server/config/defaults/webpack.config.js');
module.exports = (baseConfig, env) => {
  const config = genDefaultConfig(baseConfig, env);
  const lib_path = path.join( __dirname ,  "../src/lib" );
  config.resolve.alias['@lib-path'] = lib_path;
  return config;
};
```
this setting has solved the issue.

https://github.com/storybooks/storybook/issues/3258
Please see this in detail

DmitryEfimenko:
The workaround by @morninng almost worked with few adjustments:
I'm using `@storybook/angular` of version `4.0.0-alpha.10`. Because of that, the following line of code didn't work:
```
const genDefaultConfig = require('@storybook/angular/dist/server/config/defaults/webpack.config.js');
```
This file does not exist.
I had to patch `baseConfig` object and adjust `baseConfig.entry.styles`. For some reason the path there was not right.

igor-dv:
@DmitryEfimenko , can you please share your result for sake of clarity

[...]

PR #4162, opened 2018-09-11
Add support for "baseUrl" and "paths" from angular-cli

igor-dv:
Issue: #4136, #2718

## What I did
Added a part of the `resolve.modules` from angular-cli to SB webpack config


app/angular/src/server/angular-cli_config.js
```diff
@@ -3,6 +3,42 @@ import fs from 'fs';
...
+    const tsConfigDirName = path.dirname(tsConfigPath);
+    basicOptions.options.baseUrl = path.resolve(tsConfigDirName, baseUrl);
+  }
+
+  return basicOptions;
+}
+
+function getAngularCliParts(cliWebpackConfigOptions) {
+  // eslint-disable-next-line global-require, import/no-extraneous-dependencies
+  const ngcliConfigFactory = require('@angular-devkit/build-angular/src/angular-cli-files/models/webpack-configs');
```
kroeder: `ngcliConfigFactory` or `ngCliConfigFactory`?

app/angular/src/server/angular-cli_config.js
```diff
@@ -31,16 +67,16 @@ export function getAngularCliWebpackConfigOptions(dirToSearch) {
     project.sourceRoot
   );
 
+  const projectRoot = path.resolve(dirToSearch, project.root);
+  const tsConfigPath = path.resolve(dirToSearch, 'src/tsconfig.app.json');
```
kroeder: Why exactly the tsconfig in `src`? What about the ones in maybe `projects/lib-a/tsconfig.lib.json`? 
I just ask to make sure it's not necessary.

kroeder: Mh, it might be that angular already has taken care of this use case. I've read the initial issue and there was a workaround where the issue opener added paths to the tsconfig. This is exactly what angular does when adding new projects:

```
{
  "compilerOptions": {
    [...]
    "paths": {
      "test-library": [
        "dist/test-library"
      ],
      "test-library/*": [
        "dist/test-library/*"
      ]
    }
  }
}
```

igor-dv: Yes, but "paths" are not working with Storybook. You need to provide them as "alias" to webpack. Or to use for example - https://github.com/dividab/tsconfig-paths-webpack-plugin

It dosn't solve the "baseUrl" issue though like in #4136

kroeder (approved):

PR #4168, opened 2018-09-11
CLI refactor

Keraito:
Refactoring the CLI as an initial step to make a richer CLI:  https://github.com/storybooks/storybook/issues/1108#issuecomment-417921251. 

It might be needed to introduce some kind of way to persist data down the road for stuff like `--use-npm`, kinda depends on how much options are shared across the different commands.

Running `storybook` will display the help menu with all the commands and options.


tmeasday (approved):
LGTM!

lib/cli/bin/generate.js
```diff
@@ -1,201 +1,49 @@
...
+import { codeLog } from '../lib/helpers';
 
 const logger = console;
 
+if (process.argv[1].includes('getstorybook')) {
+  logger.log(chalk.yellow('WARNING: getstorybook is deprecated.'));
+  logger.log(chalk.yellow('To install storybook, use the following command instead:\n'));
+  codeLog(['storybook init']);
+  logger.log();
+  process.exit(0);
```
igor-dv: The depreciation should actually work. I assume `getstorybook` should do what `storybook init` does.

lib/cli/bin/generate.js
```diff
@@ -1,201 +1,49 @@
...
+program
+  .command('start')
+  .description('Start the Storybook server')
+  .option('-N --use-npm', 'Use NPM to start the Storybook server')
+  .action(options => yarnSpawnSync(options, ['run', 'storybook']));
 
-let projectType;
+program
+  .command('build')
+  .description('Build the Storybook server')
```
igor-dv: maybe something like "Build the Storybook static app" ?

lib/cli/lib/initiate.js
```diff
@@ -0,0 +1,192 @@
...
+  const CRNA_DISCUSSION =
+    'https://github.com/storybooks/storybook/blob/master/app/react-native/docs/manual-setup.md';
+
+  const runGenerator = () => {
+    switch (projectType) {
+      case types.ALREADY_HAS_STORYBOOK:
+        logger.log();
+        paddedLog('There seems to be a storybook already available in this project.');
+        paddedLog('Apply following command to force:\n');
+        codeLog(['getstorybook -f']);
```
igor-dv: getstorybook ?

lib/cli/lib/initiate.js
```diff
@@ -0,0 +1,192 @@
...
+
+      case types.RIOT:
+        return riotGenerator(npmOptions)
+          .then(commandLog('Adding storybook support to your "riot.js" app'))
+          .then(end);
+
+      default:
+        paddedLog(`We couldn't detect your project type. (code: ${projectType})`);
+        paddedLog(
+          "Please make sure you are running the `getstorybook` command in your project's root directory."
```
igor-dv: getstorybook ?

lib/cli/lib/initiate.js
```diff
@@ -0,0 +1,192 @@
...
+        );
+        paddedLog(
+          'You can also install storybook for plain HTML snippets with `getstorybook --html` or follow some of the slow start guides: https://storybook.js.org/basics/slow-start-guide/'
```
igor-dv: getstorybook ?

lib/cli/lib/initiate.js
Keraito: Oops slipped through, will change

lib/cli/bin/generate.js
Keraito: Sounds logical, will do

lib/cli/package.json
```diff
@@ -18,7 +18,8 @@
   "license": "MIT",
   "author": "Storybook Team",
   "bin": {
-    "getstorybook": "./bin/index.js"
+    "getstorybook": "./bin/index.js",
```
ndelangen: We could also decide to keep these 2 separate, and keep all backwards compatibility in the getstorybook.js file.

That might be easier to remove?

ndelangen:
This is great!

libetl:
That appeals to me, despite of the few comments I left

libetl (changes requested):
tiny changes please (at least the --riot parameter)

lib/cli/lib/initiate.js
```diff
@@ -0,0 +1,192 @@
...
+            logger.log();
+          });
+      }
+
+      case types.REACT_NATIVE:
+        return reactNativeGenerator(npmOptions)
+          .then(commandLog('Adding storybook support to your "React Native" app'))
+          .then(end);
+
+      case types.METEOR:
```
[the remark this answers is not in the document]
Keraito: I'm planning on performing an overhaul where possible of the CLI code, mainly regarding the generator parts as there is quite some duplicate code. But I want to leave that for another PR, as this PR is mainly to refactor the CLI to be able to handle more commands than just `getstorybook`. :)

lib/cli/bin/generate.js
```diff
@@ -1,201 +1,50 @@
...
-
-// Update notify code.
-updateNotifier({
-  pkg,
-  updateCheckInterval: 1000 * 60 * 60, // every hour (we could increase this later on.)
-}).notify();
+program
+  .command('start')
+  .description('Start the local Storybook server')
+  .option('-N --use-npm', 'Use NPM to start the Storybook server')
```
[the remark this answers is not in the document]
Keraito: When the CLI was made, I don't think NPM had the lockfile yet (i think), that's why the flag was added to the command. On the other hand, we can also argue that what should happen when both files are present (for whatever reason). Although I definitely agree with your solution if we want to prioritize NPM over Yarn, or the other way around which happens now, it would be a great idea

app/riot/README.md
```diff
@@ -15,7 +15,7 @@ So you can develop UI components in isolation without worrying about app specifi
 ```sh
 npm i -g @storybook/cli
 cd my-app
-getstorybook
+storybook init --riot
```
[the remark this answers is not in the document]
Keraito: I will address this in another PR

lib/cli/bin/generate.js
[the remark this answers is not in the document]
igor-dv: also `npm-shrinkwrap.json` 😅

lib/cli/package.json
Keraito: I tried this locally, but when I change the entrypoint for `getstorybook` to something else the command is not detected, not sure what's happening here 🤔 the symlink is present tho

lib/cli/lib/initiate.js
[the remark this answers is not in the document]
libetl: ok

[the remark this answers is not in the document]
libetl: but can you just remove --riot for now as the riot README.md would be wrong for a while ?

libetl:
thanks

libetl (approved):

Keraito:
Anything else blocking this? 😊

igor-dv (approved):

PR #4173, opened 2018-09-13
Presets - API generalization 

igor-dv:
Issue: 
The presets API, in the beginning, was to expose these methods: babel, webpack, preview & manager. Each one is about to extend a relevant part in the SB configuration.

After that, `babelDefault` was added to solve the babel 6 compatibility problems and remove the complexity of `babel-merge`. 

## What I did
From now, the core will expose presets loader that only knows to do the `apply`. Then core will wrap the `apply` with `babel`, `webpack`, `preview` & `manager`. Since presets object is passed deeper to the presets chain, any other preset will be able to extend it with the new functionality.

### DIFF
--- a/lib/core/src/server/config.js
+++ b/lib/core/src/server/config.js
@@ -3,6 +3,15 @@ import { logger } from '@storybook/node-logger';
 import loadPresets from './presets';
 import serverRequire from './serverRequire';
 
+function wrapCorePresets(presets) {
+  return {
+    babel: (config, args) => presets.apply('babel', config, args),
+    webpack: (config, args) => presets.apply('webpack', config, args),
+    preview: (config, args) => presets.apply('preview', config, args),
+    manager: (config, args) => presets.apply('manager', config, args),
+  };
+}
+
 function customPreset({ configDir }) {
   const presets = serverRequire(path.resolve(configDir, 'presets'));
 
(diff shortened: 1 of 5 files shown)


pksunkara:
@igor-dv I still don't get the difference between `babelDefault` and `babel`. Is it like `babelDefault` is for babel@6?

pksunkara (approved):

igor-dv:
One of the steps in babel config combination is `loadCustomBabelConfig.js`

```js
export default function(configDir, getDefaultConfig) {
  const babelConfig = loadFromPath(path.resolve(configDir, '.babelrc'));

  if (babelConfig) {
    // If the custom config uses babel's `extends` clause, then replace it with
    // an absolute path. `extends` will not work unless we do this.
    if (babelConfig.extends) {
      babelConfig.extends = path.resolve(configDir, babelConfig.extends);
    }
    return babelConfig;
  }

  return isBabelLoader8() ? getDefaultConfig() : {};
}
```

it tries to load a custom babelrc from the `.storybook` dir, if it's not existing, and user is using babe 7, we are providing our default babel configuration.

Default babel config is also located in `core` (`babe.dev.js` and `babel.prod.js`)

There are some apps, (react, mithril and vue) that have to extend the **defulat** config with their plugins/presets, and we want to extend only **our** default config, and not the one, user provided in a `.storybook` dir. 

In the beginning, I wanted to merge the defaults anyway with what user provides, and @Hypnosphi mentioned that it could break if user uses babel 6.

PR #4183, opened 2018-09-16
Upgrade bootstrap in docs

igor-dv:
Issue:
image

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -24,7 +24,7 @@
     "@storybook/addons": "^3.4.8",
     "@storybook/react": "^3.4.8",
     "babel-loader": "^6.4.1",
-    "bootstrap": "^3.3.7",
+    "bootstrap": "^4.1.3",
     "gatsby": "^1.9.273",
     "gatsby-link": "^1.6.45",
     "gatsby-plugin-sharp": "^1.6.48",
--- a/docs/src/components/Footer/index.js
+++ b/docs/src/components/Footer/index.js
@@ -7,7 +7,7 @@ import './style.css';
 const Footer = () => (
   <div id="footer" className="row">
     <div className="col-md-12">
-      <div className="row logos">
+      <div className="logos">
         <div className="col-xs-12">
           <center>
             Maintained by the <Link to="/basics/community/">Storybook Community</Link>.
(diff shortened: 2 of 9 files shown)


Hypnosphi:
Looks like some migration is needed: https://deploy-preview-4183--storybooks.netlify.com

igor-dv:
💩 .. I've checked it in a dev mode, and it looked ok... Maybe we need to add chromatic on this? (@tmeasday 😉)

igor-dv:
Fixed few things, but it still looks a bit different due to breaking changes. 
Also, I don't know yet how to fix search box

Hypnosphi:
So they fixed the vulnerability only in v4? There's no patch for v3?

igor-dv:
As far as I checked in the releases, they don't... We have the latest v3

tmeasday:
What are you thinking @igor-dv? ;) Do the docs have a storybook?

igor-dv:
Ah yeah. Docs have a setup but it doesn't work. Can we configure chromatic to snapshot only the homepage screen and one of the inner screens? Just to check that layout is not broken.

tmeasday:
We can probably do a similar iframe back that we do with the built example
apps?

Sent from Astro <https://www.astro.ai> for iOS


On 17 Sep 2018 at 5:29 pm, Igor <notifications@github.com> wrote:


Ah yeah. Docs have a setup but it doesn't work. Can we configure chromatic
to snapshot only the homepage screen and one of the inner screens? Just to
check that layout is not broken.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/storybooks/storybook/pull/4183#issuecomment-421914056>,
or mute the thread

igor-dv:
It's possible, but docs are not bootstrapped with the examples, AFAIK.

igor-dv:
I've patched the search box to use classes from v3. I am not so happy with this, so if anyone familiar with bootstrap better would like to take this, that'd be great.

ndelangen:
https://5b9faf1a3813f05aeb0eb2a2--storybooks.netlify.com/

ndelangen:
Looks fine to me

ndelangen (approved):

PR #4194, opened 2018-09-18
Remove snapshots part of CLI tests

Hypnosphi:
Issue: those snapshots never helped us catch a bug. They only fail when babel or react releases a new version, which is quite annoying


pksunkara:
Looks good. Could you please rebase and fix circleci tests too?

igor-dv (approved):

PR #4199, opened 2018-09-18
FIX regression devtool, in case of cross domain parent, window.parent might throw

ndelangen:
I caused a regression, this fixes it.

The regression occurs when the iframe and parent are on different domains or the manager frame itself is iframed in on a different domain.

calling window.parent might throw, so a try-catch is added.

@tmeasday

### DIFF
--- a/lib/core/src/server/templates/base-manager-head.html
+++ b/lib/core/src/server/templates/base-manager-head.html
@@ -9,8 +9,12 @@
 </style>
 
 <script>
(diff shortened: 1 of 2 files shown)


tmeasday:
Seems reasonable to me! Not sure we need the console.log, but no harm I guess

tmeasday (approved):

pksunkara (approved):

PR #4209, opened 2018-09-21
Remove unneeded top level dependencies

pksunkara:


package.json
```diff
@@ -11,16 +11,14 @@
     "addons/storyshots/*",
     "app/*",
     "lib/*",
-    "examples/marko-cli",
-    "examples/angular-cli",
+    "examples/*-cli",
     "examples/cra-kitchen-sink",
     "examples/html-kitchen-sink",
```
pksunkara: Could have done `examples/*-kitchen-sink` too if not for `crna-kitchen-sink` package.

lib/cli/test/fixtures/riot/package.json
```diff
@@ -12,6 +12,7 @@
     "riot": "^3.11.2"
   },
   "devDependencies": {
+    "cross-env": "^5.2.0",
```
Hypnosphi: Does it mean that user should install it before running `getstorybook`?

package.json
```diff
@@ -92,8 +80,6 @@
     "concurrently": "^4.0.1",
     "cross-env": "^5.2.0",
     "danger": "^3.8.4",
-    "enzyme": "^3.5.0",
-    "enzyme-adapter-react-16": "^1.3.0",
```
Hypnosphi: why are those unneeded?

lib/cli/test/fixtures/riot/package.json
pksunkara: Yes, see this package.json. It's something the user is supposed to consciously configure and use. Not something that should be a dependency of storybook package.

Hypnosphi: So storybook doesn't work without it? Why? We should fix or at least document it.

Hypnosphi (changes requested):
https://github.com/storybooks/storybook/pull/4209#discussion_r220089348

lib/cli/test/fixtures/riot/package.json
pksunkara: No, storybook works without it. I am saying this is something people use in their web apps which the fixture here is.

Hypnosphi: > No, storybook works without it

Let's not add it to the fixture without need then

pksunkara: `vue-cli` doesn't add it. I don't know why this fixture has it. You can see it here, https://github.com/storybooks/storybook/pull/4209/files#diff-a3134227a935db48acff355fe7fbedfdR8

Hypnosphi (approved):
Looks like that fixture actually uses cross-env

PR #4213, opened 2018-09-21
Cli gestorybook fix

Keraito:
Issue: part of #4212

When running `getstorybook`, the process would not wait until initiating Storybook was finished which is an async function, making the process terminate early before finishing.
I returned and awaited the generator, but cuz of that also need to wrap the code for `storybook init` in its own block to prevent behaviour from overflowing. Getting rid of the depreciation at a later stage is basically removing the `if` statement and case.

@shilman 	could you give this a run?


ndelangen (approved):

pksunkara (approved):

pksunkara:
cli-latest-cra is failing because of recent release in cra.

PR #4218, opened 2018-09-22
Fix #4217, exit build when successful

### LINKED ISSUE #4217: build-storybook not exiting after build completes
- Running build-storybook works but does not exit the process making it impossible to proceed to in a CI process or script.

### Steps to reproduce

- Run 
`npx build-storybook`

screen shot 2018-09-22 at 9 21 32 am

- Current alpha version
- master branch

### Affected platforms 
All

### Acceptance criteria

When the build-storybook script completes it should exit

nelsonomuto:
Issue:
#4217 
## What I did
Added the `process.exit()` line so that the process will exit after a build

## How to test

Does this need an update to the documentation? no

### DIFF
--- a/lib/core/src/server/build-static.js
+++ b/lib/core/src/server/build-static.js
@@ -76,6 +76,7 @@ export function buildStatic({ packageJson, ...loadOptions }) {
       process.exitCode = 1;
     }
     logger.info('Building storybook completed.');
+    process.exit();
   };
   const compiler = webpack(config);
   if (program.watch) {


nelsonomuto:
how do I label this, I don't have the ability to add the label

ndelangen:
Just a note that using process.exit([number]) is not recommended practice.

The reason this is problematic is because writes to process.stdout in Node.js are non-blocking and may occur over multiple ticks of the Node.js event loop. Calling process.exit(), however, forces the process to exit before those additional writes to stdout can be performed.

Rather than calling process.exit() directly, the code should set the process.exitCode and allow the process to exit naturally by avoiding scheduling any additional work for the event loop:

If you can you should set the exit code and allow Node to exit gracefully:

process.exitCode = 1;

This is what we're doing.

What's holding the process open is the question.

ndelangen:
Pasting this here as info for later debugging session:

> You can call process._getActiveRequests() to get a list of active I/O requests and process._getActiveHandles() to get a list of open handles/file descriptors.

contra:
@ndelangen Was this addressed anywhere? Possible to reopen this?

PR #4222, opened 2018-09-24
Fix broken colorpicker for knobs

EloB:
Right now this error triggers when trying to open colorpicker but with this fix it least works but we get a warning.

knobs-color-broken

**Current issue**
```
Uncaught TypeError: Cannot read property 'contains' of undefined
    at HTMLDocument.<anonymous> (velocity.js:13)
```

**After this fix issue**
```
Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in ColorPicker (created by ColorType)
    in div (created by Context.Consumer)
    in Styled(div) (created by ForwardRef)
```

### DIFF
--- a/addons/knobs/src/components/types/Color.js
+++ b/addons/knobs/src/components/types/Color.js
@@ -28,10 +28,13 @@ class ColorType extends React.Component {
     document.addEventListener('mousedown', this.handleWindowMouseDown);
   }
 
-  shouldComponentUpdate(nextProps) {
+  shouldComponentUpdate(nextProps, nextState) {
     const { knob } = this.props;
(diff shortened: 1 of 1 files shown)


Keraito (commented):
Looks good to me, at least it works now. Any idea if we could get rid of the new warning as well?

EloB:
@Keraito I found the issue inside the `react-color` library.

https://github.com/casesandberg/react-color/pull/542

ndelangen (approved):

PR #4225, opened 2018-09-24
Don't copy license anymore since lerna 3 does it

pksunkara:
Lerna 3 copies the license into individual packages before publishing. So, we don't need this anymore


ndelangen (approved):
Interesting that lerna started doing this by default

PR #4236, opened 2018-09-27
CLI add command

Keraito:
Issue: More CLI enhancements.

## What I did
Added an `add` command to the storybook CLI to add addons to the Storybook project.

## How does it work 
1. Syntax: `storybook add <addon>`. This can be either an official storybook addon like `knobs` or a random npm package. It can also include a tag like `knobs@alpha`, as long as it works with `yarn` or `npm`.
2. Check if it's an official Storybook addon by running `yarn info @storybook/addons-<addon> version` (or npm equivalent).
	- If not: check if it's an npm package by executing `yarn info <addon> version`.
		- If not: display error that it's an invalid addon/npm package.
		- If so: proceed to step 3 with `isOfficialAddon = false`.
	- If so: proceed to step 3 `isOfficialAddon = true`.
3. Check `isOfficialAddon`:
	- If it's false: install it as an devDependency by running `yarn add -D <addon>`*.
	- If it's true: Check the `package.json` for installed Storybook packages.
		- If any found: take the version of the first** package found and run `yarn add -D @storybook/addons-<addon>@<version>`.
		- If none found: run `yarn add -D @storybook/addons-<addon>`.
4. Register the installed addon by read the `.storybook/addons.js` and put a register statement after the last register occurrence.

*Maybe adjusting the package.json and then running `yarn`/`npm install` is a better long term solution?
**Works under the assumption that the user has the same versions for every installed official Storybook package. My plan is to add some checks on this and warn the user about it in the future.

Please tell me what you guys think, both feature and implementation wise!


Keraito:
image

I wouldn't really consider this duplicate code of ~18 lines, CodeFactor 😨

ndelangen:
It's fine

lib/cli/bin/generate.js
```diff
@@ -21,7 +22,7 @@ if (process.argv[1].includes('getstorybook')) {
     .option('-s --skip-install', 'Skip installing deps')
     .option('-N --use-npm', 'Use npm to install deps')
     .option('-p --parser <babel | babylon | flow>', 'jscodeshift parser')
-    .option('-h --html', 'Add storybook for HTML')
+    .option('-t --type <type>', 'Add Storybook for a specific project type')
```
Keraito: I just noticed this change should not be in this PR.. will fix

Keraito:
Trying to pick things up again from where I left off 😄 @shilman @ndelangen could I get your thoughts/reviews on this PR? Maybe we can also talk on Discord about what we wanted to work on regarding the CLI.

shilman (approved):
Looks great!

shilman:
@Keraito Can you resolve the conflicts?

Keraito:
will do @shilman !

Keraito:
Should be good now @shilman

wuweiweiwu (approved):
NICE!

Issue #4135, opened 2018-09-07
@storybook/addon-knobs 4.0.0-alpha.20 Throw Error


mhauens:
```
Module not found: Error: Can't resolve 'moment' in '/Users/mahauens/develop/wh-components/node_modules/react-datetime'
 @ ./node_modules/react-datetime/DateTime.js 6:10-27
 @ ./node_modules/@storybook/addon-knobs/dist/components/types/Date/index.js
 @ ./node_modules/@storybook/addon-knobs/dist/components/types/index.js
 @ ./node_modules/@storybook/addon-knobs/dist/components/Panel.js
 @ ./node_modules/@storybook/addon-knobs/dist/register.js
 @ ./node_modules/@storybook/addon-knobs/register.js
 @ ./.storybook/addons.js
 @ multi ./.storybook/addons.js ./node_modules/@storybook/react/node_modules/@storybook/core/dist/server/config/polyfills.js ./node_modules/@storybook/react/node_modules/@storybook/core/dist/client/manager/index.js

ERROR in ./node_modules/react-datetime/src/DaysView.js
Module not found: Error: Can't resolve 'moment' in '/Users/mahauens/develop/wh-components/node_modules/react-datetime/src'
 @ ./node_modules/react-datetime/src/DaysView.js 5:10-27
 @ ./node_modules/react-datetime/src/CalendarContainer.js
 @ ./node_modules/react-datetime/DateTime.js
 @ ./node_modules/@storybook/addon-knobs/dist/components/types/Date/index.js
 @ ./node_modules/@storybook/addon-knobs/dist/components/types/index.js
 @ ./node_modules/@storybook/addon-knobs/dist/components/Panel.js
 @ ./node_modules/@storybook/addon-knobs/dist/register.js
 @ ./node_modules/@storybook/addon-knobs/register.js
 @ ./.storybook/addons.js
 @ multi ./.storybook/addons.js ./node_modules/@storybook/react/node_modules/@storybook/core/dist/server/config/polyfills.js ./node_modules/@storybook/react/node_modules/@storybook/core/dist/client/manager/index.js
```

- @storybook/react 4.0.0-alpha.20
- @storybook/addon-knobs eact 4.0.0-alpha.20
```
"devDependencies": {
	"@storybook/addon-a11y": "^4.0.0-alpha.20",
	"@storybook/addon-actions": "^4.0.0-alpha.20",
	"@storybook/addon-backgrounds": "^4.0.0-alpha.20",
[...]

PR #4244, opened 2018-09-27
REPLACE react-datetime with native date/time pickers

ndelangen:
Issue: #4135

## What I did

I replaced the react-datetime components by a custom component using the native date & time picker provided by the browser.

Effects are:
- UX is dependent on the browser
- We no longer have a dependency on moment, so the bundle is smaller

This is based on 3.4 branch

I will do a similar thing for master


ndelangen:
```
FAIL  lib/core/src/client/preview/client_api.test.js
  ● preview.client_api › storiesOf › add › should replace stories when adding the same story

    Expected value to be:
      [Function mockConstructor]
    Received:
      [Function mockConstructor]
    
    Difference:
    
    Compared values have no visual difference.

      312 |           // v3 returns the same function we passed in
      313 |           if (jest.isMockFunction(entry.stories[0].render)) {
    > 314 |             expect(entry.stories[0].render).toBe(stories[0]);
      315 |           } else {
      316 |             expect(entry.stories[0].render()).toBe('story2');
      317 |           }
      
      at Object.<anonymous> (lib/core/src/client/preview/client_api.test.js:314:46)
```

I think this might be an existing problem with 3.4 ?

I didn't change anything related to this 😕

ndelangen:
it is:
https://circleci.com/gh/storybooks/storybook/65125

ndelangen:
Danger also
Netlify's config is missing marko app
And this branch still does cli tests that were removed in master

kylemh:
Wow this will be so much nicer when it's in. Also, one less dependency!

Seeing a lot of issues with trying to make edits to the knob itself via the text input rather than the date picker.

Using up and down arrows to select any input sometimes ends up with the incorrect date being rendered:
arrows-months

Attempting to type out a year results in immediate re-render, the wrong knob value, and the wrong display value:
typing-years

ndelangen:
going to have to make them uncontrolled inputs then. Since right now, when a non-valid date is entered, it's immediately forced into a valid date again, making it impossible to edit freely.

ndelangen:
Hey @kylemh I think I got it working this time.

Might extract this into it's own npm package maybe

ndelangen:
https://deploy-preview-4244--storybooks-official.netlify.com/

kylemh:
Awesome - all of the gif issues resolved. The last issue that I can see is the ability to type and key to non-existent dates, like February 30th of any year.

ndelangen:
That's browser behavior... I dunno maybe they allow it to give you the option of jumping to say 31th of January to 29th of February without having to switch the day first...

The dropdown will only allow you to select real dates in Chrome, but when editing the text, they allow any number < 32 for days in combo with any number < 13 for months. years allow up to 999999, but no negative and chrome will only show the last 4 digits.

kylemh:
Seems fine to me. At the end of the day, this for UI demonstration purposes.

Maybe just ensure we outline the fact that there's no validation beyond what the browser provides in the knobs README for that knob type. Otherwise, looks great to me 👍

ndelangen:
Can I get a review @storybooks/team

kylemh (approved):

addons/knobs/src/components/types/Date/index.js
```diff
@@ -1,36 +1,119 @@
...
+        <input
+          style={styles}
+          type="time"
+          id={`${name}time`}
+          ref={el => {
+            this.timeInput = el;
+          }}
+          onChange={this.onTimeChange}
+        />
+        {!valid ? <div>invalid</div> : null}
```
wuweiweiwu: can we just use `!valid &&....`?

wuweiweiwu (approved):
LGTM! Looks awesome!

addons/knobs/package.json
```diff
@@ -24,7 +24,6 @@
     "moment": "^2.21.0",
```
libetl: will we be able to get rid of moment ?

libetl: even in the parent workspace ?

pksunkara:
No other comments other than what others mentioned.

addons/knobs/src/components/types/Date/index.js
Hypnosphi: or even `valid || ...`

ndelangen: Can you guys find a fixable eslint rule for this?

I personally prefer ternaries over the `&&` or `||` pattern. But I don't care enough about it to fight eslint.

addons/knobs/package.json
libetl: thanks

ndelangen:
Going to close this is favour of the one on master, which should imminently be released as 4.0.0

PR #4245, opened 2018-09-27
FIX react-scripts-latest-fixture by installing expo globally

ndelangen:
Issue: master build is broken

## What I did

I edited the `cli-latest-cra`-test so they use npx directly and the CRNA one uses expo-cli directly

### DIFF
--- a/lib/cli/test/test_latest_cra.sh
+++ b/lib/cli/test/test_latest_cra.sh
@@ -3,27 +3,14 @@
 # exit on error
 set -e
 
-# create CR(N)A fixtures using latest versions of creators
-function create_fixture {
-  if [ $2 == "react_native_scripts" ]
-  then
-    echo "create-react-native-app requires node >=6. Checking..."
-    ../../node_modules/.bin/check-node-version --node '>=6' || return 0
-  fi
-
-  # use `npx` rather then `yarn create` to avoid installing global deps
-  ../../node_modules/.bin/npx $1 $2
-}
-
 # remove and recreate `cra-fixtures` directory
 rm -rfd cra-fixtures
 mkdir cra-fixtures
 cd cra-fixtures
 
-create_fixture create-react-app react-scripts-latest-fixture
+../../node_modules/.bin/npx create-react-app react-scripts-latest-fixture
 
-npm install -g expo-cli
-create_fixture create-react-native-app react-native-scripts-latest-fixture
+../../node_modules/.bin/npx expo init react-native-scripts-latest-fixture
 
 cd ..
 ./run_tests.sh -f cra-fixtures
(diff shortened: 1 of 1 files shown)


ndelangen:
Still asks to install expo-cli even when already installed globally... 😠

Hypnosphi:
Do they have a `-y` option?

ndelangen:
I do not know, I have asked on their slack for support..

Hypnosphi:
https://github.com/react-community/create-react-native-app#create-react-native-app

> You can now use `expo init` to create your project. See Quick Start in the Expo documentation for instructions on getting started using Expo CLI.
>
> The `create-react-native-app` command is provided for backwards compatibility.

Looks like we should just use `expo init` instead

Hypnosphi:
> `function create_fixture`

By the way this one isn't actually needed, we never run our tests in old node

ndelangen:
I just wanted to do a quick fix while I was waiting for something else.

Since I was hoping to be the quick hero who fixed master.

Anyhow, it sounds like you have a waay better picture of what's going on and should happen then I do about this @Hypnosphi 😄 

Do you want to take over this PR?

Hypnosphi:
No, please just do the changes I mentioned

ndelangen:
Opened a forum post here:
https://forums.expo.io/t/cl-initi-in-ci-times-out-because-waiting-for-input-this-command-requires-expo-cli/14255

I do not know how to fix this 😭

Hypnosphi (approved):

Hypnosphi:
Looks like `expo` just [proxies](https://github.com/expo/expo/blob/41458d1de91544c41097818db21e44e9aa84688c/packages/expo/bin/cli.js#L8) to `expo-cli`

This works without prompts:
```
npx expo-cli init -t blank my-project
```

Hypnosphi:
But the problem is, the project that you get doesn't have `react-native-scripts` dependency

Hypnosphi:
We should probably detect `expo` dependency instead

Gongreg:
Even if we manage to pass 
```
Do you want to install it globally [Y/n]? Too long with no output (exceeded 10m0s)
```
We get another issue:

```
[16:32:15] Input is required, but expo is in non-interactive mode.
Required input:
> Choose a template:
```

Hypnosphi:
@Gongreg that's why I added `-t blank`

> But the problem is, the project that you get doesn't have `react-native-scripts` dependency

Hypnosphi:
@ndelangen let's do the following:

Change this to detect `expo` instead of `react-native-scripts` (and probably rename the constant to `types.EXPO`):
https://github.com/storybooks/storybook/blob/master/lib/cli/lib/detect.js#L46-L51

Use `../../node_modules/.bin/npx expo-cli init -t blank react-native-scripts-latest-fixture` in `test_latest_cra.sh`

Issue #4180, opened 2018-09-14
SVGs are broken by quotes


cainlevy:
### Bug summary

SVGs used by my components are broken when rendered in Storybook. The data URI appears to be wrapped in double quotes.

### Steps to reproduce

1. Create a component like:

```tsx
import MySVG from 'static/my.svg';

const MyComponent = () => <img src={MySVG} />;
```

2. Note that the component renders fine in a normal Webpack setup.
3. Note that the component renders a broken image with Storybook's Webpack config.

The problem is that Storybook uses svg-url-loader but does not configure it with `noquotes=true`. 

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/react 4.0.0-alpha.3

### Affected platforms 

- svg-url-loader 2.3.2

cainlevy:
If there is a use case for SVGs wrapped in double quotes (CSS?) then I think that use case should be forced to add the quotes. Otherwise `svg-url-loader` is not drop-in compatible with `url-loader`.

Brian-McBride:
The problem with svg-loader is that if you enable the noquotes you break others things (Like Polymer). 

The pull request #4260 removes svg-url-loader from the webpack. Here is the value of that module: 
>svg-url-loader loads SVGs as UTF-8 encoded data urls. The result is smaller and faster to parse than base64.

For Storybook, I'm not sure dealing with turning on/off noquotes is worth the inline size. In your production webpack build you can tweak for super optimized builds based on your project anyway. Greater support for Storybook makes sense over nickling over minor performance gains?

Issue #4066, opened 2018-08-24
svg-url-loader works not properly in v4.0.0-alpha.16


taksenov:
### Bug or support request summary

Hello, im use StoryBook in my TypeScript project. Yesterday I noticed that the SVG-images are loading incorrectly, bug is present on image:  http://joxi.ru/D2P4QR8IqdZ062

### Please specify which version of Storybook and optionally any affected addons that you're running

My dependencies presents below:

```json
  "dependencies": {
    "axios": "^0.18.0",
    "classnames": "^2.2.6",
    "connected-react-router": "^4.4.1",
    "history": "^4.7.2",
    "lodash": "^4.17.10",
    "moment": "^2.22.2",
    "qs": "^6.5.2",
    "react": "^16.4.2",
    "react-custom-scrollbars": "4.2.1",
    "react-dom": "^16.4.2",
    "react-dropzone": "4.2.5",
    "react-loadable": "5.3.1",
    "react-markdown": "3.1.4",
    "react-modal": "^3.5.1",
    "react-native-listener": "1.0.2",
    "react-redux": "^5.0.7",
    "react-redux-toastr": "git+https://github.com/taksenov/react-redux-toastr.git",
    "react-router": "^4.3.1",
    "react-router-dom": "^4.3.1",
    "react-select": "^1.3.0",
    "react-textarea-autosize": "5.2.1",
    "react-transition-group": "^2.4.0",
    "recharts": "^1.1.0",
    "redux": "^4.0.0",
    "redux-actions": "^2.6.1",
    "redux-form": "^7.4.2",
    "redux-saga": "^0.16.0",
    "redux-thunk": "^2.3.0",
    "reselect": "^3.0.1",
    "sanitize.css": "^7.0.0"
  },
  "devDependencies": {
    "@storybook/addon-actions": "^4.0.0-alpha.16",
    "@storybook/addon-knobs": "^4.0.0-alpha.16",
    "@storybook/addon-links": "^4.0.0-alpha.16",
    "@storybook/addon-notes": "^4.0.0-alpha.16",
    "@storybook/addons": "^4.0.0-alpha.16",
    "@storybook/react": "^4.0.0-alpha.16",
    "@tygen/html": "^0.1.0",
    "@tygen/reflector": "^0.1.0",
    "@types/classnames": "^2.2.6",
    "@types/enzyme": "^3.1.13",
    "@types/enzyme-adapter-react-16": "^1.0.3",
    "@types/jest": "^23.3.1",
[...]

PR #4260, opened 2018-10-02
Fix broken SVGs -  issue #4066 / #4180

Brian-McBride:
The webpack currently uses both file-loader and svg-url-loader.

file-loader will handle .svg files just fine. svg-url-loader is causing images to not load properly. This might be identified by @cainlevy in issue 4180:
> svg-url-loader is not drop-in compatible with url-loader.

Unless there is a compelling reason to use svg-url-loader instead of just leveraging the currently utilized url-loader, this change seems to fix the issues and can result in removing the dependency on svg-url-loader from the repo completely.

Issue:

## How to test

### DIFF
--- a/lib/core/src/server/config/webpack.config.default.js
+++ b/lib/core/src/server/config/webpack.config.default.js
@@ -33,15 +33,11 @@ export function createDefaultWebpackConfig(storybookBaseConfig) {
           ],
         },
         {
-          test: /\.(ico|jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2)(\?.*)?$/,
+          test: /\.(svg|ico|jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2)(\?.*)?$/,
           loader: require.resolve('file-loader'),
           query: {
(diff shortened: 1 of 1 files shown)


cainlevy:
if `svg-url-loader` is valuable, then i would recommend configuring it with `noquotes: true`.

https://github.com/storybooks/storybook/issues/4180

Brian-McBride:
So, the suggestion is to change line 44 in this file:
https://github.com/storybooks/storybook/blob/b2b73596f9dd97b4ba8c03340bd36f868e836772/lib/core/src/server/config/webpack.config.default.js#L44  to:

```
    loader: require.resolve('svg-url-loader?noquotes'),
```

@cainlevy Have you tested this?

cainlevy:
Effectively, yes. I patched the webpack config in my app with:

```
module.exports = (baseConfig, env, config) => {
  const svgLoader = config.module.rules
    .find((r) => r.loader && r.loader.indexOf('svg-url-loader') !== -1);
  svgLoader.query = {noquotes: true};
}
```

ndelangen (approved):

Issue #4179, opened 2018-09-13
Broken with react-native@0.57 due to packager (local-cli) API change 


mgcrea:
If you are reporting a bug or requesting support, start here:
### Bug or support request summary

Bug

### Steps to reproduce

On a fresh project (`react-native init`):

```
$ yarn storybook
yarn run v1.9.4
$ storybook start -p 7007
=> Loading custom .babelrc from project directory.
=> Loading custom addons config.
=> Using default webpack setup based on "Create React App".
Scanning 1172 folders for symlinks in /Users/olivier/Projects/joonca/joonca-components/node_modules (12ms)

React Native Storybook started on => http://localhost:7007/


  error: unknown option `--projectRoots'

webpack built 91ec773834c48e7f9dab in 2184ms
Hash: 91ec773834c48e7f9dab
Version: webpack 3.12.0
Time: 2184ms
```

And another error after that:

```
ERROR in ./storybook/addons.js
Module build failed: Error: Couldn't find preset "module:metro-react-native-babel-preset" relative to directory "/Users/olivier/Projects/joonca/joonca-components/storybook"
```

Looks like this commit is the culprit:

https://github.com/facebook/react-native/commit/c5ce7626977a7650e8dce335f2ce78b94839a2a8#diff-4a892a6b8856f0253e2de7094e585ae6

### Please specify which version of Storybook and optionally any affected addons that you're running

```
    "@storybook/addon-actions": "^3.4.10",
    "@storybook/addon-links": "^3.4.10",
    "@storybook/addons": "^3.4.10",
    "@storybook/react-native": "^3.4.10",
```

### Affected platforms 

react-native

resting:
Getting the same error.
Works with RN v0.55.4 though.

jacquesdev:
This worked for me https://github.com/facebook/react-native/issues/20876

cball:
@jacquesdev mind providing more detail on the steps you took? I was able to get a project to build correctly using storybook v4.0.0-alpha.21, `.babelrc.js` and explicitly requiring `metro-react-native-babel-preset` but Storybook didn't seem to function correctly after that.

kdawgwilk:
@cball I am in the same situation as you. I can get things to build correctly with `v4.0.0-alpha.21` but the packager doesn't appear to run and so the simulator just gives me a `No bundle URL present` red screen

griable:
[...]

PR #4261, opened 2018-10-02
Removing the packager from storybook.

Gongreg:
Issue: #4179

By having packager logic inside storybook we are prone for breaking things.
We have to support metro, haul, typescript and other cases. All of these things are responsibility of the user.

To avoid it I've completely removed packager logic from storybook.

What is more starting from v5.0 the default behaviour will be to skip the server altogether and include StorybookUI in your app so this is a first step to that direction.

I've added some info how to bring back old behaviour @petrbela suggested it.


igor-dv:
I think it's rather an important breaking change, so having a note in the MIGRATION.md will be a good thing

Gongreg:
@igor-dv Definitely.

app/react-native/src/bin/storybook-start.js
```diff
@@ -48,61 +39,3 @@ server.listen(...listenAddr, err => {
...
-  let roots = [projectDir];
-
-  if (program.root) {
-    roots = roots.concat(program.root.split(',').map(root => path.resolve(root)));
-  }
-
-  try {
-    // eslint-disable-next-line global-require
-    require('babel-register')({
-      presets: [require.resolve('babel-preset-flow')],
```
igor-dv: do `babel-register` and `babel-preset-flow` still need to be installed?

Gongreg: Good question. What were they used for in the first place?

igor-dv: I have no idea =)

igor-dv (approved):

app/react-native/src/bin/storybook-start.js
Gongreg: Ah but it is flow and some other things. I will remove it. 
Thanks @igor-dv .

igor-dv:
@Gongreg, is anything missing here ?

Gongreg:
Nothing that I am aware of. We will do a blog post regarding this update. But it is not part of the PR

jqn:
@igor-dv one thing that is missing from this is some instructions on how to switch from the storybook UI to the react native UI. The old behavior kept this separate but I don't see how this makes sense 
`module.exports = __DEV__ ? StorybookUI : App;` if I want to switch back and forth. there has to be a better way to switch. Also if I'm missing something here please let me know.

Gongreg:
@jqn You could have a storybook tab in bottom nav menu. Or inside admin panel. Or even the same switch you just wrote about. It shouldnt be more work to switch this way than killing packager and starting the new one. Or if you want to you can still start the packager old way, you just have to do it yourself. We will release a blogpost/update storybook website with information when 4.0 is released. Or you can check docs folder in @storybook/react-native code. (I am on mobile so it is hard to link it)

kdawgwilk:
Or you can start the RN packager using the storybook dir: `react-native start --projectRoot storybook`

kdawgwilk:
Actually I just tried to give this a go and with RN >= 0.56 this throws an error `error: unknown option `--projectRoot'`

Gongreg:
It may be `--projectRoots`. From 0.57 it is `--projectRoot`. That is exactly the reason why we removed the packager. We don't want to maintain it.

kdawgwilk:
It is `--projectRoots` but also giving a relative path doesn't work as expected and instead I need to give it a absolute path e.g. 
```
--projectRoots `pwd`/storybook
```

esbenvb:
@kdawgwilk @Gongreg keep in mind that `--projectRoot` is not implemented in 0.57, but only documented. 0.57.1 fixes that
https://github.com/facebook/react-native/commit/fb109e9120e4366e40d7468625fcd6b14a5fe1e9

PR #4262, opened 2018-10-02
babel/minify: use `builtIns: false`

Hypnosphi:
See https://github.com/babel/minify/issues/904#issuecomment-424367995


igor-dv:
Does it solve somve open issue?

Hypnosphi:
I've got error from https://github.com/babel/minify/issues/904 on my project

igor-dv (approved):

Issue #4272, opened 2018-10-03
Pass serializers to addon-storyshots


tobilen:
I'm not sure whether to call this a bug or feature request.

# My Problem
I am snapshotting a component that includes a javascript date in its props. When a date object is serialized, it applies the timezone to the printed string, which means a serialized date will be different depending on the machines timezone.
The solution to this problem is to apply a custom serializer that strips the timezone from dates ( https://www.npmjs.com/package/jest-date-serializer ). However, my specified serializers aren't applied by `addon-storyshots`.

# Analysis
Currently, `snapshotSerializers` specified in the jest configuration get ignored by addon-storyshots if the `multiSnapshot` option is used. This is because the subdependency (`jest-specific-snapshot`) uses its own serializers.

We currently already have an option to pass a single serializer, which is just a function that gets applied to the rendered tree before it is passed to jest for snapshot comparison.

# Solution
it would be cool if instead of a single serializer we could pass an array of serializers that gets applied to the jest configuration, just like the default `snapshotSerializers` option.

igor-dv:
The v4 change is merged.

igor-dv:
I am going to close this since it's merged into the v4.

PR #4273, opened 2018-10-03
Add snapshot serializer option to Storybook v3.4

tobilen:
Issue: #4272

## What I did
* deprecate the `serializer` option.
* introduce a `snapshotSerializers` option that expects an array of serializers. They get applied to jest before snapshots are created.

## How to test

Run snapshot tests of the official example.


addons/storyshots/README.md
```diff
@@ -374,35 +374,26 @@ initStoryshots({
...
-    "snapshotSerializers": [
-      "enzyme-to-json/serializer"
-    ]
-  }
-```
+For that, you can pass an enzyme-compatible snapshotSerializer (like [enzyme-to-json](https://github.com/adriantoine/enzyme-to-json), [jest-serializer-enzyme](https://github.com/rogeliog/jest-serializer-enzyme) etc.) with the `snapshotSerializer` option (see below). 
+
 
-### `serializer`
+### `snapshotSerializers`
```
igor-dv: You should probably leave `serializer` in docs, while it's still supported, but mention that is deprecated.

addons/storyshots/src/test-bodies.js
```diff
@@ -7,6 +7,13 @@ export const snapshotWithOptions = options => ({
   renderTree,
   snapshotFileName,
 }) => {
+  if (options.snapshotSerializers) {
+    options.snapshotSerializers.forEach(serializer => {
```
igor-dv: The serializers will be added during the test. Is it ok? I think it's not the best place to set them

addons/storyshots/src/react/renderTree.js
```diff
@@ -2,6 +2,13 @@
 import reactTestRenderer from 'react-test-renderer';
 
 function getRenderedTree(story, context, { renderer, serializer, ...rendererOptions }) {
+  if (serializer) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'The "serializer" option of @storybook/addon-storyshots  has been deprecated. Please use "snapshotSerializers: [<your serializer>]" in the future.'
+    );
+  }
+
```
igor-dv: so, we will need to change this line in the end: `return serializer ? serializer(tree) : tree;`, right ?

I would do some thing like this:

```js
if (serializer) {
  console.warn('blah blah');
  return serializer(tree);
}

return tree;
```

that way it will be clearer what to remove after the deprecation.

igor-dv (changes requested):
Looks cool.

please fix linter/tests issues.

tobilen:
hey @igor-dv,

thanks for the feedback! i've integrated everything you mentioned. especially moving the ´addSerializer` calls to the initialization function feels like a good change.

as for the tests, the one failing unit test also fails on the base branch. not sure what to do about that 😕

addons/storyshots/src/test-bodies.js
tobilen: yeah, this is not ideal. i've moved serializer addition it to the initialization function

addons/storyshots/README.md
tobilen: done

tobilen:
yeah i did. 4 is still in alpha, and i would have liked to see it as a 3.x release since the migration to 4 seems to require some changes to our codebase. could we merge this on 3 and do a separate PR for v4?

igor-dv:
@shilman, can we do this ?

examples/cra-kitchen-sink/src/storyshots.test.js
```diff
@@ -3,7 +3,7 @@ import path from 'path';
 
 import { render as renderer, configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
-import serializer from 'enzyme-to-json';
+import { createSerializer } from "enzyme-to-json";
```
igor-dv: this will still break the linter ( " vs ' )

igor-dv:
ping @shilman here again

igor-dv:
@tobilen, I think having this in the v3.4 is already less relevant

PR #4275, opened 2018-10-03
removed yarn workspaces and added packages to lerna.json

dakebl:
Issue: https://github.com/storybooks/storybook/issues/4204

## What I did
I removed the yarn workspaces config from package.json and added the directory paths to lerna.json

## How to test
I have manually tested the `publish:alpha` npm script, which finished without error. 

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -22,5 +22,5 @@
     "lib/cli/test/run/*"
   ],
   "registry": "https://registry.npmjs.org",
-  "version": "4.0.1-alpha.0"
+  "version": "4.0.0-alpha.23"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -18,8 +18,7 @@
     "examples/official-storybook",
     "examples/riot-kitchen-sink",
     "examples/svelte-kitchen-sink",
-    "examples/vue-kitchen-sink",
-    "lib/cli/test/run/*"
+    "examples/vue-kitchen-sink"
   ],
   "registry": "https://registry.npmjs.org",
   "version": "4.0.0-alpha.23"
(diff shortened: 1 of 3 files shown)


gabrielcsapo:
@dakebl is there a performance change in this respect? When you would run `yarn` would it pull in the lerna dependencies?

dakebl:
Hey @gabrielcsapo, sorry for the late response. It looks like a received a false positive when I raised this PR due to my setup. I've since worked on it a little further and it looks like yarn workspaces and lerna can play nicely together once lerna.json has the packages array added to it. 

To test this, I:
- removed all of the node_modules with `lerna clean`
- ran `yarn` to install and link the packages
- ran `yarn publish:alpha -- --npm-skip`. Which seems to be correctly generating the relevant release in my forked repo. 

Without the npm publishing portion of the script however, I can't verify the performance of the script or detect any further errors that happen during the process. Would you be able to test this?

lerna.json
```diff
@@ -6,6 +6,21 @@
...
+    "lib/*",
+    "examples/*-cli",
+    "examples/cra-kitchen-sink",
+    "examples/html-kitchen-sink",
+    "examples/mithril-kitchen-sink",
+    "examples/official-storybook",
+    "examples/riot-kitchen-sink",
+    "examples/svelte-kitchen-sink",
+    "examples/vue-kitchen-sink",
+    "lib/cli/test/run/*"
```
pksunkara: This needs to be removed. That was the issue in #4204 in the first place.

pksunkara (changes requested):

PR #4280, opened 2018-10-04
[docs] add maintainers guide to help define process

gabrielcsapo:
Issue: Just joined the project as a maintainer and closed an issue by accident

## What I did

Started a conversation for what the maintainers guide should look like and have

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,54 @@
+This document will document some of the processes that members of the documentation team should adhere to.
+
+# PR Process
+
+1. Triage with the correct [label](#labels)
+2. If there a change related to it ensure it has been published and tested before closing
+
+# Labels
+
+| label name  | purpose |
+|:--------------:|:------------|
+| accessibility | |
+| addon:(name) | |
+| app:(name) | |
+| api:(name) | |
+| cleanup | Minor cleanup style change that won't show up in release changelog |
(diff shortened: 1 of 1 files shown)


pksunkara (changes requested):
There's a CONTRIBUTING.md which you should use for this.

igor-dv:
I think it will be good to separate the contribution and maintenance. 
For example, things like adding `merged` label to the issues instead of closing them, is for maintaners, but not for the one time PRs.

gabrielcsapo:
@pksunkara right as @igor-dv said this would be more for onboarding maintainers and socializing process. I think we should add more to it, just very new to know about any existing process that exists

Keraito:
Definitely, like the idea of this and agree with the above comments of keeping it separated from the `CONTRIBUTING.MD`. When I started out stuff like what @igor-dv 	mentioned was definitely unknown to me. Having a kind of established guideline that can be referred to would be great imo

gabrielcsapo:
@pksunkara thoughts on the comments about adding this?

pksunkara (approved):

gabrielcsapo:
Merging, we can keep adding to this document as new processes are defined.

igor-dv:
I think it's not really useful like this. I actually thought you will fill it a bit more before merging 🤣

PR #4282, opened 2018-10-04
Support providing options as a function, for story-specific options.

jondot:
Hi,
Adding support for providing options as a function, for story-specific options.
This doesn't break any existing infra/code, and is tested.

Thanks!

### DIFF
--- a/addons/storyshots/storyshots-core/README.md
+++ b/addons/storyshots/storyshots-core/README.md
@@ -185,6 +185,21 @@ initStoryshots({
 })
 ```
 
+Provide a function to have story-specific options:
+
+
+```js
+initStoryshots({
+  test: snapshotWithOptions(story =>({
+    createNodeMock: (element) => {
+      if(story.name == 'foobar') {
+        return null
+      }
+      return element
+    },
+  })),
+})
+```
 
 ### StoryShots for async rendered components
 
--- a/addons/storyshots/storyshots-core/src/test-bodies.js
+++ b/addons/storyshots/storyshots-core/src/test-bodies.js
@@ -1,12 +1,15 @@
 import 'jest-specific-snapshot';
 
+const isFunction = obj => !!(obj && obj.constructor && obj.call && obj.apply);
+const optionsOrCallOptions = (opts, story) => (isFunction(opts) ? opts(story) : opts);
+
 export const snapshotWithOptions = (options = {}) => ({
   story,
   context,
   renderTree,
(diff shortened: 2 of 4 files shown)


addons/storyshots/storyshots-core/README.md
```diff
@@ -185,6 +185,21 @@ initStoryshots({
 })
 ```
 
+Provide a function to have story-specific options:
```
igor-dv: It looks like this is related to a previous section (Using `createNodeMock` to mock refs), also please mention that it works with `multiSnapshotWithOptions` (it should, right ?)

addons/storyshots/storyshots-core/src/test-bodies.js
```diff
@@ -1,12 +1,15 @@
 import 'jest-specific-snapshot';
 
+const isFunction = obj => !!(obj && obj.constructor && obj.call && obj.apply);
+const optionsOrCallOptions = (opts, story) => (isFunction(opts) ? opts(story) : opts);
```
igor-dv: Why is it better than just `typeof opts === 'function'` ?

addons/storyshots/storyshots-core/stories/storyshot.snapshotWithOptionsFunction.test.js
```diff
@@ -0,0 +1,8 @@
+import path from 'path';
+import initStoryshots, { snapshotWithOptions } from '../src';
+
+initStoryshots({
+  framework: 'react',
+  configPath: path.join(__dirname, '..', '.storybook'),
+  test: snapshotWithOptions(() => ({})),
```
igor-dv: What does this example test?

addons/storyshots/storyshots-core/src/test-bodies.js
jondot: Basically just the function underscore uses. Preferred taking something that worked at-scale than going with the naive implementation (from what I read at some cases objects can return 'function').

addons/storyshots/storyshots-core/stories/storyshot.snapshotWithOptionsFunction.test.js
jondot: It tests that when you give it a function it still works (by providing options via the function). Without implementing the function/object detection this would fail

igor-dv (approved):
LGTM

PR #4284, opened 2018-10-04
chore(deps): #4267 upgrade lodash to latest

StephM-W:
Issue: Fix for #4267 

## What I did
Upgraded lodash to ^4.17.11, updated imports and mocks (from lodash.xxx to lodash/xxx)

## How to test
```yarn test --core```


danielduan (approved):

wuweiweiwu (approved):
LGTM!

igor-dv:
As far as I remember, when we tried the es6 version last time, it broke the users' code in Storyshots, because they needed to configure jest to transpile `lodash` from the `node_modules`.

Stephanemw:
Would the es6 lodash module be more appropriate in that case? https://www.npmjs.com/package/lodash-es

Stephanemw:
> As far as I remember, when we tried the es6 version last time, it broke the users' code in Storyshots, because they needed to configure jest to transpile `lodash` from the `node_modules`.

Do we have a testcase in the examples that exhibits this breakage? Otherwise, I suppose we could always request the various 2-year old lodash.xxxx packages be upgraded to address that CVE...

igor-dv (approved):

PR #4285, opened 2018-10-04
fix #2384: support string-only component for vue

### LINKED ISSUE #2384: vue storybook install fails
The example yields
```
Failed to execute 'createElement' on 'Document': The tag name provided ('<my-button :rounded="true">story as a function template</my-button>') is not a valid name.
    
      Error: Failed to execute 'createElement' on 'Document': The tag name provided ('<my-button :rounded="true">story as a function template</my-button>') is not a valid name.
```

yaodingyd:
Issue: #2384

## What I did

Add support for string only component

### DIFF
--- a/app/vue/src/client/preview/render.js
+++ b/app/vue/src/client/preview/render.js
@@ -20,6 +20,10 @@ export default function render({
   Vue.config.errorHandler = showException;
 
   const component = story();
+  
+  if (typeof component === 'string') {
+    component = { template: component };
+  }
 
   if (!component) {
     showError({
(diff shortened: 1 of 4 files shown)


app/vue/src/client/preview/render.js
```diff
@@ -20,6 +20,10 @@ export default function render({
   Vue.config.errorHandler = showException;
 
   const component = story();
+  
+  if (typeof component === 'string') {
+    component = { template: component };
```
igor-dv: component is a `const`, you can't reassign it, can you please fix it ? 
I think we need an example story as well.

igor-dv (changes requested):

app/vue/src/client/preview/render.js
yaodingyd: updated

igor-dv (approved):
Thanks

PR #4305, opened 2018-10-06
Decorators for using Storybook with Angular

klimentru1986:
Issue: Decorators for using Storybook with Angular. 

## Why i did
Angular use decorators. And this approach easier for Angular developers. 

Now you can write Storybook like Angular 

## What I did
1. Create decorator for creating story
```sh
@NgStory({
  kind: 'Story with decorator',
  name: 'with some emoji',
  parameters: {
    moduleMetadata: {
      declarations: [Button],
    },
  },
})
export class WithSomeEmojiStory {}
```
2. Create decorator for module with stories
```sh
@NgStoryModule({
  name: 'Decorator Module',
  stories: [StorybookStory, WithTextStory, WithSomeEmojiStory],
})
export class StoryModule {}
```
3. Create story module bootstrap function
```sh
ngBootstrapStoryModule(StoryModule);
```
4. Create example in `examples/angular-cli/src/stories-decorator-api`
5. Create config for using with decorator
```sh
/** configuration with using decorators @NgStoryModule and @NgStory */
function loadStoriesDecoratorConfig() {
  // put welcome screen at the top of the list so it's the first one displayed
  require('../src/stories-decorator-api/main');

  // automatically import all story ts files that end with *.stories.ts
  const req = require.context('../src/stories-decorator-api', true, /main\.ts$/);
  req.keys().forEach(filename => req(filename));
}

configure(loadStoriesDecoratorConfig, module);
```


igor-dv:
Hey @klimentru1986, thanks for the PR. It reminds me of the POC I've created some time ago https://github.com/igor-dv/storybook-annotations-api  🤣 

Anyway, I think that the Storybook monorepo is not the place to add this additional API. 
I see it wraps a Storybook public api, so I would reommend extract it to a different package out of the monorepo. Something like `@storybook/angular-x` 😄 We will be able to extend it with more angular helpers/utils that will probably come.
Also, we will be able to maintain and release it whenever we want and it will probably work with previous versions of SB =)

WDYT ?

Tagging here people that where active in angular contribution/discussion: @alterx , @Quramy, @kroeder @dereklin @jogelin, @andrei-ilyukovich

klimentru1986:
it's wrapper made in angular style. 

For example:

angular module
```sh
@NgModule({
  imports: [BrowserModule]
})
export class AppModule {}
```

angular component
```sh
@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})
export class AppComponent {}
```

angular app bootstrap
```sh
platformBrowserDynamic().bootstrapModule(AppModule);
  ```
If the API will be in SB repository, it will improve the use SB in Angular community. 
And Angular developers can use it instead current API by default in future.

Decorators API don't crush current API
New API development will be parallel to the current.

And I can create another angular example app in SB repository for decorators Api. 😉 

#### I think the Angular way is cool 😁

kroeder:
This is awesome! It might actually solve a linting issue where demo components are marked red for not being declared in a module because they are internally declared by storybook

ndelangen:
@kroeder would you be able to help @klimentru1986 to create this package in a separate repo as @igor-dv has suggested we do?

You should have all the right in this organisation.

klimentru1986:
I have a big problem with storysource loader. In [parse-helpers](https://github.com/storybooks/storybook/blob/master/addons/storysource/src/loader/parse-helpers.js) because new parse types not `Literal` and `StringLiteral` And application crash when i start it outside storybook bild.

@igor-dv @ndelangen please close this request if you categorically don't wont apply it. 
I think is best sollution in this situation.

igor-dv:
Storysource will probably not work with this API since the stories AST now is too different. @klimentru1986, why are you against having it as a separate package?

ndelangen:
@kroeder what can we do?

kroeder:
Let's schedule this after the ts migration and put it into a package `@storybook/story-decorators` if possible

Issue #4289, opened 2018-10-04
make react a implicit dependency of @storybook


gabrielcsapo:
Right now we make clients who are not using react bundle react, which seems a little confusing.

gabrielcsapo:
@shilman @igor-dv continuing discussion here

tmeasday:
This seems like the right thing to do to me.

igor-dv:
Yes. Agree.

igor-dv:
With the same effort, let's maybe add `babel-loader` + `@babel/core` to the angular app since they don't really need babel themselves.

gabrielcsapo:
@igor-dv you think we should do this in the same cleanup?

Hypnosphi:
@igor-dv I'm afraid we can't do that
https://github.com/storybooks/storybook/issues/3335#issuecomment-396078461

Hypnosphi:
> If we made react and react-dom a direct dep of all app/* packages

But we already have eact and react-dom a direct dep of all app/* packages except react

Hypnosphi:
@gabrielcsapo which particular package doesn't have an explicit dependency on react at the moment?

igor-dv:
> I'm afraid we can't do that

@Hypnosphi, are you sure it's relevant to the Angular apps as well? Everything there is TS...

igor-dv:
> which particular package doesn't have an explicit dependency on react at the moment?

looks like ember app missed it

Hypnosphi:
@igor-dv babel supports TS now, plus it can appear in `node_modules` because of being a dependency of some other third party package. What you suggest means relying on npm hoisting

Hypnosphi:
> ember app missed it

I see. @gabrielcsapo please add those dependencies to ember app

gabrielcsapo:
Sorry for the delay just published a PR

ndelangen:
Can we close this then? PR seems to be merged.

PR #4309, opened 2018-10-07
[ember] add dependencies to root application

gabrielcsapo:
Issue: #4289

## What I did

Add dependencies to the top level @storybook/ember application

### DIFF
--- a/app/ember/package.json
+++ b/app/ember/package.json
@@ -28,12 +28,14 @@
     "@babel/runtime": "^7.0.0",
     "@ember/test-helpers": "^0.7.25",
     "@storybook/core": "4.0.0-alpha.24",
+    "common-tags": "^1.8.0",
+    "react": "^16.4.2",
+    "react-dom": "^16.4.2",
     "global": "^4.3.2"
   },
   "peerDependencies": {
     "babel-loader": "^7.0.0 || ^8.0.0",
     "babel-plugin-ember-modules-api-polyfill": "^2.4.0",
-    "common-tags": "^1.8.0",
     "ember-cli-htmlbars-inline-precompile": "^1.0.3",
     "ember-source": "^3.4.0"
   }


app/ember/package.json
```diff
@@ -28,12 +28,15 @@
     "@babel/runtime": "^7.0.0",
     "@ember/test-helpers": "^0.7.25",
     "@storybook/core": "4.0.0-alpha.24",
+    "common-tags": "^1.8.0",
+    "moment": "^2.22.2",
```
igor-dv: I think `moment` is not needed here. It comes from knobs AFAIR.

app/ember/package.json
```diff
@@ -28,12 +28,15 @@
     "@babel/runtime": "^7.0.0",
     "@ember/test-helpers": "^0.7.25",
     "@storybook/core": "4.0.0-alpha.24",
+    "common-tags": "^1.8.0",
```
igor-dv: Why is this moved?

gabrielcsapo: It is used in @storybook/ember and that's it.

app/ember/package.json
gabrielcsapo: will remove, it was being required when I tried to install

app/ember/package.json
igor-dv: Doesn't ember itself depend on this dep?

gabrielcsapo: no, it is used for outputting error information in the render function

igor-dv (approved):

pksunkara (approved):

fran-worley:
@igor-dv since adding react and react-dom to my app dependencies it is now being recognised as a react app by `storybook init`. 

Does the order of if statements in the app detection need reworking so ember is detected before react?

igor-dv:
@Keraito is working on a CLI improvement where you will be able to pass a flag of the wanted framework. Probably `react` is one of the first apps to be checked...

fran-worley:
@igor-dv I figured a simple solution would be to move the react clause to the end [here](https://github.com/storybooks/storybook/blob/master/lib/cli/lib/detect.js#L57-L62)

As I side note I see that ember isn't included [here](https://github.com/storybooks/storybook/blob/master/lib/cli/lib/detect.js#L117) is that likely to cause issues for full ember support?

igor-dv:
>  I figured a simple solution would be to move the react clause to the end here

yes, it will probably help, want to PR ?

> As I side note I see that ember isn't included here is that likely to cause issues for full ember support?

I am less familiar with the CLI part. @Keraito @gabrielcsapo do you have something to say about this ?

fran-worley:
@igor-dv compiling PR now.

> As I side note I see that ember isn't included here is that likely to cause issues for full ember support?

Having now added 'ember' to that list, it is correctly detecting when storybook is already setup so I think it needs to be in there and I'll add to my PR in a separate commit.

PR #4314, opened 2018-10-07
Type safe select

ca057:
Issue:

This solves the case in the `select`-knob when you pass in an array or an object where the values are not strings but the prop the knob is used for expects another type (e.g. a number), see for example issue #4106 

## What I did
The passed in options of the `select`-knob are stored in the local state, so they are not converted to strings when setting them as a value of the `option`-element. On a selection, the original value will be received from the state.

## How to test

Does this need a new example in the kitchen sink apps? yes - see added story for testing.


addons/knobs/src/components/types/Select.js
```diff
@@ -3,26 +3,61 @@ import PropTypes from 'prop-types';
...
+  [safeToString(value)]: value,
+});
+
 class SelectType extends Component {
-  renderOptionList({ options }) {
-    if (Array.isArray(options)) {
-      return options.map(val => this.renderOption(val, val));
-    }
-    return Object.keys(options).map(key => this.renderOption(key, options[key]));
+  static getDerivedStateFromProps({ knob }) {
```
ndelangen: I think we can instead use memoization here to achieve the same performance.

What do you think? It's usually recommended to use memoization over getDerivedStateFromProps.

examples/cra-kitchen-sink/src/stories/type-safe-select.stories.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
ndelangen: Should move this into the official example for addon knobs, the cra example should be a lean and simple showcase our compatibility with CRA, not demo every possible addon/options. That's what the official example is for 😄

ndelangen (changes requested):

ndelangen:
Awesome work @ca057 !

Thank you for improving the knobs addon, this makes a ton of sense.

What do you think about my comment of the usage of getDerivedStateFromProps? Happy to keep it if you feel this is a better solution.

ca057:
Hey @ndelangen !

Thanks for the feedback! I’m happy to improve that 👍

igor-dv:
Any plans on finishing this PR?

ca057:
> Any plans on finishing this PR?

Hey - yes! I have it open for the whole weekend to continue it again. Sorry, I was out the last month. Let me update my PR to include the latest changes and work on the comments from @ndelangen .

ca057:
@igor-dv I checked out the latest version of Storybook and saw that this feature was added in the meantime by another PR. Can you check the latest version please and confirm this? I would then close this PR. Thanks for your patience.

ndelangen:
Yeah this seems to be fixed already, thank you @ca057 for your time and energy.
Bit sad your PR didn't get merged, but want to let you know I truly appreciate the effort you put in.

PR #4318, opened 2018-10-08
Fixes to cli ember support (followup on #4309)

fran-worley:
In reference to comment: https://github.com/storybooks/storybook/pull/4309#issuecomment-427808091

This PR corrects app detection when 'react' has also been added to package.json and fixes detection of storybook configuration.

### DIFF
--- a/lib/cli/lib/detect.js
+++ b/lib/cli/lib/detect.js
@@ -122,6 +122,7 @@ function isStorybookInstalled(dependencies, force) {
       'polymer',
       'mithril',
       'riot',
+      'ember',
     ];
     if (
       supportedFrameworks.reduce(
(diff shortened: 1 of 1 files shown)


lib/cli/lib/detect.js
```diff
@@ -23,6 +23,13 @@ function detectFramework(dependencies) {
     return types.VUE;
   }
 
+  if (
```
gabrielcsapo: odd that order matters...

gabrielcsapo (approved):

gabrielcsapo:
Thank you @fran-worley !

lib/cli/lib/detect.js
fran-worley: @gabrielcsapo I had to add react to my apps package.json as a temporary work around and from that point it was determined to be a react app. In switching them around, it checks for ember first, react second and correctly determines the type. 

Figured it was an easy fix to something that could catch others out.

PR #3903, opened 2018-07-20
[onDeviceUI] Add ability to render addons in React Native

Gongreg:
# Issue 
Currently the only way to use addons in React Native is to display them in browser (that means you need to have a server running). This PR allows users to create addons which work in RN app itself.

You can test out the new functionality here! https://expo.io/@gongreg/crna-kitchen-sink.

40x

This is the main PRs for this functionality. Relevant PRs: 

#4326
#4327
#4328
#4340

## Changes

* #### Updated the UI.
  * Moved top bar to the bottom, added border to make it more distinct.
  * Bar now has three buttons: Navigator, preview, addons. Clicking on either navigator or addons makes the preview minimized.
  * You can swipe on the bar to select different panels.
  * You can press on minimized preview to maximize it.
  * Added a button on the right to hide the bar completely.

* #### `onDeviceUI` is now true by default.
You can disable it by passing onDeviceUI as false.

* #### Two new options for initial UI display
  *  isUIOpen - display bottom bar initially (default - true).
  *  tabOpen - which panel to display initially (navigator(-1), preview(0) or addons(1) (default is 0).

* #### Safe area view hack (for iphone x) removed.
Now it uses SafeAreaView component. It requires RN 0.50.

* #### Reduced amount of rerenders happening when using onDeviceUI.
Before using onDeviceUI made two rerenders on story change events.

* #### Channel is created immediately.
Before it was impossible to use addons channel if you didn't render storybookUI component. Now the channel is created as soon as you call getStorybookUI.

## Testing
I've updated CRNA example to support on device addons.
I've also published expo example: https://expo.io/@gongreg/crna-kitchen-sink

## Part of other PR's

* #### Addon registration
cli now creates rn-addons.js file and imports it in entry storybook file.

Current storybook architecture works in two parts. Manager and preview. Adding addons to onDeviceUI removes this separation. 

This separation was necessary to support emitting events inside decorators.
Without it now we get warning that renders are not "pure" (they are calling setState while setState is already happening).

By using async channel we avoid this issue since events are emitted only after render has finished.

* #### Handles websocket error without crashing.
If storybook tries to connect to non existing storybook server it will not throw red screen anymore.

* #### Created new on device addons
  * Backgrounds
  * Knobs
  * Notes

* #### Unified template for Storybook installation in React Native.
Both CRNA and RN Vanilla will use same template when installing storybook.

PR #4326, opened 2018-10-09
[onDeviceUI] Updated channel to support async option

Gongreg:
This is a part of PR #3903.

* Async option for the channels. If set to true emit is called in setImmediate.

Current storybook architecture works in two parts. Manager and preview. Adding addons to onDeviceUI removes this separation.

This separation was necessary to support emitting events inside decorators.
Without it now we get warning that renders are not "pure" (they are calling setState while setState is already happening).

By using async channel we avoid this issue since events are emitted only after render has finished.

* Websockets channel does not crash RN if connection fails. Added optional onError argument to call on connection fail.

### DIFF
--- a/lib/channel-websocket/src/index.js
+++ b/lib/channel-websocket/src/index.js
@@ -3,15 +3,13 @@
 import { WebSocket } from 'global';
 import Channel from '@storybook/channels';
 
-const logger = console;
-
 export class WebsocketTransport {
-  constructor({ url }) {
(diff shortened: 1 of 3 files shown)


igor-dv (approved):

ndelangen (approved):

Gongreg:
@igor-dv, @ndelangen lets merge this one? Nothing is blocking it.

PR #4327, opened 2018-10-09
[onDeviceUI] Adding on device addons.

Gongreg:
This PR is part of #3903.

This adds 3 on device addons: Notes, Knobs, Backgrounds.
They are definitely not the cleanest and up to date addons.

Looking forward @ndelangen suggested that we have to think of the way to display addons as webviews in the app. But for now these are good enough to display that we can have addons in the app.


addons/ondevice-backgrounds/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@storybook/addon-ondevice-backgrounds",
+  "version": "4.0.0-alpha.20",
```
igor-dv: missed the version update

addons/ondevice-backgrounds/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "url": "https://github.com/storybooks/storybook.git"
+  },
+  "license": "MIT",
+  "main": "dist/index.js",
+  "jsnext:main": "src/index.js",
+  "scripts": {
+    "prepare": "node ../../scripts/prepare.js"
+  },
+  "dependencies": {
+    "@storybook/addons": "4.0.0-alpha.20",
```
igor-dv: like everywhere

addons/ondevice-knobs/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "license": "MIT",
+  "main": "dist/index.js",
+  "jsnext:main": "src/index.js",
+  "scripts": {
+    "prepare": "node ../../scripts/prepare.js"
+  },
+  "dependencies": {
+    "@storybook/addons": "4.0.0-alpha.20",
+    "@storybook/addon-knobs": "4.0.0-alpha.20",
```
igor-dv: didn't get where this in use

addons/ondevice-knobs/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "deep-equal": "^1.0.1",
+    "prop-types": "^15.6.2",
+    "react-native-color-picker": "^0.4.0",
+    "react-native-modal-datetime-picker": "^5.1.0",
+    "react-native-modal-selector": "0.0.27",
```
igor-dv: should this be stuck to the exact version?

igor-dv (approved):
Mostly versions mismatch.

addons/ondevice-knobs/package.json
Gongreg: A person who uses ondevice-knobs will be still importing addon-knobs inside their app. So if he doesn't install it directly we still want him to have it in his node_modules. @igor-dv.

igor-dv: Yes, but users will need to install it explicitly. In that case, you probably want it to be a peer

Gongreg: Alright, thanks :)

Gongreg:
Nothing seems to be blocking this PR too @ndelangen, @igor-dv.

igor-dv:
Let's merge if after the #3903 is ready? We won't use addons without the support for addons, right?

Gongreg:
Sure.

ndelangen:
merge when ready

ndelangen (approved):

PR #4336, opened 2018-10-10
Fix Array.js value to string

do7be:
Issue: https://github.com/storybooks/storybook/issues/4299

## What I did

I fixed Array.js `value` types for `react-text-area-autosize`

https://github.com/andreypopp/react-textarea-autosize/blob/master/src/index.js#L21

## How to test

I did only test on browser. I didn't find that error.


ndelangen (approved):

ndelangen:
Thank you @do7be !

PR #4338, opened 2018-10-10
Try to fix cli tests

igor-dv:
Issue: CLI tests are red already a few days

## What I did
I've found that it's related to the test/fixture/riot. There are few deps there marked as `latest` - that can be flaky.
Removing the `uglifyjs-webpack-plugin` seems fixes the problem, though, I am not sure it's a real solution.


libetl (approved):
looks good to me.

PR #4340, opened 2018-10-10
[onDeviceUI] Adding React Native documentation in docs site.

Gongreg:
This is documentation that we should use starting from React Native storybook v4.
Moving it to a separate PR since I am not sure when we will release v4.


ndelangen (approved):

ndelangen:
Perfect! 👏

ndelangen:
merge when ready

PR #4343, opened 2018-10-10
[improvement] Add support for sass type files

tomepejo:
Issue:

## Add support for sass type files

Current configuration only supported sass files.

## How to test

Add component in the storybook that has sass external file.


igor-dv:
Just to clarify, don't you use `angular-cli` ?

tomepejo:
Yes, I use the `angular-cli`.

Maybe this is not the only place that this support for `.sass` files should be added.
Please point me if you know any other place in the configuration where there's only support for `.scss` files currently.

The fix for us currently is inside of the `webpack.config.js` file in the `.storybook` folder is:
```
    {
      test: /\.sass$/,
      use: [require.resolve('raw-loader'), require.resolve('sass-loader')],
    }
```

igor-dv:
Storybook integrates with angular-cli style rules, and angular-cli do have `sass` defined. I am asking because it's strange that you even needed to put sass to the extended webpack.config.js

ndelangen:
@igor-dv it seems we're adding .scss though, why not .sass?

igor-dv:
These styling rules are only needed for the default usage, because most of the users are using angular-cli. And we are overriding styling rules with the rules from angular-cli.

Anyway, this change is ok.

@tomepejo, we need now to fix the exclusion. Take a look [here](https://github.com/storybooks/storybook/blob/master/app/angular/src/server/angular-cli_config.js#L109)

ndelangen (dismissed):

ndelangen:
merge when ready

app/angular/src/server/angular-cli_config.js
```diff
@@ -110,7 +110,7 @@ export function applyAngularCliWebpackConfig(baseConfig, cliWebpackConfigOptions
   // because @angular-devkit/build-angular created rules have include/exclude for global style files.
   const rulesExcludingStyles = baseConfig.module.rules.filter(
     rule =>
-      !rule.test || (rule.test.toString() !== '/\\.css$/' && rule.test.toString() !== '/\\.scss$/')
+      !rule.test || (rule.test.toString() !== '/\\.css$/' && rule.test.toString() !== '/\\.sass$/' && rule.test.toString() !== '/\\.scss$/')
```
igor-dv: I think it won't work now. Here is a string comparison, while we need a regex matching.

Hypnosphi: Yes. please replace with something like `!rule.test.test('.sass')` etc

Hypnosphi (changes requested):

ndelangen:
Need help applying those changes @tomepejo ?

tomepejo:
> Need help applying those changes @tomepejo ?

Yes. I don't see the reason to create a regex comparison when just extending the current condition should resolve the issue. Can you please explain? Also, if easier.. please just apply the fix right away. Thank you.

igor-dv (approved):

Issue #4249, opened 2018-09-28
Feature Storybook Node API


ar53n:
If you are reporting a bug or requesting support, start here:
### Bug or support request summary

Hello, I propose to do `Node API`.
Like a `webpack`
```
const Storybook = require('storybook')

const storybookInstance = new Storybook(config)
```

igor-dv:
That's nice. I even have a few stashed changes toward this =)

igor-dv:
@Keraito, does this issue somehow collides with your work on CLI ?

Keraito:
@igor-dv 	currently not as the CLI is basically just running terminal commands on the background 😅 Having this kind of API to do it codewise would be cool tho for both the CLI and what is suggested here.

PR #4344, opened 2018-10-10
CLI-less api for node

igor-dv:
Issue: #4249

## What I did

There are cases when people want to run storybook from their code. 
I've exposed the "standalone" api, so now it will be possible to do something like this:

```js
import build from '@storybook/react/standalone';

build({
  mode: 'dev',
  port: 9009,
  configDir: './storybook',
  frameworkPresets: [],
});
```

### DIFF
--- a/app/angular/src/server/index.js
+++ b/app/angular/src/server/index.js
@@ -1,5 +1,4 @@
 import { buildDev } from '@storybook/core/server';
-
 import options from './options';
 
 buildDev(options);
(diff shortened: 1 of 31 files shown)


ndelangen (approved):

igor-dv:
Will document it later 😇 I swear

transitive-bullshit:
@igor-dv I wasn't aware this was so recent of an addition, but I've found it **extremely** useful.

Thanks!!

PR #4382, opened 2018-10-12
Navigator panel had 1 px visible line on screen.

Gongreg:


Hypnosphi:
You can just commit directly to master instead, to avoid triggering extra CI builds

PR #4411, opened 2018-10-13
Use postcss.config.js when present

kylemh:
Issue: #2455 

## What I did
- Add `find-up` to dependencies to search for the nearest `postcss.config.js`
- Use the nearest `postcss.config.js` to determine postcss-loader options if present, maintain previous defaults if not.

## How to test
- Use local storybook core against a test repository using a postcss config with an option not present in the default set (such as `postcss-import`).

**Is this testable with Jest or Chromatic screenshots? Only if a custom serializer renders all styles.
Does this need a new example in the kitchen sink apps?** 
No.

**Does this need an update to the documentation?** 
Yes. Describe that custom postcss.config.js should be handled just like custom babel configurations. Just wanted to ensure this is the correct direction before I add any changes.


lib/core/src/server/config/webpack.config.default.js
```diff
@@ -19,16 +24,18 @@ export function createDefaultWebpackConfig(storybookBaseConfig) {
...
-                plugins: () => [
-                  require('postcss-flexbugs-fixes'), // eslint-disable-line
-                  autoprefixer({
-                    flexbox: 'no-2009',
-                  }),
-                ],
-              },
+              options: postcssConfig
+                ? {
+                    postcssConfig,
```
igor-dv: should it be `{ ...postcssConfig }` ?

I think will be better something like this:

```ja
postcssConfig || {
   // flexbug blah blah
}
```

lib/core/src/server/config/webpack.config.default.js
```diff
@@ -19,16 +24,18 @@ export function createDefaultWebpackConfig(storybookBaseConfig) {
...
+                  }
+                : {
+                    plugins: () => [
+                      require('postcss-flexbugs-fixes'), // eslint-disable-line global-require
```
igor-dv: I understand that something won't work if this plugin will be missing?

igor-dv (commented):
We should have an example for this I think

lib/core/src/server/config/webpack.config.default.js
kylemh: The plugins are currently available, correct? I didn't remove dependencies for that reason.

igor-dv: I mean, should we somehow merge it to the user's config?

kylemh:
Gonna wait on #4405 before I move forward.

The only work left to be done will be to integrate with those changes + provide an example + update docs 👍 

I'm prepared to make the moves at any moment in my work day.

lib/core/src/server/config/style.config.js
```diff
@@ -0,0 +1,42 @@
+import { logger } from '@storybook/node-logger';
+import autoprefixer from 'autoprefixer';
+import findUp from 'find-up';
+
+export default function getStyleConfig(baseConfig) {
+  // null if file is not found
+  const customPostcssConfig = findUp.sync('postcss.config.js', {
```
igor-dv: Will be very nice if you could asyncify this... Need to add some additional `await`s in the call chain though. Not critical one, bit since we already support async/await that be great =)

kylemh:
Suggested spot for documentation:
 A section after https://storybook.js.org/configurations/custom-babel-config/ called "other common custom configurations" to leave room for how `postcss.config.js` gets consumed and - eventually - how `.browserlistrc` gets consumed

lib/core/src/server/config/style.config.js
```diff
@@ -0,0 +1,47 @@
+import { logger } from '@storybook/node-logger';
+import autoprefixer from 'autoprefixer';
+import findUp from 'find-up';
+
+export default async function getStyleConfig(baseConfig) {
+  // null if file is not found
+  const customPostcssConfig = await findUp.sync('postcss.config.js', {
```
igor-dv: it should be just `findUp(...)` without the sync

kylemh: 🤦‍♂️  fixed

kylemh:
@ndelangen thoughts on deleting this branch PR and starting from the new API changes?

kylemh:
Gonna take another stab at this in a different PR since direction has changed a bit

PR #4413, opened 2018-10-14
Improve Environment loading

### LINKED ISSUE #4171: Alpha Version Static Export Build Script Does Not Respect Env Vars

**What did you expect to happen?**

Running `build-storybook` would respect environment variables such as `NODE_ENV` and `STORYBOOK_*`

**What happened instead?**

Running `build-storybook` seems to not respect environment variables such as `NODE_ENV` and `STORYBOOK_*`

**NOTE**: `start-storybook` does behave as expected.

### Steps to reproduce

Here is a repo where you can test this using the provide configs and NPM scripts:
https://github.com/jhoffmcd/storybook-test-build-script

- @storybook/react 4.0.0-alpha.21

### Affected platforms 

Developer tooling?

pksunkara:
Fixes #4171 

We made our own dotenv loader at https://github.com/storybooks/lazy-universal-dotenv


ndelangen (approved):

lib/core/src/server/config/utils.js
```diff
@@ -29,8 +30,10 @@ export function loadEnv(options = {}) {
       env[name] = JSON.stringify(process.env[name]);
     });
 
+  const { stringified } = getEnvironment({ nodeEnv: JSON.parse(env.NODE_ENV) });
```
kylemh: I'm trying to put in an honest review, but I'm afraid you'll have way more context than me.

This line and 36 read as if you're specifically pulling out NODE_ENV and only applying that to environments. Is there anything to be worried about with other env vars not being passed, like `JEST_ENV`?

kylemh (approved):

lib/core/src/server/config/utils.js
pksunkara: Not all env vars get passed in Webpack. That was what the PR was about. Anything from the system, only `STORYBOOK_*` is allowed. You can add everything else to `.env` files.

PR #4426, opened 2018-10-15
[ember] create ember entry point to official storybook

gabrielcsapo:
## What I did

- Adds an entry point to the ember example app


Hypnosphi:
It should be a symlink not a text file

pksunkara:
My bad, github shows symlink like this, so thought this diff contained a symlink. Should have looked at the details.

pksunkara:
@gabrielcsapo Can you make a PR fixing this? Thanks.

Issue #4296, opened 2018-10-05
Full-control webpack config with default results in duplicate style loader rules


Yogu:
We use storybook with angular in full-control webpack.config mode (to configure ts aliases). In v4.0.0-alpha.20 this worked fine, but sincd v4.0.0-alpha.21, we get the error "Module build failed: Unknown word" while loading CSS file. This is because `module.rules` includes css rules twice - once from angular-cli and once from the storybook's default webpack. I guess this should be migitgated by `applyAngularCliWebpackConfig` which extracts `rulesExcludingStyles` for this reason. However, this function is no longer called to generate the default config passed to our custom webpack function.

I guess this problem was introduced with presets (#4027), but I have not looked further into it.

### Steps to reproduce

If the description is not clear enough, I can try to make an MCVE.

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/angular 4.0.0-alpha.21 through alpha.24

jfsiii:
These symptoms sound just like what I'm seeing in a React project, although we are pinned at `alpha.16`.

We use the [full control + default mode](https://storybook.js.org/configurations/custom-webpack-config/#full-control-mode--default) and our setup looks roughly like

```javascript
import ourConfig from 'path/to/config'

module.exports = (baseConfig, env, defaultConfig) => {
  const merged = {
    // keep storybook's defaults
    ...defaultConfig,

    // add our rules (required for SASS loaders, js/happypack, etc)
    module: {
      rules: [
        ...defaultConfig.module.rules,
        ...ourConfig.module.rules
      ]
    },

    // add our plugins (babel transforms, etc)
    plugins: [...defaultConfig.plugins, ...ourPluginsFiltered ],

    // our changes to `resolve`
    resolve: { ...defaultConfig.resolve, ...ourConfig.resolve }
  }

  return merged
}
```

In our case the error is coming from `postcss-loader` 

![storybook unknown word css error](https://user-images.githubusercontent.com/57655/46564079-02b9bf80-c8ba-11e8-9f59-98b0897f5ef8.png)


which is added by `defaultConfig.module.rules`. I switched the rules section to use `baseConfig.module.rules`, leaving `defaultConfig` for the other parts and things work as expected.

Definitely a workaround but posting in case it unblocks someone else.

igor-dv:
@Yogu, I don't know what is MCVE is, but if it will help me to reproduce that'd be great. Also, please share your custom webpack.config for Storybook.

[...]

Issue #4230, opened 2018-09-25
Angular 6 setup errors - starting from the docs - Module build failed (from ./node_modules/postcss-loader/lib/index.js):


KolbySisk:
### Bug or support request summary

I'm going to share my journey to get storybook angular to work again in my project. It was working fine when I was running Angular 5, but after updating to 6 ... it stoped. 

So I'm starting with a fresh Angular 6.0.3 project. 

Following the Storybook documentation, after a basic setup, I get the error:
```javascript
node_modules/webpack/lib/DefinePlugin.js:42
                compiler.hooks.compilation.tap(
                               ^
TypeError: Cannot read property 'compilation' of undefined
```

After digging around the current issues I learned that this was a problem with webpack that arose when updating from a5 to 6. Was told that this was fixed in the `4.0.0-alpha.11+` release.

~ During this point, depending on the alpha version, I've gotten an error about `moment` not being installed. Assuming this is fixed now with the latest alpha release since I didnt see it with `.22`. ~

Now I'll try installing the latest alpha `4.0.0-alpha.22`. I get the error:
`Error: Cannot find module 'babel-loader/package.json`

I can't find any threads mentioning this error, so I'll just take a guess and install `babel-loader`
Cool, seemed to do something. Now I get the error:
`Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@babel/core'`

ok, `npm i @babel/core`

Now I get:
```javascript
ERROR in ./src/styles.css (./node_modules/raw-loader!./node_modules/postcss-loader/lib??embedded!./node_modules/@storybook/core/node_modules/style-loader!./node_modules/css-loader??ref--17-1!./node_modules/@storybook/core/node_modules/postcss-loader/src??postcss!./src/styles.css)
Module build failed (from ./node_modules/postcss-loader/lib/index.js):
Syntax Error

(2:1) Unknown word
```
After removing `/src/styles.css` from my `angular.json` file it appears to be working. So now the only problem appears to be how can I include these styles in the `angular.json`?

nunos7:
Hi, i have a similar problem...
I use in my project angular 6.1.0 and @storybook/angular 4.0.0-alpha.22, when i try to start storybook i get this error:

info @storybook/angular v4.0.0-alpha.22
info
info => Loading presets
TypeError: Cannot read property 'build' of undefined

igor-dv:
@KolbySisk, 
[...]

Issue #4169, opened 2018-09-12
Single Configuration File - A.K.A. storybook.config.js - monoconfig


igor-dv:
_let's discuss the "Single Configuration File" feature_ 

I will update this template with the agreed changes. The inspiration is from [here](https://github.com/storybooks/SBNext/blob/master/examples/react/storybook.config.js)

## API
The `storybook.config.js` will be at a root level, but we also may consider it to be configurable with the cli argumetns (for example for people that will have a few configurations).

### structure
```ts
interface SbConfg {
  entries?: Array<string> | string
  addons?: Array<string>
  presets?: Array<string> | Array<object>
  webpack?: object | function
  babel?: object | funciton
};
```

#### `storybook.config.js` exposes an object
```ts
module.exports = SbConfg;
```

#### `storybook.config.js` exposes a fuction
```ts
module.exports = (env, mode) => SbConfg  // anything else ?
```

### entries
Let's not call this field `stories`, because semantics will maybe changed

#### globs:
```js
{
  entries: 'src/**/*.story.js'
}
```

#### array of globs:
```js
{
  entries: ['src/**/*.stories.js', 'src/**/*.examples.js']
}
```

#### path to file:
```js
{
  entries: 'path/to/stories.js' // it may have a `require.context` or something else
}
```

#### combination
```js
{
  entries: ['path/to/stories.js',  'src/**/*.stories.js']
}
```

- I am not sure, but maybe regex is also relevant.
- we should add a sensible default, so it will work without any configuration

### addons

An array of addons in the following structure:

```js
{
  addons: ['@storybook/addon-X', '@storybook/addon-Y']
}

```

We will register the addons by ourselves (no need of `addons.js`). The question is what to do with the addon's defaults, for example with what we setup today in `config.js` ? we can use the same pattern as in `presets` 🤷‍♂️ 

#### array of strings
```js
{
  presets: ['@storybook/preset-scss', '@storybook/preset-typescript']
}
```

#### array of objects
```js
{
  presets: [
    {
      preset: '@storybook/preset-scss',
      options: { }
    }
  ]
}
```

#### combination
```js
{
  presets: [
    {
      preset: '@storybook/preset-scss',
      options: { }
    },
    '@storybook/preset-typescript'
  ]
}
```

#### object
Will be merged to the final SB webpack config
```js
{
  webpack: {
    module: {
      rulese: []
    },
    // other stuff
  }
}
```

#### function
Like today, but we may change if something is redundnat
```js
{
[...]

PR #4431, opened 2018-10-15
Expose webpackFinal extension

igor-dv:
Issue: 
`angular-cli` preset should filter out all the styling rules in order to integrate its rules instead. 
Currently, angular-cli preset is kicked-in in the middle of the presets chain because it's a framework preset.

The last preset in the chain adds default rules or execute the custom webpack config. We need the ability for other presets to clean the webpack config after all the manipulations.

image

Related #4296

## What I did
I've added the `webpackFinal` step.

### DIFF
--- a/lib/core/src/server/config.js
+++ b/lib/core/src/server/config.js
@@ -7,6 +7,7 @@ function wrapCorePresets(presets) {
   return {
     babel: (config, args) => presets.apply('babel', config, args),
     webpack: (config, args) => presets.apply('webpack', config, args),
+    webpackFinal: (config, args) => presets.apply('webpackFinal', config, args),
     preview: (config, args) => presets.apply('preview', config, args),
     manager: (config, args) => presets.apply('manager', config, args),
   };
@@ -34,7 +35,9 @@ function getWebpackConfig(options, presets) {
     manager: presets.manager([], options),
   };
 
-  return presets.webpack({}, { ...options, babelOptions, entries });
+  const webpackConfig = presets.webpack({}, { ...options, babelOptions, entries });
+
+  return presets.webpackFinal(webpackConfig, options);
 }
 
 export default options => {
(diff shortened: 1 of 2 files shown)


igor-dv:
Looks like it should also fix the initial issue from #4230

lib/core/src/server/config.js
```diff
@@ -7,6 +7,7 @@ function wrapCorePresets(presets) {
   return {
     babel: (config, args) => presets.apply('babel', config, args),
     webpack: (config, args) => presets.apply('webpack', config, args),
+    webpackFinal: (config, args) => presets.apply('webpackFinal', config, args),
```
kroeder: Since I am not familiar with the new presets: Can developers access the presets from outside? 
I was just thinking that `webpack` seems irrelevant because `webpackFinal` is the real working webpack configuration

Do we even need 2 presets or just apply `webpackFinal` on `webpack` in the background and only offer 1 webpack preset? Or is this some sort of "if anything went wrong, `webpack` is your safety net"?

igor-dv: > Can developers access the presets from outside 

It's possible with the [custom presets](https://github.com/storybooks/storybook/blob/master/lib/core/src/server/config.js#L15) (currently experimental, but will be not experimental after the #4169)

Presets are chained. Let's say you have:

``` 
preset-a
  webpack
  webpackFinal
 
preset-b
  webpack

preset-c
  webpack
  webpackFinal
```
they will be called in the next order: 

1. preset-a.webpack -> preset-b.webpack -> preset-c.webpack
2. preset-a.webpackFinal -> preset-c.webpackFinal

That way some earlier defined preset can override the configuration of the later defined preset, of course if the later one doesn't have its own extension.

I was thinking about having something like `pitch` api in webpack loaders (they have it in the opposite order).

So it will be like this:

```
preset-a.webpack 
       preset-c.webpack
              preset-d.webpack
              preset-d.webpackBack
       preset-c.webpackBack     
preset-a.webpackBack
```
It can be supported in the [core level](https://github.com/storybooks/storybook/blob/master/lib/core/src/server/presets.js#L44)

kroeder (approved):
Looks fine, thanks for clarifying presets

PR #4452, opened 2018-10-16
bug-fix-jest-addon-undefined-file-extension

ghost:
Previously the code was such that the file name would always have an extension that read `undefined` this fixes that.

Issue:

## What I did
We updated from v4.0.0-alpha.24 to v4.0.0-rc.0 and the unit tests no longer showed up for the jest addon. We dug into the code and quickly realized that the file extension will always be undefined in line 7 of the code because it isn't an object but a string.

## How to test
Without this fix, see that jest test results do not show up, then update it and the test results should be available. (not sure if there were unit tests to make sure of this implementation)

### DIFF
--- a/addons/jest/src/index.js
+++ b/addons/jest/src/index.js
@@ -2,9 +2,9 @@ import addons from '@storybook/addons';
 import deprecate from 'util-deprecate';
 import { normalize } from 'upath';
 
-const findTestResults = (testFiles, jestTestResults, jestTestFilesOptions) =>
+const findTestResults = (testFiles, jestTestResults, jestTestFilesExt) =>
   Object.values(testFiles).map(name => {
-    const fileName = `${name}${jestTestFilesOptions.filesExt}`;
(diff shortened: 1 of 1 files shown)


renaudtertrais:
I am not sure to really understand the problem.
How to reproduce the issue ?
`jestTestFilesOptions ` should be an object no ?

human-readable:
@renaudtertrais correct If you look at the code below

You will notice that on line 22 (https://github.com/storybooks/storybook/blob/master/addons/jest/src/index.js#L22 ) the last parameter is being sent in as a string (the file extension name).

However, on line 7 (https://github.com/storybooks/storybook/blob/master/addons/jest/src/index.js#L7) it is clearly expecting this to be an object which would make it undefined every time and the reason our tests stopped being included on the Tests tab in storybook.

Hope that explanation is a little bit more clear and gives insight to my change

renaudtertrais (approved):

wuweiweiwu (approved):
Nice fix

PR #4473, opened 2018-10-17
docs: update note about require.context

eddiemonge:
Updated the docs to be more clear about how to replace webpack's require.context


wuweiweiwu (approved):
LGTM! its much clearer now

PR #4480, opened 2018-10-18
Update docs to use docgen loader instead of plugin

leandrooriente:
## What I did

Updated documentation to use `react-docgen-typescript-loader` intead of `react-docgen-typescript-plugin`. 

Docgen plugin was deprecated and should be replaced by the custom loader. The plugin slows down storybook development a lot (in our case from 2 minutes to 10 seconds), so it's quite important to remove it from your documentation. 

## How to test

Should I provide a small example of a storybook running with TS and docgen loader?


leandrooriente:
We also use `cache-loader` in any custom loader that we add to our custom webpack config.
Not sure if you should recommend it, but it improves a lot the build time

ndelangen (approved):

ndelangen:
> We also use cache-loader in any custom loader that we add to our custom webpack config.
Not sure if you should recommend it, but it improves a lot the build time

This PR is good to go, IMHO. But adding a section about cache-loader would be great I think.

leandrooriente:
If I'm not wrong, the next topic already covers it: https://storybook.js.org/configurations/typescript-config/#using-typescript-with-the-tsdocgen-addon

Hypnosphi:
@leandrooriente please resolve conflicts

ndelangen (approved):

PR #4498, opened 2018-10-19
fix loader config for cursor images

komiyamast:
## What I did
fix loader config for cursor images

## How to test

* Core & Examples

* All passed


igor-dv (approved):
Thanks

PR #4507, opened 2018-10-22
Addon-Knob: Add missing key property in Panel

janssen92:
Issue:
Missing key in iterator for two level knobs. 

## What I did

Added the key, similar to the id.

## How to test

Open Storybook with knob addon and multi level options, should ne throw an error anymore.

### DIFF
--- a/addons/knobs/src/components/Panel.js
+++ b/addons/knobs/src/components/Panel.js
@@ -171,7 +171,7 @@ export default class Panel extends PureComponent {
         {groupIds.length > 0 ? (
           <TabsState>
             {Object.entries(groups).map(([k, v]) => (
-              <div id={k} title={v.title}>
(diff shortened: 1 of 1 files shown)


janssen92:
What do I need to do in order to make the failing check green?

igor-dv (approved):

wuweiweiwu (approved):

L0stSoul:
hi! 

Seems like this fix is not in latest alfa release: https://github.com/storybooks/storybook/releases/tag/v4.1.0-alpha.13

as far as i understand your guideline:

> Every week: release patch releases off master to the NPM tag latest, and alpha releases off next to the NPM tag next (as e.g. 4.1.0-alpha.0).

it should be there. so i'm a bit confused.

PR #4522, opened 2018-10-23
FIX knobs select to actually support rich values

ndelangen:
Fixes: https://github.com/storybooks/storybook/issues/4402

### DIFF
--- a/addons/knobs/src/components/types/Select.js
+++ b/addons/knobs/src/components/types/Select.js
@@ -1,32 +1,32 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 
 import { Select } from '@storybook/components';
 
-class SelectType extends Component {
-  renderOptionList({ options }) {
-    if (Array.isArray(options)) {
-      return options.map(val => this.renderOption(val, val));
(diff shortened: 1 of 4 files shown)


ndelangen:
@sapkra This should truly fix the issue.

The problem was indeed that trying to set a rich value as a value on `<select>` doesn't work.

So In this PR I'm mapping back to the object-key (which should always be a string).

Also simplified the code a bit.

ndelangen:
If you'd like to review the code perhaps?

addons/knobs/src/components/types/Select.js
```diff
@@ -1,32 +1,32 @@
...
-      <Select value={knob.value} onChange={e => onChange(e.target.value)} size="flex">
-        {this.renderOptionList(knob)}
-      </Select>
-    );
-  }
-}
+const SelectType = ({ knob, onChange }) => {
+  const { options } = knob;
+  const entries = Array.isArray(options)
+    ? options.reduce((acc, k) => Object.assign(acc, { k }), {})
```
Hypnosphi: it should be `{[k]: k}` instead of `{k}`

PR #4530, opened 2018-10-23
ADD timeout to update check

ndelangen:
This way if for some reason the update check 'hangs', we ignore it and continue without it.

### DIFF
--- a/lib/core/src/server/build-dev.js
+++ b/lib/core/src/server/build-dev.js
@@ -77,7 +77,7 @@ function applyStatic(app, options) {
   }
 }
 
-const updateCheck = async () => {
+const updateCheck = async version => {
   let result;
   const time = Date.now();
   try {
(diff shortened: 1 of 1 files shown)


lib/core/src/server/build-dev.js
```diff
@@ -77,15 +77,19 @@ function applyStatic(app, options) {
...
     const fromCache = await cache.get('lastUpdateCheck', { success: false, time: 0 });
 
     // if last check was more then 24h ago
     if (time - 86400000 > fromCache.time) {
-      const fromFetch = await fetch('https://storybooks.netlify.com/versions.json');
+      const fromFetch = await Promise.race([
+        fetch(`https://storybook.js.org/versions.json?current=${version}`),
+        // if fetch is too slow, we won't wait for it
+        new Promise((res, rej) => global.setTimeout(rej, 1500)),
+      ]);
```
igor-dv: Won't the ` const data = await fromFetch.json();` fail in case of a timeout?

ndelangen: hence the `try {} catch () {}`:
https://github.com/storybooks/storybook/pull/4530/files/c7bf9f2faa78112e4dd13f606a237d2d04be2430#diff-3aeb5d20c1ea4bab7a14ff9327d11f0dR83

igor-dv (approved):

PR #4539, opened 2018-10-24
Fix the node-logger version in app/react

igor-dv:
Issue: https://github.com/storybooks/storybook/pull/4405#discussion_r227603277


shilman (approved):
LGTM!

Issue #4191, opened 2018-09-17
Storybook 4 not compatible with react 15


Sujimoshi:
### Bug
Expected: Min version of react is set to `react >= 15.0.0` , but real minimal version of `react` set by `@emotion/core` which need `react >= 16.3.0`
Using react 15 produce error on loading storybook
![image](https://user-images.githubusercontent.com/10153611/45650818-30d88a80-bad8-11e8-985d-b0ce61a745ad.png)

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/react 4.0.0-alpha.21
- react 15

### Screenshots / Screencast / Code Snippets (Optional)
```js
// @storybook/core/package.json
"dependencies": {
    "@emotion/core": "0.13.0",
},
"peerDependencies": {
    "react": ">=15.0.0",
    "react-dom": ">=15.0.0"
},
```
```js
// @emotion/core/package.json
"peerDependencies": {
    "react": ">=16.3.0"
},
```

Sujimoshi:
It's frustrating, since I can not use the react 15 and webpack 4 in my application if it use storybook

ndelangen:
Hey @Sujimoshi thanks for the issue.

We want to help you as much as wel can. I hope you understand we're doing the best we can to stay backwards compatible AND allow users to use the latest and greatest AND have a nice experience doing it ourselves.

Right now this is these thoughts on the matter:
We're moving forward with React 16 compatibility only for the moment. We didn't want to exclude React 15, it just happened with us ourselves updating to the modern tools/libs.

This may sounds strange, but React is the only lib that has this problem because of how we're using webpack & set up the package dependencies. We want to change this.

Mid term, we want to split the manager & preview more and this would allow you to run any version of react, and us to run our own.

Short term there are a couple of options:
- Keep using the alpha.20
- Upgrade to React 16
- Find some compatibility lib
- Assist us with achieving the above faster. I'd be super happy to have a chat about it in detail.
- Assist us in bringing back compatibility using some other method

I understand this is not exactly the news you were hoping to get, Let's help each other!

Sujimoshi:
Isn't alpha.20 have the same dependencies?

Sujimoshi:
Isn't alpha.20 have the same dependencies?

Sujimoshi:
I am not familiar with storybook source code, but why not just remove react from peerDeps, and allow storybook webpack bundle two versions of react

brochington:
[...]

PR #4590, opened 2018-10-26
Core separate manager from preview

ndelangen:
I've separated the manager from the preview webpack build.

The manager is now build in a separate child process.

this better isolates the user's code, because the webpack config is no longer shared between manager & preview.

This makes it possible for 2 versions of react to exist: 1 in preview, 1 in manager.
Or the preview react to be swapped with a react compatible lib like preact.

Also, previously conflicting webpack loaders are now possible.
We no longer need to blacklist/whitelist modules in loaders to make sure they do not double apply.

In essence: the users can now be given far more control of the webpack config.

---

This ties into presets.

Preset have the ability to add babel|webpack config to the preview and manager separately.


lib/cli/test/fixtures/react_babel_config_js/package.json
```diff
@@ -15,6 +15,6 @@
     "@babel/plugin-proposal-optional-chaining": "7.0.0",
     "@babel/core": "7.1.0",
     "babel-loader": "8.0.2",
-    "@storybook/react": "^4.0.0-alpha.17"
+    "@storybook/react": "^4.0.0-rc.6"
```
Hypnosphi: It's probably my mistake that those dependencies are in fixtures. They should be added by our CLI in test runs instead

Hypnosphi:
> This makes it possible for 2 versions of react to exist: 1 in preview, 1 in manager.

Is this something not implemented here yet?

igor-dv:
> Is this something not implemented here yet?

How can we support #4191 now 🤔 ?

app/riot/src/server/framework-preset-riot.js
```diff
@@ -15,5 +15,12 @@ export function webpack(config) {
         },
       ],
     },
+    resolve: {
+      ...config.resolve,
+      alias: {
+        ...config.resolve.alias,
+        'riot-compiler': 'riot-compiler/dist/es6.compiler',
```
igor-dv: I am not sure what happened (maybe env changes) but riot-compiler is now resolved to riot-compiler/lib/compiler which is a node version and riot app is failing.
This is change is a workaround, but must be fixed in this PR

lib/core/src/server/manager/manager-webpack.config.js
```diff
@@ -0,0 +1,80 @@
...
+      }),
+      new HtmlWebpackHarddiskPlugin(),
+      new webpack.DefinePlugin({ 'process.env': stringified }),
+      new CaseSensitivePathsPlugin(),
+      new Dotenv({ silent: true }),
+    ],
+    module: {
+      rules: [
+        {
+          test: /\.jsx?$/,
```
igor-dv: Assuming we have a manager client code already transpiled, we can remove this rule, right?

ndelangen: Likely, but users could load local un-compiled addons, in addons.js

igor-dv: 😞

ndelangen:
> > This makes it possible for 2 versions of react to exist: 1 in preview, 1 in manager.
> 
> Is this something not implemented here yet?

It could work in theory already, But we still have to change 'react' to a regular dependency.

lib/core/src/server/manager/manager-webpack.config.js
igor-dv: Should we even handle this usecase ?

ndelangen: I think we should, it makes it easier to get started writing addons?

.babelrc.js
```diff
@@ -0,0 +1,46 @@
+module.exports = {
```
libetl: should not it be babel.config.js ?

app/riot/src/server/framework-preset-riot.js
libetl: oops, I don't know what I can do about it.

.babelrc.js
ndelangen: Got this from here:
https://babeljs.io/docs/en/config-files

Seems to work..

ndelangen:
@tmeasday @shilman any idea what's wrong with chromatic here?

It seems consistent, but I can't reproduce locally

app/riot/src/server/framework-preset-riot.js
igor-dv: @ndelangen, this issue was not addressed but introduced in this PR...

lib/core/src/server/config/utils.js
```diff
@@ -2,38 +2,37 @@ import path from 'path';
...
-
-export const excludePaths = [path.resolve('node_modules')];
-
 export const nodeModulesPaths = path.resolve('./node_modules');
-
-export const nodePaths = (process.env.NODE_PATH || '')
-  .split(process.platform === 'win32' ? ';' : ':')
-  .filter(Boolean)
-  .map(p => path.resolve('./', p));
+export const excludePaths = [nodeModulesPaths];
```
igor-dv: Can we revert this changes in favor of #4706 ?

lib/core/src/server/manager/manager-webpack.config.js
```diff
@@ -0,0 +1,82 @@
+import webpack from 'webpack';
+import Dotenv from 'dotenv-webpack';
+import HtmlWebpackPlugin from 'html-webpack-plugin';
+import HtmlWebpackHarddiskPlugin from '@ndelangen/html-webpack-harddisk-plugin';
+import CaseSensitivePathsPlugin from 'case-sensitive-paths-webpack-plugin';
+
+import { version } from '../../../package.json';
```
igor-dv: we should provide the version to the child process.

lib/core/src/server/config/utils.js
ndelangen: I'm unsure exactly what it'd be that I'd have to change here.

igor-dv:
@ndelangen, I will revert all the NODE_PATH changes today. 
Anyway, there is also this - https://github.com/storybooks/storybook/pull/4590#discussion_r228752008 comment.

igor-dv:
> @ndelangen, I will revert all the NODE_PATH changes today

Actually I think it will be easier to just merge the #4706 before

ndelangen:
Should be good now

libetl:
Congratulations

PR #4616, opened 2018-10-29
Clearer descriptions for state addons

CyrusZei:
removed duplicated addons 

Issue:

## What I did
I removed duplicated addons

## How to test
There are two of them 

Does this need an update to the documentation? yes


docs/src/pages/addons/addon-gallery/index.md
```diff
@@ -130,10 +130,6 @@ Add ability to customize styles in the story preview area
 
 Embed [Figma](https://figma.com) designs in a storybook panel.
 
-### [State](https://github.com/Sambego/storybook-state)
-
-Manage state inside a story. Update components when this state changes.
-
```
Hypnosphi: Those are actually two different addons by different authors and with different APIs. To make it more visible, we can use "State by Sambego" and "State by dump247" titles

Hypnosphi (dismissed):

docs/src/pages/addons/addon-gallery/index.md
CyrusZei: oh, I just noticed that now, so sorry for this. 

and maybe change the description so that the two first sentence is not identical? Just a thought

ndelangen:
Thank you for this PR @CyrusZei Would you be able to change this so both are preserved but distict?

Long term I think it'd be great if we could merge these 2 solutions into one somehow.
What do you think @Sambego @dump247 ?

Sambego:
That's something worth considering but this will probably also cause some issues.

Users who prefer one API or the other will remain with an unmaintained addon, unless we find a way to cleanly merge the 2 addons keeping both APIs. If so we should add a deprecation warning to our addons, giving users enough time to switch to the newly created/merged addon.

ndelangen:
@dump247 @Sambego can you give a pitch on why your API is (in your opinion) the better one?

I think we should consolidate on 1 API. Mainly to save ourselves the maintenance nightmare of having to maintain too many usecases in code, but also to keep distraction low for end-users.

I'm curious what brought you to create the API you did, so we can access which one we should pick going forward. Does that sounds reasonable?

Sambego:
Hey @ndelangen sure, here are some if the reasons why I like the API of `@sambego/storybook-state`.

- It uses the component based approach which we know and love from React
- It has an available decorator to use with the storybook `addDecorator` method of preferred
- It passes the state as props to the components being used inside the stories. This allows for an easy "plug and play" experience if the components use Reacts passing of props architecture.
- It allows for extra flexibility by letting you parse the state before passing it as props in the story.
- It allows for extra flexibility by accepting functions next to components to pass along props in a more controlled way.

This plugin is based upon my needs at the time of creation. Some improvements based on issues and pull requests from people in the community are also added since the first version.

Some extra points in favor if my library are that the code is fully unit tested and I've setup a Travis CI pipeline to test the code before merging any future pull requests. Since today we also have Typescript typings, thanks to a community contribution.

ndelangen (approved):

Sambego:
Hey @ndelangen I've joined your discord channel, feel free to get in contact whenever you have some time :)

PR #4619, opened 2018-10-29
chore(core): update css-loader to latest

StephM-W:
Issue:
css-loader 1.0.0 depends on vulnerable versions of lodash packages

## What I did
updated to latest css-loader that addresses these transitive dependency issues

## How to test
yarn test


igor-dv (approved):

lib/core/package.json
```diff
@@ -47,7 +47,7 @@
     "commander": "^2.19.0",
     "common-tags": "^1.8.0",
     "core-js": "^2.5.7",
-    "css-loader": "^1.0.0",
+    "css-loader": "^1.0.1",
```
igor-dv: Looks like `css-loader` is used in some other packages too. Can you update it there as well?

Stephanemw: Sure thing, happy to!

PR #4622, opened 2018-10-30
Resources addon to dynamically add/remove css

nm123github:
Issue: I think there might be use-cases where folks want to quickly hop between different css resources at runtime. Here is an addon that will help with that.

## What I did
demo


nm123github:
This might seem similar to recent [resources-addon PR](https://github.com/storybooks/storybook/pull/4152) but is slightly different. This one allows you to switch between css files/themes predefined at compile time. This was a usecase that folks had asked and might be a good idea to deliver just that in this addon.

The other addon (which is more of a codeinjector-addon) will be a different PR, (most likely will update the previous one).

addons/cssresources/src/CssResourcePanel.js
```diff
@@ -0,0 +1,155 @@
+import { document, DOMParser } from 'global';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { monoFonts } from '@storybook/components';
+import styled from '@emotion/styled';
+import 'pretty-checkbox/dist/pretty-checkbox.css';
+
+import css from 'react-syntax-highlighter/languages/prism/css';
```
igor-dv: There some performance improvement @Hypnosphi did in storysource addon for the `react-syntax-highlighter`. Take a look there

addons/cssresources/src/CssResourcePanel.js
```diff
@@ -0,0 +1,155 @@
...
+            <div style={{ padding: 10 }}>
+              <div style={{ 'margin-bottom': 10 }} className="pretty p-svg p-curve p-smooth">
+                <input
+                  id={`cssresources${i}`}
+                  type="checkbox"
+                  checked={picked}
+                  onClick={this.handleChange.bind(this, i)}
+                />
+                <div className="state p-success" style={{ 'font-size': '14px' }}>
+                  <svg className="svg svg-icon" viewBox="0 0 20 20">
```
igor-dv: Can you extract this svg to a separate component?

addons/cssresources/src/index.js
```diff
@@ -0,0 +1,23 @@
+import addons, { makeDecorator } from '@storybook/addons';
+
+export const withCssresources = makeDecorator({
```
igor-dv: withCssresources => withCssResources

igor-dv (approved):

ndelangen (approved):

PR #4623, opened 2018-10-30
[documentation] Update URL to babel config in default-config

quetzaluz:
Issue: Documentation refers to a file which no longer exists

## What I did
- Update documents to refer to similar dev config file. I am open to recommendations on how to revise this in a different way. But for now linking to a 404 is not helpful in this documentation and this is a potential fix.
## How to test
- Evaluate the documentation update update


shilman (approved):
LGTM! Thanks! 🥇

PR #4649, opened 2018-10-30
Fix a bug with buildStaticStandalone resolving too early

transitive-bullshit:
Issue: `buildStaticStandalone` returns a `Promise` for its build completing, but that Promise doesn't take into account the actual compiler work, meaning that dependents that want to create static builds have no way of reliably knowing when the build completes.

## What I did

I made a fairly simple change to `build-static.js` that only resolves the `buildStaticStandalone` async result once the build completes or errors.

## How to test

You can reproduce this via a standalone static build like this one:

```js
const build = require('@storybook/react/standalone')
const packageJson = require('./package.json')

const configDir = './.storybook'
const outputDir = './storybook-static'

build({
  mode: 'static',
  packageJson,
  configDir,
  outputDir,
  port: 60001,
  host: 'localhost'
}).then(() => {
  console.log('storybook build should be finished', outputDir)
}).catch((err) => {
  console.error('storybook build error', err)
})
```

Note that this is a bugfix for https://github.com/hydrateio/docz-plugin-storybook/pull/4

cc @mergebandit


igor-dv:
CI is oddly red, can you please merge from master?

transitive-bullshit:
@igor-dv thanks for the quick response; from what I can tell, my branch is sync'ed with latest `master`. I just pulled latest and merged again to be sure. The latest commit from master before this PR is `d33ac612f9e32dce715ab3c1a1d5ef452ed4488e` which matches the latest commit from master on this upstream repo.

I've never seen this type of [error](https://circleci.com/gh/storybooks/storybook/74361?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) before but CI seems to think that the `yarn.lock` file is dirty.

Not sure how to proceed from here but happy to make any changes you guys would like to see.

mergebandit:
FYI - it looks like this is failing due to the use of [yarn 1.9](https://github.com/yarnpkg/yarn/issues/6440) - the integrity entries are being removed from each of the entries in the `yarn.lock` file. Example:

image

igor-dv:
Probably CircleCI issue. TeamCity is passing. I will rerun CircleCI.

mergebandit:
FYI - the `scripts/repo-dirty-check.js` is what is failing, suggesting that you should run `yarn bootstrap --reset --core`. Running this on a clean checkout of this branch does not resolve this issue.

igor-dv:
This test is checking that yarn.lock is updated after some packages were added/changed. But since there were not any pkgs changes here, IMO it should not fail...

transitive-bullshit:
I'm guessing something changed on circleci's end with respect to the version of Yarn it is running. @igor-dv is a fresh circleci build passing on master if run today?

igor-dv:
It was a node issue (@Hypnosphi solved it in CI).

igor-dv (approved):

PR #4652, opened 2018-10-30
Allow for angular-devkit/core version aligned with Angular v7

emilio-martinez:
Issue:

## What I did

Set @storybook/angular's peer dependency range to allow for Angular ^7.0.0. There's a [test fixture for it](https://github.com/storybooks/storybook/blob/v4.0.0/lib/cli/test/fixtures/angular-cli-v7/package.json#L17) already. Additionally, I was able to successfully run and build Storybook v4 with a Angular CLI v7-generated project.

This change simply removes the peer dependency warning when installing @storybook/angular.

## How to test

There's a [test fixture for it](https://github.com/storybooks/storybook/blob/v4.0.0/lib/cli/test/fixtures/angular-cli-v7/package.json#L17) already.

### DIFF
--- a/app/angular/package.json
+++ b/app/angular/package.json
@@ -44,7 +44,7 @@
     "zone.js": "^0.8.26"
   },
   "peerDependencies": {
-    "@angular-devkit/core": "^0.6.1",
+    "@angular-devkit/core": "^0.6.1 || >=7.0.0",
     "@angular/common": ">=6.0.0",
     "@angular/compiler": ">=6.0.0",
     "@angular/core": ">=6.0.0",


app/angular/package.json
```diff
@@ -44,13 +44,13 @@
...
   "peerDependencies": {
-    "@angular-devkit/core": "^0.6.1",
-    "@angular/common": ">=6.0.0",
-    "@angular/compiler": ">=6.0.0",
-    "@angular/core": ">=6.0.0",
-    "@angular/forms": ">=6.0.0",
-    "@angular/platform-browser": ">=6.0.0",
-    "@angular/platform-browser-dynamic": ">=6.0.0",
+    "@angular-devkit/core": "^0.6.1 || >=7.0.0",
+    "@angular/common": ">=6.0.0 || >=7.0.0",
```
igor-dv: Doesn't `>=6.0.0` range include `>=7.0.0` as well ? IMO we  need to fix only `@angular-devkit/core`

emilio-martinez: Good point! Done

igor-dv (approved):

Issue #4641, opened 2018-10-30
@storybook/angular shouldn't depend on zone.js


emilio-martinez:
**Describe the bug**
Currently, @storybook/angular lists zone.js within its dependencies; however, (as far as I can tell) in practice @storybook/angular doesn't actually depend on it. Of course, [Angular does](https://angular.io/guide/npm-packages#support-packages) but perhaps it's worth removing it as a @storybook/angular dependency to avoid conflicts.

**To Reproduce**
See https://github.com/storybooks/storybook/blob/v4.0.0/app/angular/package.json#L44.

**Expected behavior**
 @storybook/angular shouldn't depend on zone.js if it doesn't use it.

**Screenshots**
N/A

**Code snippets**
N/A

**System:**
 - OS: MacOS
 - Device: Macbook Pro 2017
 - Browser: Chrome
 - Framework: Angular
 - Addons: none
 - Version: 4.0.0

**Additional context**
N/A

kroeder:
@igor-dv @storybook/angular does not depend on it. Wasn't it core or something? We already had a discussion about zone.js

emilio-martinez:
Sorry, re-read my description and realized it was very odd/poorly written 😅updated.

igor-dv:
It's used in polyfills

https://github.com/storybooks/storybook/blob/master/app/angular/src/client/preview/angular-polyfills.js#L53

Do we really need these polyfills, though ?

kroeder:
Well, this time zone.js isn't a dependency of a dependency (like webpack is just a dep of angular core). zone.js is a direct dep of angular - maybe we can put zone.js as peer dep and dev dep for development this time?

emilio-martinez:
Peer dependency makes sense. The way I see it, Storybook should be pulling the version of zone.js already being used in the host Angular project anyway, unless there's a specific task Storybook is using zone.js for that differs from the rest of the project.

igor-dv:
Wanna PR this ?

emilio-martinez:
For sure. Opened!

PR #4663, opened 2018-10-31
Set zone.js as a peer and dev dependency

emilio-martinez:
Issue: #4641

## What I did
Set zone.js as a peer and dev dependency for @storybook/angular.

## How to test
Behavior shouldn't change since Storybook should be pulling the version of zone.js already being used in Angular projects anyway.


app/angular/package.json
```diff
@@ -51,6 +50,10 @@
     "@angular/forms": ">=6.0.0",
     "@angular/platform-browser": ">=6.0.0",
     "@angular/platform-browser-dynamic": ">=6.0.0",
-    "babel-loader": "^7.0.0 || ^8.0.0"
+    "babel-loader": "^7.0.0 || ^8.0.0",
+    "zone.js": "^0.8.26"
+  },
+  "devDependencies": {
+    "zone.js": "^0.8.26"
```
igor-dv: @kroeder should we add zone.js in cli?

kroeder: You mean when running `sb init`? No, there's no angular without zone.js installed. `ng new my-app` already does add zone.js, angular does not work without it - We do not need to take care of this

https://medium.com/@MertzAlertz/what-the-hell-is-zone-js-and-why-is-it-in-my-angular-2-6ff28bcf943e

> Let’s do an experiment. Go and take your favorite Angular 2 app, and don’t include zone.js. See what happens, I’ll wait here. Notice anything interesting? None of the bindings are working!

emilio-martinez:
CLI tests seem to have tripped while testing Vue
> Module build failed (from ./node_modules/vue-loader/index.js):
> TypeError: Cannot read property 'vue' of undefined

igor-dv (approved):

PR #4666, opened 2018-10-31
Delete sfc_vue fixture to fix master

igor-dv:
Issue: 

Master is red because of the old `vue` fixture that already does not work with the new sb@cli version.
Any new `vue` cli tests are welcomed.


Hypnosphi:
Did we intentionally break that usecase?

igor-dv:
Definitely, agree with you, removing this is not a real solution. I feel that vue needs a bit more maintenance love. 
Anyway, having this failing is confusing other PRs. The deletion could be alway reverted when anyone will invest some time in solving the real problem.
But controlling this from CI is much better 👍

Hypnosphi:
https://github.com/storybooks/storybook/pull/4671

Issue #4007, opened 2018-08-12
@storybook/addon-viewport defaultViewport not working


tquinlan1992:
### Bug or support request summary

Each time I reload the web browser the viewport always starts at responsive when I was hoping the setup I have in .storybook/config.js should make it default to the `iphone6` viewport.

### Steps to reproduce

package.json
```
...
  "scripts": {
    "build": "build-storybook -c .storybook -o .out",
     "start": "start-storybook -p 9001 -c .storybook",
  },
  "devDependencies": {
    "@storybook/addon-actions": "^4.0.0-alpha.16",
    "@storybook/addon-centered": "^4.0.0-alpha.16",
    "@storybook/addon-info": "^4.0.0-alpha.16",
    "@storybook/addon-links": "^4.0.0-alpha.16",
    "@storybook/addon-storysource": "^4.0.0-alpha.16",
    "@storybook/addon-viewport": "^4.0.0-alpha.16",
    "@storybook/addons": "^4.0.0-alpha.16",
    "@storybook/react": "^4.0.0-alpha.16",
    "typescript": "3.0.1"
}
...
```

.storybook/addons.js
```
import '@storybook/addon-storysource/register';
import '@storybook/addon-actions/register';
import '@storybook/addon-links/register';
import '@storybook/addon-viewport/register';
```

.storybook/config.js
```
import { configure, addDecorator } from '@storybook/react';
import centered from '@storybook/addon-centered';
import { configureViewport } from '@storybook/addon-viewport';

configureViewport({
  defaultViewport: 'iphone6'
});
 
addDecorator(centered);

function loadStories() {
  require('../stories/index.tsx');
  require('../stories/App/index.tsx');
  // You can require as many stories as you need.
}

configure(loadStories, module);
```

.storybook/webpack.config.js
```
const path = require("path");
const TSDocgenPlugin = require("react-docgen-typescript-webpack-plugin");


module.exports = (baseConfig, env, config) => {
  config.module.rules.push({
    test: /\.(ts|tsx)$/,
    loader: require.resolve("awesome-typescript-loader")
  },
  {
    test: /\.tsx?$/,
    include: [
      path.resolve(__dirname, "../stories")
    ],
    loaders: [
      {
        loader: require.resolve('@storybook/addon-storysource/loader'),
        options: { parser: 'typescript' }
      }
    ]
  });
  config.plugins.push(new TSDocgenPlugin()); // optional
  config.resolve.extensions.push(".ts", ".tsx");

[...]

PR #4683, opened 2018-11-01
Addon Viewport - fix "defaultViewport" configuration

igor-dv:
Issue: #4007

## What I did
1. `configureViewport` was failing if "viewports" field was not defined
2. `defaultViewport` was not persistent with navigation

P.S. Navigating between stories will always reset to the "defaultViewport" value, IMO it's a strange behavior when manually changing the viewport and navigating to a different story.

P.P.S the branch is called `addon-jest/fix-configureviewport` but I meant `addon-vieport/fix-configureviewport`


shilman (approved):
LGTM!

PR #4702, opened 2018-11-03
feat(storiesNav): deep theming for stories nav panel

mbellary-chwy:
## Feature:
deep theming for stories nav panel

## What I did

- deep theming for stories nav panel
- updating README file with all the deep theme variables supported
- updating `HeadingLink` component to support html strings for `brand`. So that `image` can be used for `brand`
- placing `props` to last in `chevronRight` component, so that it overrides the theme props passed


lib/ui/src/modules/ui/components/stories_panel/stories_tree/tree_decorators.js
```diff
@@ -2,35 +2,50 @@ import { decorators } from 'react-treebeard';
...
 import MenuItem from '../../menu_item';
 import treeNodeTypes from './tree_node_type';
 import { highlightNode } from './tree_decorators_utils';
 
 function noop() {}
 
-function ToggleDecorator({ style }) {
+function ToggleDecorator({ style, theme }) {
   const { height, width, arrow } = style;
+  const { treeArrow } = theme;
```
ndelangen: This looks messy, but it doesn't matter too much, since that's all getting cleanup up with the UI overhaul.

ndelangen (approved):

lib/ui/src/modules/ui/components/stories_panel/stories_tree/tree_decorators.js
MansoorBashaBellary: Sure. Lets cleanup with UI overhaul

Issue #4274, opened 2018-10-03
[3.4.11 - react - info - a11y] Stories render w/ page scrolled down instead of @ page start


nicholaai:
### Bug or support request summary

This problem is with storybook react / addon-info / addon-a11y. I'm also using styled-components v4, but I've adjusted my code to use a plain old button and this problem persists. 

When switching to a story (named Component Summary), I expect for the viewport to be at the top of the screen. Instead, the story is in a scrolled position, showing the lower half of the story.

### Screenshots

When clicking on a story I expect to be greeted by the following story:

![screen shot 2018-10-03 at 11 16 56 am](https://user-images.githubusercontent.com/8497422/46430504-f4bf3f80-c6fd-11e8-83e2-79fe3d6a0fff.png)

Instead, page loads with the story scrolled down  :
![screen shot 2018-10-03 at 11 17 13 am](https://user-images.githubusercontent.com/8497422/46430514-fab52080-c6fd-11e8-854a-29b2df6fbe86.png)

I found #3326 and #3342, which are kind of related, but not exactly what my issue is. I've check, and 3.4 also has those changes to the codebase, even tho it states that the changes are fixed in alpha version.

I'm currently using
```js
setDefaults({
    inline: true
})
```
to have my component's render inline with my story info. 

If I switch inline to `false` and click the `show info` a couple things happen

The initial render of the story is in a scrolled down position. I've tested this by narrowing the height of my window.

Expected:
![screen shot 2018-10-03 at 11 42 15 am](https://user-images.githubusercontent.com/8497422/46431797-9b590f80-c701-11e8-8438-b0dc28a7255f.png)

Actual:
![screen shot 2018-10-03 at 11 44 17 am](https://user-images.githubusercontent.com/8497422/46431858-b88dde00-c701-11e8-8198-3484dfde5dae.png)

If I click the `show info` button, the button the story renders as you would expect (non-scrolled).

![screen shot 2018-10-03 at 11 40 19 am](https://user-images.githubusercontent.com/8497422/46431653-37cee200-c701-11e8-807f-b928892700e4.png)

If I remove check117 by removing: 
```js
addDecorator(checkA11y);
```
then all stories render as expected. 

### Code
My config: 
```js
import * as React from 'react';
import { normalize } from 'polished';
import { configure, addDecorator } from '@storybook/react';
import { setDefaults } from '@storybook/addon-info';
import { checkA11y } from '@storybook/addon-a11y';
[...]

PR #4704, opened 2018-11-03
Better a11y

hipstersmoothie:
Issue: #4274

## What I did

When using info addon in `inline` mode with the a11y addon the page would jump to the bottom. This is because of axe. For it to work properly it needs to scroll to the bottom of the page. You can pass an option to scroll back but that would would look bad too.

* add id to story root in info addon
* in a11y addon check for info wrapper and use that to run axe if present
* just do axe on root element's children cause we don't care about the root's accessibility

if the content is long the page will still scroll to the bottom, but this mostly solves the problem. I have tested this in my own stories and it work for components in the viewport.

## How to test

Is this testable with Jest or Chromatic screenshots? No.


igor-dv (commented):
`story-root` id sounds me too generic to be in info addon. Also having `a11y` know about the inner implementation of `info` addon, makes them coupled... 

🤔 can we think of a more generic solution?

hipstersmoothie:
> story-root id sounds me too generic to be in info addon.

how about `.info-story-root`? That would be more specific to the into addon.

> Also having a11y know about the inner implementation of info addon, makes them coupled...

It inherently has to be coupled. The problem with running axe on the whole page is that is has to scroll no matter what. It creates a pretty bad user experience. If the user is using `inline: true` the whole page is displayed. So the really does mean that the a11y addon needs to know the space in info where the story code actually lies.

> 🤔 can we think of a more generic solution?

The was my attempt at a more generic implementation. The id on the root of the story in the info addon doesn't callout the a11y addon in particular. This id could be the target of other plugins or even the user targeting it to style it. So in my opinion it already is pretty generic. I also believe it to be a nice feature that the info addon exposes where it renders the story code. 

I am open to suggestions but for me this accomplishes:

- few code changes
- low likelyhood for bugs
- generic

what other way would you have the info addon expose the story code div? @igor-dv

ndelangen:
Long term I think this is the wrong approach, but for right now, I don't have strong objections really.

Having a preferred and fallback id selector for addon-a11y doesn't sound terrible, and adding an extra id to the html of addon-info is fine too.

Long term we want to move away from addon-info all together and move the documentation people are doing in addon-info to an addon that renders in the manager.

That way the story can render in isolation.

ndelangen:
If you change the id to `info-story-root` this should be good IMHO. Do you agree @igor-dv ?

Would you like to help with the docs in the manager approach @hipstersmoothie ?

igor-dv:
If it solves currently the problem I would say it's ok. What about other addons that may wrap a story, though?

ndelangen (approved):

hipstersmoothie:
Nice!

PR #4710, opened 2018-11-04
fix(react-native): prevent the Navigation bar to stay underneath the keyboard

estevaolucas:
Issue: N/A

## What I did

Wrapping react-native `<OnDeviceUI>` with a [`<KeyboardAvoidingView>`](https://facebook.github.io/react-native/docs/keyboardavoidingview) prevent the bottom navigation bar to stay underneath the keyboard when a TextInput field is focused. It's an issue only on iOS.

Solution tested on Android and iOS.

before:
before

after:
after

## How to test

Does this need a new example in the kitchen sink apps? `<TextField>` component created to demo this fix (45d81aa6d1d96e0f3cbe8fb81c06aef74e4f8f80. In case it's not needed to have this new component and story, just let me know that I can remove it without any problem)


Gongreg:
Hey @elucaswork. I am a little bit afraid of just wrapping everything with KeyboardAwareView, since then the preview size changes when opening the keyboard. And then how users should play around with components which are already wrapped with KeyboardAwareView?

Maybe it is okay if keyboard hides the nav bar?
I thought the issue is that keyboard hides story list (navigator panel).

I need your opinion on this one :)

elucaswork:
@Gongreg good point. Nested `KeyboardAwareView` can be a problem.

The problem happening in my project right now: even adding a decorator with `KeyboardAvoidingView`, won't make the navigation bar visible because `AbsolutePositionedKeyboardAwareView` height will be 100%, ignoring the keyboard height.
 
```
stories
  .addDecorator(getStory => (
    <KeyboardAvoidingView
      behavior={Platform.OS === "android" ? null : "position"}
    >
      {getStory()}
    </KeyboardAvoidingView>
  ))
```

I created this PR with the idea of avoiding components with `TextInput` to make user getting stuck in Storybook without the option of changing the visible storybook or close the keyboard.

It's not a problem for regular `TextInput`, with `default` `keyboardType` or `email-address`, where the user has a return key on the keyboard. Because with the return key, the user can close the keyboard by tapping on it.

The problem only happens when the keyboard doesn't have a return key (when `keyboardType` is `number-pad`, `decimal-pad`, `numeric`
, `phone-pad`)

simulator screen shot - iphone x - 2018-11-05 at 12 56 28

What I can try to do, to properly fix this issue (which is an edge-case) is change remove `KeyboardAvoidingView` and change `AbsolutePositionedKeyboardAwareView` to be aware of the keyboard, by decreasing its size when the keyboard is visible.

elucaswork:
@Gongreg thanks for your thoughts here.

> I would prefer to leave keyboard handling to the story itself if possible, since storybook shouldn't interrupt to reproduce components same way as they are in the app.

Makes sense for me. I'm gonna close it. It's a very edge case issue. And in case later I come with a better approach, I can reopen/open a new PR. 👍

Issue #4681, opened 2018-11-01
Support custom `react-scripts` packages


mrmckeb:
**Is your feature request related to a problem? Please describe.**
The ability to automatically use `react-scripts` is great. However, it only supports `react-scripts` named `react-scripts`. 

**Describe the solution you'd like**
Facebook supports the flag `--scripts-version` when using CRA, which they then use to set up CRA. I can see that storybook is hardcoded to this. Script versions can have different package names than `react-scripts`, but often (like my package) don't change much and work fine with storybook.

**Describe alternatives you've considered**
I can rename my package to `react-scripts` as a workaround.

**Are you able to assist bring the feature to reality?**
Yes, I am happy to. I'm not sure which path to go down:
- Flag during startup
- Environment variable
- Try to understand the symlink path for react-scripts (in .bin)

igor-dv:
> Flag during startup

IMO it's not an ultimate solution

> Environment variable

Sounds better to me.

@chadfawcett what do you think ?

chadfawcett:
I think I'd prefer an environment variable over a flag, but I like the idea of referencing the symlink.

@mrmckeb Do you know if the `/.bin/react-scripts` path stays the same? Or does it also get renamed to `/.bin/react-custom-scripts`? I haven't used this feature myself, and couldn't get a clear answer from the issue discussions or docs.

mrmckeb:
@igor-dv @chadfawcett Thanks for your time. I'll create a quick PR now with the `symlink` solution and we can discuss. 

It's definitely `/.bin/react-scripts` which is helpful. CRA leaves the scripts the same in `package.json` too, so it still looks like standard CRA (such as `react-scripts build`).

mrmckeb:
We can move the conversation to #4712 if you would like. I've created an example there, resolving the path of the `react-scripts` package in use from `/.bin/react-scripts`.

PR #4712, opened 2018-11-04
Add support for custom `react-scripts` packages

mrmckeb:
Issue: #4681

## What I did
Added support for `react-scripts` packages with names other than `react-scripts`. 

## How to test
[Create a fork](https://medium.com/@denis.zhbankov/maintaining-a-fork-of-create-react-app-as-an-alternative-to-ejecting-c555e8eb2b63) of `react-scripts` with a different name (or use an existing), and use that for your `--scripts-version` when running `create-react-app`.


chadfawcett (approved):
Looks good to me!

igor-dv (approved):
We need some CLI test fixture for this IMO

ndelangen (approved):

mrmckeb:
Certainly @igor-dv and @ndelangen, I'll take a look at doing this over the weekend. Hopefully it'll be fairly straight-forward to test.

And thanks for your review too @chadfawcett.

igor-dv:
I've played a bit with this in #4836 and It doesn't find for me the `react-scripts` pkg in the `cra-kitchen-sink` example. Might It be related either to monorepo or windows?

igor-dv:
When I am debuggin this:

```js
let reactScriptsPath;

function getReactScriptsPath() {
  if (reactScriptsPath) return reactScriptsPath;
  const appDirectory = fs.realpathSync(process.cwd());
  const reactScriptsScriptPath = fs.realpathSync(
    path.join(appDirectory, '/node_modules/.bin/react-scripts')
  );
  reactScriptsPath = path.join(reactScriptsScriptPath, '../..');
  return reactScriptsPath;
}
```

in `cra-kitchen-sink`, I see that
```
reactScriptsScriptPath = 'root\examples\cra-kitchen-sink\node_modules\.bin\react-scripts'
```

Then

```
reactScriptsPath = 'root\examples\cra-kitchen-sink\node_modules'
```

And after when we are tring to use `isReactScriptsInstalled that does 

```
require(path.join(getReactScriptsPath(), 'package.json'));
```
path to `package.json` is wrong and doesn't point to the package.json in the `react-scripts`

mrmckeb:
And does it work for you outside of that environment @igor-dv? And are you using WSL? I'm not home now, but can take a look later.

igor-dv:
Even tests (cra-config) are failing locally for me, but passing in CI. Probably a windows thing =(

PR #4713, opened 2018-11-05
feat(react-native): persist last displayed story

estevaolucas:
Issue: N/A

## What I did

I've implemented this feature from the [react-native roadmap](https://medium.com/storybookjs/whats-new-in-storybook-4-0-react-native-741c7f481bbb
):
> Easily select initial story: We could use async storage to remember last selected story. It would help users who cannot use Hot Reload to have similiar functionality just by using Live Reload.

The last displayed store will be persisted just in case storybook is not running.

## How to test

Is this testable with Jest or Chromatic screenshots? N
Does this need a new example in the kitchen sink apps? N
Does this need an update to the documentation? N

### DIFF
--- a/app/react-native/src/preview/index.js
+++ b/app/react-native/src/preview/index.js
@@ -1,7 +1,7 @@
 /* eslint-disable react/no-this-in-sfc, no-underscore-dangle */
 
 import React from 'react';
-import { NativeModules } from 'react-native';
+import { AsyncStorage, NativeModules } from 'react-native';
 import parse from 'url-parse';
 import addons from '@storybook/addons';
 
@@ -11,6 +11,8 @@ import createChannel from '@storybook/channel-websocket';
 import { StoryStore, ClientApi } from '@storybook/core/client';
 import OnDeviceUI from './components/OnDeviceUI';
 import StoryView from './components/StoryView';
+
+const STORAGE_KEY = 'lastOpenedStory';
 
 export default class Preview {
   constructor() {
@@ -65,7 +67,7 @@ export default class Preview {
         const port = params.port !== false ? `:${params.port || 7007}` : '';
 
         const query = params.query || '';
-        const { secured } = params;
+        const { secured, shouldPersistSelection } = params;
         const websocketType = secured ? 'wss' : 'ws';
         const httpType = secured ? 'https' : 'http';
 
@@ -75,7 +77,7 @@ export default class Preview {
           url,
           async: onDeviceUI,
           onError: () => {
-            this._setInitialStory();
+            this._setInitialStory(shouldPersistSelection);
 
             setInitialStory = true;
           },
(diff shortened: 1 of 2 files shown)


Gongreg:
Hey, I believe this one should use a toggle in getStorybookUI. Maybe not everyone wants to use this behaviour. 

Also it would be great to have an option to pass in initial story as an option manually to getStorybookUI.

Gongreg:
Also I've just created an issue #4716 . Seems like I broke old behaviour in v4. It is a little bit relevant to this task.

elucaswork:
> Also it would be great to have an option to pass in initial story as an option manually to getStorybookUI.

@Gongreg I just added a `shouldPersistSelection` option. Please let me know what you think. 557d6f48254efa62d405a68b83091a1eefbe000f

> Hey, I believe this one should use a toggle in getStorybookUI. Maybe not everyone wants to use this behaviour.

I don't think I get it. Isn't the `shouldPersistSelection` enough in this case?

Gongreg:
Hey @elucaswork, I think I would even make this option true by default. This is a cool feature.

What I meant about separate option (like initial story) is so that users could control the initial story themselves out of storybook scope. Now they have to get the channel and emit the initial story. But it can be done separately, it was just an idea :)

In the long run we just have to make sure that there is this priority:

Server, manually setting up, reading from storage.

If user is running server, just allow the server to set the story, if not and user manully passes initial story, set to that story, if not try to read from storage and set it, and if not just open up the first story.

I hope the idea is clear :)

So to finish this PR I think just setting option to true by default should be enough.

app/react-native/src/preview/index.js
```diff
@@ -133,15 +135,22 @@ export default class Preview {
     channel.emit(Events.GET_CURRENT_STORY);
   }
 
-  _setInitialStory = () => {
-    const story = this._getInitialStory();
+  _setInitialStory = async (shouldPersistSelection: boolean) => {
```
Gongreg: I may be off but since when JS supports type definitions?

app/react-native/src/preview/index.js
```diff
@@ -133,15 +135,22 @@ export default class Preview {
...
+    let story = this._getInitialStory();
+
+    if (shouldPersistSelection) {
+      const value = await AsyncStorage.getItem(STORAGE_KEY);
```
Gongreg: I suggest to also have some kind of check to make sure that we don't get garbage from AsyncStorage, I had few cases where I didn't get what I expected

app/react-native/src/preview/index.js
elucaswork: oops. good catch. 🙄

elucaswork:
@Gongreg thanks. got it now. Nice idea. 

I just implemented `initialSelection` option here, in the same PR. let me know what you think about it.

```
getStorybookUI({
  initialSelection: { kind: 'Knobs', story: 'with knobs' },
})
```

Gongreg:
Looks great! :)

examples-native/crna-kitchen-sink/.babelrc
```diff
@@ -1,5 +1,5 @@
 {
-  "presets": ["babel-preset-expo"],
+  "presets": ["react-native"],
```
Gongreg: The only question I have what was changed here?

Gongreg (approved):

examples-native/crna-kitchen-sink/.babelrc
elucaswork: `npm run server` is broken right now because of this config. I just removed it from here. I'm to open a PR with it.

Gongreg: @elucaswork  Do you know why is it broken?

elucaswork: It was an issue on `expo` `30.0.1`

It fails with `npm run storybook`
screen shot 2018-11-07 at 8 40 55 am

But I just checked out that it's fixed on `31.0.4`. So #4728 will address it.

Gongreg:
@elucaswork it might be useful to update console logs to be more explicit about what is wrong.

PR #4720, opened 2018-11-05
Fix tests in node 11 by using `Array#sort` correctly

Hypnosphi:
Sorting algorithm has [changed](https://twitter.com/mathias/status/1036626116654637057) in latest V8, and it doesn't work with our improper usage (returning boolean instead of number) anymore

## How to test

```
nvm install 11
yarn test --core
```


igor-dv (approved):

Hypnosphi:
See also https://github.com/nodejs/node/pull/22754#issuecomment-423452575

PR #4752, opened 2018-11-08
React native on device package

Gongreg:
Issue: React Native storybook brings a lot of dependencies which are unnecessary for React Native itself. Also all of them are used only if you use storybook server. We had a discussion with @ndelangen and as a first step we want to release separate package which only has on device logic. 

## What I did

Created react-native-on-device package, created client-api package.


igor-dv:
Closing this in favor of #4942

PR #4758, opened 2018-11-09
TS Migration: addon-notes

kroeder:
Issue: N/A

## What I did

@ndelangen and I talked about doing a POC for TypeScript in Storybook

This POC needs to answer the following questions:
- [x] How does the transition from *.js to *.ts looks like for an example addon (I took addon-notes)
  - I had to configure a couple of things that are now available for the next packages we might move to `ts`
  - Renaming `*.js` to `*.ts` also might cause a couple of red lines. Mostly imports that must be used differently or type mismatches.
  - Rewriting the code base took me about 1-2 hours. Most of the time I worked on this I had issues with the initial configuration
- [x] CI must pass with everything else working as usual
  - I guess except `Danger` everything's green 🙂 
- [x] eslint must play along with tslint. They should not get in conflict
  - It seems that `tslint` was already configured. I have not noticed any weird linting errors
- [x] Performance of `yarn dev` should not get significantly slower as it is right now
  -  Just running `yarn dev:ts` is super fast, `yarn dev` is slow as usual ;)

## What I found out
- `tslint.json` already exists in the root directory, must check if it works
- `yarn dev:ts` already searches for `tsconfig.json` files inside all workspaces. If it found one, it compiles the project to `.js`
- `.tsx` files can be compiled to `.js` as well

## Some side notes
- The current implementation has room for e.g. outsourcing the interfaces in a seperate file or even re-use them in other files but since I'm not familiar with the actual implementation I decided to not make this as clean as possible in the POC
- Some things might be not correct because I am not familiar with react but I've added todos on everything I need to figure out


kroeder:
Currently working on a fix for imports

It should import from
`import { withNotes } from '@storybook/addon-notes'` 
but instead this happens
`import {withNotes} from '@storybook/addon-notes/dist/addon';`

addons/notes/src/public_api.ts
```diff
@@ -0,0 +1,3 @@
+export * from './addon';
+export * from './addon-react';
```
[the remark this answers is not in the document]
kroeder: It's actually `import { withNotes } from '@storybook/addon-notes'` or for react `import { WithNotes } from '@storybook/addon-notes'` - Kinda bad as well

addons/notes/package.json
```diff
@@ -30,6 +31,10 @@
     "marked": "^0.5.1",
     "prop-types": "^15.6.2"
   },
+  "devDependencies": {
+    "@types/jest": "^23.3.9",
+    "@types/react": "^16.7.1"
+  },
```
kroeder: Maybe these 2 @types should be moved into the root repo package.json as well? 
Otherwise they have to be installed for every typescript package

igor-dv: makes sense. Can we also configure them in the root `tsconfig`?

kroeder: I moved both types in the root package.json. I also removed prop-types, after adding typescript they were not used anymore. I also moved most of the tsconfig stuff to the root tsconfig

addons/notes/src/__tests__/index.ts
```diff
@@ -1,5 +1,5 @@
 import addons from '@storybook/addons';
-import { withNotes } from '..';
+import { withNotes } from '../addon';
```
igor-dv: this file should be renamed to index.test.ts as well. BTW, why rename `index` to `addon` at all?

addons/notes/src/addon.ts
```diff
@@ -16,7 +18,13 @@ export const withNotes = makeDecorator({
     const storyOptions = parameters || options;
 
     const { text, markdown, markdownOptions } =
-      typeof storyOptions === 'string' ? { text: storyOptions } : storyOptions;
+      typeof storyOptions === 'string'
+        ? {
+            text: storyOptions,
+            markdown: undefined,
```
igor-dv: Is this mandatory?

tsconfig.json
```diff
@@ -1,11 +1,19 @@
 {
   "compileOnSave": false,
   "compilerOptions": {
+    "baseUrl": ".",
+    "declaration": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "skipLibCheck": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
igor-dv: Will you fix the "any" things as part of the POC?

kroeder: I will refactor lots of stuff as soon as @ndelangen and whoever wants to join the discussion are happy with the stuff happening here 🙂

addons/notes/src/addon.ts
kroeder: my IDE said, they are missing - normally it should not be necessary.
I guess there is a better way, e.g. 
```? {
  text: storyOptions
} as StoryOptions```

There may be an even better way, I'll check this later

File: addons/notes/src/__tests__/index.ts (hunk 3 above)
Comment (kroeder) [reply addons/notes/src/__tests__/index.ts]: I did this when there was a index.ts and a react.ts 
I talked to Norbert and asked him what's the difference and that I find it kinda confusing for an outsider to really know what's going on

Now, we have just 1 file (react.ts was removed), this may not be necessary anymore but still: `index.ts` doesn't tell you anything about what this file does - let's discuss if you disagree 🙂

Comment (igor-dv) [reply addons/notes/src/__tests__/index.ts]: I can't argue against If it helps to dive into the code=) I just don't like unneeded code movements, and I do like the consistency across the code base =)

Comment (kroeder) [reply addons/notes/src/__tests__/index.ts]: Maybe some people should talk about a naming convention before we touch every addon? I mean we can leave them as they are - it's just a good moment to do stuff like this in one refactoring process

**Comment by kroeder** [comment]
@ndelangen I added comments to my initial post. Feel free to check the changes and ask me questions - or lets talk later this week 🙂

File: addons/notes/src/addon.ts [hunk 6]
```diff
@@ -1,8 +1,10 @@
 import addons, { makeDecorator } from '@storybook/addons';
-import marked from 'marked';
+import { MarkedOptions, parse as marked } from 'marked';
 
-function renderMarkdown(text, options) {
-  return marked(text, { ...marked.defaults, ...options });
+// todo why not just calling marked directly instead of wrapping it?
+// todo removed { marked.defaults, options } merge. I believe this was not necessary, have to check
```
apastuhov: There is no need in the wrapper or merging options. You use exposed `parse` attr.

[parse is actually an alias](https://github.com/markedjs/marked/blob/f8d556a1cbcab1e5c4eb0acea066316b48a9403c/lib/marked.js#L1610) and inside of `parse` it is [merging options](https://github.com/markedjs/marked/blob/f8d556a1cbcab1e5c4eb0acea066316b48a9403c/lib/marked.js#L1547)

So you can write just `import { parse as renderMarkdown } from 'marked';`

addons/notes/src/addon.ts
apastuhov: I did not found a class or interface of `StoryOptions`. Anyway, without specified options it will fail because it expect that values from that object. You can define it as default values, like:
```
const { text, markdown = undefined, markdownOptions = undefined } = { text: "" };
```
Here is an example in a [playground](http://www.typescriptlang.org/play/#src=declare%20var%20storyOptions%3A%20any%3B%0D%0A%0D%0A%0D%0Aconst%20%7B%20text%2C%20markdown%2C%20markdownOptions%20%3D%20undefined%20%7D%20%3D%0D%0A%20%20%20%20%20%20typeof%20storyOptions%20%3D%3D%3D%20'string'%0D%0A%20%20%20%20%20%20%20%20%3F%20%7B%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20text%3A%20storyOptions%2C%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%0D%0A%20%20%20%20%20%20%20%20%3A%20storyOptions%3B)

addons/notes/src/register.tsx
```diff
@@ -1,18 +1,44 @@
...
-  constructor(...args) {
-    super(...args);
+export class Notes extends React.Component<NotesProps, NotesState> {
+
+  // todo check how react typescript does this, especially unmounted
+  stopListeningOnStory: any; // todo check correct definition
+  unmounted: any; // todo check correct definition
+
+  constructor(...args: any) {
+    super(args /* todo: This was ...args before, check if this was a bug */);
```
apastuhov: Typescript is _smart_, so:
```
super(args);      // will be transformed to _super.call(this, args)
super(...args);   // will be transformed to _super.apply(this, args)
```
So your change will cause a bug. If you want to use spread - write 
```
constructor(...args: any[])
```
Anyway - there is no need to define a constructor with React, see my comment bellow

addons/notes/src/register.tsx
```diff
@@ -1,18 +1,44 @@
...
     this.state = { text: '' };
```
apastuhov: There is no need to define the state in the constructor with React, or bind methods. You can just write it as a property:
```
public state: NotesState = { text: '' };
// ...
private onAddNotes = (args) => { /* ... */ }
```
instead of:
```
constructor(...args: any[]) {
   this.state = { text: '' };
   this.onAddNotes = this.onAddNotes.bind(this);
}
// ...
onAddNotes(args) { /* ... */ }
```
As result - usually there is no need to define constructor explicitly

tsconfig.json
apastuhov: In some cases you can use type `unknown` it is like `any` but type-safe.
[new `unknown` type in TS 3+](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type)

apastuhov (changes requested):
Nice start! I've added a few comments about React-TS specific stuff.

addons/notes/src/index.ts
```diff
@@ -16,7 +12,13 @@ export const withNotes = makeDecorator({
...
 
     const { text, markdown, markdownOptions } =
-      typeof storyOptions === 'string' ? { text: storyOptions } : storyOptions;
+      typeof storyOptions === 'string'
+        ? {
+            text: storyOptions,
+            markdown: undefined,
+            markdownOptions: undefined,
+          }
+        : storyOptions;
```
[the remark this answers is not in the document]
apastuhov: You are right. You can leave it as it exists. Also, do not lint TS code with ESLint (there are a lot of small conflicts like this), and update configuration of your IDE 😉

tsconfig.json
apastuhov: Yes, it is better not to use `any`.
And for `unkown is not defined` - maybe you have an old version of TS installed localy? Or made a typo and defined it as a variable and not a type?

addons/notes/src/register.tsx
```diff
@@ -1,18 +1,44 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
 import addons from '@storybook/addons';
 
 import styled from '@emotion/styled';
 
+interface NotesChannel {
```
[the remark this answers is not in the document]
kroeder: Discussed this in Discord with igor: We leave this as is and migrate @storybook/channels next, then fix this as well

tsconfig.json
kroeder: `noImplicitAny` is true again

addons/notes/src/addon.ts
kroeder: Done

addons/notes/src/index.ts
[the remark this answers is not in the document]
kroeder: I know that 😄 After switching to my notebook it seems the latest version of webstorm is causing this issue.

igor-dv (approved):
Lgtm

kroeder:
As discussed here https://stackoverflow.com/questions/41746028/proptypes-in-a-typescript-react-application#comment80415295_43187969 I should add prop types again because mom-typescript consumers don't profit from definition files

apastuhov (approved):
LGTM, otherwise there is one small comment to improve.

Issue #4587, opened 2018-10-26
Add TypeScript support via Babel 7


GeeWee:
Seeing as adding TypeScript support via Babel 7 is getting a lot of traction (see e.g. https://github.com/facebook/create-react-app/pull/4837 for a high profile example), it would be nice for Storybook to support this possibility out of the box.

What I expect to be possible, is for me to add the proper values to my .babelRc and for Storybook to then be able to compile TypeScript to JS.

*Example .babelrc*
```json
{
  "presets": [
    ["@babel/preset-react", {"development": true}],
    ["@babel/preset-typescript"],
    ["@babel/env", { "modules": false }]
  ],
  "plugins": [
    "@babel/proposal-class-properties",
    "@babel/proposal-object-rest-spread"
  ]
}
```

However this does not work. Storybook only applies these rules to javascript files in the webpack configuration. You also have to create a corresponding webpack configuration file:
```js
module.exports = (storybookBaseConfig, configType) => {
	storybookBaseConfig.resolve.extensions.push('ts', 'tsx');
	storybookBaseConfig.module.rules[0].test = /\.(mjs|jsx?|tsx?)$/;
	return storybookBaseConfig;
};
```

Would it be possible to include the typescript files in the test and the extensions per default? I can't see any downside for people not using typescript, as nothing will change for them.

Itrulia:
I am also trying to add support for that via Babel. How did you get around the `Error: Requires Babel "^7.0.0-0", but was loaded with "6.26.3".` error?

I already installed @babel/core instead of babel-core

GeeWee:
(I updated the configuration above. I have it working with it, except for decorators which require an additional plugin)

Hypnosphi:
Great idea @GeeWee , PR is welcome

Everspace:
Trying to do this just yesterday with a create-react-app with Typescript enabled and encountered that baseurl isn't handled well, or is isn't compatible with create-react-app's required parameters (such as noemit).

mrmckeb:
Now that this is official, it would be great to add this out-of-the-box.

mrmckeb:
Just adding this resolved the issue for me.
```
module.exports = (baseConfig, configType, config) => {
	config.resolve.extensions.push('.ts', '.tsx');
	config.module.rules[0].test = /\.(mjs|jsx?|tsx?)$/;
	return config;
};
```

However, the docs suggest the following:
```
module.exports = (baseConfig, env, config) => {
[...]

PR #4763, opened 2018-11-10
Adding documentation on  CRA2 Integration

mohamedmansour:
When Babel 7 is used, usually within CRA2, it conflicts with the current Storyboard TypeScript integration since it relies on JSX to be `react`. Babel 7 relies it to be `preserve`. Added documentation so that we use Babel defaults.

- Updating documentation to support create-react-app 2+ (CRA2) official TypeScript support. 
- Updating the correct wording from `Typescript` to `TypeScript`

### DIFF
--- a/docs/src/pages/configurations/typescript-config/index.md
+++ b/docs/src/pages/configurations/typescript-config/index.md
@@ -1,11 +1,13 @@
 ---
 id: 'typescript-config'
-title: 'Typescript Config'
+title: 'TypeScript Config'
 ---
 
-This is a central reference for using Storybook with Typescript.
+This is a central reference for using Storybook with TypeScript.
 
-## Dependencies you may need
+## Setting up TypeScript with awesome-typescript-loader
+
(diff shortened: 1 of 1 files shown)


mohamedmansour:
@shilman  I believe we should remove the other documentation regarding, and just focus on Babel 7?

igor-dv (commented):
Thanks ! I am hesitating to add this into docs, though, because it's rational just to support TS by default like in CRA 🤔 CC #4587

igor-dv (approved):
NVM. Will remove it later, after the support is implemented =)

mohamedmansour:
@igor-dv do you mean if the app is TypeScript, storybook would be smart enough to imply it would be TypeScript? That would be super great

ndelangen (approved):

kushalmahajan:
If the doc was updated it's still giving the error. I'm using CRA2 and followed the guide under section babel-loader? Anything that should I do further?

PR #4765, opened 2018-11-10
Add-on actions: Specify sort function that compares number-like strings as number

notherno:
Issue:

If action logger displays array with 10 or more items, the order of keys is broken because `Array.sort()` 
 sorts an array alphabetical order by default.

(Ex: [0, 1, 2, 3, ..., 10, 11] is displayed as [0, 1, 10, 11, 2, ..., 9])

image

## What I did

Specify a sorting function that tries to parse object key as a number.

## How to test

I added a unit test and can be checked on an official Storybook at.

http://localhost:9011/?selectedKind=Addons%7CActions&selectedStory=All%20types&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Try `yarn build-packs` and run `yarn start`.

I ran on Chrome dev-console (iframe.html) with following script.

```
window.test = new Array(35).fill(null)
```

and show `window` to log test data.

image

### DIFF
--- a/addons/actions/src/components/ActionLogger/index.js
+++ b/addons/actions/src/components/ActionLogger/index.js
@@ -7,6 +7,18 @@ import { ActionBar, ActionButton } from '@storybook/components';
 
 import { Actions, Action, Wrapper, InspectorContainer, Countwrap, Counter } from './style';
 
+const castIfNumber = subject => {
+  const num = Number(subject);
+  return Number.isNaN(num) ? subject : num;
+};
+
+const sortObjectKeys = (a, b) => {
+  if (a === b) {
+    return 0;
+  }
+  return castIfNumber(a) < castIfNumber(b) ? -1 : 1;
+};
+
 const ActionLogger = withCSSContext(({ actions, onClear }, { theme }) => (
   <Wrapper>
     <Actions>
@@ -16,7 +28,7 @@ const ActionLogger = withCSSContext(({ actions, onClear }, { theme }) => (
           <InspectorContainer>
             <Inspector
               theme={theme.addonActionsTheme || 'chromeLight'}
-              sortObjectKeys
+              sortObjectKeys={sortObjectKeys}
               showNonenumerable={false}
               name={action.data.name}
               data={action.data.args || action.data}
--- a/addons/actions/src/components/ActionLogger/index.js
+++ b/addons/actions/src/components/ActionLogger/index.js
@@ -2,22 +2,10 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import Inspector from 'react-inspector';
 import { withCSSContext } from '@emotion/core';
-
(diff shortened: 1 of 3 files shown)


rhalff:
The fix seems good, however I think it's patching an underlying bug somewhere.

Can you determine whether the data property correctly contains the array you are providing?
Or is `window.test` already an object when it is being passed in.

```js
<Inspector
  theme={theme.addonActionsTheme || 'chromeLight'}
  sortObjectKeys={sortObjectKeys}
  showNonenumerable={false}
  name={action.data.name}
  data={action.data.args || action.data} <-- is window.test still an array?
```
I currently do not have a storybook development environment setup.

I do not see the same behavior here: http://xyc.github.io/react-inspector/ which further indicates there is a bug in reviving arrays.

notherno:
@rhalff
Thank you for your reply.

It would be a lack of explanation of my part.

This behavior is due to the prop of `react-inspector`, `sortObjectKeys` (which takes `boolean` or `function`).

If this prop is given, it sorts keys (`Object.getOwnPropertyNames(data)`) of an object and **also an array**.

https://github.com/xyc/react-inspector/blob/fa337f48d1d1b63757970f35032f37b328673be0/src/object-inspector/ObjectInspector.js#L35-L39

This part says that:

* If `sortObjectKeys` is `undefined` or `false`, the order of `Object.getOwnPropertyNames(data)` is used to define an order of keys.
* If `true` is given, it sorts keys with `sort()` to display.
* If a `function` is given, that function is passed to `sort()` and you can customize the order of keys.

The playground you referred to doesn't use `sortObjectKeys` prop, therefore original key order of  `Object.getOwnPropertyNames(data)` is adopted and this behavior doesn't appear.

https://github.com/xyc/react-inspector/blob/270e53efbb98be441cf49c84ce38907265992450/src/CodePlayground.js#L115

To fix an order, there are 2 choices:

1. Stop using `sortObjectKeys` to use a natural order of keys
2. Customize the key order with a sorting function

Using a natural order breaks a current behavior of actions add-on, so I implemented a sorting function.

notherno:
Finally, I found it more proper that `react-inspector` should take original key order of arrays regardless of `sortObjectKeys` prop.

So I'll make a PR for `react-inspector` to fix this issue.

rhalff:
@notherno thanks a lot for the thorough analysis, fixing react-inspector seems to be the best option to me also. I'll keep this PR open just in case.

notherno:
I've just submitted a PR.

https://github.com/xyc/react-inspector/pull/66

ndelangen:
@notherno @rhalff Can we close this PR, if the fix is going to be in the underlaying lib?

rhalff:
Closed in favor of https://github.com/xyc/react-inspector/pull/66

PR #4773, opened 2018-11-12
Vue / Knobs - optimize on force render

igor-dv:
Issue: #4516

## What I did
I've added the optimization that should partially address the #4516 problem.

### This solves:
In case the knobs are bounded to `data()` function:

```js
storiesOf('Addon|Knobs', module)
  .addDecorator(withKnobs)
  .add('Simple', () => {
    const name = text('Name', 'John Doe');
    const age = number('Age', 44);

    return {
      components: { SimpleKnobExample },
      template: '<simple-knob-example :name="name" :age="age" />',
      data() {
        return { name, age };
      },
    };
  })
```

we can easily extract this data and replace it in the existing story

### This doesn't solve
The case where the story returns only template:

```js
storiesOf('Addon|Knobs', module)
  .addDecorator(withKnobs)
  .add('All knobs', () => {
    /*a ton of knobs here*/
    return {
      template: `
          <div style="border:2px dotted ${colour}; padding: 8px 22px; border-radius: 8px">
            <h1>My name is ${name},</h1>
            <h3>today is ${new Date(today).toLocaleDateString('en-US', dateOptions)}</h3>
            <p>${stockMessage}</p>
            <p>Also, I have:</p>
            <ul>
              ${items.map(item => `<li key=${item}>${item}</li>`).join('')}
            </ul>
            <p>${salutation}</p>
          </div>
        `,
    };
```

I don't know how to force update the existing story with this template without recreating the story 🤷‍♂️ 

CC, @y-nk @pksunkara, would like to hear your thoughts if there are other use-cases when we can just update the component and not destroy it.


igor-dv:
Example:

https://deploy-preview-4773--storybooks-vue.netlify.com/?selectedKind=Addon%7CKnobs&selectedStory=Simple&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

app/vue/src/client/preview/render.js
```diff
@@ -1,12 +1,49 @@
...
+
+  if (!data) {
+    return false;
+  }
+
+  Object.entries(data).forEach(([key, value]) => {
+    lastStory[key] = value;
+  });
+
+  lastStory.$forceUpdate();
```
y-nk: I don't believe this will be necessary since Vue as observers for all values of component internal states ; but that won't kill performances either for most of the case, so it's ok to leave it if you feel it's safer.

y-nk:
with permission of @igor-dv , i proposed to solve this. 

### What I did

First i went back to defining the most simple usecase, as it would be in React. Because Vue and React are different, we can't really translate, but what we know is that stories are returning components, and knobs are inserted into properties of those components (for the most part).

Then, I've been recoding the render function. The main idea is to keep the root container alive, and use an outside variable to pass props to the component's story. In order not to break HMR, we still need to recreate a new story everytime it refreshes, but only not if `forceRender` is `true`, which seems to link to when knobs are modified (it's `undefined` when you save the file and HMR reloads it).

igor-dv (commented):
I've fixed a few minor things. I can't approve since it's my PR

@ndelangen @pksunkara

y-nk (approved):

PR #4791, opened 2018-11-15
REFACTOR dev-serve to output less noise

ndelangen:
This moves the logging of the computation-time to build-dev.

This also removed the webpack stats output on build, if all is good there is 0 output.
If there are errors, they are logged.

### DIFF
--- a/lib/core/src/server/build-dev.js
+++ b/lib/core/src/server/build-dev.js
@@ -5,7 +5,7 @@ import favicon from 'serve-favicon';
 import path from 'path';
 import fs from 'fs';
 import chalk from 'chalk';
-import { logger } from '@storybook/node-logger';
+import { logger, colors } from '@storybook/node-logger';
 import fetch from 'node-fetch';
 import Cache from 'file-system-cache';
 import opn from 'opn';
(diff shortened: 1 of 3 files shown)


lib/core/src/server/dev-server.js
```diff
@@ -63,11 +68,19 @@ export default async function(options) {
...
 
   const iframeStartTime = process.hrtime();
+  let iframeTotalTime;
   const iframeCompiler = webpack(iframeConfig);
   const devMiddlewareOptions = {
-    noInfo: true,
     publicPath: iframeConfig.output.publicPath,
-    watchOptions: iframeConfig.watchOptions || {},
+    watchOptions: {
+      aggregateTimeout: 1,
```
kroeder: What's the reason behind setting this to 1?

Webpack itself sets it to 300 per default
https://webpack.js.org/configuration/watch/#watchoptions-aggregatetimeout

lib/core/src/server/build-dev.js
```diff
@@ -205,7 +209,11 @@ export async function buildDevStandalone(options) {
     console.log(
```
kroeder: Has not changed but I'm wondering why this is a console.log instead of `logger.info`

kroeder (commented):
@ndelangen I can't test it right now; I just reviewed the code 🙂

lib/core/src/server/dev-server.js
ndelangen: get faster HMR

lib/core/src/server/build-dev.js
ndelangen: because logger.info adds a green 'info' before what's logged

kroeder (approved):

Issue #3905, opened 2018-07-20
[Feature request] Add nesting of story tree in React Native


Gongreg:
### Work summary
It would help a lot to display story list as a full screen list instead of half screen. Since now the story names are sometimes too long to properly fit the sidebar.

Some people have requested functionality to display a nested tree instead of one level list. It shouldn't be too difficult to implement (I hope).

The main concern I have is that adding multiple level tree can be a performance issue in RN.

### Where to start

You need to setup CRNA example and try to work on App/OnDeviceUI files. 

### Acceptance criteria

* Stories displayed in full screen instead of a sidebar.
* Stories are nested.

### Who to contact

You can contact me or anyone from contributors in slack.

sesm:
@Gongreg did you try using a `/` separator in story names? It does the job for me. But alas, this feature is not documented.

sesm:
@Gongreg regarding nested stories, did you try to use a `/` separator in story names? It does the job for me. But alas this feature is not documented.

Gongreg:
Hey @sesm, it does work but it works differently than in browser. What I am thinking about is implementing full tree view just like in the browser.

![image](https://user-images.githubusercontent.com/3867635/43571023-87331a5c-9644-11e8-9189-21c23ab58d8d.png)

Something like this but inside the simulator view.

ndelangen:
@Gongreg We could decide to do one of these things:

1. We re-implement the treeview in RN
   We can likely re-use a ton, because of the way I wrote it 
   (separating the visual from state)
2. We write the UI of the treeview in such a way it's renderable in RN
3. We embed the treeview in a webview in RN and communicate with it over a channel
   We'd re-use the code in it's entirety and complexity is lowest

Isakdl:
Any progress on this?

Gongreg:
Hey, there was no progress from my side. @ndelangen, would it be possible to create a bundle with treeview now? Or the new treeview is still WIP?

lancesnider:
> @Gongreg regarding nested stories, did you try to use a `/` separator in story names? It does the job for me. But alas this feature is not documented.

It doesn't seem to do anything for me, other than make the name longer. Am I missing something?
![Screen Shot 2019-03-18 at 8 07 19 PM](https://user-images.githubusercontent.com/3202211/54577602-a9ea6900-49b9-11e9-9ef1-9b1d29c1a89a.png)

shilman:
@lancesnider This feature has not been implemented, which is why the issue is still open

PR #4806, opened 2018-11-19
feat(react-native): Add ability to filter story list from React-Native

estevaolucas:
Issue: N/A

## What I did

1) Add the feature of filtering the story list from React-Native.
2) Wrap storybook view with `KeyboardAvoidingView` to make the story list present the whole list while the filter bar is open. (Basically doing this PR #4710, but adding an option, `shouldDisableKeyboardAvoidingView`, to disable this `KeyboardAvoidingView`, preventing the problem with nested `KeyboardAvoidingView`)
3) Remove `KeyboardAvoidingView` from `OnDeviceUI/addons/wrapper.js`. Looks like it's not working, as expected, and know with `KeyboardAvoidingView` as Storybook wrapper, it's no longer needed;
4) Close keyboard when moves from story list or addons view to the preview.

<br><br><p align="center"></p>

## How to test

- Is this testable with Jest or Chromatic screenshots? N
- Does this need a new example in the kitchen sink apps? N
- Does this need an update to the documentation? Y

### DIFF
--- a/app/react-native/src/preview/components/StoryListView/index.js
+++ b/app/react-native/src/preview/components/StoryListView/index.js
@@ -116,9 +116,10 @@ export default class StoryListView extends Component {
     const { data } = this.state;
 
     return (
-      <React.Fragment>
+      <View style={style.flex}>
         <TextInput
           clearButtonMode="while-editing"
+          disableFullscreenUI
           onChangeText={this.handleChangeSearchText}
           placeholder="Filter"
           returnKeyType="search"
@@ -142,7 +143,7 @@ export default class StoryListView extends Component {
           sections={data}
           stickySectionHeadersEnabled={false}
         />
-      </React.Fragment>
+      </View>
     );
   }
 }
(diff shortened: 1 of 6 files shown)


Gongreg:
Hey @elucaswork, great job!. Did you test it with android? Both landscape and portrait? Does switching between them with keyboard open works correctly? :)

app/react-native/src/preview/components/StoryListView/index.js
```diff
@@ -88,24 +116,33 @@ export default class StoryListView extends Component {
...
-          />
-        )}
-        renderSectionHeader={({ section: { title } }) => (
-          <SectionHeader title={title} selected={title === selectedKind} />
-        )}
-        keyExtractor={(item, index) => item + index}
-        sections={data}
-        stickySectionHeadersEnabled={false}
-      />
+      <React.Fragment>
```
Gongreg: Maybe lets not use fragment yet? They are only added in React 16.2. It will break backwards compatibility.

Gongreg (changes requested):

Gongreg:
Since you've been contributing a lot lately, I've added you @elucaswork to the Storybook team! Welcome :)

Gongreg:
Lets see how it will do :)

Regarding webview for story list we have to wait till @ndelangen finishes working on refactoring of it so we could use it too.

The idea would be to create a mini bundle with code necessary to run the sidebar, then load it in a page and use postmessages to communicate between webview and react native part.

Gongreg:
#3905 is the issue about it.

PR #4800, opened 2018-11-16
Update the Starter Guide for React

tobiasthorin:
Issue: N/A

## What I did
- Correct some grammar.
- Change some language to be more clear and accommodating.
- Remove the inclusion of the actions add on in the code example, since this prevents successful compilation if the guide if followed exactly.

## How to test
- Follow the steps in the React Starter Guide
- Storybook should now start without problems on the first try

PR #4807, opened 2018-11-19
Correct the name of the dependency @babel/core

tobiasthorin:
## What I did
Corrected the name of the dependency @babel/core, which I missed in the previous pull request #4800.


igor-dv (approved):
👍

PR #4824, opened 2018-11-20
Add TypeScript support for react-scripts

mrmckeb:
Issue: #4587

## What I did
Added support for TypeScript when using `react-scripts@>=2.1.0`.

## How to test
Create a new Create React App project and pass in the `--typescript` flag.
```sh
npx create-react-app my-app --typescript
```
Install storybook (this build) and run `storybook`. You should not see any errors.


igor-dv (commented):
Very cool! Thanks for taking it.
Do you want to add CLI tests for this in a separate PR?

mrmckeb:
@igor-dv Definitely, I can take a look at that over the weekend.

I've squashed and pushed those changes, thanks for the review.

igor-dv (approved):

Vinnl:
Thanks for this work! Do I understand it correctly that now, using Storybook 4.1, we can:

- Use Storybook for React components written in TypeScript (this PR).
- Write Storybook configuration in TypeScript (#4902).
- NOT write stories themselves in TypeScript.

And follow-up in case the latter is false: is there anything that needs to be done to enable writing stories in TypeScript, other than renaming them from `.jsx` to `.tsx`?

mrmckeb:
Yes, all of my team's stories are `.tsx` and it works great.

Vinnl:
@mrmckeb Hmm, and you've removed your custom webpack.config.js and don't have a TypeScript loader in your own dependencies?

I'm asking because I'm running into

> Module parse failed: Unexpected token (17:22) You may need an appropriate loader to handle this file type.

So it appears that for some reason, without my own Webpack config, Storybook's config does not look for TypeScript files.

mrmckeb:
Hi @Vinnl, can you let me know what version of `react-scripts` and `@storybook/react` you're using? Thanks!

mrmckeb:
Hmm, where are you stories located? In component directories?

Perhaps you could push up a quick test repo that I could play with? Something very rough.

Vinnl:
@mrmckeb The stories are in `/stories`, next to the components which are in `/src`. The project is open source, so it can be seen [here](https://gitlab.com/Flockademic/plaudit/tree/57-use-svg-component-and-native-typescript-in-storybook/client). (Note that this work is done in the branch `57-use-svg-component-and-native-typescript-in-storybook`, and that there are several projects in this repo - the relevant one is in `/client`.)

mrmckeb:
Would it be OK if I look over the weekend @Vinnl? Sorry for the delay.

mrmckeb:
I just confirmed with the latest packages, and it is still working for me locally @Vinnl.

image

```
    "react-scripts": "2.1.1",
```

```
    "@storybook/addon-a11y": "^4.1.1",
    "@storybook/addon-actions": "^4.1.1",
    "@storybook/addon-knobs": "^4.1.1",
    "@storybook/addon-links": "^4.1.1",
    "@storybook/addons": "^4.1.1",
    "@storybook/react": "^4.1.1",
    "@types/storybook__addon-actions": "^3.4.1",
    "@types/storybook__addon-knobs": "^3.4.1",
    "@types/storybook__addon-links": "^3.3.3",
    "@types/storybook__react": "^4.0.0",
```

Vinnl:
Thanks for working on this at all @mrmckeb - you don't owe me anything :)

But you were entirely correct! Moving the stories into the `src/` folder solved the issue, and the stories get compiled just like my regular components now. Thanks a bunch!

igor-dv:
@mrmckeb we probably need to add `stories` dir into the `include` as well 🤦‍♂️  SB cli creates this dir when you execute `sb init`...

mrmckeb:
I actually feel that this is not a good default. I much prefer stories to be alongside components - perhaps that should be the default? This results in less jumping between folders, and a smaller chance that people will forget to update stories after changing components (which I've seen a lot).

Regardless, this is a very easy fix if we want to add it ;)

igor-dv:
Yeah, that is how most of the people work with stories, IMO. Though I am not sure it's easy to do this with the cli at least now...

Issue #4821, opened 2018-11-20
Build error from slow start - babel-core vs @babel/core


andyhall:
Hi,

I'm following the [slow start](https://storybook.js.org/basics/guide-vue/) instructions, which start with:

```sh
npm i --save-dev babel-core
npm i --save-dev babel-loader
```

When I later run the script to start storybook, it fails out with:

```
ERROR in ./.storybook/config.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@babel/core'
 babel-loader@8 requires Babel 7.x (the package '@babel/core'). 
If you'd like to use Babel 6.x ('babel-core'), you should install 'babel-loader@7'.
```

Do I need a different `core` or a different `loader`, or is something else going on?

(Or even better, is there any way to get a "built" version of storybook that's already transpiled and doesn't need Babel to be present at runtime? I assume not but if there is that would be ideal!)

Thanks!

igor-dv:
The is a bit outdated. You should use `@babel/core`.

y-nk:
@andyhall i'm gonna update the wiki page as we speak, but meanwhile you can use [this post](https://github.com/storybooks/storybook/issues/4505#issue-372454482) as a up-to-date slow start instruction set for vue.

andyhall:
@y-nk Those steps work, thanks! Feel free to close this when the doc issue is resolved.

y-nk:
@andyhall PR to update doc is on its way. Thank you for reporting !

PR #4825, opened 2018-11-21
Update guide for Vue to match current requirements

y-nk:
Issue:

docs is not up to date regarding Guide for Vue.

## What I did

Fixing the doc, added dependencies as they should be. It is separated in 2 parts to highlight dev and peer dependencies.

It will answer to #4505 #4475 and #4821 

### Side note

As @igor-dv mentioned on Discord, `babel-preset-vue` is probably not necessary. As seen in this package's homepage, it's a preset to transform JSX in Vue projects, but Vue.js team advices to use template instead for the majority of cases (see [here](https://vuejs.org/v2/guide/render-function.html#Basics)). As we can extend webpack configuration to add loaders easily, it would make more sense to remove `babel-preset-vue` from the dependencies and advice advanced users of Vue.js to add it themselves following [existing SB documentation](https://storybook.js.org/configurations/custom-webpack-config/).

### DIFF
--- a/docs/src/pages/basics/guide-vue/index.md
+++ b/docs/src/pages/basics/guide-vue/index.md
@@ -37,8 +37,8 @@ Make sure that you have `vue`, `babel-core`, `babel-loader` in your dependencies
 
 ```sh
 npm i --save vue
-npm i --save-dev babel-core
-npm i --save-dev babel-loader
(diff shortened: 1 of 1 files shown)


igor-dv (approved):
Lgtm

PR #4837, opened 2018-11-23
Fix snapshots to use "emotion/snapshot-serializer"

igor-dv:
Issue: emotion snapshot serializer was not applied to storyshots. Serializers should be provided directly to initStorysthos method.


shilman (approved):
Good catch!!

Issue #4759, opened 2018-11-09
v4 no longer shows correct component in Story source


alisonjoseph:
Our users rely on storybook to know how to use our components. They use the "show info" to see the component name annd prop types table. Since upgradinng to v4 we are gettinng questions about what componennt to use since it is no lonnger visible. Is this by design or a bug? and is there any way to fix. Thank you! 🙌 

This is what we get w/ v3

<img width="584" alt="screen shot 2018-11-09 at 8 09 11 am" src="https://user-images.githubusercontent.com/2753488/48267111-0f6c8e80-e3f7-11e8-8310-50bc89d4edad.png">

vs after upgrading to v4. 
<img width="669" alt="screen shot 2018-11-09 at 8 09 24 am" src="https://user-images.githubusercontent.com/2753488/48267106-0bd90780-e3f7-11e8-9f62-23b74f71c1e6.png">

MorganLester:
Looks like you need to add displayName and that resolves it.

vpicone:
@MorganLester The weird thing is that requirement isn't new but it worked fine before (presumably through @babel/preset-react's "@babel/plugin-transform-react-display-name"). Wondering why that preset might not work as expected when running build-storybook in v4.

igor-dv:
I can assume there is something to do with babel minification configuration

vpicone:
@igor-dv The minification configuration of storybook itself? This is the .storybook/.babelrc for the project. It's only an issue comes from the statically built storybook:
`
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {
          "browsers": ["last 1 versions", "Firefox ESR"]
        }
      }
    ],
    "@babel/preset-react"
  ],
  "plugins": [
    "dev-expression",
    "@babel/plugin-syntax-dynamic-import",
    "@babel/plugin-syntax-import-meta",
    "@babel/plugin-proposal-class-properties",
    "@babel/plugin-proposal-json-strings",
    [
      "@babel/plugin-proposal-decorators",
      {
        "legacy": true
      }
    ],
    "@babel/plugin-proposal-function-sent",
    "@babel/plugin-proposal-export-namespace-from",
    "@babel/plugin-proposal-numeric-separator",
    "@babel/plugin-proposal-throw-expressions",
    "@babel/plugin-proposal-export-default-from",
    "@babel/plugin-proposal-logical-assignment-operators",
    "@babel/plugin-proposal-optional-chaining",
    [
[...]

Issue #4717, opened 2018-11-05
CRA's webpack configuration is mangling function names (and removing prop types) in static build


jansav:
**Describe the bug**
When webpack configuration is used directly when CRA, it causes `build-storybook` to mangle function names. This causes issues that code source and prop types are not presented correctly.

As #3879 says, it could be solved by adding `.displayName` to each component, but since CRA is now removing prop types in production builds (https://github.com/facebook/create-react-app/pull/3818), prop types are not visible in static version of storybook.

**To Reproduce**
1. Create app using Create React App
2. Install storybook
3. Create functional component and stories for it
4. Run`npm run build-storybook` (without having custom webpack configuration)
5. Serve static storybook and see results.

**Expected behavior**
Function name is used as display name. Prop types are visible in page.

**System:**
 - Framework: React, react-scripts (2.1.1)
 - Addons: **Info**, knobs, options
 - Version: 4.0.2

**Additional context**
It might be that this is something that storybook can't really do anything else than suggest **not** using CRA's webpack configuration directly. However I would like to see if there is any suggestions how to solve this issue without using full control mode.

jansav:
Any tips?

davidchristie:
I recently hit this issue when upgrading `@types/react` to version 16.7.6.

Dropping back to 16.4.18 fixed it for me: https://github.com/davidchristie/social-network/pull/216/commits/9d6f0046d5a277e2020aba4794a09f6fe8318472

PR #4838, opened 2018-11-23
Core - do not mangle on production build

igor-dv:
Issue:
#4717
#4759

## What I did
So, the "production" mode is adding by default the ["terser-webpack-plugin" ](https://github.com/webpack/webpack/blob/e1df721fd77634b5a22b4349c7d7ea15f0a2188e/lib/WebpackOptionsDefaulter.js#L306) minimizer that mangles by default, and it breaks the info addon.

We need to turn the mangling off.


lib/core/src/server/preview/iframe-webpack.config.js
```diff
@@ -95,6 +95,7 @@ export default ({
         chunks: 'all',
       },
       runtimeChunk: true,
+      minimize: false,
```
kroeder: Why not minimize in production? Isn't this what you want for production?

kroeder (approved):
Looks good

shilman (approved):
LGTM - thanks for fixing this!

Issue #4845, opened 2018-11-25
[cra-kitchen-sink] required index.html file missing


maacky:
**Describe the bug**
Due to index.html missing in 'examples/cra-kitchen-sink/public', `yarn start` in 'examples/cra-kitchen-sink' throws `Could not find a required file`. The file was removed in e93d9f3 (FIX chromatic by removing some index.html files and correcting the location of static files).

**To Reproduce**
1. Go to 'examples/cra-kitchen-sink'
2. Run `yarn`
3. Run `yarn start`

**Expected behavior**
An example CRA app starts.

**Are you able to assist bring the feature to reality?**
Sure, as fixing the bug requires adding a simple .html file to inject the app into. However, since the file seems to have been deleted on purpose, I'm not sure if it wasn't done with some other solution in mind. If adding the file is the way to go, ping me and I'll be happy to fix the bug.

igor-dv:
Yeah, probably we need this file back in the cra example. Do you want to return it @maacky  ?

maacky:
@igor-dv, sure, let me do it.

PR #4856, opened 2018-11-26
restore index.html in cra-kitchen-sink

maacky:
Issue: #4845 

## What I did
Restore the missing index.html file in the CRA example app.


ndelangen:
Seems to break Chromatic still?

ndelangen:
I added a filter for index.html files when copying over files from static dirs.

ndelangen (approved):

ndelangen:
Thank you @maacky!

PR #4870, opened 2018-11-27
Add try, catch for IE11 as it breaks script execution

hmelenok:
Issue:
More details - https://github.com/storybooks/storybook/issues/4854, IE11 breaks on change url with security error 
## What I did
Add try/catch for this case
## How to test
Try to load 
- Is this testable with Jest or Chromatic screenshots? - **I think not**
- Does this need a new example in the kitchen sink apps? - **Nope**
- Does this need an update to the documentation? - **I not sure about that**


ndelangen:
OK, so something is causing a security issue in IE11, would be nice to actually prevent the security issue, but meanwhile merging this can do no harm 👌

ndelangen (approved):

ndelangen:
Thank you @hmelenok !

PR #4872, opened 2018-11-27
Properly sets initial story out of Async Storage.

Gongreg:
Issue: Fixing an issue with feature added in #4713.

Scenario:

Start the storybook without server running.

There is possibility that websocket connection will fail before the component is fully rendered and after component has started to render. In that case it fails to select initial story.

Now I pass in function to onDeviceUI to get initial story. The drawback is that the story is not set immediately. But it is better than nothing.


estevaolucas (approved):
LGTM

PR #4876, opened 2018-11-27
Updated readme to not instruct use to save-dev

grp06:
currently the readme says that this should be saved in dev dependencies, which its shouldn't

Issue:

## How to test

### DIFF
--- a/addons/centered/README.md
+++ b/addons/centered/README.md
@@ -7,7 +7,7 @@ Storybook Centered Decorator can be used to center components inside the preview
 ### Usage
 
 ```sh
-npm install @storybook/addon-centered --save-dev
+npm install @storybook/addon-centered --save
 ```
 
 #### As a decorator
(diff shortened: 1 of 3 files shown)


ndelangen (approved):

ndelangen:
Should be good to merge

@shilman this is ok to merge directly into master?
Or should it be merged into `next` and labelled `patch`?

addons/centered/README.md
```diff
@@ -7,7 +7,7 @@ Storybook Centered Decorator can be used to center components inside the preview
 ### Usage
 
 ```sh
-npm install @storybook/addon-centered --save-dev
+npm install @storybook/addon-centered --save
```
igor-dv: That's not true. Mostly people are using storybook in their dev deps, since Storybook is a tool for their projects, but not the dependency they rely within the application. That means all the SB packages are also, mostly, should be installed into dev deps.

igor-dv (changes requested):

addons/centered/README.md
ndelangen: I see, yeah it depends on how someone uses storybook right?

If they feel it's a deliverable / deployable unit, then it's part of their regular dependencies.

TBH for most apps I've worked with it doesn't really matter too much, for libs obviously it matters a lot.

I suggest to keep it with `--save-dev` as @igor-dv suggests.

What do you think @grp06 ?

wuweiweiwu (approved):
I don't feel too strongly about either `--save` or `--save-dev`

ndelangen:
Thanks for contributing @grp06, I closed the PR, because we seem to mostly agree the --save-dev is warranted here. But want to thank you regardless for the effort you put in!

PR #4883, opened 2018-11-28
Reject webpack-related issues

ranyitz:
## The Problem

If there is a webpack-related issue, you won't be able to see it, and instead, you'll see the following error:

image

Which does not help with finding the solution to the problem.

__________

> The err object will not include compilation errors and those must be handled separately using stats.hasErrors() which will be covered in detail in Error Handling section of this guide. The err object will only contain webpack-related issues, such as misconfiguration, etc.

[From webpack docs](https://webpack.js.org/api/node/#webpack-)

______

## The Solution

When there is an error, reject it instead of the stats.


hipstersmoothie (approved):

igor-dv (approved):
Thanks

PR #4899, opened 2018-12-01
fix(): deprecations with replacing setOptions with withOptions

mbellary-chwy:
Issue:

Storybook options getting reset when setOptions used in storybooks

## What I did

- replaced `setOptions` with `withOptions` as `setOptions` is deprecated in 4.0
- used correct syntax of `addParameters` in stories to update options

## Before this fix:

- Running storybooks of ADDONS -> a11y -> Default, the options are getting reset. Storybooks are breaking

## After this fix:

Storybooks will work fine after this fix

### DIFF
--- a/examples/vue-kitchen-sink/src/stories/__snapshots__/custom-decorators.stories.storyshot
+++ b/examples/vue-kitchen-sink/src/stories/__snapshots__/custom-decorators.stories.storyshot
@@ -8,13 +8,17 @@ exports[`Storyshots Custom|Decorator for Vue render 1`] = `
     style="border: medium solid blue;"
   >
     <div
-      style="border: medium solid red;"
+      style="border: medium solid blue;"
     >
-      <button
-        class="button"
+      <div
+        style="border: medium solid red;"
       >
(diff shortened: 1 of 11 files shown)


examples/vue-kitchen-sink/src/stories/__snapshots__/custom-decorators.stories.storyshot
```diff
@@ -8,13 +8,17 @@ exports[`Storyshots Custom|Decorator for Vue render 1`] = `
     style="border: medium solid blue;"
   >
     <div
-      style="border: medium solid red;"
+      style="border: medium solid blue;"
```
igor-dv: Looks decorators are mixed up here like in #4846

igor-dv: Ah, that's in vue examples, where the problem is found...

igor-dv (approved):

ndelangen:
Awesome @MansoorBashaBellary

PR #4902, opened 2018-12-02
Add CRA TypeScript support for .storybook folder

mrmckeb:
Issue: https://github.com/storybooks/storybook/issues/4587#issuecomment-442485939

## What I did
Allowed the use of `.ts` and `.tsx` files in the `.storybook` folder. This was not in scope initially, but was added on request.

## How to test
When using with CRA and TypeScript, rename your `config.js` to `config.tsx` in the `.storybook` folder.

### DIFF
--- a/app/react/src/server/cra-config.js
+++ b/app/react/src/server/cra-config.js
@@ -62,7 +62,15 @@ export const getRules = extensions => rules =>
   }, []);
 
 const getStyleRules = getRules(cssExtensions.concat(cssModuleExtensions));
-const getTypeScriptRules = getRules(typeScriptExtensions);
+const getTypeScriptRules = typescriptRules => {
+  const rules = getRules(typeScriptExtensions)(typescriptRules);
+  // Add support for using TypeScript in the `.storybook` folder.
+  const babelRule = rules[0];
+  // This line can be removed when this issue is resolved: https://github.com/storybooks/storybook/issues/4903
+  if (babelRule.include !== 'string') return rules;
+  babelRule.include = [babelRule.include, path.join(babelRule.include, '../.storybook')];
+  return rules;
+};
 
 export function getCraWebpackConfig(mode) {
   const pathToReactScripts = getReactScriptsPath();
--- a/app/react/src/server/cra-config.js
+++ b/app/react/src/server/cra-config.js
@@ -2,7 +2,6 @@ import fs from 'fs';
 import path from 'path';
 import semver from 'semver';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
-
 import { normalizeCondition } from 'webpack/lib/RuleSet';
 
 const cssExtensions = ['.css', '.scss', '.sass'];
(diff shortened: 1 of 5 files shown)


mrmckeb:
@igor I found a strange issue, unrelated, where this config is loaded twice - you can see that it's happening because it logs the action twice in the terminal.
image

I've [safeguarded against this for now](https://github.com/storybooks/storybook/pull/4902/files#diff-eda2a92e48b4df8907d95774b054a271R69), but we should look to resolve it - unfortunately I don't have time this weekend. I'll create an issue for it.

app/react/src/server/cra-config.js
```diff
@@ -62,7 +62,14 @@ export const getRules = extensions => rules =>
   }, []);
 
 const getStyleRules = getRules(cssExtensions.concat(cssModuleExtensions));
-const getTypeScriptRules = getRules(typeScriptExtensions);
+const getTypeScriptRules = typescriptRules => {
+  const rules = getRules(typeScriptExtensions)(typescriptRules);
+  // Add support for using TypeScript in the `.storybook` folder.
+  const babelRule = rules[0];
+  if (babelRule.include !== 'string') return rules;
+  babelRule.include = [babelRule.include, path.join(babelRule.include, '../.storybook')];
```
igor-dv: We need to use the `configDir` prop coming from the preset in order to put the reall config dir that user could provide via the cli. It should be already a resolved path.

igor-dv (changes requested):

igor-dv:
> I found a strange issue, unrelated, where this config is loaded twice 

Yeah, it is a "known" thing. We apply presets on `defaultConfig` and `baseConfig`because we don't know what user will choose. The fix could be only to have a defaultConfig, but for now it's a breaking change and should be discussed...

app/react/src/server/cra-config.js
mrmckeb: @igor-dv I've added this in the same way that the `react-native` app manages it, but using `commander`. Please let me know what you think.

app/react/src/server/cra-config.js
```diff
@@ -1,10 +1,17 @@
 import fs from 'fs';
 import path from 'path';
 import semver from 'semver';
+import program from 'commander';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
-
 import { normalizeCondition } from 'webpack/lib/RuleSet';
 
+program
```
igor-dv: Oh, that is not what I meant 😅 , sorry if it was out of context (I was on my phone)

[Here](https://github.com/storybooks/storybook/blob/next/app/react/src/server/framework-preset-cra.js#L4) is the preset entry poring to the webpack configuration for CRA. The second parameter that is passed is `options` (it was not in use in this preset)

You need to do something like this:

```js
export function webpackFinal(config, { configDir }) {
  if (!isReactScriptsInstalled()) {
    logger.info('=> Using base config because react-scripts is not installed.');
    return config;
  }

  logger.info('=> Loading create-react-app config.');

  return applyCRAWebpackConfig(config, configDir);
}
```

we can just pass the full `options` object in case there will be other needed props to use in future.

mrmckeb: Will update now, thanks!

app/react/src/server/__tests__/cra-config.test.js
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs';
-import { getReactScriptsPath } from './cra-config';
+import { getReactScriptsPath, getTypeScriptRules } from '../cra-config';
```
igor-dv: let's not move the tests into the `__tests__` dir, we have this in a few places, but it's mostly the inconsistency, rather the desired way (IMO).

app/react/src/server/cra-config.js
```diff
@@ -62,7 +61,16 @@ export const getRules = extensions => rules =>
...
 
 const getStyleRules = getRules(cssExtensions.concat(cssModuleExtensions));
-const getTypeScriptRules = getRules(typeScriptExtensions);
+
+export const getTypeScriptRules = (webpackConfigRules, configDir) => {
+  const rules = getRules(typeScriptExtensions)(webpackConfigRules);
+  // Add support for using TypeScript in the `.storybook` (or config) folder.
+  const babelRule = rules[0];
+  // Resolves an issue where this config is parsed twice (#4903).
+  if (typeof babelRule.include !== 'string') return rules;
```
igor-dv: I think the problem here is that you are mutating the initial object, instead of just returning a new instance

mrmckeb: Thanks @igor-dv, this is now resolved.

app/react/src/server/__tests__/cra-config.test.js
mrmckeb: This is now resolved, thanks for the feedback.

ndelangen (approved):

igor-dv (approved):
LGTM 💪

PR #4904, opened 2018-12-03
Remove use of const

jrwebdev:
Issue:

Usage of `const` is causing runtime errors in older browsers

## What I did

Removed use of `const`

## How to test

no
no
no


ndelangen (approved):

ndelangen:
Good solution @jrwebdev!

PR #4908, opened 2018-12-03
Document configuration options for Storyshots test

### LINKED ISSUE #4875: Document configuration options for Storyshots test
I use storyshots to do a simple smoke test for my stories. As my components use **ref** and **portal**, the default react-test-renderer does not work for me.

After digging through the source code, i found that it is indeed possible to overwrite the renderer, contrary to what the documentation says about the `test` option: https://github.com/storybooks/storybook/tree/master/addons/storyshots/storyshots-core#test

```js
import initStoryshots, { renderWithOptions } from '@storybook/addon-storyshots';
import { mount } from 'enzyme';

// Ensure the stories are rendered without an error
initStoryshots({
  test: renderWithOptions({
    // Overwrite the default react test renderer, as it does not support
    // refs and portals, both of which are heavily used by antd
    renderer: mount,
  }),
});
```

The documentation could be improved by specifying the options that can be given to `renderWithOptions` and `snapshotWithOptions`, especially this hidden functionality of overwriting the renderer.

spawnia:
Resolves #4875

Just a docs update.

### DIFF
--- a/addons/storyshots/storyshots-core/README.md
+++ b/addons/storyshots/storyshots-core/README.md
@@ -434,11 +434,27 @@ If you are running tests from outside of your app's directory, storyshots' detec
 
 ### `test`
 
-Run a custom test function for each story, rather than the default (a vanilla snapshot test). Setting `test` will take precedence over the `renderer` option. See the exports section below for more details.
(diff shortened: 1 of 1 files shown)


igor-dv (approved):
Thanks !

Issue #4903, opened 2018-12-02
create-react-app config is loading twice


mrmckeb:
**Describe the bug**
The create-react-app config is loading twice, but should not be.

**To Reproduce**
1. Run storybook alongside an application based on create-react-app.
2. Observe terminal output.

**Expected behavior**
This config should be loaded once only.

**Screenshots**
![image](https://user-images.githubusercontent.com/5043083/49340758-ff8a4800-f64c-11e8-854d-50ca79855899.png)

**System:**
 - OS: Windows 10 - Ubuntu on WSL
 - Device: Surface Book 2
 - Browser: N/A
 - Framework: React
 - Addons: a11y, actions, knobs, links
 - Version: 4.1.0-alpha.8

**Additional context**
I discovered this whilst working on #4902, as I was getting a strange error and then discovered that the config was loading twice. I've worked around this issue on that branch, for now, and commented so that line can be remove when this issue is resolved.

igor-dv:
As I've mentioned here - https://github.com/storybooks/storybook/pull/4902#issuecomment-443510766

We are applying presets on `baseConfig` and `defaultConfig` independently (it's not an issue introduced only in presets feature, it was also before, but now it's more visible):

```js
module.exports = (baseConfig, mode, defaultConfig) => { /*...*/ }
```
The reason for this, is that we don't know what user will choose.

The possible solution could be:

### Changing them into funcs:

```js
module.exports = (baseConfig, mode, defaultConfig) => {
  const config = baseConfig(); // let's say the presets will be applyed only upon the invocation.
}
```

☝️ this can be backward compatible with a deprecation message.

### Reducing the `baseConfig` at all (my prefered actually)

The difference between the `base` and `default` configs are only a few css/images rules added by default. IMO this is more a historical setup, and we can just use only one. Also, since we are trying to make things work out of the box with the tools like CRA and angular-cli (and more in the future).

So it could be like this:

```js
module.exports = (config, mode) => { /* ... */ }
```

☝️ this is a breaking change

igor-dv:
Will be a part of the v5

PR #4927, opened 2018-12-05
Core - Simplify custom webpack config 

igor-dv:
Issue: Partially related to #4903

Today we have an option to extend  webpack config by providing the function like this:

```js
module.exports = (basicConfig, mode, defaultConfig) => { /*...*/ }
```

Where `defaultConfig` is a few rules bigger than a `basicConfig`. I think it brings more confusion rather a benefit.

## What I did

1. Reduced the `basicConfig` and changed to the signature to:

```js
module.exports = ({ config, mode }) => { /*...*/ }
```

Why object? Because it's much easier to extend and deprecate in the future. 

2. Removed a minor prop (`defaultConfigName`) that is passed forme the frameworks to core. This prop is kinda not needed.

### DIFF
--- a/app/angular/src/server/options.js
+++ b/app/angular/src/server/options.js
@@ -2,7 +2,6 @@ import packageJson from '../../package.json';
 
 export default {
   packageJson,
-  defaultConfigName: 'angular-cli',
   frameworkPresets: [
     require.resolve('./framework-preset-angular.js'),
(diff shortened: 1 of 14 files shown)


ndelangen:
Shall we merge this into the overhaul-ui branch?

Hypnosphi:
What is the recommended way of getting only the necessary part of default config now?

Issue #3999, opened 2018-08-10
moment and emotion requirement undocumented for addon-knobs


eps1lon:
Adding `addon-knobs` requires `moment` and `emotion` installed which is not documented. It does produce a warning in yarn however which might be sufficient. Just filing this issue to be sure.

### Steps to reproduce
- list dependencies as described below
- `yarn`
    ```bash
   warning "@storybook/addon-knobs > react-emotion@9.2.6" has unmet peer dependency "emotion@^9.1.3".
    warning "@storybook/addon-knobs > react-datetime@2.15.0" has unmet peer dependency "moment@>=2.16.0".
   ``` 
- `yarn start storybook`
   ```bash
   ERROR in ./node_modules/react-datetime/DateTime.js
   Module not found: Error: Can't resolve 'moment' in '/home/eps1lon/Development/src/js/react-blend-circles/node_modules/react-datetime'
    @ ./node_modules/react-datetime/DateTime.js 6:10-27
    @ ./node_modules/@storybook/addon-knobs/dist/components/types/Date/index.js
    @ ./node_modules/@storybook/addon-knobs/dist/components/types/index.js
    @ ./node_modules/@storybook/addon-knobs/dist/components/Panel.js
    @ ./node_modules/@storybook/addon-knobs/dist/register.js
    @ ./node_modules/@storybook/addon-knobs/register.js
    @ ./.storybook/addons.js
    @ multi ./.storybook/addons.js ./node_modules/@storybook/core/dist/server/config/polyfills.js ./node_modules/@storybook/core/dist/client/manager/index.js
   ```

### Please specify which version of Storybook and optionally any affected addons that you're running

- @storybook/react 4.0.0-alpha.16
- @storybook/addon-knobs 4.0.0-alpha.16

Possible solutions:
- list `moment` and `emotion` as a peer to ´addon-knobs`
- list `moment` and `emotion` as optional to `addon-knobs` and allow tree shakeable import so that import e.g. `number` does not pull in `date`
- list `moment` and `emotion` as direct dependency

igor-dv:
@ndelangen, is this somehow missed from the latest theme/deps changes?

wuweiweiwu:
Can verify that this is an issue. I think moment should be a direct dep since its required by `react-datetime`

eps1lon:
[...]

PR #4928, opened 2018-12-05
fix: Add missing peer dependencies to addon-knobs

eps1lon:
Fixes `Uncaught TypeError: Object(...) is not a function` when using `@storybook/addon-knobs` and `@emotion/core@^10.x`

Issue:
- #3999

The combination of the above packages created the following dependency tree:
```
├─ @emotion/core@10.x
├─ @storybook/addon-knobs@4.x
│  └─ @emotion/styled@0.x
```
However `@emotion/styled` has a peer dependency to `@emotion/core@0.x`. I didn't find any method in npm or yarn to add `@emotion/core@0.x` while still being able to use `@emotion/core@10.x`. This is probably the reason missing peer dependencies should be added immediately instead of relying on consuming packages to do this.

## What I did
- Add `@emotion/core@0.x` to dependencies of `@storybook/addon-knobs`.

## How to test

  No
  No
  No

### DIFF
--- a/addons/knobs/package.json
+++ b/addons/knobs/package.json
@@ -21,6 +21,7 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "dependencies": {
+    "@emotion/core": "^0.13.1",
     "@emotion/styled": "^0.10.6",
     "@storybook/addons": "4.1.0-alpha.10",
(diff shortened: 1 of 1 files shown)


wuweiweiwu (approved):
LGTM!

igor-dv (approved):

PR #4929, opened 2018-12-05
Move 'core-js/modules/es6.promise' into storyshots for angular

igor-dv:
Issue: https://github.com/storybooks/storybook/pull/4900#discussion_r238841278


shilman (approved):
YAY

PR #4961, opened 2018-12-10
add forwardRef support to addon-info

### LINKED ISSUE #4787: addon-info doesn't handle React.forwardRef appropriately
**Describe the bug**
When forwarding a ref to a child component, the component shows up as `<Unknown>`. You can set the display name of the wrapper component to the name of the child component, but this means less legible devtools.

**To Reproduce**
1. Export a component with a React.forwardRef wrapper:
2. View the story source section of addon-info

*Note:* If you set the displayName of the wrapper component, addon-info will use that. However, simply using the displayName from the child makes for less readable dev tools, it looks like component nested within itself.

**Expected behavior**
Story source should ignore the wrapper component `<ForwardRef>` and display the child before displaying `<Unknown>`

**Code snippets**
### `<Unknown>` - currently happening w/o intervention
```javascript
import React, { forwardRef } from 'react';

const withForwardRef = Component => {
  const forwarder = forwardRef((props, ref) => (
    <Component {...props} innerRef={ref} />
  ));
  return forwarder;
};

export default withForwardRef;
```
screen shot 2018-11-13 at 1 39 18 pm

### `<ForwardRef(Component)>` - good for dev tools bad for story
```javascript
import React, { forwardRef } from 'react';

const withForwardRef = Component => {
  const forwarder = forwardRef((props, ref) => (
    <Component {...props} innerRef={ref} />
  ));
  forwarder.displayName = `ForwardRef(${Component.displayName || Component.name})`; // use display name if provided
  return forwarder;
};

export default withForwardRef;
```
screen shot 2018-11-14 at 10 36 00 am

### `<Component>` (duplicate child) good for story, bad for dev tools
```javascript
import React, { forwardRef } from 'react';

const withForwardRef = Component => {
  const forwarder = forwardRef((props, ref) => (
    <Component {...props} innerRef={ref} />
  ));
  forwarder.displayName = Component.displayName || Component.name; // use display name if provided
  return forwarder;
};

export default withForwardRef;
```
screen shot 2018-11-14 at 10 38 06 am

yjcxy12:
Issue: closes #4787

## What I did
- added `react-is` to dependencies in `addon-info`.
- in both `Node` and `Story` checks if the element is `forwardRef` element and parses accordingly.
- add new example of `forwardRef` component in kitchen sink app.

## How to test

Yes.
Yes.
Don't think so.

### DIFF
--- a/addons/info/package.json
+++ b/addons/info/package.json
@@ -30,6 +30,7 @@
     "nested-object-assign": "^1.0.1",
     "prop-types": "^15.6.2",
     "react-addons-create-fragment": "^15.5.3",
+    "react-is": "^16.6.1",
     "react-lifecycles-compat": "^3.0.4",
     "util-deprecate": "^1.0.2"
   },
--- a/addons/info/src/components/Node.js
+++ b/addons/info/src/components/Node.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { isForwardRef } from 'react-is';
 import PropTypes from 'prop-types';
 import Props from './Props';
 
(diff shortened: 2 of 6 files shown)


igor-dv:
In [docs](https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools) they mentioned to just use a `dislpayName`. 
Is your solution better ?

ndelangen (approved):

ndelangen:
@yjcxy12 this is awesome! thank you!

ndelangen:
This code throws when the component uses a react hook.

See this issue: https://github.com/storybooks/storybook/issues/6848

I've been able to make some adjustments to make that use-case work, I'll open a PR momentarily. Will you be able to review @yjcxy12 

Perhaps a solution based on displayName would be better indeed @igor-dv

PR #4973, opened 2018-12-11
[docs] Update TS config docs

cosmith:
Hi, I updated the example code setting up react-docgen-typescript-loader after having fought with it for quite some time. 

I'm also wondering if recommending react-docgen-typescript-webpack-plugin is really necessary since configuring directly with react-docgen-typescript-loader is the same amount of code... 

```js
const path = require("path");

module.exports = (baseConfig, env, config) => {
    config.module.rules.push({
        test: /\.(ts|tsx)$/,
        include: path.resolve(__dirname, "../"),
        use: [require.resolve("ts-loader"), require.resolve("react-docgen-typescript-loader")],
    });
    config.resolve.extensions.push(".ts", ".tsx");

    return config;
};
```

### DIFF
--- a/docs/src/pages/configurations/typescript-config/index.md
+++ b/docs/src/pages/configurations/typescript-config/index.md
@@ -145,34 +145,39 @@ Please refer to the [react-docgen-typescript-loader](https://github.com/strothj/
 
 Additional annotation can be achieved by setting a default set of info parameters:
 
-```js
+```ts
+import {addDecorator} from "@storybook/react";
+import {withInfo} from "@storybook/addon-info";
+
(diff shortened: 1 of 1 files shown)


wuweiweiwu (approved):
LGTM

wuweiweiwu:
^ hm interesting. Can you open a feature request?

cosmith:
@wuweiweiwu do you mean a feature request about the `react-docgen-typescript-webpack-plugin` recommendation?

ndelangen (approved):

wuweiweiwu:
@cosmith yep!

PR #4977, opened 2018-12-11
TypeScript migration: @storybook/channels

kroeder:
Issue: N/A

## What I did

Migrated `@storybook/channels` to TypeScript

## How to test

Tests are included - currently almost 100% coverage. Test for `async: true` is missing


lib/channels/src/index.ts
```diff
@@ -1,27 +1,60 @@
...
+  // generates a random 13 character string
+  return Math.random()
+    .toString(16)
+    .slice(2);
+};
+
+export class Channel {
+  private _sender = generateRandomId();
+  private _listeners: ListenersKeyValue = {};
+  private readonly _async: boolean = false;
```
igor-dv: The readonly field that has a value can be changed in a ctro 😱 ??

kroeder: `readonly` can only be set directly or in the constructor, if you meant that?

igor-dv: I just see this `_async` prop is also defined in ctor.

kroeder:
Rewrote all tests in order to reflect the actual usage of this package
Before, there were lots of tests on private class members

Coverage is almost 100% - I don't know how to test `async: true` (see todo comment in the code)

Also, the whole package is type safe - no `any` usage except for generics as default value (that makes them optional)

lib/channels/src/index.ts
```diff
@@ -0,0 +1,152 @@
+export type ChannelHandler<TEventArgs = any> = (event: ChannelEvent<TEventArgs>) => void;
```
Keraito: > Also, the whole package is type safe - no `any` usage except for generics as default value (that makes them optional)

Some questions because my knowledge is limited on TS: Is this what you mean with these default values for generics?  Why exactly do we need to provide a default value for a generic type for the args of this `ChannelHandler`? Sorry if this is answered somewhere else already (since I know that the TS channel is active), but just asking 😄

lib/channels/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export interface ChannelTransport<TEventArgs = any> {
+  send(event: ChannelEvent<TEventArgs>): void;
+  setHandler(handler: ChannelHandler<TEventArgs>): void;
+}
+
+export interface ChannelEvent<TEventArgs = any> {
```
Keraito: Continuing on above's comment: if it is necessary to have default values for generic types and this is the implementation of it, do we have to drill it down every interface and type?

lib/channels/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+  type: string; // eventName
```
Keraito: Are the possible values for this `type` field of a limited set? If so it might be better to turn this into a Enum. This depends on whether we have full knowledge on the possible values for this field and also if that's something of importance of course.

lib/channels/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+  from: string;
```
Keraito: Same goes for the `from` field similar to the `type` field, see comment above.

lib/channels/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+  constructor({ transport, async = false }: ChannelArgs = {}) {
+    this.isAsync = async;
+    if (transport) {
+      this._transport = transport;
+      this._transport.setHandler(event => this._handleEvent(event));
+    }
+  }
+
+  get hasTransport() {
+    return !!this._transport;
```
Keraito: I personally prefer the following way of enforcing the return of a boolean based on the existence of an object due to readability. this is minor though. Choose whichever you prefer :)

```suggestion
    return Boolean(this._transport);
```

lib/channels/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      this._events = {};
+    } else if (this._events[eventName]) {
+      delete this._events[eventName];
+    }
+  }
+
+  removeListener(eventName: string, listener: Listener) {
+    const listeners = this.listeners(eventName);
+    if (listeners) {
+      this._events[eventName] = listeners.filter(l => l !== listener);
```
Keraito: Not sure how this works in TS, but how does it handle the equality of two typed objects (`Listener`)? Is there something like an equality method that we can override that would make this more controlled?

lib/channels/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+export interface Listener<TEventArgs = any> {
+  (...args: TEventArgs[]): void;
+
+  ignorePeer?: boolean;
+}
+
+interface EventsKeyValue {
+  [key: string]: Listener[];
```
Keraito: What does this `[key: string]` part do? 😮

Keraito (commented):
Awesome PR and work! Have some question regarding different aspects of TS and implementation. Would definitely like to help out with this! 😃

lib/channel-websocket/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "license": "MIT",
   "main": "dist/index.js",
-  "jsnext:main": "src/index.js",
+  "jsnext:main": "src/index.ts",
```
ndelangen: I think we need to remove this line possibly?

ndelangen: This is related to tree-shaking, we need to make sure we don't blow up the bundle when removing this.

lib/channel-websocket/src/index.ts
```diff
@@ -1,45 +1,57 @@
...
+  send(event: any) {
     if (!this._isReady) {
       this._sendLater(event);
     } else {
       this._sendNow(event);
     }
   }
 
-  _sendLater(event) {
+  private _sendLater(event: any) {
```
ndelangen: remove all the prexix _

lib/channels/src/index.ts
ndelangen: Would be nice, but I don't think it's possible, since addons are able to use the channel too.

lib/channels/src/index.ts
ndelangen: Yeah should be either `manager` or `preview`

lib/channels/src/index.ts
ndelangen: 🆗

lib/channels/src/index.ts
ndelangen: I prefer `!!` too

ndelangen (approved):

ndelangen:
We need to NOT test this file:
`examples/angular-cli/src/test`

ndelangen:
hmm
```
Number of tests 149 is 84% less than 913 in build #5452
[16:25:46]	Build failure condition: Anchor value for number of tests is taken from <latest successful build>. 
Fail build if number of tests is 20% less than the anchor value
```

lib/channels/src/index.ts
kroeder: Just creating a key-value type

https://www.typescriptlang.org/play/#src=interface%20KeyVal%20%7B%0D%0A%20%20%20%20%5Bkey%3A%20string%5D%3A%20number%3B%0D%0A%7D%0D%0A%0D%0Aconst%20keyValue%3A%20KeyVal%20%3D%20%7B%0D%0A%20%20%20%20someKey%3A%206%2C%0D%0A%20%20%20%20anotherKey%3A%2010%2C%0D%0A%20%20%20%20notPossible%3A%20%22foo%22%0D%0A%7D

key obviously is a string but the important part is the type of `value`

gaetanmaisse (approved):

ndelangen:
Ready to merge?

lib/channels/src/index.ts
kroeder: I thought a lot about the generics and came to the conclusion that we should not use generics for channels because it just won't work. It's error by design

My initial thoughts were that you have 1 event with multiple listeners and all of them share the same data-interface but here's the thing, no one is stopping you from doing this:

```typesciprt
channel.addListener<{ id: string }>('event1', data => data.id);
channel.addListener<{ name: string }>('event1', data => data.name);

// Just won't do, there's no correct interface anymore 
channel.listeners<???>('event1');
```

I will remove them for now, thanks for asking this!

lib/channels/src/index.ts
kroeder: Default values are not necessary but prevent a breaking change. Without a default the consumer is forced to pass in an interface. However, as mentioned above I'm going to remove them

lib/channel-websocket/package.json
kroeder: I added this todo to the organization task

kroeder:
> Ready to merge?

Yes, I still need feedback from you regarding the `from` property - see discussions!

PR #4988, opened 2018-12-13
Make applyPresets function signature more consistent

pksunkara:
Just something I wanted to fix while reading this part of the code


igor-dv (approved):

ndelangen:
The test failing in TeamCity aren't failing on my local machine 😞

PR #5007, opened 2018-12-15
Make to work on create react app v2 storybook v4.1.1

pavinthan:
To fix 
```ERR! Error: Cannot find module '/{path}/node_modules/{react-scripts-provider}/react-scripts/config/webpack.config.dev'
ERR!     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:603:15)```

### DIFF
--- a/app/react/src/server/cra-config.js
+++ b/app/react/src/server/cra-config.js
@@ -83,12 +83,12 @@ export function getCraWebpackConfig(mode) {
 
   const craWebpackConfig =
     mode === 'production' ? 'config/webpack.config.prod' : 'config/webpack.config.dev';
-  
+
   const craWebpackConfigFile = path.join(pathToReactScripts, craWebpackConfig);
 
   let pathToWebpackConfig;
   if (fs.existsSync(craWebpackConfigFile)) {
-      pathToWebpackConfig = require.resolve(craWebpackConfigFile)
+    pathToWebpackConfig = require.resolve(craWebpackConfigFile);
   }
 
   if (!fs.existsSync(pathToWebpackConfig)) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by kylemh** [comment]
Confused about this...

I'm using the same versions you are and am having no issue.

**Comment by pavinthan** [comment]
@kylemh CRA v2 have move `webpack.config.{env}.js` to `webpack.config` so `require.resolve(path.join(pathToReactScripts, craWebpackConfig)) ` will throw error so just added a condition only if `fs.existsSync(path.join(pathToReactScripts, craWebpackConfig))` exists resolve the file otherwise It's get `path.join(pathToReactScripts, 'config/webpack.config')`


this case could apply for custom react-scripts as well, please advise me further 🙏

File: app/react/src/server/cra-config.js [hunk 1]
```diff
@@ -83,8 +83,13 @@ export function getCraWebpackConfig(mode) {
 
   const craWebpackConfig =
     mode === 'production' ? 'config/webpack.config.prod' : 'config/webpack.config.dev';
+  
+  const craWebpackConfigFile = path.join(pathToReactScripts, craWebpackConfig);
 
-  let pathToWebpackConfig = require.resolve(path.join(pathToReactScripts, craWebpackConfig));
+  let pathToWebpackConfig;
+  if (fs.existsSync(craWebpackConfigFile)) {
```
igor-dv: at this point `craWebpackConfigFile` is not a real path and fs can't really find it. I think that's why all the tests are failing.

appbak3r:
Having the same issue #5077

mrmckeb:
This is a duplicate of #5074.

PR #5011, opened 2018-12-16
Add TypeScript CRA example

### LINKED ISSUE #5006: CRA2 TypeScript example
Would be really helpful for many newcomers to have some good base example for CRA + TypeScript case as it's getting to be very popular in react community.

I can send a PR if you agree with including it 👍

mucsi96:
Issue:

## What I did

I have created an example incorporating CRA2, TypeScript, TSLint, Jest, Enzyme

image

Closes #5006


mucsi96:
@shilman Can you help me with broken build? Looks like its totally not related to the changes I have made. Would be very helpful to move this forward

ndelangen (approved):

igor-dv:
This PR is containing changes from wrong commits (probably needs a merge from `next`).
IMO we don't need this as an example app, considering maintenance cost. Better having it as a CLI test. If this more like having a working example, it can also be done out of the monorepo..

mucsi96:
I can see that storybook itself is migrating over TypeScript. Its a very popular setup. Because of that I thought it makes sence to have a similar example setup as well. It looks easy now but I spent half a day to make it working as it required quite a bit of investigation. IMO if this is out of monorepo less people will find it. What do you think?

igor-dv:
@mucsi96, mind solve the conflicts + fix the linter?

PR #5015, opened 2018-12-17
Include macro instructions for Storyshots docs

kylemh:
Issue:

## Background

Setting up Storybook in CRA2 with Storyshots is currently impossible without ejecting and using a plugin, or inlining a function which could be made into a library.

https://github.com/storybooks/storybook/issues/4479 seems to cover the pain well.

I've published a macro for CRA2 users to easily mock/polyfill `require.context` in a test environment, and updated the documentation for Storyshots pointing out how to do so.

## How to test

- `npx create-react-app test`
- Setup Storybook in `test`
- Integrate Storyshots addon
- Follow my instructions (in this PR)


tmeasday (approved):
LGTM!

PR #5016, opened 2018-12-17
ADD alias to react & react-dom

### LINKED ISSUE #5014: Storybook 4.1 renders using SB's `react-dom`, not the app's
**Describe the bug**
`@storybook/react` calls `ReactDOM.render` using the version of `react-dom` (16.6) storybook ships with, not the version that the user's app uses. This means if the app uses `react-dom` internally (e.g. `ReactDOM.findDOMNode()`), all heck breaks loose:

```
Unable to find node on an unmounted component.
    invariant@http://localhost:7000/vendors~main.b5d3e831a4562a230f16.bundle.js:43768:24
findCurrentFiberUsingSlowPath@http://localhost:7000/vendors~main.b5d3e831a4562a230f16.bundle.js:47871:39
<snip>
```

**To Reproduce**
1. Check out https://github.com/tmeasday/sb-react-dom-version-mismatch
2. Run the storybook. Observe error

Not this is a modified version of CRA that depends on `react-dom@16.5`. If you allow the app to use `react-dom@16.6` (and thus share it with SB), the problem does not occur.

**Expected behavior**
The preview window should render with the user's version of `react-dom`

**Screenshots**
image

ndelangen:
fixes: https://github.com/storybooks/storybook/issues/5014

### DIFF
--- a/lib/core/package.json
+++ b/lib/core/package.json
@@ -71,6 +71,7 @@
     "redux": "^4.0.1",
     "regenerator-runtime": "^0.12.1",
     "resolve": "^1.8.1",
+    "resolve-cwd": "^2.0.0",
     "semver": "^5.6.0",
     "serve-favicon": "^2.5.0",
     "shelljs": "^0.8.2",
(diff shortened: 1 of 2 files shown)


tmeasday (approved):
Seems like the right solution. Let me try it in the next alpha

tmeasday:
@ndelangen it seems like tests are failing?

PR #5018, opened 2018-12-17
Typescript-Migration: @storybook/addons

kroeder:
Issue: N/A

## What I did

Working on the TS-Migration with @ndelangen and @igor-dv


tsconfig.json
```diff
@@ -9,13 +9,15 @@
     "noImplicitAny": true,
     "jsx": "react",
```
kroeder: jsx allows `react` or `react-native` 
Should there be a check for special react native build steps?

like `if(buildReactNative) { tsconfig.jsx = 'react-native'; }` ? 

See https://www.typescriptlang.org/docs/handbook/jsx.html

scripts/prepare.js
```diff
@@ -38,8 +38,12 @@ function logError(type, packageJson) {
...
-babelify({ errorCallback: () => logError('js', packageJson) });
-removeTsFromDist();
-tscfy({ errorCallback: () => logError('ts', packageJson) });
+if (packageJson && packageJson.types && packageJson.types.indexOf('d.ts') !== -1) {
+  tscfy({ errorCallback: () => logError('ts', packageJson) });
+} else {
+  babelify({ errorCallback: () => logError('js', packageJson) });
+  removeTsFromDist();
+  tscfy({ errorCallback: () => logError('ts', packageJson) });
+}
```
kroeder: Pure ts projects may cause unexpected behavior when it runs through babelify as well
In my case babelify copied `tsx` files into dist

package.json
```diff
@@ -18,7 +18,7 @@
     "bootstrap": "node ./scripts/bootstrap.js",
     "bootstrap:crna-kitchen-sink": "npm --prefix examples-native/crna-kitchen-sink install",
     "bootstrap:docs": "yarn install --cwd docs",
-    "build-packs": "lerna exec --scope '@storybook/*' --parallel -- \\$LERNA_ROOT_PATH/scripts/build-pack.sh \\$LERNA_ROOT_PATH/packs",
+    "build-packs": "lerna exec --scope '@storybook/*' -- \\$LERNA_ROOT_PATH/scripts/build-pack.sh \\$LERNA_ROOT_PATH/packs",
```
kroeder: I had to remove --parallel in order to fix `native-smoke-test`
The ts build of `addons` always failed because it seemed that `channels` wasn't build but `addons` relied on `channels` so it could not build as well

I found this  https://github.com/lerna/lerna/issues/1681 and thought "give it a shot", though I'm not sure if this is okay? I spent a lot of time on fixing this, if this is not the right fix I definitely need help for another way.

ndelangen: 🆗

ndelangen (approved):

lib/addons/src/make-decorator.ts
```diff
@@ -1,23 +1,41 @@
...
+
+export type StoryGetter = (context: StoryContext) => any;
+
+export type StoryWrapper = (
+  getStory: StoryGetter,
+  context: StoryContext,
+  settings: WrapperSettings
+) => any;
+
+type MakeDecoratorResult = (...args: any) => any;
```
leoyli: I believe this line wasn't correctly typed actually... It's not simply saying there is no type (`any`) but TSLint should point out an error saying args must be an array, so it should be `any[]`.

PR #4885, opened 2018-11-29
First POC of example file format... fun times

tmeasday:
Issue: #4848

This branch is a proof-of-concept / experimentation with the example format as documented [here](https://gist.github.com/tmeasday/4de20eab47226a870ab1025642ba848c) and discussed in #4848.

## What I did

1. Added a `importAllExamples` that allows you to use the example format from https://gist.github.com/tmeasday/4de20eab47226a870ab1025642ba848c in official example.

2. Converted the first example over :)

## How to test

Run official storybook for now

Issue #4848, opened 2018-11-26
New story file format


tmeasday:
As part of a push towards [a better documentation mode](
https://github.com/storybooks/storybook/issues/4341), and to make stories more general and more easily consumed by other tools (thus making things like a [Docz plugin](https://github.com/hydrateio/docz-plugin-storybook) easier), we are considering a new file format for stories that is less tied to SB APIs:

The current proposal is a file `Component.examples.js`, which in it's simplest form looks like:

```js
import React from 'react';

import Component from 'somewhere';

export default Component;

export const variant1 = () => <Component variant="1" />;
export const variant2 = () => <Component variant="2" />;
export const variant3 = () => <Component variant="3" />;
````

More detail is proposed [in this gist](https://gist.github.com/tmeasday/4de20eab47226a870ab1025642ba848c) as well as some discussion with @igor-dv and @transitive-bullshit with regards to some thoughts about the format.

The primary use cases we are considering include (please reach out if you have others that are relevant):

1. Defining a set of components and stories to appear inside storybook / docs automatically
2. Making it easy to render one story, or the metadata of one component at a time inside a `mdx` file.
3. Making it easy to re-use a story to drive a test case in a tool like Jest.
4. Making it easy for external tools to catalogue a list of stories inside a project w/o needing the storybook runtime.

If you have other proposals, this is the place to reach out with them. Some alternatives:

## Proposal B: A more modern view on the current format 
[by @transitive-bullshit here](https://gist.github.com/tmeasday/4de20eab47226a870ab1025642ba848c#gistcomment-2767279):

```js
import { Story, Stories } from '@storybook/react'
import Button from './Button'

export default new Stories(Button, [
  new Story('with text', () => (
    <Button>Hello Button</Button>
  )),

  new Story('with emoji', () => (
    <Button>
      <span role='img' aria-label='so cool'>
        :+1: 
      </span>
    </Button>
  ))
])
```

donaldpipowitch:
> As part of a push towards [a better documentation mode](#4341)

FYI: The link points to this issue, because #4341 is treated as an anchor.

tmeasday:
After talking with @transitive-bullshit about his format's pros+cons we've agreed to have a go with the format in the gist to get a feel for it. PR/branch is here: https://github.com/storybooks/storybook/pull/4885 -- all welcome to get involved / make suggestions

PR #5058, opened 2018-12-20
Use new story format in official storybook

tmeasday:
Moving #4885 over here -- for #4848.

As we work on the new UI, let's write our stories in the new format.

@ndelangen would you be interested in pairing on improving this `importAll` function? How's your HMR? This works as is but makes use of `window` and is a bit of a hack.


examples/official-storybook/config.js
```diff
@@ -60,8 +62,60 @@ addParameters({
   },
 });
 
-function importAll(req) {
-  req.keys().forEach(filename => req(filename));
+// TODO -- somehow don't use window for this cross HMR storage
+const { previousExports = {} } = window; // eslint-disable-line no-undef
```
ndelangen: You can use module.hot.data:

https://webpack.js.org/api/hot-module-replacement/#dispose-or-adddisposehandler-

```
dispose (or addDisposeHandler) 
Add a handler which is executed when the current module code is replaced. This should be used to remove any persistent resource you have claimed or created. If you want to transfer state to the updated module, add it to given data parameter. This object will be available at module.hot.data after the update.

module.hot.dispose(data => {
  // Clean up and pass data to the updated module...
});
```

tmeasday: OK, that worked, thanks @ndelangen

tmeasday:
So I think this is OK to merge. In terms of a user API, I think we should either export a function:

```js
importAll(require.context(...), module);
```

or modify the configure API to also accept an array of contexts:

```js
configure([ require.context(...), require.context(...)], module);
```

What do you think @ndelangen?

ndelangen (approved):

examples/official-storybook/stories/core/parameters.stories.js
```diff
@@ -0,0 +1,28 @@
...
+export default {
+  title: 'Core|Parameters',
+  decorators: [fn => fn({ parameters: { decoratorParameter: 'decoratorParameter' } })],
+  parameters: {
+    chapterParameter: 'chapterParameter',
+  },
+};
+
+// I'm not sure what we should recommend regarding propTypes? are they a good idea for examples?
+// Given we sort of control the props, should we export a prop type?
```
Hypnosphi: Technically, story itself is not a React component: decorators call it as a function, and propTypes have no effect in that case

ndelangen: In react they are eventually invoked as a React Component in the renderer though?

Hypnosphi: Not the inner story function but `decoratedStory`

PR #5079, opened 2018-12-24
Update CRA Webpack configuration path

coka:
Issue: #5077

## What I did

I have updated `cra-config.js` to handle the [Webpack configuration merge](https://github.com/facebook/create-react-app/pull/5722) in CRA 2.1.2. This breaks projects running an older version of CRA. Should `cra-kitchen-sink` be updated to reflect this?

Related PR: #5074

## How to test

Run Storybook in a project with version 2.1.2 of `react-scripts` in its dependencies.

### DIFF
--- a/app/react/src/server/cra-config.js
+++ b/app/react/src/server/cra-config.js
@@ -81,7 +81,14 @@ export const getTypeScriptRules = (webpackConfigRules, configDir) => {
 export function getCraWebpackConfig(mode) {
   const pathToReactScripts = getReactScriptsPath();
 
-  const craWebpackConfig = 'config/webpack.config';
+  let craWebpackConfig = 'config/webpack.config';
+
+  // Needed for backward compatibility.
+  // https://github.com/facebook/create-react-app/pull/5722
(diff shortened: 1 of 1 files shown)


app/react/src/server/cra-config.js
```diff
@@ -81,8 +81,7 @@ export const getTypeScriptRules = (webpackConfigRules, configDir) => {
 export function getCraWebpackConfig(mode) {
   const pathToReactScripts = getReactScriptsPath();
 
-  const craWebpackConfig =
-    mode === 'production' ? 'config/webpack.config.prod' : 'config/webpack.config.dev';
+  const craWebpackConfig = 'config/webpack.config';
```
appbak3r: I guess I shouldn't do that due to backward compatibility with older versions of CRA and react-scripts

coka: I've added a check for this. Let's see if CI passes.

mrmckeb:
This is a duplicate of #5074.

PR #5101, opened 2018-12-26
Fix a11y addon, add ability to change context

### LINKED ISSUE #4889: addon-a11y not working on latest next
**Describe the bug**
`@storybook/addon-a11y` not reporting a11y test results on `next` branch . It shows `0 passes` and `0 violations` on any story regardless of decorator being set or not.

**To Reproduce**
1. Checkout `next` branch
2. Start the `official-storybook` kitchen sink app or install/register add-on to any other app.
2. Open any of the `a11y` stories with decorator set.
3. Go to `Accessibility` panel.
4. See 0 results on both passes/violations.

**Expected behavior**
Should display amount of test passes or violations respectively in each tab.

**Screenshots**
screen shot 2018-11-29 at 10 19 17 am

**Code**
It's related to the `runA11yCheck` function on the addon's declaration and the DOM being selected to run the aXe tests on:

```js
const runA11yCheck = () => {
  const infoWrapper = document.getElementById('story-root').children;
  const wrapper = document.getElementById('root');

  axe.reset();
  axe.configure(axeOptions);
  axe
    .run(infoWrapper || wrapper)
    .then(results => channel.emit(CHECK_EVENT_ID, results), logger.error);
};
```

**System:**
 - Browser: any
 - Framework: all
 - Addons: a11y
 - Version: 4.0.9

**Additional context**
Tried this on some of the other kitchen sink apps with same result.

igor-borisov:
Issue:#4889

A11y addon requires to have Info addon and also info add-on must have settings 

```
// info addon settings
{
        inline: true, // where the components are inlined
        source: false,
}
``` 

## What I did
So I removed this requirement and also add ability to configure context for a11y checking.

### DIFF
--- a/addons/a11y/src/index.js
+++ b/addons/a11y/src/index.js
@@ -6,22 +6,28 @@ import { logger } from '@storybook/client-logger';
 
 import { CHECK_EVENT_ID, REQUEST_CHECK_EVENT_ID } from './shared';
 
-let axeOptions = {};
+const config = {
+  axeOptions: {},
+  contextElementId: '',
+};
 
 export const configureA11y = (options = {}) => {
-  axeOptions = options;
+  Object.assign(config, options);
 };
 
 const runA11yCheck = () => {
   const channel = addons.getChannel();
-  const infoWrapper = document.getElementById('story-root').children;
-  const wrapper = document.getElementById('root');
+
+  let axeContext = config.contextElementId;
+  if (!axeContext) {
+    const infoWrapper = document.getElementById('story-root');
+    const wrapper = document.getElementById('root');
+    axeContext = (infoWrapper && infoWrapper.children) || wrapper;
+  }
 
   axe.reset();
-  axe.configure(axeOptions);
-  axe
-    .run(infoWrapper || wrapper)
-    .then(results => channel.emit(CHECK_EVENT_ID, results), logger.error);
+  axe.configure(config.axeOptions);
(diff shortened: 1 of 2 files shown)


igor-dv (commented):
Currently, this is a breaking change. Wanna think of backward compatibility? Otherwise, it will be released only with v5

addons/a11y/src/index.js
```diff
@@ -6,22 +6,29 @@ import { logger } from '@storybook/client-logger';
...
+  config.axeOptions = axeOptions;
+  config.contextElementId = contextElementId;
 };
 
 const runA11yCheck = () => {
   const channel = addons.getChannel();
-  const infoWrapper = document.getElementById('story-root').children;
-  const wrapper = document.getElementById('root');
+
+  let axeContext = config.contextElementId;
```
igor-dv: let's extract the `axeContext` discovery to a fnction getAxeContext() or something like this.

igor-dv (approved):

dallasclark:
I see the info addon requires `{ inline: true }`, is it possible to enable a11y without having to display info inline?

Thanks !

igor-borisov:
> I see the info addon requires `{ inline: true }`, is it possible to enable a11y without having to display info inline?
> 
> Thanks !

Not for actual version of a11y addon, but after approving this pull request we can use it without inline info option. I had the same issue and trying to fix it by this set of commits.

PR #5107, opened 2018-12-27
import utils directly to prevent import cycles

koenpunt:
when using `@storybook/addon-actions` in a React Native project there constantly is a warning about an import cycle.

By importing the util function directly this cycle is broken, and thus makes the warning disappear.

~**Edit**; while working on this, it turns out there's more than one cycle. My fix is not sufficient yet, and I think I should take a different approach, where I make sure the utils do not import the files causing the cycle instead of the other way around~

I've updated all imports to be direct, instead of using the index, which prevents the cycle.


rhalff (approved):
Seems fine to me, thanks!

ndelangen (approved):

Issue #5069, opened 2018-12-21
Storyshots: Unable to run async job before initStoryshots


gidztech:
**Describe the bug**
I am trying to connect to an existing Chromium instance with [addon-storyshots-puppeteer](https://www.npmjs.com/package/@storybook/addon-storyshots-puppeteer#specifying-a-custom-puppeteer-browser-instance). The docs provide the following snippet, which is meant to achieve this.

```javascript
import initStoryshots from '@storybook/addon-storyshots';
import { imageSnapshot } from '@storybook/addon-storyshots-puppeteer';
import puppeteer from 'puppeteer';
 
(async function() {
    const customBrowser = await puppeteer.connect('ws://yourUrl');
 
    initStoryshots({
      suite: 'Image storyshots',
      test: imageSnapshot({ storybookUrl: 'http://localhost:6006', customBrowser }),
    });
})();
```
First of all, the [`puppeteer.connect`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerconnectoptions) function takes an object, where you pass `browserWSEndpoint`, not  just a string. However, once that is corrected there's another problem.

When you run the test, Jest will not wait for the async operation to complete and will attempt to run test blocks before they are created. This results in the error `Your test suite must contain at least one test.`, followed by some more errors thrown in `initStoryshots` as a consequence.

My next attempt was to use a `beforeAll` hook:

```javascript
describe('Storyshots', () => {
    let testFn;

    beforeAll(async () => {
        const customBrowser = await puppeteer.connect({ browserWSEndpoint: 'ws://yourUrl' });
        testFn = imageSnapshot({
            storybookUrl: 'http://localhost:6006',
            customBrowser
        });
    });

    initStoryshots({
        framework: 'react',
        suite: 'Storyshots',
        test: testFn
    });
});
```
However, this still doesn't work and the test blocks that are generated are executed before the `beforeAll` hook has completed. 

**Expected behavior**
There should be a way to run an async job before `initStoryshots` is executed, so that we can pass in config to `initStoryshots`. In this case, using an already running instance of Chromium.

**System:**
 - OS: MacOS
 - Device: MacBook Pro 2014
 - Browser: N/A
 - Framework: React (but not important to this problem)
 - Addons: [addon-storyshots, addon-storyshots-puppeteer]
 - Version: [4.1.2]

gidztech:
[...]

PR #5130, opened 2019-01-02
Storyshots: Fix async issue with obtaining custom Puppeteer instance …

gidztech:
…(#5069#)

Issue: The `customBrowser` config option for storyshots-puppeteer doesn't work because Jest doesn't wait for async functions to complete in test files that aren't part of the actual test function. You will get a `Your test suite must contain at least one test.` error because it doesn't find the tests in the current stack.

## What I did
Instead of trying to do the async operation outside of the test function, we can pass the function to `initStoryshots` to later be called inside the `beforeAll` hook, where Jest will happily wait.

I changed `customBrowser` to `getCustomBrowser`, which is an async function that returns the `browser` instance. If that function isn't provided in config, the normal launch behaviour remains.

### DIFF
--- a/addons/storyshots/storyshots-puppeteer/README.md
+++ b/addons/storyshots/storyshots-puppeteer/README.md
@@ -148,7 +148,7 @@ initStoryshots({
 
 ### Specifying a custom puppeteer `browser` instance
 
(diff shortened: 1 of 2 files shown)


ndelangen:
@gidztech I think this is a breaking change, right?

ndelangen (approved):

gidztech:
@ndelangen Technically a breaking change because of the API change, but it wasn't actually working before.

ColCh:
@ndelangen this is a bug fix.

For now, it's impossible to use custom puppeteer instance

PR #5150, opened 2019-01-05
FIX highlightColor

ndelangen:
Issue:
screen_shot_2018-12-21_at_10 32 42_am_480

after fix:
screenshot 2019-01-05 at 12 03 04
screenshot 2019-01-05 at 12 01 17

### DIFF
--- a/lib/ui/src/modules/ui/components/stories_panel/stories_tree/tree_decorators.js
+++ b/lib/ui/src/modules/ui/components/stories_panel/stories_tree/tree_decorators.js
@@ -103,7 +103,7 @@ ContainerDecorator.defaultProps = {
 };
 
 function HeaderDecorator(props) {
-  const { style, node, ...restProps } = props;
+  const { style, node, theme, ...restProps } = props;
 
   let newStyle = style;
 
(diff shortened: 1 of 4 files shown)


Hypnosphi (approved):

Hypnosphi:
I can't spot a difference in failing screenshots. Is it a Chromatic bug?

shilman:
@ndelangen do we need to patch this into `next`?

ndelangen:
yes

lib/ui/src/modules/ui/components/stories_panel/stories_tree/tree_decorators.js
```diff
@@ -114,7 +114,7 @@ function HeaderDecorator(props) {
     };
   }
 
-  const name = highlightNode(node, style);
+  const name = highlightNode(node, style, theme.highlightColor, 'black');
```
Hypnosphi: Can you please make text color customizable here as well? `black` doesn't play well with our `highlightColor` =(
screenshot 2019-01-11 at 20 25 19

PR #5151, opened 2019-01-05
Migrate @storybook/client-logger to TypeScript

dandean:
Issue: #5030 

## What I did

Migrated `@storybook/client-logger` to TypeScript

## How to test

Existing tests were retained in `lib/client-logger/src/index.test.ts`


ndelangen (approved):

lib/client-logger/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "license": "MIT",
   "main": "dist/index.js",
-  "jsnext:main": "src/index.js",
+  "types": "src/index.d.ts",
```
kroeder: This should be `dist/index.d.ts` I think?

dandean: Fixed this – thanks for pointing it out.

kroeder (approved):

PR #5207, opened 2019-01-10
UI story_panel toggle bug when multiple stories has same name and path

yjcxy12:
Issue:
When two stories with different root has same namespace and name, toggle one will toggle another. This causes problems when component libraries choose to render same component with different theme, and separate theme with different roots.

It is reproducible in `official-storybook` example. If `Core|Events.Force re-render` story is clicked, `Addons|Events` opens.

## What I did
Pass down `selectedRoot` to `story_panel` component, and pass down `selectedHierarchy` to `story_tree` component only if the `selectedRoot` matches the root.

## How to test

Yes.
No.
No.

### DIFF
--- a/lib/ui/src/modules/ui/components/stories_panel/__snapshots__/index.stories.storyshot
+++ b/lib/ui/src/modules/ui/components/stories_panel/__snapshots__/index.stories.storyshot
@@ -363,8 +363,8 @@ exports[`Storyshots UI|stories/StoriesPanel when open on mobile device 1`] = `
 </div>
 `;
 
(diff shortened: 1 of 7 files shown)


tmeasday:
Thanks so much for this PR @yjcxy12. Actually this code has been reworked in the `release/5.0` which will be merged to `next` imminently. 

Any chance you could check to see if the bug still applies on that branch or perhaps reapply this change to that branch? Sorry to have led you down a smal deadend here

yjcxy12:
@tmeasday This appears to be fixed in `release/5.0` branch.

Is there a point to merge this into v4 as hotfix? If not, please feel free to close this.

tmeasday:
@shilman should we patch this to the 4.x branch?

PR #5210, opened 2019-01-10
Add @emotion/core dependency to addon-a11y

BPScott:
## What I did

Fix yarn complaining about a missing peer dependency 

`warning "@storybook/addon-a11y > @emotion/styled > @emotion/styled-base@0.10.6"
has unmet peer dependency "@emotion/core@0.x.x".`

The same sort of problem as #5135 but for a different package.

## How to test

N/A


BPScott:
Looks like #5241 has superseeded this

PR #5212, opened 2019-01-11
Add explicit return types to @storybook/addons

dandean:
Issue: #5130 

## What I did

This patch adds explicit types to all functions in `lib/addons/index.ts` (aka, `@storybook/addons`).

I was working on migrating a different module to TypeScript when I found that the `lib/addons` module was providing an object which was typed as `any` instead of `AddonStore`. This made it a bit difficult to use the module without having to go poke through source code.

Rather than just adding the type to that one export I added types to all of the functions in the module, even when `void`. Making the types explicit like this helps ensure that someone doesn't accidentally change the return type as the code is evolved.

## How to test

Not really testable; we just get better type information when using the module.


ndelangen:
```
Summary of all failing tests
 FAIL  addons/notes/src/__tests__/index.ts
  ● Test suite failed to run

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    addons/notes/src/__tests__/index.ts:9:23 - error TS2339: Property 'mockReturnValue' does not exist on type '() => Channel'.

    9     addons.getChannel.mockReturnValue(channel);
                            ~~~~~~~~~~~~~~~
    addons/notes/src/__tests__/index.ts:21:23 - error TS2339: Property 'mockReturnValue' does not exist on type '() => Channel'.

    21     addons.getChannel.mockReturnValue(channel);
                             ~~~~~~~~~~~~~~~
    addons/notes/src/__tests__/index.ts:33:23 - error TS2339: Property 'mockReturnValue' does not exist on type '() => Channel'.

    33     addons.getChannel.mockReturnValue(channel);
                             ~~~~~~~~~~~~~~~
    addons/notes/src/__tests__/index.ts:48:23 - error TS2339: Property 'mockReturnValue' does not exist on type '() => Channel'.

    48     addons.getChannel.mockReturnValue(channel);
                             ~~~~~~~~~~~~~~~
```

Is a jest mocking compatibility issue with jest & typescript.

What's the recommended approach here?

ndelangen (approved):

PR #5236, opened 2019-01-14
FIX issue with accidentally thinking things are roots

tmeasday:
Issue: #5227 

This was a mix up in the change of format for the story store


shilman (approved):
Thanks! 🙇

PR #5241, opened 2019-01-14
MIGRATE to wrap emotion into a @storybook/theming package

ndelangen:
- This ensures a single version
- This adds a good place to add themes
- Addons can take a dependency on something storybook
- We could migrate to another styling lib without asking all users to migrate with us


tmeasday (approved):
LGTM. Should we include this in 5.0? Probably just about in time to scrape in before RC / feature freeze

ndelangen:
Yes, should be in 5.0.0

I also plan on moving the router into it's own package too.

shilman (commented):
@ndelangen Why did all the storyshots change?

Do you have anything else in mind before we feature freeze 5.0 and restrict PRs to bugfixes only (aside from the RN core refactor, which I know is going to take a little time cc @Gongreg @igor-dv)?

ndelangen:
@shilman the storyshots change because the emotion jest snapshot thing no longer functions.

I personally think this is not a HUGE deal, though I'd like to fix it in the future. For the sake of debuggability.

PR #5252, opened 2019-01-15
REFACTOR router into @storybook/router

ndelangen:
- ensure we can migrate to a different router without updating all users
- ensure single version
- easier api
- separate from @storybook/component makes it lighter


lib/router/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+  location: object;
+  navigate: (to: string) => void;
+  viewMode?: string;
+  storyId?: string;
+}
+interface MatchingData {
+  match: null | { path: string };
+}
+
+interface QueryLocationType {
```
apastuhov: If an interface describes prop or state - add it as a suffix. So you will differentiate interfaces of objects/models and interfaces which describes communications between components.

lib/router/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+  children: (a: RenderData) => React.ReactNode;
+}
+interface QueryMatchType {
+  path: string;
+  startsWith: boolean;
+  children: (a: MatchingData) => React.ReactNode;
```
apastuhov: `child` or something else instead of `a` will increase readability

lib/router/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+  startsWith: boolean;
+  hideOnly: boolean;
+  children: (a: RenderData) => React.ReactNode;
+}
+
+const queryNavigate = (to: string) => {
+  navigate(`?path=${to}`);
+};
+
+const QueryLink = ({ to, children, ...rest }: { to: string; children: React.ReactNode }) => (
```
apastuhov: Use type for `{ to: string; children: React.ReactNode }`

lib/router/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+QueryLocation.displayName = 'QueryLocation';
+
+const QueryMatch = ({ children, path: targetPath, startsWith = false }: QueryMatchType) => (
+  <QueryLocation>
+    {({ path: urlPath, ...rest }) => {
+      let match;
+      if (!urlPath) {
+        return null;
+      }
+      if (startsWith) {
```
apastuhov: maybe refactor all if conditions and extract them to some function?

lib/router/src/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const storyDataFromString: (path: string) => StoryData = memoize(1000)((path: string | undefined | null) => {
+  const result: StoryData = {
+    viewMode: undefined,
+    storyId: undefined,
+  };
+
+  if (path) {
+    const [, p1, p2] = path.match(/\/([^/]+)\/([^/]+)?/) || [undefined, undefined, undefined];
```
apastuhov: Extract regexp to constant
```
const reg = /\/([^/]+)\/([^/]+)?/;
const [, p1 = undefined, p2 = undefined] = path.match(reg);
```

apastuhov: And give proper names `viewMode` and `storyId`;

apastuhov (changes requested):

lib/router/src/router.tsx
ndelangen: ✅

ndelangen: Should I place the interfaces in a different file?

lib/router/src/router.tsx
ndelangen: ✅

lib/router/src/utils.ts
ndelangen: ✅

tmeasday (changes requested):
LGTM apart from readme problem

lib/router/README.md
```diff
@@ -0,0 +1,6 @@
+# Storybook Theming
```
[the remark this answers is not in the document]
ndelangen: fixed it and added a bunch of docs

examples/official-storybook/package.json
```diff
@@ -9,7 +9,7 @@
     "generate-addon-jest-testresults": "jest --config=tests/addon-jest.config.json --json --outputFile=stories/addon-jest.testresults.json",
     "graphql": "node ./graphql-server/index.js",
     "image-snapshots": "yarn run build-storybook && yarn run do-image-snapshots",
-    "storybook": "cross-env STORYBOOK_DISPLAY_WARNING=true DISPLAY_WARNING=true start-storybook -p 9011 -c ./ -s built-storybooks"
+    "storybook": "cross-env STORYBOOK_DISPLAY_WARNING=true DISPLAY_WARNING=true start-storybook -p 9011 -c ./ -s built-storybooks --no-dll"
```
tmeasday: How did this sneak into this PR? Not a big deal I guess

lib/ui/src/components/nav/bar.js
```diff
@@ -36,7 +36,7 @@ const BarLi = styled.li(({ active, theme }) => ({
 
 export const BarItem = ({ path, children, active }) => (
   <BarLi active={active}>
-    <Router.Link to={path}>{children}</Router.Link>
+    <Link.Link to={path}>{children}</Link.Link>
```
tmeasday: Should this just be `Link`? It is below

tmeasday (changes requested):

ndelangen:
Good to merge then?

lib/router/src/router.tsx
apastuhov: If you plan to export interfaces and reuse them in other source files - then "yes", otherwise - if an interface is a private to a component - no need to move.

lib/router/README.md
```diff
@@ -0,0 +1,68 @@
+# Storybook Router
+
+Storybook Router is a wrapper library for reach/router.
+It ensures a single version of the router is used everywhere.
+
+It also includes some ready to use utils to read the path, query, viewMode and storyId from location.
+
+It also includes Components:
```
apastuhov: Do you need all that documentation for properties? If you have interfaces and definitions, seems like a duplicating. Also it will be better to write JSDoc for attributes. For example, this one is not obvious from code:

```
- **children** - a function that will receive a location when it changes
```

lib/router/src/router.tsx
```diff
@@ -0,0 +1,109 @@
...
+  navigate: (to: string) => void;
+  viewMode?: string;
+  storyId?: string;
+}
+interface MatchingData {
+  match: null | { path: string };
+}
+
+interface QueryLocationProps {
+  children: (a: RenderData) => React.ReactNode;
```
apastuhov: rename

lib/router/src/router.tsx
```diff
@@ -0,0 +1,109 @@
...
+interface QueryMatchProps {
+  path: string;
+  startsWith: boolean;
+  children: (matchingData: MatchingData) => React.ReactNode;
+}
+interface RouteProps {
+  path: string;
+  startsWith: boolean;
+  hideOnly: boolean;
+  children: (a: RenderData) => React.ReactNode;
```
apastuhov: rename

lib/router/README.md
apastuhov: The same for component documentation? Why do not keep it as a part of the component?

A clickable element that will `navigate` to a new location/path.

lib/router/src/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const queryFromString = memoize(1000)(s => qs.parse(s, { ignoreQueryPrefix: true }));
+export const queryFromLocation = (location: { search: string }) => queryFromString(location.search);
+export const stringifyQuery = (query: object) => qs.stringify(query, { addQueryPrefix: true, encode: false });
+
+export const getMatch = memoize(1000)((current: string, target: string, startsWith: boolean = true) => {
+  if (!current) {
+    return null;
+  }
+  if (startsWith) {
```
apastuhov: I can be wrong, but I would rewrite it like:
```
const startsWithTarget = startsWith && current.startsWith(target);
const currentIsTarget = typeof target === 'string' && current === target;
const matchTarget = target && current.match(target);

if (startsWithTarget || currentIsTarget || matchTarget) {
    return { path: current }
}
return null;
```

apastuhov (dismissed):
Few moments to improve, I think it will be the last feedback from my side.

lib/router/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const QueryLink = ({ to, children, ...rest }: { to: string; children: React.ReactNode }) => (
+  <Link to={`?path=${to}`} {...rest}>
+    {children}
+  </Link>
+);
+QueryLink.displayName = 'QueryLink';
+
+const QueryLocation = ({ children }: QueryLocationType) => (
```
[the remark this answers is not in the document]
apastuhov: I got your point, thanks for the explanation. Maybe I've just used to write 1 component per 1 file with a bunch of documentation.

tmeasday:
@ndelangen this seems like a bug? 
https://github.com/storybooks/storybook/pull/5252#discussion_r248122476

tmeasday (approved):

lib/router/src/router.tsx
ndelangen: ✅

lib/router/src/router.tsx
ndelangen: ✅

lib/router/src/utils.ts
ndelangen: ✅

lib/router/README.md
```diff
@@ -0,0 +1,68 @@
+# Storybook Router
+
+Storybook Router is a wrapper library for reach/router.
```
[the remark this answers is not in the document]
ndelangen: what's the typo? 😄

lib/router/README.md
ndelangen: I guess because we link to this readme, and so a search on the readme or github will show up this useful info?

If you think this documentation is redundant, because the typings already cover this info; should I just remove it?

ndelangen: I moved the docs into code comments

ndelangen:
Whitespace in toolbar is ugly, should fix

shilman (approved):
Awesome, looks great aside from the visual regression that you already noted. 👍 

As a side note, and as a separate PR, it would be great to move `id.js` and `pathToId.js` and their associated tests into this package.

ndelangen:
Going to fix the whitespace bu, then merge

lib/router/README.md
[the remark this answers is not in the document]
apastuhov: reach? I thought here should be `react`?)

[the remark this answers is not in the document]
ndelangen: nope :) that's a different package

PR #5257, opened 2019-01-16
FIX storyshots using the wrong parameter for disabling

tmeasday:
It should be `disable: true` not `disabled: true`.

(I think this was my fault, sorry)


Hypnosphi:
Is it a breaking change?

tmeasday:
I think this is unreleased, no?

shilman (approved):
LGTM

tmeasday:
Just to confirm the feature has not yet been released: https://github.com/storybooks/storybook/commit/e0dee7eb5c5b5f93ea41bbd052847d3cac49a105

PR #5289, opened 2019-01-18
Add a persistence API to context/state -- use for keyboard shortcuts

tmeasday:
For #5271 

This PR implements the persistence API for core and demonstrates it working for keyboard shortcuts.


lib/ui/src/core/store.js
```diff
@@ -1,33 +1,50 @@
...
-  };
-};
-
-export default createStore;
+  setState(patch, persistence = 'none') {
+    this.upstreamSetState(patch);
+
+    if (persistence !== 'none') {
+      const storage = persistence === 'session' ? sessionStorage : localStorage;
+      update(storage, patch);
```
ndelangen: I think some throttling needs to happen.

Writing to localStorage is an expensive, synchronous, blocking action!
And so is JSON.stringify.
Is our state a serialisable object btw?

We can use the second callback argument setState has:

```js
this.write = data => localStorage.setItem(key, JSON.stringify(data));
this.persist = throttle(() => this.getState(), 1000):
this.upstreamSetState(patch, persist);
```

How often to we need to persist? Maybe max once a second?

I like how this makes persistence just an automatic thing.

Would be cool to also persist stuff to the FS someday.

ndelangen: Do we WANT to store/persist the entire global state?

It would mean we'd show the stories in the explorer faster, but they may not be accurate at all anymore.

ndelangen: Your setState call on store means users can no longer get a callback of when the setState has been fully applied by react.

Since you're taking the second argument of setState, and using it for the persistence enum.

lib/ui/src/core/store.js
```diff
@@ -1,33 +1,50 @@
-// simple store implementation
-const createStore = initialState => {
-  let state = initialState;
-  const listeners = [];
+// TODO -- make this TS?
 
-  function notify() {
-    listeners.forEach(l => l());
+import { localStorage, sessionStorage } from 'global';
```
ndelangen: We might be better off using indexedDB, since we'd not have to stringify and parse.

ndelangen:
Might we want to use https://github.com/localForage/localForage

lib/ui/src/core/store.js
tmeasday: OK, a few things here:

> 1. I think some throttling needs to happen.

You are probably right.

> 2. Do we WANT to store/persist the entire global state?

I don't think so. I was just imagining we would persist things that users would expect (settings permanently, things like knobs or UI choices on the session). Other state like stories don't need to be persisted, that's just making trouble for ourselves.

> 3. Would be cool to also persist stuff to the FS someday.
 
Yeah, also to figure out a way to do it per-storybook (right now if you run two storybooks on the same port they will clobber each other no doubt) and across all storybooks (as the persisted state will be attached to a given port). But we can add that.

> 4. Your setState call on store means users can no longer get a callback of when the setState has been fully applied by react.

Yes, I neglected this. What do you think the best approach is? We could just allow either as second argument (enum or callback) or three arguments if desired. 

Or we could always just have three arguments `setState(patch, persistence, callback)`. 

What do you think?

lib/ui/src/core/store.js
tmeasday: I don't really know if it's a big deal. Easy to change after the fact as this is hidden behind an API.

lib/ui/src/core/store.js
tmeasday: Oh and:

> 5. Is our state a serialisable object btw?

I think it's the responsibility of the caller to ensure that what they are persisting is serializable. If they aren't persisting, they can do what they want of course.

tmeasday:
> Might we want to use https://github.com/localForage/localForage

Definitely something we might want to consider in the future. Again we can change pretty easily later.

What do you think about this API:

```
async setState(patch, persistence)
```

Rather than a callback, the caller can just await the result?

tmeasday:
@ndelangen one more thing -- we may need a way to control the way state is merged. 

For example, shortcuts are stored in the form `{ shortcuts: { 'fullScreen': ['X'] } }`

If you merge the above with the existing value (`{ shortcuts: { 'fullScreen': ['F'] } }` say) our merge algorithm results in `{ shortcuts: { 'fullScreen': ['F', 'X'] } }` .

Do you think perhaps we should add an option to `setState` to influence how it behaves?

```js
store.setState({ shortcuts: { 'fullScreen': ['X'] } }, { persistence: 'permament', mergeArrays: false })

// or

store.setState({ shortcuts: { 'fullScreen': ['X'] } }, { persistence: 'permament', maxMergeDepth: 2 })
```

What do you think?

Hypnosphi:
@tmeasday Can `patch` be a function from previous state like in React?

tmeasday:
@Hypnosphi I think it can, yeah. Here is an API we are settling on, based on this and another conversation with @ndelangen 

```typescript
// Excuse my made up typescript
class Store {
  async setState(patch: object | function, [callback: function], options: { persistence?: null | 'session' | 'permanent' }): object
}

// So you can call
const newState = await store.setState({ a: 'b' })
// or the traditional
store.setState({ a: 'b' }, newState => {});

// You can also pass an function to mutate state
store.setState(oldState => ({ a: 'b' }))

// You can pass a set of options (right now only `persistence` as a last parameter
store.setState({a: 'b'}, { persistence: 'session }

// Within state, you can control merging with `Store.mergeState: boolean | 'noChildren'`
store.setState({ a: {..., [Store.mergeState]: false } }); // <- this means overwrite the old value of `a`
store.setState({ a: { b: [...], [Store.mergeState]: 'noChildren' } }) // <- this means merge `a` but overwrite `a.b`

// You can remove keys using `Store.removeState`
store.setState({ a: { b: Store.removeState } }) // <- This means unset `a.b` in the state
store.setState({ a: { b: Store.removeState } }, { persistence: 'session') // <- This means unset `a.b` in the state and ensure it is deleted from the session
```

I'm going to play around with it, let me know if you have any objections or good ideas

tmeasday:
OK, discussed this with @ndelangen again and I think there is some issues with the merging part of the above. Will come back to this tomorrow, I'll update my code snippet above.

tmeasday:
Ok, here is what I am thinking now.

1. We retain the "augmented" version of the react `setState` API (ability use async, persistence option)

2. As `setState` doesn't do any merging (below the first level), we will not do so either.

3. If a module/addon wants to merge persisted state with default state at startup, we will make the persisted state available to the init function and it can do so there. 

I will try and finish the above and demonstrate with the shortcuts module.

tmeasday:
This branch is ready to review and merge now

ndelangen (approved):

PR #5295, opened 2019-01-18
FIX issue where storybook wouldn't be able to navigate when hosted as a file://

ndelangen:
Issue: https://github.com/storybooks/storybook/issues/5274

## What I did
I added the current pathname as a base to anything we route/navigate to.
This seems to work 🤷‍♂️

### DIFF
--- a/lib/ui/src/index.js
+++ b/lib/ui/src/index.js
@@ -1,3 +1,4 @@
+import { document } from 'global';
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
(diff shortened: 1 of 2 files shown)


lib/ui/src/index.js
```diff
@@ -16,6 +17,13 @@ HelmetProvider.displayName = 'HelmetProvider';
 
 const Container = process.env.XSTORYBOOK_EXAMPLE_APP ? React.StrictMode : React.Fragment;
 
+const base =
+  document.location.protocol === 'file:'
+    ? `${document.location.origin}${document.location.pathname}`
+    : '/';
+
+console.log({ base });
```
tmeasday: This looks like it's not needed (all the changes in this file?)

tmeasday (approved):
Seems reasonable apart from the extra bit

lib/ui/src/index.js
ndelangen: will remove ✅

PR #5302, opened 2019-01-19
RE-ADD a 'configureA11Y' which is deprecated for backwards compatibility

ndelangen:
Issue: https://github.com/storybooks/storybook/issues/5235#issuecomment-455731433


tmeasday (approved):
🙌

PR #5305, opened 2019-01-19
Storyshots: restore emotion snapshots

ndelangen:
Issue: -I had broken the emotion snapshot serializer with my @storybook/theming PR-

## What I did
I've added the serializer to the storyshots and main jest config.


tmeasday (approved):
LGTM

PR #5320, opened 2019-01-21
ADD explicit stories for validating the decorator arguments feature

ndelangen:
We had a story for this, but it was "lost" when migrating to the experimental new story format.

I want to make sure this renders properly in the current story format too, so I added a story.

In order to snapshot the full parameters, I had to mkae them consistent across platforms. So I remove the cwd from the filepath in storyshots.


examples/official-storybook/stories/core/decorator-agument.stories.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+storiesOf('Core|Parameters/Decorator aguments', module)
+  .addDecorator(fn => fn({ decoratorArgument: true }))
+  .add('default', data => <pre>Parameters are {JSON.stringify(data, null, 2)}</pre>);
```
tmeasday: Boo! Why not the new format?

tmeasday:
@ndelangen sorry this one got buried. What are your thoughts on it re:5.0 -- shall we just merge after?

ndelangen:
I ended up removing/filtering the fileName parameter in next already, I'll close this.

PR #5321, opened 2019-01-21
REFACTOR treeview utils so they are memoizable && memoize treeview utils && ADD fuzzy search

ndelangen:
Issue: - fuzzy search was a feature in V4, we broke it in V5, this brings it back

I REFACTORED the utils a bit, to make it memoizable.

### DIFF
--- a/lib/components/src/treeview/treeview.js
+++ b/lib/components/src/treeview/treeview.js
@@ -125,9 +125,7 @@ const calculateTreeProps = memoize(50)((input, filter, selectedId, extraExpanded
     {}
   );
 
-  const selectedAncestorIds = selectedId
-    ? getParents({ id: selectedId, dataset }).map(i => i.id)
-    : [];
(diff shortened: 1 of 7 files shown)


lib/components/src/treeview/treeview.js
```diff
@@ -125,9 +125,7 @@ const calculateTreeProps = memoize(50)((input, filter, selectedId, extraExpanded
     {}
   );
 
-  const selectedAncestorIds = selectedId
-    ? getParents({ id: selectedId, dataset }).map(i => i.id)
-    : [];
+  const selectedAncestorIds = selectedId ? getParents(selectedId, dataset).map(i => i.id) : [];
```
tmeasday: What's going on on this line? Is there a prettier issue somewhere between you and me? ;)

tmeasday (approved):
LGTM.

Do we need to update the treeview's behaviour so it expands the matched entries? I've noticed that it's annoying that often the results of the search are buried deep inside nested groups.

PR #5340, opened 2019-01-23
ADD a STORY_UNCHANGED event

tmeasday:
Emit when the STORY_RENDER event is received but nothing has changed, so we aren't re-rendering.

This helps tools like Chromatic detect with certainty when a story has finished rendering


shilman (approved):
YAY 🎉

PR #5359, opened 2019-01-24
Shortcuts refactor

tmeasday:
Issue: Shortcuts code was spread over the place a bit and didn't properly reflect changes via the new persistence layer.

## What I did

 1. Created a single file `ui/src/libs/shortcut.js` that is responsible for turning events into shortcuts and printing them to the screen.

2. Refactored all the event binding and interpreting code to be handled by the `ui/src/core/shortcuts.js` module (mediated by the above).

3. Reworked the `settings/shortcuts.js` page to only look for a single keypress shortcut (no combinations allowed)

4. Change the way the preview sends shortcuts -- instead of interpreting them in the preview, instead it sends a stripped down `KeyboardEvent` over the channel on a new `PREVIEW_KEYDOWN` event (after filtering for events inside text fields).

## Notes / Qns

A. Is there a better way to access the apis of other modules from a given module's api than this junk: 

https://github.com/storybooks/storybook/blob/7c545cacb47b05a95653c341d64eb805e5adb628/lib/ui/src/core/shortcuts.js#L66

(passing `fullApi` in) -- @ndelangen?

B. Does it matter that I bind the keydown event handler outside of `componentDidMount` now?

C. I added typescript support to the `official-storybook`. Would have been nice to do with a preset :wink (@shilman  / @igor-dv)

## How to test

Please try it out in official storybook. It should all work well now.

There are still some weirdnesses with rendering shortcuts that @Jessica-Koch is working on (try option-something on a mac). Also I am not sure if it works properly in older browsers -- this is something else she was looking at (falling back to `keyCode` when `event.key` is not defined).

The above things are not new to this branch however.


ndelangen (approved):
Aside from the ts-loader, which I'm going ask @igor-dv about, this looks great!

ndelangen:
I'm also unsure how to better deal with the core 'modules' needing the full api..

tmeasday:
> I'm also unsure how to better deal with the core 'modules' needing the full api..

Let's punt that for the immediate term. But I think we need something.

PR #5370, opened 2019-01-25
Versions use persistence

tmeasday:
## What I did

Versions API now uses persistence API.
Also persists dismissed state of notification


shilman:
I'm doing some tweaking on the version logic and testing it end-to-end and will merge once that's ready. @domyen any low-hanging fruit here on the UI (looking pretty sharp IMHO)?

storybook

shilman:
@tmeasday I think there's a bug in `'getLatestVersion works'`:

**What I did:**

1. Create a `versions.json` file with `5.0.0-alpha.9`, served on my localhost
2. Update the code to query that file, for testing purposes
3. Update `versions.json` with `5.0.0-alpha.10`
4. Refresh the browser

**What I expected:**

1. Initially, `5.0.0-alpha.9` will show in about page
2. After updating `5.0.0-alpha.10` will be the latest version

**What actually happened:**

`5.0.0-alpha.10` gets **appended** to the versions state:

storybook

Then this code returns `5.0.0-alpha.9` as the latest version:

```js
      Object.values(versions).find(v => v.tag === 'latest');
```

shilman (changes requested):
see comment

Hypnosphi (approved):

tmeasday:
@shilman -- fixed the issue.

I'm wondering if I should perhaps normalize the version info a little so I don't need to do all this work, something like:

```
{
  versions: {
    current: '3.0.0',
    tags: {
      latest: '4.0.0',
    },
    info: {
       '3.0.0': { ... }
    },
}
```

shilman:
@tmeasday that representation makes a lot of sense to me 👍

tmeasday:
Yeah, you are probably right. I'll change it.

shilman (approved):
Great! I'll add the semver logic in a separate PR

PR #5373, opened 2019-01-26
FIX blurring an empty field restores default

Jessica-Koch:
Issue:
Focusing an element nullifies input but it is never restored when tabbing out of element
## What I did
- restore teh default shortcut if shortcut field is empty on blur
## How to test

### DIFF
--- a/lib/ui/src/settings/shortcuts.js
+++ b/lib/ui/src/settings/shortcuts.js
@@ -101,16 +101,15 @@ class ShortcutsScreen extends Component {
 
     this.setState({
       activeFeature: focusedInput,
-      shortcutKeys: { ...shortcutKeys, [focusedInput]: { shortcut: null, error: false } },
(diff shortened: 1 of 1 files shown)


lib/ui/src/settings/shortcuts.js
```diff
@@ -101,16 +101,15 @@ class ShortcutsScreen extends Component {
 
     this.setState({
       activeFeature: focusedInput,
-      shortcutKeys: { ...shortcutKeys, [focusedInput]: { shortcut: null, error: false } },
+      shortcutKeys: { ...shortcutKeys, [focusedInput]: { shortcut: [], error: false } },
```
tmeasday: Hmm this null was intentional.

tmeasday (changes requested):
Hmm, I'm not quite sure this is the fix :/

tmeasday:
Definitely see the bug though

PR #5375, opened 2019-01-26
FIX keybindings were pointing to null events

Jessica-Koch:
Issue:

## What I did
Some of the storybook keybindings were mapped to mistyped shortcuts, I fixed the incorrect mapping and also passed additonal keyboard event properties in the event object that I will need.

## How to test


tmeasday (approved):
LGTM

PR #5403, opened 2019-01-29
Update index.md

canercandan:
setting `module` to `commonjs` ends up with a undefined `createElement` error, though setting it to `esnext` did make the loading of the react components work

Issue:

## How to test

### DIFF
--- a/docs/src/pages/configurations/typescript-config/index.md
+++ b/docs/src/pages/configurations/typescript-config/index.md
@@ -47,7 +47,7 @@ The above example shows a working Webpack config with the TSDocgen plugin also i
 {
   "compilerOptions": {
     "outDir": "build/lib",
-    "module": "commonjs",
+    "module": "esnext",
     "target": "es5",
     "lib": ["es5", "es6", "es7", "es2017", "dom"],
     "sourceMap": true,


docs/src/pages/configurations/typescript-config/index.md
```diff
@@ -47,7 +47,7 @@ The above example shows a working Webpack config with the TSDocgen plugin also i
 {
   "compilerOptions": {
     "outDir": "build/lib",
-    "module": "commonjs",
+    "module": "esnext",
```
wuweiweiwu: i believe if we're targeting `es5` then module generation is using `commonjs` according to typescript docs

https://www.typescriptlang.org/docs/handbook/compiler-options.html

Can you post a reproduction?

canercandan: @wuweiweiwu just tried those values:

```
    "module": "commonjs",
    "target": "es5",
```

and getting an error tough it works with the previous values (esnext and esnext for both module and target)

ndelangen: I'm not sure what's right, we're using all of the above in the codebase...

screenshot 2019-02-14 at 17 07 47

Lets get some more typescript folk in here..

@kroeder @gaetanmaisse @apastuhov 

What do you guys thinks is the correct value for our docs?

gaetanmaisse: @canercandan Do you use dynamic module imports in your project? 

I did not succeed in reproducing your problem neither using React sample project nor a real world angular app. Can you create a minimal example reproducing this issue?

---

@ndelangen I will investigate in the next few days about the different values we are using in SB codebase.

ndelangen: Thanks, I'll close this PR in the meantime then. Please re-open if this is something we want to continue

PR #5413, opened 2019-01-30
FIX the layout issue where panels still sort-of appear on non story viewmodes

ndelangen:
Desktop layout had panels still semi-showing when in info viewmode


tmeasday (approved):
LGTM

PR #5437, opened 2019-01-31
Simplified examples for decorators in docs

martnst:
Since this page is explaining Addons,  I was distracted / confused by the `@storybook/addon…` imports in the examples on decorators. Eventually I realized that those are actually not what the examples are about.

### DIFF
--- a/docs/src/pages/addons/introduction/index.md
+++ b/docs/src/pages/addons/introduction/index.md
@@ -30,7 +30,6 @@ Then we can use it when writing stories.
 
 ```js
 import { storiesOf } from '@storybook/react';
-import { action } from '@storybook/addon-actions';
 
 import Center from './center';
 import Button from './button';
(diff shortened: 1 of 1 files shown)


wuweiweiwu (commented):
I don't this these changes will make the examples more understandable. 

These docs illustrate how addons / decorators work. And your code changes will remove the usage of addons

martnst:
I am new to storybook and was reading _Intro to Addons_ the very first time. The first sections talks about Decorators and how to use `addDecorator` vs. using Wrapper Components. It does not explain anything about `import { action } from '@storybook/addon-actions';` or `import { linkTo } from '@storybook/addon-links';` yet. However the examples include those at this point already, which I personally did find rather **confusing** than helping. I assume others people will experience the same. Hence my PR to simplify those examples to focus on Decorators. 

You are right it is removing the Addon code examples from the page, but actually they should go into the seconds section of that page, which is about the actual [Native] Addons, but has no examples yet.

ndelangen:
I'll improve this page after the V5 release

PR #5452, opened 2019-02-01
Fix/tool addons

### LINKED ISSUE #5416: @storybook/addon-backgrounds@next does not work
**Describe the bug**
I'm trying to use storybook@next. After I installed & set `@storybook/addon-background`, I faced two problems.

1. Background tool is not displayed occasionally.
2. When I tried to set a background (by clicking a color button), it emits the following error.
    ```
    TypeError: _this.tooltipRef.current.scheduleHide is not a function
    ```

**To Reproduce**
1. `git clone https://github.com/Ailrun/storybook-issue5416`
1. `npm i`
1. `npm start`
1. Check that there is no background tool just like the screenshot.
  image
1. (While running `npm start`) uncomment line 5-21 of `index.js`.
   (FYI, I already set the global parameters in `.storybook/config.js`)
1. Check that there is a background tool.
   (Weirdly...)
1. Open the web dev tool.
1. Click the background tool button.
1. Select `grey` (or any other colors).
1. See the error.

**Expected behavior**
Always display the tool, and do not emit any errors.

**Screenshots**
image

**System:**
 - OS: Manjaro (Kernel 4.18)
 - Device: Desktop
 - Browser: Firefox 64
 - Framework: React
 - Addons: Backgrounds
 - Version: 5.0.0-alpha.10

ndelangen:
fixes: https://github.com/storybooks/storybook/issues/5416

I removed the need for tool addons to find the iframe, instead they can inject global styles, hwihc is marginally better then setting inline styles.

Long term i want to add a addon-type to wrap the preview component, so this can be done very cleanly.

@Ailrun care to review?

### DIFF
--- a/addons/backgrounds/src/Tool.js
+++ b/addons/backgrounds/src/Tool.js
@@ -14,29 +14,21 @@ const iframeId = 'storybook-preview-background';
 
 const getState = memoize(10)((props, state) => {
   const data = props.api.getCurrentStoryData();
-  const list = data && data.parameters && data.parameters[PARAM_KEY];
+  const list = (data && data.parameters && data.parameters[PARAM_KEY]) || [];
 
-  return list && list.length
-    ? list.reduce(
-        (acc, { name, value, default: isSelected }) => {
-          acc.items.push({ name, value });
+  const items = list.length
+    ? list.map(({ name, styles: value, id }) => ({ name, value, id }))
+    : list;
 
-          if (isSelected && state.selected !== 'transparent') {
-            if (!list.find(i => i.value === state.selected)) {
-              acc.selected = value;
-            }
-          }
-          return acc;
-        },
-        {
-          items: [],
-          selected: state.selected,
-        }
-      )
-    : {
-        items: [],
-        selected: 'transparent',
-      };
+  const selected =
+    state.selected === 'responsive' || list.find(i => i.id === state.selected)
+      ? state.selected
+      : list.find(i => i.default) || 'responsive';
+
+  return {
+    items,
+    selected,
+  };
 });
 
 export default class BackgroundTool extends Component {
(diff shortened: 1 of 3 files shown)


ndelangen:
Weird the image tests work on my machine..

ndelangen:
darnit

addons/viewport/src/Tool.js
```diff
@@ -1,121 +1,120 @@
...
+  const list = toList(data && data.parameters && data.parameters[PARAM_KEY]);
+
+  return list && list.length
+    ? list.reduce(
+        (acc, { name, styles: value, id }) => {
+          acc.items.push({ name, value, id });
+
+          if (state.selected !== 'responsive') {
+            if (!list.find(i => i.id === state.selected)) {
+              acc.selected = id;
```
tmeasday: I don't understand this code. Does it keep setting `acc.selected` if `state.selected` isn't in the list? It seems like it'll do this for every item in the list.

I think this code is overly complex for what it is trying to do -- or maybe I don't understand it? In either case I think it needs to be clearer. I *think* this is what it shoudl probably look like:

```js
const items = list && list.length && list.map(({ name, styles: value, id }) => ({name, value, id});

// It actually will be the *last* element in the list in the code above I think, but I doubt this is what you meant?
const selected = (state.selected === 'responsive' || list.find(i => i.id === state.selected))  ? state.selected : list[0].id;

return {
  list,
  selected,
  isRotated: state.isRotated,
};
```

addons/backgrounds/src/Tool.js
```diff
@@ -1,26 +1,25 @@
...
-const getState = (props, state) => {
+const getState = memoize(10)((props, state) => {
   const data = props.api.getCurrentStoryData();
   const list = data && data.parameters && data.parameters[PARAM_KEY];
 
   return list && list.length
     ? list.reduce(
         (acc, { name, value, default: isSelected }) => {
-          acc.backgrounds.push({ name, value });
+          acc.items.push({ name, value });
```
tmeasday: Same comment here as for the other addon

tmeasday (changes requested):
I'm the first to do a complicated list comphrension but I think this one might be a step too far ;)

ndelangen:
@tmeasday I improved to code, thanks!

PR #5457, opened 2019-02-01
FIX for image-snapshot being broken && FIX snapshots

ndelangen:
by actually serving content (no file://)


tmeasday:
@ndelangen could be a memory issue but here is what chrome logs:

https://circleci.com/gh/storybooks/storybook/95407

Googling implied maybe it was `puppeteer@1.12` but downgrading to `1.11` doesn't seem to have helped.

PR #5534, opened 2019-02-10
Fix type of function `makeDecorator` from `lib/addons`

gaetanmaisse:
## What I did

Fix type of function `makeDecorator` from `lib/addons`.

`MakeDecoratorResult` is type of object returned by `makeDecorator` and not `makeDecorator` itself.
This fix also helps a lot in addons TS migration because correct types are now inferred 😃.

Also remove some `any` typings linked to `makeDecorator` in `addon/cssresources`

### DIFF
--- a/addons/cssresources/src/index.ts
+++ b/addons/cssresources/src/index.ts
@@ -67,7 +67,7 @@ export const withCssResources = makeDecorator({
   skipIfNoParametersOrOptions: true,
   allowDeprecatedUsage: false,
 
-  wrapper: (getStory: any, context: any, { options, parameters }: any) => {
+  wrapper: (getStory, context, { options, parameters }) => {
     const storyOptions = parameters || options;
     addons.getChannel().on(EVENTS.SET, setResources);
 
--- a/lib/addons/src/make-decorator.ts
+++ b/lib/addons/src/make-decorator.ts
@@ -26,18 +26,18 @@ type MakeDecoratorResult = (...args: any) => any;
 interface MakeDecoratorOptions {
   name: string;
   parameterName: string;
-  allowDeprecatedUsage: boolean;
-  skipIfNoParametersOrOptions: boolean;
+  allowDeprecatedUsage?: boolean;
+  skipIfNoParametersOrOptions?: boolean;
   wrapper: StoryWrapper;
 }
 
-export const makeDecorator: MakeDecoratorResult = ({
+export const makeDecorator = ({
   name,
   parameterName,
   wrapper,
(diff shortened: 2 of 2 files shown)


addons/cssresources/src/index.ts
```diff
@@ -67,7 +67,7 @@ export const withCssResources = makeDecorator({
   skipIfNoParametersOrOptions: true,
   allowDeprecatedUsage: false,
 
-  wrapper: (getStory: any, context: any, { options, parameters }: any) => {
+  wrapper: (getStory, context, { options, parameters }) => {
```
Keraito: This seems unrelated to your PR, but good catch anyways 👍

lib/addons/src/make-decorator.ts
```diff
@@ -31,13 +31,13 @@ interface MakeDecoratorOptions {
...
 
-export const makeDecorator: MakeDecoratorResult = ({
+export const makeDecorator = ({
   name,
   parameterName,
   wrapper,
   skipIfNoParametersOrOptions = false,
   allowDeprecatedUsage = false,
-}: MakeDecoratorOptions) => {
+}: MakeDecoratorOptions): MakeDecoratorResult => {
```
Keraito: Good catch 👍

lib/addons/src/make-decorator.ts
```diff
@@ -31,13 +31,13 @@ interface MakeDecoratorOptions {
   wrapper: StoryWrapper;
 }
 
-export const makeDecorator: MakeDecoratorResult = ({
+export const makeDecorator = ({
   name,
   parameterName,
   wrapper,
   skipIfNoParametersOrOptions = false,
   allowDeprecatedUsage = false,
```
Keraito: Both `skipIfNoParametersOrOptions` and `allowDeprecatedUsage` have default values set here, so you shouldn't have to set those in the tests explicitly. But I leave it up to you.

Keraito (approved):

lib/addons/src/make-decorator.ts
gaetanmaisse: Ok, I understand why I had to explicitly set them in test, just because they are not declared as optional in `MakeDecoratorOptions`. I will make them optional and revert my changes on tests.

addons/cssresources/src/index.ts
gaetanmaisse: Just found it when I was looking for `makeDecorator` usage, so I cleaned it ;)

ndelangen (approved):

PR #5549, opened 2019-02-11
Update getStorybook to return story parameters

timhaines:
Issue:
v4 of Storybook [introduced story parameters](https://github.com/storybooks/storybook/pull/2679), but getStorybook doesn't return them.  This PR updates getStorybook so it does return parameters.

This allows in-browser consumers to access story parameters with a more supportable `window['__STORYBOOK_CLIENT_API__'].getStorybook()` instead of `window['__STORYBOOK_CLIENT_API__']['_storyStore']['_data']`

## What I did
Updated the getStorybook function to return parameters

## How to test
Unit tests

### DIFF
--- a/lib/client-api/src/client_api.js
+++ b/lib/client-api/src/client_api.js
@@ -226,9 +226,9 @@ export default class ClientApi {
     this._storyStore.getStoryKinds().map(kind => {
       const fileName = this._storyStore.getStoryFileName(kind);
 
-      const stories = this._storyStore.getStories(kind).map(name => {
-        const render = this._storyStore.getStoryWithContext(kind, name);
-        return { name, render };
+      const stories = this._storyStore.getStories(kind).map(story => {
+        const render = this._storyStore.getStoryWithContext(kind, story.name);
+        return { name: story.name, render, parameters: story.parameters };
       });
 
       return { kind, fileName, stories };
--- a/lib/client-api/src/client_api.js
+++ b/lib/client-api/src/client_api.js
@@ -226,9 +226,9 @@ export default class ClientApi {
     this._storyStore.getStoryKinds().map(kind => {
       const fileName = this._storyStore.getStoryFileName(kind);
 
-      const stories = this._storyStore.getStories(kind).map(story => {
-        const render = this._storyStore.getStoryWithContext(kind, story.name);
-        return { name: story.name, render, parameters: story.parameters };
+      const stories = this._storyStore.getStories(kind).map(({ name, parameters }) => {
+        const render = this._storyStore.getStoryWithContext(kind, name);
+        return { name, render, parameters };
       });
 
       return { kind, fileName, stories };
--- a/lib/client-api/src/client_api.js
+++ b/lib/client-api/src/client_api.js
@@ -226,10 +226,12 @@ export default class ClientApi {
     this._storyStore.getStoryKinds().map(kind => {
       const fileName = this._storyStore.getStoryFileName(kind);
 
-      const stories = this._storyStore.getStories(kind).map(({ name, parameters }) => {
-        const render = this._storyStore.getStoryWithContext(kind, name);
-        return { name, render, parameters };
-      });
+      const stories = this._storyStore
+        .getStoriesWithParameters(kind)
+        .map(({ name, parameters }) => {
+          const render = this._storyStore.getStoryWithContext(kind, name);
+          return { name, render, parameters };
+        });
 
       return { kind, fileName, stories };
     });
--- a/lib/client-api/src/story_store.js
+++ b/lib/client-api/src/story_store.js
@@ -214,8 +214,7 @@ export default class StoryStore extends EventEmitter {
 
     return Object.keys(this._legacydata[key].stories)
       .map(name => this._legacydata[key].stories[name])
-      .sort((info1, info2) => info1.index - info2.index)
-      .map(info => ({ name: info.name, parameters: info.parameters }));
+      .sort((info1, info2) => info1.index - info2.index);
   }
 
   getStoryFileName(kind) {
--- a/lib/client-api/src/story_store.js
+++ b/lib/client-api/src/story_store.js
@@ -206,7 +206,7 @@ export default class StoryStore extends EventEmitter {
       .map(info => info.kind);
   }
 
-  getStories(kind) {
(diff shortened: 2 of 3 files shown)


lib/client-api/src/story_store.js
```diff
@@ -215,7 +215,7 @@ export default class StoryStore extends EventEmitter {
     return Object.keys(this._legacydata[key].stories)
       .map(name => this._legacydata[key].stories[name])
       .sort((info1, info2) => info1.index - info2.index)
-      .map(info => info.name);
+      .map(info => ({ name: info.name, parameters: info.parameters }));
```
timhaines: I have reviewed the codebase to ensure this getStories function isn't used outside of the callers updated in this PR, so don't consider this a breaking change.

Keraito: Wouldn't it be more beneficial at this point to just remove the map? You're basically just getting two properties and destructuring them into a new object. What more is there in these info objects?

lib/client-api/src/client_api.js
```diff
@@ -226,9 +226,9 @@ export default class ClientApi {
     this._storyStore.getStoryKinds().map(kind => {
       const fileName = this._storyStore.getStoryFileName(kind);
 
-      const stories = this._storyStore.getStories(kind).map(name => {
-        const render = this._storyStore.getStoryWithContext(kind, name);
-        return { name, render };
+      const stories = this._storyStore.getStories(kind).map(story => {
+        const render = this._storyStore.getStoryWithContext(kind, story.name);
+        return { name: story.name, render, parameters: story.parameters };
```
Keraito: You can destructure `name` and `parameters` out of `story` or spread the `story` object in the return value right?

Keraito (commented):
From a code perspective there's not much wrong with this, tagging @tmeasday to judge the behaviour of this as he was responsible for the tagged PR.

lib/client-api/src/story_store.js
timhaines: @Keraito thanks for the review.  The info objects also contain index and fn.  I chose to make the minimal possible change to what `getStories` returns.  You're right, index and fn could optionally be returned too, and callers could do with them what they please.  Let me know if that's the preference, and I'll update this PR.

lib/client-api/src/client_api.js
timhaines: @Keraito I'm not sure how that's done.  Could you provide an example?

Keraito: Either 
1. `const stories = this._storyStore.getStories(kind).map(({name, parameters}) => {` and then you don't have to do `story.name`/`story.parameters` and just use `name` and `parameters` instead, which means your return statement becomes `return { name, render, parameters }`.
2. or your return statement becomes `return { render, ...story }` and it will insert the `name` and `parameters` properties automatically in the return object. 

If the `story` contained more properties, those would be added as well to the return object (which is not too great), but in this case it only contains name and parameters anyways. Either of the above options is fine tho, pick whichever you like. Docs on `...`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

lib/client-api/src/story_store.js
Keraito: Personally, I don't mind too much, as either is fine. This also kinda influences my other comment, so maybe it's better to leave it at the changes you have made. Don't think the additional map has severe performance implications, but don't quote me on that 😆

timhaines: Thanks @Keraito - have pushed a commit that implements both of your suggestions.

lib/client-api/src/client_api.js
timhaines: Have gone with option 1, thanks.

lib/client-api/src/story_store.js
```diff
@@ -214,8 +214,7 @@ export default class StoryStore extends EventEmitter {
 
     return Object.keys(this._legacydata[key].stories)
       .map(name => this._legacydata[key].stories[name])
-      .sort((info1, info2) => info1.index - info2.index)
-      .map(info => info.name);
+      .sort((info1, info2) => info1.index - info2.index);
```
timhaines: Both callers to this modified `getStory` function have been updated in this PR.

ndelangen:
Hey @timhaines In v5 there's a new api for getting the stories data complete with parameters:

https://github.com/storybooks/storybook/blob/next/lib/client-api/src/story_store.js#L111-L123

Long term I'd like to remove the getStorybook api.

Would be happy to merge this PR though in the meantime.

lib/client-api/src/story_store.js
tmeasday: Is this a breaking change though? There could be other external consumers of this API I guess.

tmeasday (changes requested):
I think returning extra data in `getStorybook()` is good, but do we need to change the return value of `getStories()`? That part doesn't seem as necessary?

PS We use code something like 

```js
.getStorybook()
  .map(({ kind, stories }) =>
    stories.map(({ name }) => {
      const { parameters } = storyStore.getStoryAndParameters(kind, name);
```

Although I'm looking at our code and it says "Annoying to have to do this, we should support this in .getStorybook" so it looks like this change is long overdue ;)

timhaines:
@tmeasday to address your valid concern, I've left `getStories` behavior intact, and have introduced `getStoriesWithParameters`.  

Note that `getStories` is only tested indirectly, as `getStoriesWithParameters` is now.

timhaines:
@ndelangen thanks for the heads up on those two functions.  I don't think they can be used as is in place of getStories, as they don't preserve the ordering in the same way getStories returns stories.

It's interesting to hear you're planning on deprecating and removing the getStorybook API long-term. Is there anything documented about how you'd like to replace the functionality?

lib/client-api/src/story_store.js
timhaines: Resolve so getStories behavior is unchanged.

timhaines:
Closing this in favor of adopting raw() over improving getStorybook().

ndelangen:
Thanks tim, if were missing any functionality in either `.raw` or `.extract` please let me know.

The `.extract` is for storybook itself, to transfer the stories to the manager.
`.raw` is for integration purposes such as storyshots.

PR #5564, opened 2019-02-13
Fix: Prioritise programatic configuration

loklaan:
Issue:

All config options, except for `configDir` and `outputDir`, prioritise the programatic "`loadOptions`".

This causes unexpected results such as defaulting to `"./.storybook"` even though the programatic config has them declared explicitly/differently.

## What I did

I made sure that these two options also prioritise correctly.

## How to test

  **Not sure, I haven't look at tests in the repo before!**
  **No, it probably resembles expected behaviour.**
  **No, it probably resembles expected behaviour.**


ndelangen (approved):
Wow, that's a nice bug you found there!

loklaan:
Thanks @ndelangen!

PR #5565, opened 2019-02-13
Fix: Avoid false-negative checks for port

loklaan:
Issue:

When a port is provided via env vars, we're usually told by the CLI that it's unavailable, only to be suggested the same port next.

The usage of these `SBCONFIG_CONFIG_<NAME>` env vars is only intended for internals, as far as I'm aware. Nevertheless, when using the `SBCONFIG_CONFIG_PORT` we're left with an unexpected behaviour.

## What I did

Make sure the `SBCONFIG_CONFIG_PORT` is cast to a number, so as not to trip on any explicit (`===`) equality tests further down the code.

## How to test

  **Not sure, I haven't looked at the test suit in the repo sorry :(**
  **No, as it resembles expected behaviour and was rarely encountered.**
  **No, as it resembles expected behaviour and was rarely encountered.**

### DIFF
--- a/lib/core/src/server/cli/dev.js
+++ b/lib/core/src/server/cli/dev.js
@@ -57,6 +57,10 @@ async function getCLI(packageJson) {
     configDir: 'SBCONFIG_CONFIG_DIR',
   });
 
+  if (typeof program.port === 'string' && program.port.length > 0) {
+    program.port = parseInt(program.port, 10);
+  }
+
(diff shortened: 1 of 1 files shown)


ndelangen (approved):
You find the weirdest bugs 🎖

loklaan:
Aw thanks @ndelangen! That's legitimately such a nice thing to say. 🎉 Adding it to my bio everywhere haha

PR #5573, opened 2019-02-13
fix for when outputDir is an absolute path

petersendidit:
Issue:  Absolute path output in build-storybook doesn't work as expected #5334

## What I did
Added a check for `outputDir` to see if its a absolute path already before trying to resolve it into one.
## How to test
Set `--output-path` to an absolute path

- Is this testable with Jest or Chromatic screenshots? not that I know of


ndelangen (approved):

ndelangen:
Nice find, and nice solution, Thank you @petersendidit!

PR #5599, opened 2019-02-15
Fix ie 11

tmeasday:
Issue: Latest version didn't work in IE11 as the `URL` class is not properly supported.

In this case we actually didn't need that prop for these stories. In general we should use something like https://github.com/jsdom/whatwg-url if we want a `URL` in a story.

https://github.com/storybooks/storybook/issues/5590

## What I did

Removed `location` from stories, and fixed the mobile layout to not pass it unnecessarily to the preview.

## How to test

Test this branch on IE11 - https://fix-ie-11--storybooks-official.netlify.com


shilman (approved):
WFM

PR #5610, opened 2019-02-15
Use the correct props to start the tooltip open

tmeasday:
Issue: #5591

The `startOpen` prop is for the stateful `WithTooltipState` (ie. the one that can open and close).

To get an always open one, use `tooltipOpen` on the `WithTooltip`. I assume this is what you wanted @domyen?

PS It seems backwards to export the stateless one as the default @ndelangen. But I guess you are against default exports full stop...


domyen:
LGTM Thanks @tmeasday!

domyen (approved):

tmeasday:
Not sure what's happening with the build but I will merge and fix if it continues

PR #5629, opened 2019-02-17
Fix unkind addon-ondevice-knobs readme.

nitaking:
Issue: #4839

## What I did
Fixed addon-ondevice-knobs's readme.
Instead of importing from `addon-ondevice-knobs`, I explicitly used importing` addon-knobs`.


nitaking:
On the next occasion, we need to unify it so that it can be used with `* -ondevice- *` import.

shilman (approved):
Much better!!! 🎉

PR #5669, opened 2019-02-19
Fix/story fn consistency

ndelangen:
Issue: in a previous PR I changed the meaning of `story` on the context, this reverts that, deals with the consequences of that reversion and cleans up the api a bit.

## What I did

- CLEANUP naming of story, storyFn in client_api & story_store
  - MOVE getDecorated into storyStore
  - ADD getOriginal as the clear method for getting the undecorated storyFn
  - RENAME getStory to storyFn
- CLEANUP naming of story, storyFn in apps
- FIX storyshots after story changed to storyFn
- RENAME `component` to `element` where applicable for consistency.
- RENAME `story` to `fn` inside our decorators

### DIFF
--- a/addons/actions/src/preview/withActions.ts
+++ b/addons/actions/src/preview/withActions.ts
@@ -56,11 +56,11 @@ const actionsSubscription = (...args: any[]) => {
   return lastSubscription;
 };
 
-export const createDecorator = (actionsFn: any) => (...args: any[]) => (fn: () => any) => {
+export const createDecorator = (actionsFn: any) => (...args: any[]) => (storyFn: () => any) => {
   if (root != null) {
     addons.getChannel().emit(Events.REGISTER_SUBSCRIPTION, actionsSubscription(actionsFn, ...args));
   }
-  return fn();
+  return storyFn();
 };
 
 export const withActions = createDecorator(actions);
--- a/addons/actions/src/preview/withActions.ts
+++ b/addons/actions/src/preview/withActions.ts
@@ -56,11 +56,11 @@ const actionsSubscription = (...args: any[]) => {
   return lastSubscription;
 };
 
-export const createDecorator = (actionsFn: any) => (...args: any[]) => (story: () => any) => {
+export const createDecorator = (actionsFn: any) => (...args: any[]) => (fn: () => any) => {
   if (root != null) {
     addons.getChannel().emit(Events.REGISTER_SUBSCRIPTION, actionsSubscription(actionsFn, ...args));
   }
-  return story();
+  return fn();
 };
 
 export const withActions = createDecorator(actions);
--- a/app/vue/src/client/preview/render.js
+++ b/app/vue/src/client/preview/render.js
@@ -18,7 +18,7 @@ const root = new Vue({
 });
 
 export default function render({
-  story,
+  storyFn,
   selectedKind,
   selectedStory,
   showMain,
@@ -28,9 +28,9 @@ export default function render({
 }) {
   Vue.config.errorHandler = showException;
 
-  const component = story();
+  const element = storyFn();
 
-  if (!component) {
+  if (!element) {
     showError({
       title: `Expecting a Vue component from the story: "${selectedStory}" of "${selectedKind}".`,
       description: stripIndents`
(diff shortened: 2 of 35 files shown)


addons/actions/src/preview/withActions.ts
```diff
@@ -56,11 +56,11 @@ const actionsSubscription = (...args: any[]) => {
   return lastSubscription;
 };
 
-export const createDecorator = (actionsFn: any) => (...args: any[]) => (story: () => any) => {
+export const createDecorator = (actionsFn: any) => (...args: any[]) => (fn: () => any) => {
```
tmeasday: Is `fn` a bit unclear - I think `storyFn()` makes more sense, no?

lib/client-api/src/story_store.js
```diff
@@ -134,28 +134,48 @@ export default class StoryStore extends EventEmitter {
...
+      id,
       kind,
-      name: storyName,
-      story,
+      name,
+      story: name, // legacy
+    };
+
+    // immutable original storyFn
+    const getOriginal = memoize(1)(() => original);
```
tmeasday: What is the point of memoizing a function that returns a reference? Doesn't this just waste memory?

lib/client-api/src/story_store.js
```diff
@@ -134,28 +134,48 @@ export default class StoryStore extends EventEmitter {
...
+
+    // lazily decorate the story when it's loaded
+    const getDecorated = memoize(1)(() => applyDecorators(getOriginal(), getDecorators()));
```
tmeasday: `applyDecorators` just wraps the function right? Why not just run this right away and store the result?

tmeasday (changes requested):
I think we should use `storyFn` everywhere -- people will read our decorators and being as clear and consistent as possible is good.

Also I don't understand the use of memoization in the store.

lib/client-api/src/story_store.js
tmeasday: I guess the idea is be "lazy"? I'm not quite sure that composing functions is very expensive

lib/client-api/src/story_store.js
ndelangen: true I guess

lib/client-api/src/story_store.js
ndelangen: I founds that applyDecorators is getting kinda expensive when run on all stories on init.. It delays the rendering of the first meaningful paint on large storybooks.

You don't know how many decorators a user has, nor what they do.

addons/actions/src/preview/withActions.ts
ndelangen: Yea I don't feel strongly about that. I do feel `fn` is better then `story`, but I don't really prefer `fn` over `storyFn` or vice-versa. I picked `fn` since it's a little less verbose.

If you really feel `stroyFn` > `fn`, I'll change it.

tmeasday: I think it is best to be consistent and clear. So I vote for `storyFn` everywhere

lib/client-api/src/story_store.js
tmeasday: Ok, sure, seems reasonable

ndelangen:
Corrected as by your comments

tmeasday (approved):

app/vue/src/client/preview/render.js
```diff
@@ -28,9 +28,9 @@ export default function render({
 }) {
   Vue.config.errorHandler = showException;
 
-  const component = story();
+  const element = storyFn();
```
backbone87: @ndelangen for the sake of clarity: `component` is more correct, as the return value of the story function should be a vue component. the term `element` in vue is used to refer to DOM elements

lib/client-api/src/story_store.js
```diff
@@ -261,13 +281,8 @@ export default class StoryStore extends EventEmitter {
...
 
-    const { story, parameters } = data;
-    return () =>
-      story({
-        kind,
-        story: name,
-        parameters,
-      });
+    const { story } = data;
+    return () => story();
```
Hypnosphi: @ndelangen 

So now `getStory` and `getStoryWithContext` do actually the same thing. Should we deprecate one of those two?

PR #5676, opened 2019-02-20
FIX addon-centered README to refer to /dist files

smiled0g:
Issue:
Currently `import centered from '@storybook/addon-centered/react'` results in file not found.

## What I did
Using files under `/dist` instead.

## How to test

### DIFF
--- a/addons/centered/README.md
+++ b/addons/centered/README.md
@@ -16,7 +16,7 @@ example for React:
 
 ```js
 import { storiesOf } from '@storybook/react';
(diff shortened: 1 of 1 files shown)


shilman (changes requested):
Thanks for catching this! However, I don't think the fix is correct. We shouldn't be importing things from the dist folder.

I think it should be e.g.:

```
import { centered } from '@storybook/addon-centered/react';
```

ndelangen:
I think the confusion is that the documentation on the `next` branch is for the V5 branch/release

PR #5679, opened 2019-02-20
FIX singleton module issue for manager & theme not being set correctly

ndelangen:
Issue: https://github.com/storybooks/storybook/issues/5663

## What I did
- enforce singleton modules of storybook packages for the manager
- add a safety check for theme


lib/theming/src/create.ts
```diff
@@ -177,3 +178,35 @@ export const create = (vars: ThemeVar, rest?: Rest): Theme => ({
 
   ...(rest || {}),
 });
+
+export const ensure = (input: any): Theme => {
```
shilman: I don't like this safety check. I feel like this is going to hide bugs. Can we fail fast here instead?

shilman (changes requested):

lib/theming/src/create.ts
ndelangen: Fail when anything is missing in the theme?

shilman: Yeah, so the stack trace leads you to the error check, rather than in the depths of some component where it's unclear whether it's a bug in the theme or a bug in the component.

ndelangen: I agree on your point that it may hide bugs, but I think throwing on a missing property will cause issues later if we're adding something to the theme.

In that scenario, anything we add to the theme object, will break every theme users have made.

I think some grace is appropriate here, likely the worst that is to happen is, some styling inconsistency in someone's storybook, not broken one.

ndelangen: My suggestion:

screenshot 2019-02-20 at 21 46 41

The storybook continues to function with the light theme merged in.

tmeasday (approved):
I like this, the resolve solution for singletons is about the best we can do.

Also warning on missing theme props is probably what we want.

👍

lib/theming/src/create.ts
shilman: OK, this looks better to me. Thanks!

shilman (approved):

PR #5702, opened 2019-02-21
ADD a timeout around the `/` search hotkey, so it doesn't type into the input

ndelangen:
Issue: https://github.com/storybooks/storybook/issues/5419

> * If you click anywhere in the Sidebar and press "/", you'll insert a '/' in the search box. This should not happen.

## What I did
I added a 16ms timeout before giving the input-element focus.

I know it may not be the cleanest solution, but in order to cancel the event, I'd need access TO the event, and that would require more refactoring then I'm comfortable doing at this stage of the release.

### DIFF
--- a/lib/ui/src/core/shortcuts.js
+++ b/lib/ui/src/core/shortcuts.js
@@ -101,11 +101,14 @@ export default function initShortcuts({ store }) {
           if (!showNav) {
             fullApi.toggleNav();
           }
-          const element = document.getElementById('storybook-explorer-searchfield');
 
-          if (element) {
-            element.focus();
-          }
+          setTimeout(() => {
+            const element = document.getElementById('storybook-explorer-searchfield');
+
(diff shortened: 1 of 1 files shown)


tmeasday:
Why does it need to be 16ms? Is 0 not enough?

ndelangen:
I expected the question, @tmeasday I just pick 16ms, as that over the years seems to be the magic number that seemed to always work in stupid cases like this.

It's roughly 1 animation frame.

I could do requestAnimationFrame instead, but was like 🤷‍♀️

tmeasday:
@ndelangen I guess I ask because I would have guessed 0 would be enough--just need to get the focus behind the event handling, which happens synchronously

ndelangen:
@tmeasday I changed it

shilman (approved):
Great!

PR #5705, opened 2019-02-21
Core: Clean up debug logging

ndelangen:
Issue:
- There was a debugger statement left in the postmessage-channel
- The console is noisy because of logging all messages over the channel

## What I did
REMOVE debugger statement
CHANGE log statement in channel to use logger.debug
ADD info & debug method to logger


shilman (approved):
LGTM!

PR #5710, opened 2019-02-21
Updated documentation

michael-brittlebank:
Updated documentation surrounding the nesting stories feature when also using the Storybook Options Addon

Issue: Nested stories do not work out of the box when also using the Storybook Options Addon. It wasn't until reading [this article](https://medium.com/sears-israel/story-hierarchy-in-storybook-3-2-7506b5a6ceb9) by the feature's developer that I solved why it wasn't happening like the existing documentation stated.

## What I did

Added documentation

## How to test
NA


ndelangen:
The new default in V5 is actually `|` btw, but this documentation is correct for V4

PR #5713, opened 2019-02-21
Addon-a11y: Fix buggy scroll behavior

ndelangen:
Issue: #5598

ADD options to a11y addon so it's doesn't do buggy buggy stuff
MOVE the scrollTop to after the component rendered
ADD stories to test the behavior properly


tmeasday (approved):
👏 well done @ndelangen

ndelangen:
https://github.com/storybooks/storybook/issues/5598

PR #5724, opened 2019-02-22
FIX a11y setup being undefined

ndelangen:
Issue: a11y config when not set can cause an error

## What I did
Fix it by adding a default config of `{}`


ndelangen:
@shilman

PR #5748, opened 2019-02-25
Fuzzy search improvement

ndelangen:
Issue: The fuzzy search finds pretty much everything when typing just 2 chars

## What I did
- ADD a memoized fn to get the values of dataset
- CHANGE toFiltered so a 2 char search will only find exact matches

## How to test

- Run the official example
- in the filter input type 'qs'
  expected: no results
- in the filter input type `qss'
  expected: it should find the graphQl stories

### DIFF
--- a/lib/ui/src/components/sidebar/treeview/utils.js
+++ b/lib/ui/src/components/sidebar/treeview/utils.js
@@ -3,6 +3,8 @@ import FuzzySearch from 'fuzzy-search';
 
 export const prevent = e => e.preventDefault();
 
+const toList = memoize(1)(dataset => Object.values(dataset));
+
 export const keyEventToAction = ({ keyCode, ctrlKey, shiftKey, altKey, metaKey }) => {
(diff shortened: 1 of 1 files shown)


shilman (approved):
Looks good!

@ndelangen I know this is existing behavior, but perhaps we shouldn't search filename and notes?
 - It will make the search faster
 - It is easy to highlight the results (do we even show the filename anywhere in the UI?)

ndelangen:
> @ndelangen I know this is existing behavior, but perhaps we shouldn't search filename and notes?
> 
> * It will make the search faster

I'm sceptical it will make any difference

> * It is easy to highlight the results (do we even show the filename anywhere in the UI?)

I don't feel strongly about it, I think it can be both very useful and distracting.
Maybe long term, we can use a lib that ranks search results to likeliness of match, and sort results by rank.

For now, I'd recommend we keep it and listen for feedback from users.

PR #5749, opened 2019-02-25
ADD activity to addon-background and addon-viewport tool

ndelangen:
Issue: https://github.com/storybooks/storybook/issues/5596

## What I did
ADD active indicators to addon-background tool & addon-viewport tool

## How to test
Toggle the viewport and background tool on an example

expected result:
A blue 3px border should appear when the tool is active


shilman (approved):
@ndelangen Add stories for these?

ndelangen:
We should add stories for the toolbar, when we have the toolbar extracted from preview.

PR #5753, opened 2019-02-25
fix: remove call to `forceReRender()` on STORY_CHANGED in addon knobs

gaetanmaisse:
Issue: #5382

## What I did

Fix issue https://github.com/storybooks/storybook/issues/5382

I removed call to `forceReRender()` on STORY_CHANGED in addon knobs as this call is no necessary because it's already done by SB core and was causing an error in SB Angular.

## How to test

Open Angular CLI example and check that there is no error in browser console when switching between stories using addon knobs.


shilman (approved):
Awesome! Works great for me!

PR #5790, opened 2019-02-28
add gov.uk story book react site

theatlasroom:
Issue: Noticed that gov.uk react wasnt listed

## What I did
Added gov.uk to list of examples


shilman (approved):
Thanks, looks great!

Issue #5781, opened 2019-02-27
v5 applies decorators to all the kind, no matter where it's placed


Hypnosphi:
```
import React from 'react';
import { storiesOf } from '@storybook/react';
import centered from '@storybook/addon-centered/react';

storiesOf('Stories', module)
  .add('noncentered', () => 'Hello')
  .addDecorator(centered)
  .add('centered', () => 'Hello');
```

V4: only second story is centered
V5: both stories are centered

This should be either fixed or documented in `MIGRATION.md`

tmeasday:
I'm fairly sure this was an intentional change (right @ndelangen).

Is it possible to add per-story decorators? It doesn't actually look like it is, but it would be easy to add via changing this line:

https://github.com/storybooks/storybook/blob/9f87b711b5aa9f487225c48407c2a23b7cd5df17/lib/client-api/src/client_api.js#L206

To 

```js
          getDecorators: () => [allParam.decorators, ...localDecorators, ..._globalDecorators, withSubscriptionTracking],
```

(Arguably we should make `localDecorators` and `globalDecorators` just be set via `addParameters` behind the scenes, maybe something for v5.1).

If everyone is OK with this, my `MIGRATION.md` notes would go something like:

### Decorator changes

Previously, adding decorators to a component/kind was "stateful". If you added a decorator after a story, the decorator would only apply to future stories. For example:

```js
storiesOf('Stories', module)
  .add('noncentered', () => 'Hello')
  .addDecorator(centered)
  .add('centered', () => 'Hello');
```

Moving forward, we have a new API planned that will not support this. As it is clearer, if you want to apply a decorator to subset of a kind's stories, you can use the decorator parameter;

```js
storiesOf('Stories', module)
  .add('noncentered', () => 'Hello')
  .add('centered', () => 'Hello', { decorators: [centered] });
```

What does everyone think? We could also work to make the existing behaviour deprecated.

tmeasday:
I added local decorators in https://github.com/storybooks/storybook/pull/5806

shilman:
@tmeasday I think we should deprecate the old behavior as part of the 5.0 release. Feels "tricky" to me...

tmeasday:
Deprecate or remove entirely?

shilman:
Sorry, remove entirely and document rationale in `MIGRATION.md`

ndelangen:
```js
storiesOf('Stories', module)
  .add('noncentered', () => 'Hello')
[...]

PR #5806, opened 2019-03-01
Allow local decorators via params

tmeasday:
for #5781

## What I did

I added support for local decorators via a parameter. This doesn't resolve the issue but at least it provides a palatable alternative.

### DIFF
--- a/lib/client-api/src/client_api.js
+++ b/lib/client-api/src/client_api.js
@@ -203,7 +203,12 @@ export default class ClientApi {
         },
         {
           applyDecorators: this._decorateStory,
-          getDecorators: () => [...localDecorators, ..._globalDecorators, withSubscriptionTracking],
+          getDecorators: () => [
+            ...(allParam.decorators || []),
+            ...localDecorators,
+            ..._globalDecorators,
+            withSubscriptionTracking,
+          ],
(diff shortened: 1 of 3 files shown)


lib/client-api/src/client_api.js
```diff
@@ -203,7 +203,12 @@ export default class ClientApi {
         },
         {
           applyDecorators: this._decorateStory,
-          getDecorators: () => [...localDecorators, ..._globalDecorators, withSubscriptionTracking],
+          getDecorators: () => [
+            ...(allParam.decorators || []),
```
shilman: @tmeasday not part of the change per se, but why is this called `allParam`?

shilman (approved):

tmeasday:
🤷‍♀️

lib/client-api/src/client_api.js
tmeasday: @shilman I guess it's a combination of all parameters (local global etc) in a similar way?

tmeasday: This is why I meant that `addDecorator` could just be redefined to `addDecorator(d) => addParameters({ decorators: [d] })`

PR #5829, opened 2019-03-04
Restore viewport behaviour

tmeasday:
Issue: #5757 and #5818

A few things were going on here:
 - You no longer got a set of viewports by default by just registering the addon
 - You couldn't set the `defaultViewport` any more

## What I did

(Partially) restored the old API (still via the parameter API), added warnings for no-longer supported stuff, and documented it.

NOTE: we lost the `onViewportChange` option. I'm not sure that was very widely used? Maybe we'll hear about it if so 🤷‍♂️


shilman (approved):
Thanks @tmeasday, this is awesome!

PR #5843, opened 2019-03-04
FIX base theme initialization and theme bootup

ndelangen:
Issue: #5824 #5780 

creating a theme that derived off of `light` or `dark` default themes was impractical, and wasn't working quite as users were expecting it to behave.

## What I did
ADD ability for users to specify baseTheme, and the theme createFn will merge in the appropriate themeVars.


lib/theming/src/themes/light.ts
```diff
@@ -1,36 +1,10 @@
 import { create } from '../create';
-import { color, typography, background } from '../base';
+
+import { themeVars } from './dark-vars';
```
shilman: @ndelangen should this be from `light-vars`?

shilman (changes requested):
@ndelangen please test the hell out of this?

lib/theming/src/themes/dark.ts
```diff
@@ -1,36 +1,9 @@
...
-  barTextColor: '#999999',
-  barSelectedColor: color.secondary,
-  barBg: color.darkest,
-
-  // Form colors
-  inputBg: '#3f3f3f',
-  inputBorder: 'rgba(0,0,0,.3)',
-  inputTextColor: color.lightest,
-  inputBorderRadius: 4,
+  ...themeVars,
```
tmeasday: Isn't this superfluous?

lib/theming/src/create.ts
```diff
@@ -110,76 +115,85 @@ const darkSyntaxColors = {
...
+
+export const create = (vars: ThemeVar, rest?: Rest): Theme => {
+  const inherit: ThemeVar = { ...vars, ...(base[vars.base] || base.light), ...base.light };
+  const p = pick.bind(null, inherit);
+
+  return {
+    base: vars.base,
+    color: createColors(inherit),
+    background: {
+      app: p('appBg'),
```
tmeasday: I don't really understand why this isn't just `inherit.appBg`?

lib/theming/src/create.ts
```diff
@@ -110,76 +115,85 @@ const darkSyntaxColors = {
...
-  typography: {
-    fonts: {
-      base: vars.fontBase || typography.fonts.base,
-      mono: vars.fontCode || typography.fonts.mono,
+function pick<B, K extends keyof B>(b: B, key: K): B[K] {
+  return b[key];
+}
+
+export const create = (vars: ThemeVar, rest?: Rest): Theme => {
+  const inherit: ThemeVar = { ...vars, ...(base[vars.base] || base.light), ...base.light };
```
tmeasday: Should this just be `themeVars` or something

tmeasday (changes requested):
I feel like this code would benefit from some tests but I appreciate this is a fairly late change

lib/theming/src/create.ts
tmeasday: I mean this is just a naming question, not a big deal ;)

lib/theming/src/themes/dark.ts
tmeasday: As in, wouldn't `create({ base: 'dark' })` have the exact same effect, by definition.

shilman:
When I add this line to `offcial-storybook/config.js` the app hangs:

```
    theme: create({ base: 'dark', colorPrimary: '#FF4785', colorSecondary: '#1EA7FD' }),
```

lib/theming/src/themes/light.ts
ndelangen: indeed

lib/theming/src/create.ts
ndelangen: hmm, yeah

ndelangen:
I'm writing tests @shilman

shilman:
@ndelangen A couple comments, both outside the scope of this PR:
- After testing out a buggy version, I had to do `localStorage.clear()` to get the theme to load. Seems like a bug, but maybe it's a corner case
- Seems like the theme serialization should be profiled.. pretty slow to load on my laptop

shilman (changes requested):
@ndelangen Maybe a stupid question:

```
    theme: create({ base: 'dark', colorPrimary: 'red', colorSecondary: 'green' }),
```

I did this in `official-storybook`. I see green in the resulting UI, but I don't see any red. Doesn't seem right to me?

ndelangen:
@shilman 

I'd like to remove the need to do theme serialization asap.

theme.color.primary seems to be very rarely used, @domyen came up with the naming and usage of colors.

domyen:
@ndelangen yea `theme.color.primary` is not used at all, happy for you to axe it! (the var was mirrored from storybook/frontpage, but we don't need it here)

shilman:
@domyen @ndelangen seems like an unfortunate choice of names then... I thought the point of this minimal theme var abstraction was to help us preserve backwards compatibility?

shilman:
We should be making minimal changes at this point though, so I don't really want to change it. Though it makes me cringe a bit.

domyen:
I’m talking about what’s exposed in the create() function “colorPrimary”.

We still want to have a theme.color.primary as a part of the larger theme
colors (it’s set to storybook pink).

ndelangen:
@shilman @tmeasday 

I did a refactor to ensure the least amount of data needed is transferred over the channel.

I ended up not 'creating' the full theme and then trying to transmit, instead it's much more efficient to transfer the `ThemeVars` across and construct the actual `Theme` in the manager.

This means no more hack to make sure the functions make it across, since is a plain flat object.

I know it's not what you were hoping for: it's a bigger change I wish was necessary.

ndelangen:
I wrote in such a way, users won't need to change anything.
As long as they are using `create()`..

ndelangen:
I also noticed that setOptions is called a ton, and it's was re-creating the theme on every navigation.

It'd be nice to fix this a bit better, but for now I improved it by `deepEqual(theme, updatedTheme)`.

PR #5844, opened 2019-03-04
ADD parameters to compile-js to no longer copy .ts files into dist

ndelangen:
Issue: in our dist there are typescript source files

## What I did
I added a cli parameter to babel to NOT copy these files into dist


kroeder:
I think this might be already fixed in https://github.com/storybooks/storybook/pull/5109 - I will solve my CI issues and then hopefully merge this into next 

But we can of course merge this and throw it away afterwards 🙂

kroeder (approved):

shilman (approved):
🙆‍♂️

jonrimmer:
This is not fixed, the `.ts` files are still copied. Babel's `--copy-files` argument copies anything not processed by Babel, which includes anything marked to be ignored: https://github.com/babel/babel/issues/6226

PR #5849, opened 2019-03-04
Fix code example in "writing-addons" docs

rockingskier:
Issue: Missing comma in docs.

## What I did

Added a comma to the docs.


rockingskier:
I've just noticed that `STORY_CHANGED` is import however `STORY_RENDERED` is used:

`import { STORY_CHANGED } from "@storybook/core-events";`

```
api.on(STORY_RENDERED, this.onStoryChange);
api.off(STORY_RENDERED, this.onStoryChange);
```

I'm hapy to update but not sure which is correct.

shilman (approved):
Thanks for the fix!

PR #5856, opened 2019-03-05
Ensure the theme is insta-loaded from local storage

tmeasday:
Issue: Theme was not loading from local storage, leading to a FOUC

## What I did

Read the correct key off local storage and preference persisted options to the defauls

## How to test

Set `theme: themes.dark` and reload the SB


shilman (approved):
LGTM

PR #5858, opened 2019-03-05
Tech/workaround persisted options with defaults

tmeasday:
Issue: a partial fix for #5857 

## What I did

Ensure persisted options do not apply once the real options are loaded.

See the comment on the PR.

## How to test

Set, then unset an option, reload the page.


shilman (approved):
LGTM

PR #5867, opened 2019-03-05
fix: move `react-select` dependency to 2.2

SimenB:
Issue:

Version 2.3 went from emotion 10 to emotion 9, causing issues with code dedupe for us (since yarn decided to hoist differently).

2.3 is still en semver range here, but this allows us to lock down the older version.

We're stuck on `alpha.5` due to the upgrade that landed in alpha 6

See https://github.com/JedWatson/react-select/pull/3346

## What I did

Lowered the version in package.json.

An alternative plan is to not use a lib that's larger than react-dom just to render a dropdown in an addon 😛 

## How to test

Not much to test

### DIFF
--- a/addons/knobs/package.json
+++ b/addons/knobs/package.json
@@ -35,7 +35,7 @@
     "qs": "^6.5.2",
     "react-color": "^2.17.0",
     "react-lifecycles-compat": "^3.0.4",
-    "react-select": "^2.3.0"
+    "react-select": "^2.2.0"
   },
   "peerDependencies": {
     "react": "*"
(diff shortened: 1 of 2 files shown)


ndelangen (approved):

ndelangen:
If this frees up your upgrade path, I'm ok with having a lower version of this package for a while!

SimenB:
Thank you!

SimenB:
>Would love to switch to something lighter!

I haven't tested it, but maybe https://ui.reach.tech/menu-button could do the trick for you?

PR #5884, opened 2019-03-05
Fix imported package for "dark" theme

mattfwood:
Issue:

Could not import themes from `@storybook/components`.

image

## What I did

Changed the imported example for theming to use `@storybook/theming` instead of `@storybook/components`.

I followed the documentation, but it looks like `themes` was moved to the `@storybook/theming` package or this was a typo.

## How to test


wuweiweiwu (approved):
Nice catch!

PR #5894, opened 2019-03-06
ADD a depenendency on `core-js@2` to `@storybook/core`

tmeasday:
Issue: #5831

We need it as we use `@babel/preset-env` with the `useBuiltIns: 'usage'` option.

See https://babeljs.io/docs/en/babel-preset-env#usebuiltins


tmeasday:
Huh?

tmeasday:
Probably should have looked at the delta before creating the PR 🤷‍♂️

PR #5900, opened 2019-03-06
Fix typo in example, use MyPanel component

44px:
Issue:
`render` function used undeclared `Panel` component. Seems like it should use `MyPanel` instead.

## What I did
Fixed small typo in example in "Writing Addons" article.


shilman (approved):
Thanks!!!

PR #5905, opened 2019-03-06
Readme updated to reference jsDelivr for the badge

maraisr:
Could we prompt people to use jsDelivr or alike? The github.com ones don't work outside of GitHub.


shilman (approved):
LGTM!

PR #5944, opened 2019-03-07
Theme docs fix

PascalPixel:
The doc says `export default create({...` for the theme, which gives an error when you do `import { ... } from ...`


shilman (approved):
Thanks @Superpencil 🚀

PR #5947, opened 2019-03-07
Fixed missing iframe.html in output directory

anru:
path.join is unsafe when using absolute paths after first argument.
For resolving from cwd path.resolve function is safe and makes correct results in both cases, using absolute or relative paths.

Issue:

There is no iframe.html after `build-storybook -o /any/abs/path` is successfully finished.

## What I did

`build-storybook -o /tmp/any/abs/path` 

## How to test

pass any absolute path to `-o` cli argument to `build-storybook` command.


shilman (approved):
Ouch! Great catch!!! 👯

PR #5948, opened 2019-03-07
Update slow start to include correct dependencies

eikooc:
The `@storybook/addon-actions` was missing

## What I did

Edited documentation


ndelangen:
I removed the addon-actions from the example in the docs. To keep them more focussed on what they are about.

Thanks for the PR @eikooc, really appreciated! 🙇

Issue #4324, opened 2018-10-09
Storysource addon - allow prism configuration


cheshrkat:
If you are reporting a bug or requesting support, start here:
### Bug or support request summary

Have added Storysource and was expecting to be able to configure Prism syntax highlighting - eg. choose the theme, tweak the highlight style and so on. However I can't find anything in the docs about that - I can see the Prettier settings but nothing about Prism? That all seems hard-coded. This is an issue as the chosen theme is really hard to read and clashes badly with our own design.

Not sure if I'm just missing something, or if this is a feature request...

### Steps to reproduce

n/a I think

### Please specify which version of Storybook and optionally any affected addons that you're running

- "@storybook/react": "^3.3.15",
- "@storybook/addon-storysource": "^3.4.11",

### Affected platforms 

n/a I think

### Where to start

n/a I think

### Acceptance criteria

- Ability to choose theme for Prism when using Storysource
- Ability to configure/extend the styles applied to the Story tab, eg. the highlight style

igor-dv:
BTW, @ndelangen and Co are working on overhaul UI changes, where this is addressed, but it will take some time to finish.

ndelangen:
The UI overhaul will include theming for all syntax highlighting inside storybook.

https://github.com/storybooks/storybook/pull/4086

cheshrkat:
Thanks for the advice. If I get time I'll take a look at a PR (won't be right away, just to be honest :)).

vanessayuenn:
`addon-storysource` has been deprecated. As of Storybook 9, it has been replaced by the [addon-docs code panel](https://storybook.js.org/docs/writing-docs/code-panel), which provides more snippet options. Please see [migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#storysource-addon-removed).

PR #5971, opened 2019-03-08
Add default style for story source code syntax and add style option for the addon

yjcxy12:
Issue: connects #4324

## Summary
I've noticed after V5 the code in story source panel has minimal code syntax highlighting. Where in V4 it had dark theme and was much easier to read the source.

I've created this PR for an initial discussion for story source addon code syntax highlighting.

I'm aware that as part of V5 upgrade, there's bunch of theming variables you can set for code syntax highlighting, however I find it very difficult to implement in consuming applications. Thus the PR so that users can pass in any themes within `react-syntax-highlighter/dist/styles/prism`.

Please feel free with the feedback:
- Is this something you'd consider for the story source addon?
- Any other approaches in mind?

## What I did
- add default prism style for story source plugin
- add storySource option to customise styling
## How to test
- added `Addon | Storysource` stories to `official-storybook` kitchen sink app. Navigate to the stories, and there're two stories, one with default prism theme, another with tomorrow prism theme.

Yes.
Yes.
Yes.

### DIFF
--- a/addons/storysource/src/StoryPanel.js
+++ b/addons/storysource/src/StoryPanel.js
@@ -5,6 +5,8 @@ import { Link } from '@storybook/router';
 import { SyntaxHighlighter } from '@storybook/components';
 
 import { createElement } from 'react-syntax-highlighter';
+import prism from 'react-syntax-highlighter/dist/styles/prism/prism';
+import { OPTION_KEY } from './constants';
 import { EVENT_ID } from './events';
 
 const StyledStoryLink = styled(Link)(({ theme }) => ({
@@ -159,13 +161,18 @@ export default class StoryPanel extends Component {
   };
 
   render() {
-    const { active } = this.props;
+    const { active, api } = this.props;
     const { source } = this.state;
+    const storyData = api.getCurrentStoryData();
+    const options = storyData && storyData.parameters && storyData.parameters[OPTION_KEY];
+    const finalStyle = options && options.style ? options.style : prism;
 
     return active ? (
       <StyledSyntaxHighlighter
         language="jsx"
         showLineNumbers="true"
+        style={finalStyle}
+        useInlineStyles
         renderer={this.lineRenderer}
         format={false}
         copyable={false}
--- a/addons/storysource/src/constants.js
+++ b/addons/storysource/src/constants.js
@@ -1,0 +1,1 @@
+export const OPTION_KEY = 'storySource';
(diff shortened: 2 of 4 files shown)


Armanio:
Looks good! But what about other frameworks? Vue, svelte, angular?

yjcxy12:
I believe it should work in other frameworks in the same way as it's added to the add-on source code.

Think what's missing is examples for different themes for other frameworks. I can add them in if that's required.

Armanio:
Oh, I confused: 'react-syntax-highlighter' its syntax highlighter for use in react, not react syntax highlighter. 😃

libetl:
storysource will get a big revamp soon or later.
Your change is likely to be accepted but will not stay for the long term.
Please ask me in discord#storysource for more info about that.

addons/storysource/src/StoryPanel.js
```diff
@@ -5,6 +5,8 @@ import { Link } from '@storybook/router';
 import { SyntaxHighlighter } from '@storybook/components';
 
 import { createElement } from 'react-syntax-highlighter';
+import prism from 'react-syntax-highlighter/dist/styles/prism/prism';
```
libetl: I don't think that you should use dist in the import path

libetl: However you maybe have no choice

yjcxy12: Got the example code from readme. 
Don't see any other options. https://github.com/conorhastings/react-syntax-highlighter/blob/master/README.md#prism

yjcxy12:
I have no problem with it being accepted or rejected or only merged for a short time. Whatever is best for the project.

addons/storysource/src/StoryPanel.js
libetl: ok :+1:

libetl:
in 6.x or 7.x, monaco editor will replace the current codemirror + prism.

addons/storysource/src/StoryPanel.js
ndelangen: This might be a more versatile approach to styling the code-editor then the theming api, but I'm not quite happy with how this would bundle ALL prism themes into the manager, even if the user isn't even using any.

Can we make this lazy loading somehow?

ndelangen (changes requested):

addons/storysource/src/StoryPanel.js
igor-dv: Why would even make that default ? I mean it looks ok to pass the style from the outside, and the default could be set in `config.js`.

PR #5978, opened 2019-03-09
SYNC prettier printWidth with eslint && FIX resulting linting warnings

ndelangen:
Issue: printWidth for eslint & typescript is configured differently

## What I did
I set them to the same value & fixed all resulting warnings


shilman (approved):
LGTM

PR #5986, opened 2019-03-09
fixing documentation for correct src location in angular guide

andrewevans0102:
Issue:

## What I did

I changed the code in the angular guide
from:
const req = require.context('../stories', true, /.stories.ts$/);

to be:
const req = require.context('../src/stories', true, /.stories.ts$/);

This is to match what is automatically generated when you do the initial install of the CLI into your angular project.

## How to test


gaetanmaisse:
Not sure this is relevant because the section you are updating is about **Manual setup** and not about what the initial install of the CLI does. Moreover, the next line gives some details about this: `That’ll load stories in ../stories/index.js. You can choose where to place stories, you can co-locate them with source files, or place them in an other directory.`

PR #5987, opened 2019-03-09
Updated React reference to Vue

MKRazz:
Issue: Small reference to `React` instead of `Vue` in Vue guide

## What I did
Updated the reference

## How to test
No testing needed, just a copy update.


shilman (approved):
Thanks!

PR #5988, opened 2019-03-09
Improve typings of theming lib

gaetanmaisse:
## What I did

Exported `styled` was implicitly of type `CreateStyled<any>`, now it is typed using SB Theme.
This stronger typing reveals an invalid property use in a11y addon.

## How to test

SB must compile.

### DIFF
--- a/lib/theming/src/index.ts
+++ b/lib/theming/src/index.ts
@@ -1,8 +1,9 @@
-import styled from '@emotion/styled';
+import emotionStyled, { CreateStyled } from '@emotion/styled';
 import dark from './themes/dark';
 import light from './themes/light';
+import { Theme } from './base';
 
-export { styled };
+export const styled = emotionStyled as CreateStyled<Theme>;
 
 export * from './base';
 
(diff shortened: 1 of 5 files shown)


lib/theming/src/index.ts
```diff
@@ -1,8 +1,9 @@
-import styled from '@emotion/styled';
+import emotionStyled, { CreateStyled } from '@emotion/styled';
 import dark from './themes/dark';
 import light from './themes/light';
+import { Theme } from './base';
 
-export { styled };
+export const styled = emotionStyled as CreateStyled<Theme>;
```
gaetanmaisse: @ndelangen am I right to type `styled` with `Theme` or should I had used `ThemeVars`?

addons/a11y/src/components/Report/Info.tsx
```diff
@@ -4,6 +4,7 @@ import { styled } from '@storybook/theming';
 import { Result } from 'axe-core';
 
 const Wrapper = styled.div(({ theme }) => ({
+  // Hmm sadly background of SB theme has no `bar` property :/
```
gaetanmaisse: If my typing is right then `backgroundColor` is not properly defined, any idea what could be the correct value?

lib/theming/src/index.ts
ndelangen: The components will receive the `Theme`

addons/a11y/src/components/Report/Info.tsx
gaetanmaisse: Discussing with @ndelangen `bar` should be added and it's default value id  `#fff`

PR #6016, opened 2019-03-11
fix rendering code block in @storybook/addon-info

pveyes:
Issue: #5915 

## What I did

Fix markdown rendering logic inside `@storybook/addon-info` package to handle code block. When code block syntax is passed to addon info, render will fail because `Code` component will receive `language` and `code` props instead of `props.props`.

To fix this we forward `code` props as `children` as required by `SyntaxHighlighter` component. We also render `ThemeProvider` because `SyntaxHighlighter` expect a emotion theme as seen in  https://github.com/storybooks/storybook/blob/next/lib/components/src/syntaxhighlighter/syntaxhighlighter.js#L36 or else it will fail

I'm actually not sure if this is the correct approach because by hardcoding `ThemeProvider` we'll always get light theme (using default argument of `convert` function). Any pointer to how to solve this is appreciated.

**Note** that even using this patch, I still can't use `~~~` to annotate code block (maybe because it's not supported by `marksy`?) but triple backticks ``` works fine

## How to test

- Is this testable with Jest or Chromatic screenshots? yes snapshot included

We can refer to issue #5915 for manual testing, implementation should be covered by the test included


pveyes:
There's a bunch of snapshot changes because we render code block with syntax hightlight now

olgagiza-home24 (approved):

pveyes:
rebased to latest master, ready to review

whengely:
Are changes necessary or does this just need reviewed?

ndelangen:
I see `import { SyntaxHighlighter } from '@storybook/components';` was already being used prior to this PR.. This ties the addon-info to a high version of react (because emotion 10).

Unsure

shilman:
I think we should not break older versions of react with this change. We should find a better fix and use this one for Storybook Docs where things will be running in the manager

ndelangen:
@shilman Yeah docs-mode is a few weeks away though I think, maybe sooner in alpha releases. but still..

As mentioned, it seems SyntaxHighlighter was already being used prior to this PR.. So it's not new to this PR..

It should really be this way, because it ties the addons very closely to the newest version of react we use.

I guess if this isn't a problem for users, we can use continue..

ndelangen:
Let's continue.

ndelangen:
@shilman this is good to merge, and i’m fairly sure it’s NOT a breaking change, since the dependency for SyntaxHighlighter was already there:
https://github.com/storybooks/storybook/pull/6016

addons/info/src/components/markdown/code.js
```diff
@@ -1,12 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { SyntaxHighlighter } from '@storybook/components';
+import { ThemeProvider, convert } from '@storybook/theming';
 
-// XXX: is this a bug? should it be (props) => ?
-const Code = ({ props }) => <SyntaxHighlighter bordered copyable {...props} />;
+const Code = ({ code, language, ...props }) => (
+  <ThemeProvider theme={convert()}>
```
shilman: what does this `convert()` do?

shilman:
@ndelangen it's the `ThemeProvider` that i'm worried about

addons/info/src/components/markdown/code.js
pveyes: this is to make sure we get theme object that's needed by `SyntaxHighlighter`

I also have mixed feeling about this because currently user will be forced with light theme (calling `convert` without argument will make it return light theme)

If we have a way to get currently used theme inside this `Code` component, we can just pass it to `convert` as args

artyomtrityak:
Any progress on this one? We want to use docs from source in our project and blocked by this one.

olga-giza:
@artyomtrityak
You don't need to fork or create another addon.
Here is simple walkaround
```
import SyntaxHighlighter from '@storybook/components/dist/syntaxhighlighter/syntaxhighlighter';
import { ThemeProvider } from 'emotion-theming';
import * as baseTheme from '@storybook/theming/dist/base';

const CodeBlock = ({ language = 'text', code }) => (
  <ThemeProvider theme={baseTheme}>
    <SyntaxHighlighter bordered copyable language={language}>
      {code}
    </SyntaxHighlighter>
  </ThemeProvider>
);

export const components = {
  code: CodeBlock,
};
```

and just then use it in your story:
```
import { components } from './path/to/storybook/overrides'
storyOf('Name', module).add(
      'Other name',
      () => <div>demo</div>,
      { info: { components } }
);
```

whengely:
> @artyomtrityak
> You don't need to fork or create another addon.
> Here is simple walkaround
> 
> ```
> import SyntaxHighlighter from '@storybook/components/dist/syntaxhighlighter/syntaxhighlighter';
> import { ThemeProvider } from 'emotion-theming';
> import * as baseTheme from '@storybook/theming/dist/base';
> 
> const CodeBlock = ({ language = 'text', code }) => (
>   <ThemeProvider theme={baseTheme}>
>     <SyntaxHighlighter bordered copyable language={language}>
>       {code}
>     </SyntaxHighlighter>
>   </ThemeProvider>
> );
> 
> export const components = {
>   code: CodeBlock,
> };
> ```
> 
> and just then use it in your story:
> 
> ```
> import { components } from './path/to/storybook/overrides'
> storyOf('Name', module).add(
>       'Other name',
>       () => <div>demo</div>,
>       { info: { components } }
> );
> ```

This is to add a bit if you're using typescript, make it globally applied, and for completeness.

I needed to add a typing to a d.ts file

 `declare module '@storybook/components/dist/syntaxhighlighter/syntaxhighlighter'`

install @storybook/components

then added this to my config.js withInfo addDecorator...
`addDecorator(withInfo({
  components,
  ...
}))`

ndelangen:
Before:
Screenshot 2019-06-14 at 00 47 58

---
after:
Screenshot 2019-06-14 at 00 46 33

ndelangen:
@shilman these some of the improvements you were thinking about?

Issue #6019, opened 2019-03-11
[docs] Add hint regarding addon tabs order


HerrBertling:
I'd like to add a short sentence about the order of the addon tabs:

![image](https://user-images.githubusercontent.com/1827373/54135172-9f5c2c80-4419-11e9-907f-7ce101e13210.png)

I asked within the Storybook Slack channel and got the helpful hint from @benediktvaldez that the order of imports in the `addons.js` is what produces the order of the tabs.

This wasn't obvious to me through reading the docs.

I'd add a short sentence within "Using addons" since I guess other people would search for this as well.

Hope it's okay to directly open a PR with this, shouldn't™ be a huge change :)

PR #6021, opened 2019-03-11
docs(addons): add hint about addon tab order

HerrBertling:
Issue: #6019 

which I missed and I guess would be rather helpful for others searching for this.

## What I did

Added a short paragraph about how to change the tab order. I also inserted some headlines to structure the page a bit more. Also, because of the injected paragraph, I changed the sentence about the usage since it now lacked context.

## How to test

1. Start the docs subfolder as described in its readme.
1. Navigate to "Using addons".
1. Read the sentence 😆

### DIFF
--- a/docs/src/pages/addons/using-addons/index.md
+++ b/docs/src/pages/addons/using-addons/index.md
@@ -29,7 +29,19 @@ This will register all the addons and you'll be able to see the actions and note
 
 ![Stories without notes](../static/stories-without-notes.png)
 
-Now when you are writing a story, import it and add some notes:
+## Addons tab order
+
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM tho I'm not sure this still makes sense in SB5 where the notes addon is a `TAB` type? @ndelangen?

shilman (changes requested):
On second thought I think this doesn't make sense in SB5 since addon-notes doesn't show up in the addons panel anymore. Can you use an example like `knobs` that does? cc @ndelangen

HerrBertling:
> Can you use an example like knobs that does?

Sure!

ndelangen (approved):

shilman (approved):

PR #6026, opened 2019-03-11
fix(a11y): remove bg color from reports

Armanio:
Issue: #6025 

## What I did
Remove bg from reports

## How to test
Open cra-kitchen-sink with dark theme
Open a11y addon
Show reports

no
no
no

Before:

After:

### DIFF
--- a/addons/a11y/src/components/Report/Info.tsx
+++ b/addons/a11y/src/components/Report/Info.tsx
@@ -3,11 +3,10 @@ import React, { FunctionComponent } from 'react';
 import { styled } from '@storybook/theming';
 import { Result } from 'axe-core';
 
-const Wrapper = styled.div(({ theme }) => ({
-  backgroundColor: theme.background.bar,
+const Wrapper = styled.div({
   padding: '12px',
   marginBottom: '10px',
-}));
+});
 const Help = styled.p({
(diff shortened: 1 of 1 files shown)


addons/a11y/src/components/Report/Info.tsx
```diff
@@ -3,11 +3,10 @@ import React, { FunctionComponent } from 'react';
 import { styled } from '@storybook/theming';
 import { Result } from 'axe-core';
 
-const Wrapper = styled.div(({ theme }) => ({
-  backgroundColor: theme.background.bar,
```
wuweiweiwu: I think we should be updating `theme.background.bar` in this fix instead. As we want it to be responsive to theme changes

Armanio: Yes, i thought about this, but a11y addon not have a bg on other panels.

ndelangen: indeed!

ndelangen: Looking at it, I do think there's both a bug in the application of the correct color to theme.background.bar AND this element should have the background color removed!

ndelangen (approved):

PR #6042, opened 2019-03-12
fix (addon/backgrounds): addon panel doesn't hide

### LINKED ISSUE #6041: addon-backgrounds doesn't hide the addons panel when transparent
**Describe the bug**
Addon-backgrounds overwrites the background color of the preview. However, if the color is "cleared" or set to transparent, then the preview panel is also transparent and will not show the default theme background color but instead reveal the underlying addons-panel that should be hidden. This is due to the fact that "Showing or Hiding the Addons Panel" does in fact not change the size of the addons panel but changes the size of the preview panel so it completely covers it.

**To Reproduce**
1. Create a new empty directory
2. Create the HTML Kitchensink with `npx -p @storybook/cli sb init --type html`
3. Add addon-backgrounds with `npm i -D @storybook/addon-backgrounds`
4. [Add import to config.js](https://github.com/storybooks/storybook/tree/master/addons/backgrounds#configuration): `import '@storybook/addon-backgrounds/register';`
5. Add backgrounds to /stories/index.stories.js:
```js
import { storiesOf } from '@storybook/html';
storiesOf('Demo', module)
  .addParameters({
    backgrounds: [
      { name: 'transparent', value: 'transparent', default: true },
      { name: 'facebook', value: '#3b5998' },
    ],
  })
  .add('heading', () => '<h1>Hello World</h1>');
```
6. Start the Storybook
7. Untick the "show Addons" in the "..." Dropdown or press hotkey `A`
**BUG: It still shows the Addons Panel**
8. Select "facebook" as background color in the background selector
9. Select "Clear Background" as background color in the background selector
**BUG: It still shows the Addons Panel**

**Expected behavior**
The Addon Panel should be hidden. Instead it still shows.

**Screenshots**
image

**Additional context**
Although this is technically correct (transparent being transparent), it should not be transparent because the preview panel needs one fallback background color. this should be the default background theme color.

divslinger:
Fixes #6041

Instead of turning the preview transparent, addon-backgrounds should set the background color to the theme's default.
I added the background import and set background.content as the replacement for all transparent returns


ndelangen:
This would make the preview background not respect the user's theme.

ndelangen:
I refactored this addon quite a bit in my PR for migrating the core to typescript.

I incorporated this fix into that PR: 5592373782dca0646684eb4211fd46ba74134520

I also made sure the user's theme is respected.

ndelangen:
I just pushed a little fix to this PR, because I want @glitchperfect to have this PR merged.

I really appreciate the effort.

I think this should be good to merge now.

I'll deal with the merge conflict in my api branch.

shilman (approved):
LGTM

ndelangen (approved):

PR #6043, opened 2019-03-12
[bug]Error message when removing value from number knob 

toshi1127:
Issue: #5912 

When removing the value from the "number" knob, following error appears in the console:

## What I did
Since number may be null, I assigned '' at that time.

|before|after|
| --- | --- |
|スクリーンショット 2019-03-12 21 45 31|スクリーンショット 2019-03-12 21 46 05|

## How to test

- Is this testable with Jest or Chromatic screenshots? I'm not sure what is the proper way to do it here.


shilman (approved):
LGTM, thanks!! 🥇

PR #6072, opened 2019-03-13
[Docs] Vue guide typo

viniciuskneves:
The example shows 3 stories but the text refers to just 2 stories which might be a bit confusing for newcomers.


shilman (approved):
👍

PR #6086, opened 2019-03-14
Updated Readme.md

arnaudmanaranche:
# What I did

Update the README for the addon `Viewport`.

# Issue

Link: #5970


shilman (changes requested):
Hmm I think the original docs makes more sense because `options` can contain various keys, not just `viewports`?

ndelangen (approved):

PR #6091, opened 2019-03-14
addon-notes: register as a panel or tab

hipstersmoothie:
Issue: People in my organization prefer having the notes tab in the addon panel, but in v5 it only renders as a tab.

## What I did

Expose a `register-panel` script from addon-notes. This allows the user register the notes as a panel.

## How to test

- Is this testable with Jest or Chromatic screenshots? yes, but addons seem untested

I have verified that this works in my storybook. If a user tries to use both of the registers then storybook breaks, but this is expected because you are registering the same addon twice

### DIFF
--- a/addons/notes/README.md
+++ b/addons/notes/README.md
@@ -19,7 +19,10 @@ Then create a file called `addons.js` in your storybook config.
 Add following content to it:
 
 ```js
+// register the notes addon as a tab
 import '@storybook/addon-notes/register';
+// register the notes addon as a panel
+import '@storybook/addon-notes/register-panel';
 ```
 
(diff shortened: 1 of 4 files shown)


hipstersmoothie:
@ndelangen @shilman How do you guys feel about this change?

shilman (approved):
This works for me! @ndelangen ?

frehner:
Just started using storybook, and I would love to have notes in the addon panel as well.

It's kind of confusing because currently the screenshot in the docs show it in a panel, but I guess the default for storybook v5 is a tab now?

ndelangen (approved):

PR #6095, opened 2019-03-14
Migrate @storybook/components to TS

gaetanmaisse:

## What I did

 - [x] 🏗 Migrate `@storybook/components` to TypeScript
 - [x] 🔧 Add some types and interfaces and use them to type codebase 
 - [x] ⚔️ Fight with TS compiler about anonymous functions, see https://github.com/storybooks/storybook/pull/6095#issuecomment-477480930 and https://github.com/storybooks/storybook/pull/6095/commits/5fa1be5447c657705879604c4eda7fe6fc13c0d5
 - [x] ♻️ Localy disable a md linter rule (as we need to do what is done) to (try to) fix `Lint (Storybook)` job

## TODO in another PR

 - 🛠 Refactor some imports/exports
 - 📦 Use these typings in addons


ndelangen:
Mind this PR:
https://github.com/storybooks/storybook/pull/5402

You may want to base on that, it's close to being merged.

lib/components/package.json
```diff
@@ -42,6 +44,16 @@
     "recompose": "^0.30.0",
     "simplebar-react": "^0.1.4"
   },
+  "devDependencies": {
+    "@storybook/addon-actions": "5.1.0-alpha.4",
+    "@storybook/react": "5.1.0-alpha.4",
+    "@types/js-beautify": "^1.8.0",
+    "@types/react-textarea-autosize": "^4.3.3",
```
gaetanmaisse: Maybe it should be placed in `dependencies `

ndelangen:
@gaetanmaisse how's this going? Do you need assistance?

This would be an immense step forward for our typescript migration epic.

gaetanmaisse:
@ndelangen I migrate codebase to TS and the only thing still in progress before I can put this PR in review is fixing snapshot testing: as TS compiler generates js sources that are not the same as Babel ones it makes tests failed  (some functions are anonymous in TS build). 

For instance:

Babel generated source for Scrollbar:
```js
var ScrollArea = function ScrollArea(_ref4) {
  var children = _ref4.children,
      vertical = _ref4.vertical,
      horizontal = _ref4.horizontal,
      props = _objectWithoutProperties(_ref4, ["children", "vertical", "horizontal"]);

  return _react.default.createElement(_react.Fragment, null, _ref5, _react.default.createElement(Scroll, _extends({
    vertical: vertical,
    horizontal: horizontal
  }, props), children));
};

exports.ScrollArea = ScrollArea;
```

TS generated source for Scrollbar:
```js
exports.ScrollArea = function (_a) {
    var children = _a.children, vertical = _a.vertical, horizontal = _a.horizontal, props = __rest(_a, ["children", "vertical", "horizontal"]);
    return (react_1.default.createElement(react_1.Fragment, null,
        react_1.default.createElement(theming_1.Global, { styles: ScrollAreaStyles_1.getScrollAreaStyles }),
        react_1.default.createElement(Scroll, __assign({ vertical: vertical, horizontal: horizontal }, props), children)));
};
```

With Babel compiler `ScrollArea` tag name will be used in snapshots but with TS it will be a default `Component` tag.

Set `"module":"es6"` in tsconfig fixes this problem but causes build error as externals libs can not be imported anymore... 🤷‍♂️ still investigating on this

lib/components/src/brand/StorybookIcon.tsx
```diff
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 
-export const StorybookIcon = ({ ...props }) => (
-  <svg viewBox="0 0 64 64" {...props} title="Storybook icon">
+export const StorybookIcon: FunctionComponent<{}> = ({ ...props }) => (
+  <svg viewBox="0 0 64 64" {...props}>
```
gaetanmaisse: `title` is not a valid attribute of `svg` tag but an inner tag

lib/components/src/ActionBar/ActionBar.tsx
```diff
@@ -1,5 +1,4 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { FunctionComponent } from 'react';
```
benoitdion: nit: my personal preference is to use the `FC` short-hand since that makes the lines shorter

lib/components/src/ScrollArea/ScrollAreaStyles.tsx
```diff
@@ -185,4 +187,4 @@ export const getScrollAreaStyles = theme => ({
     overflowY: 'hidden',
     overflowX: 'scroll',
   },
-});
+} as any);
```
kroeder: Why where you in need of casting `any` here?

lib/components/src/form/input/input.tsx
```diff
@@ -92,27 +92,35 @@ const validation = ({ valid, theme }) => {
...
-export const Input = styled.input(styles, sizes, alignment, validation, { minHeight: 32 });
-Input.styles = { ...styleResets, ...styles };
-Input.sizes = sizes;
-Input.alignment = alignment;
+export const Input = styled.input<any>(styles as any, sizes, alignment as any, validation, {
+  minHeight: 32,
+});
+(Input as any).styles = { ...styleResets, ...styles };
+(Input as any).sizes = sizes;
+(Input as any).alignment = alignment;
```
kroeder: Is there anything we can do about all these `any` for `styled`?

lib/components/src/tabs/tabs.tsx
```diff
@@ -122,7 +145,7 @@ export const Tabs = React.memo(
                 type="button"
                 key={id}
                 active={active}
-                onClick={e => e.preventDefault() || actions.onSelect(id)}
+                onClick={(e: any) => e.preventDefault() || actions.onSelect(id)}
```
kroeder: Isn't this just `e:Event` if you don't want to be more specific?

lib/components/src/tooltip/Tooltip.tsx
```diff
@@ -99,7 +110,16 @@ const Wrapper = styled.div(
...
 );
 
-export default function Tooltip({
+export interface TooltipProps {
+  arrowRef?: any;
+  tooltipRef?: any;
+  hasChrome?: boolean;
+  arrowProps?: any;
+  placement?: string;
+  color?: keyof Color;
```
kroeder: Can you explain what's going on here? Never saw an interface using a `keyof Color` type

lib/components/src/tooltip/WithTooltip.tsx
```diff
@@ -129,26 +128,26 @@ const WithTooltip = lifecycle({
...
       unbinders.forEach(unbind => {
         unbind();
       });
     };
   },
   componentWillUnmount() {
-    if (this.unbind) {
-      this.unbind();
+    if ((this as any).unbind) {
+      (this as any).unbind();
```
kroeder: Why do you needed to cast `this as any` ?

kroeder (approved):
Good job! 🚀

lib/components/src/ActionBar/ActionBar.stories.tsx
```diff
@@ -11,7 +11,7 @@ export default {
   Component: ActionBar,
   title: 'Basics|ActionBar',
   decorators: [
-    storyFn => (
+    (storyFn: () => JSX.Element) => (
```
leoyli: `JSX` comes from the global, which is subject to change and may not be future proof...  Here is [the reason with some tips](https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680#58e3) that I found super useful.

lib/components/src/ActionBar/ActionBar.tsx
leoyli: But with IDE this should not be too problematic I believe. I personally think shorthand is evil since it is not really reader friendly.

lib/components/src/tooltip/WithTooltip.tsx
```diff
@@ -1,26 +1,38 @@
...
+const TargetSvgContainer = styled.g<{ mode: string }>`
   cursor: ${props => (props.mode === 'hover' ? 'default' : 'pointer')};
 `;
 
+interface WithTooltipPureProps {
+  svg?: boolean;
+  trigger?: 'none' | 'hover' | 'click' | 'right-click';
+  closeOnClick?: boolean;
+  placement?: Placement;
+  modifiers?: any;
```
leoyli: maybe use `unknown` instead of `any`?!  Here is a [great explanation around `any`, `never`, and `unknown`](https://blog.logrocket.com/when-to-use-never-and-unknown-in-typescript-5e4d6c5799ad) and when to use them. In general, I only found typing function with arbitrary numbers of arguments will use `any`.

lib/components/src/tooltip/Tooltip.tsx
leoyli: `keyof` is a method(? or lookup type?) to get all member names in a type.  This gives a string subset typed with the names of all property in the `Color` type.  I.e., `keyof Color === 'primary' | 'secondary'| ...`.

leoyli:
Don't have time to dig in too detail but I think it is a great work for sure!  I've attached couple comments and referenced some great articles. Thank you very much!

lib/components/src/ActionBar/ActionBar.tsx
leoyli: This comment reminds me that typing in React is really not an easy job... Maybe not so interesting, but I think some component could be more restrict to its children with this generic type recipe: `type FCNoChildren<P> = FunctionComponent<{ children?: never } & P>;` for those component should not ever have a children (i.e. always inlined).

gaetanmaisse: I agree with @leoyli on:
> shorthand is evil since it is not really reader friendly

Especially for react newbie like me.

benoitdion: I almost always prefer longer names but this is one of the few exceptions since `FC`s are all over the place. It helps avoid line wraps when using descriptive component and prop names. I'd use it with `React.FC` though, importing just `FC` does feel like a little too much.

With all that said, if the convention so far has been to import `FunctionComponent` it makes sense to continue doing that.

lib/components/src/tabs/tabs.tsx
gaetanmaisse: Hmm yep I missed that but know that I look at this line I don't really understand it, what is
`e => e.preventDefault() || actions.onSelect(id)` supposed to do? because `e => e.preventDefault()` is always defined 🤔 

 cc @ndelangen 😇

lib/components/src/ActionBar/ActionBar.stories.tsx
gaetanmaisse: I didn't know about that, good article!

lib/components/src/tooltip/WithTooltip.tsx
gaetanmaisse: Fixed! Just as for `placement` I have used types from `popper.js` (the underlying library used for tooltips) so no more `any` here. I sometimes use `unknown` but it's not a reflex yet, soon ;)

lib/components/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "types": [
```
gaetanmaisse: @kroeder  I had to use `types` to includes types of some library as it's done like this in "parent" [`tsconfig.json`](https://github.com/storybooks/storybook/blob/next/tsconfig.json), is there a reason to do it instead of letting TS compiler use all types available in `/node_modules/@types/` based on _@types, typeRoots and types_ of https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

lib/components/src/form/input/input.tsx
gaetanmaisse: I tried to but have no solution for now... 😞

lib/components/src/tooltip/WithTooltip.tsx
gaetanmaisse:
```
Error:(139, 16) TS2339: Property 'unbind' does not exist on type '_ReactLifeCycleFunctionsThisArguments<WithTooltipPureProps, {}>'.
```
I think `WithTooltip` component need to be refactored in order to make these `bind` and `unbind` functions class properties or something like that. I keep it in mind for another PR.

lib/components/package.json
```diff
@@ -42,6 +44,17 @@
...
   },
+  "devDependencies": {
+    "@storybook/addon-actions": "5.1.0-alpha.20",
+    "@storybook/react": "5.1.0-alpha.20",
+    "@types/js-beautify": "^1.8.0",
+    "@types/react-syntax-highlighter": "^10.1.0",
+    "@types/react-textarea-autosize": "^4.3.3",
+    "@types/recompose": "^0.30.5",
+    "enzyme": "^3.9.0",
+    "jest": "^23.6.0",
```
ndelangen: must be updated to 24.7.1

lib/components/src/ActionBar/ActionBar.tsx
ndelangen: Let's keep the import as `FunctionComponent`. If desired, we can create the shorthand locally in the module.

This will allow newcomers to better understand what's going on, and find the proper documentation more easily.

lib/components/src/bar/separator.tsx
```diff
@@ -1,13 +1,12 @@
 import React, { Fragment } from 'react';
 import { styled } from '@storybook/theming';
 
-export const interleaveSeparators = list =>
+export const interleaveSeparators = (list: any[]) =>
```
ndelangen: any[] would be ReactElement[]

lib/components/src/placeholder/placeholder.tsx
```diff
@@ -15,7 +14,11 @@ const Message = styled.div`
   font-size: ${props => props.theme.typography.size.s2 - 1}px;
 `;
 
-export const Placeholder = ({ children, ...props }) => {
+// Need to define Placeholder using old fashioned `function Placeholder` to avoid TS compiler to
```
ndelangen: Is this still valid after the change to use babel for compilation?

lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
```diff
@@ -144,17 +172,7 @@ export default class CopyableCode extends Component {
...
-  children: PropTypes.node.isRequired,
-  copyable: PropTypes.bool,
-  bordered: PropTypes.bool,
-  padded: PropTypes.bool,
-  format: PropTypes.bool,
-  className: PropTypes.string,
-};
-
-CopyableCode.defaultProps = {
+(CopyableCode as any).defaultProps = {
```
ndelangen: add as static property?

lib/components/src/tabs/tabs.tsx
ndelangen: `e.preventDefault()` will return `undefined` always.

so `e.preventDefault() || actions.onSelect(id)` will call both.

it's the same as

```
(e: any) => {
  e.preventDefault();
  actions.onSelect(id);
}
```

lib/components/src/tabs/tabs.tsx
```diff
@@ -143,18 +166,7 @@ export const Tabs = React.memo(
...
-  tools: PropTypes.node,
-  selected: PropTypes.string,
-  actions: PropTypes.shape({
-    onSelect: PropTypes.func.isRequired,
-  }).isRequired,
-  absolute: PropTypes.bool,
-  bordered: PropTypes.bool,
-};
-Tabs.defaultProps = {
+(Tabs as any).defaultProps = {
```
ndelangen: can be set as a static property?

lib/components/src/tooltip/WithTooltip.tsx
ndelangen: 👍

lib/components/src/tabs/tabs.tsx
gaetanmaisse: Hmm I did not find an easy way to do it as `Tabs` is created using `React.memo`

lib/components/src/placeholder/placeholder.tsx
gaetanmaisse: I will revert my commit to try ;)

gaetanmaisse:
@ndelangen  @shilman I think we will be good to go with this one after a final check on CI results (still running when I'm writing this). Will this be part of 5.1.X or should we wait before merging?

Then I will continue with "Part II" :
 - 🛠 Refactor some imports/exports
 - 📦 Use these typings in addons
 - 🐎 Check if build can be faster

lib/components/src/placeholder/placeholder.tsx
gaetanmaisse: 🎉

lib/components/src/ActionBar/ActionBar.stories.tsx
```diff
@@ -28,7 +28,7 @@ export default {
   ],
 };
 
-export const singleItem = () => <ActionBar actionItems={[{ title: 'Clear', onClick: action1 }]} />;
```
mohsinulhaq: this space removal doesn't seem intentional. or is it?

gaetanmaisse: It wasn't, will be back in https://github.com/storybooks/storybook/pull/6500 😉

PR #6098, opened 2019-03-14
feat(storiesNav): deep theming for stories nav panel

mbellary-chwy:
Feature:

deep theming for stories nav panel

## What I did

- deep theming for stories nav panel
- updating README file with all the deep theme variables supported
- updating `LogoLink ` to support html elements

This is similar to the deep theming PR of SB 4.0:

https://github.com/storybooks/storybook/pull/4702

### DIFF
--- a/docs/src/pages/configurations/theming/index.md
+++ b/docs/src/pages/configurations/theming/index.md
@@ -88,6 +88,29 @@ export default create({
   inputTextColor: 'black',
   inputBorderRadius: 4,
 
+  // Deep theming stories nav
+  storiesNav: {
+    backgroundColor: '#128CED'
+  },
+  navHeading: {
+    alignItems: 'center'
+  },
+  treeSubHeading: {
+    color: '#F8D71C'
+  },
+  treeItemHeader: {
+    fontWeight: 500
+  },
+  activeMenuLink: {
+    fontWeight: 600
+  },
+  treeExpander: {
+    borderLeft: '3.5px solid white'
+  },
+  treeIcons: {
+    color: 'white'
+  },
+
   brandTitle: 'My custom storybook',
   brandUrl: 'https://example.com',
   brandImage: 'https://placehold.it/350x150',
--- a/lib/theming/src/base.ts
+++ b/lib/theming/src/base.ts
@@ -126,6 +126,15 @@ export interface ThemeVars {
   inputTextColor?: string;
   inputBorderRadius?: number;
 
+  // Deep theming stories nav
+  storiesNav?: object;
+  navHeading?: object;
+  treeSubHeading?: object;
+  treeItemHeader?: object;
+  activeMenuLink?: object;
+  treeExpander?: object;
+  treeIcons?: object;
+
   brandTitle?: string;
   brandUrl?: string;
   brandImage?: string;
--- a/lib/theming/src/themes/dark.ts
+++ b/lib/theming/src/themes/dark.ts
@@ -31,6 +31,15 @@ const theme: ThemeVars = {
   inputBorder: 'rgba(0,0,0,.3)',
   inputTextColor: color.lightest,
   inputBorderRadius: 4,
+
+  // Deep theming stories nav
+  storiesNav: {},
+  navHeading: {},
+  treeSubHeading: {},
+  treeItemHeader: {},
+  activeMenuLink: {},
+  treeExpander: {},
+  treeIcons: {},
 };
 
 export default theme;
--- a/lib/theming/src/themes/light.ts
+++ b/lib/theming/src/themes/light.ts
@@ -31,6 +31,15 @@ const theme: ThemeVars = {
   inputBorder: color.border,
   inputTextColor: color.darkest,
   inputBorderRadius: 4,
+
+  // Deep theming stories nav
+  storiesNav: {},
+  navHeading: {},
+  treeSubHeading: {},
+  treeItemHeader: {},
+  activeMenuLink: {},
+  treeExpander: {},
+  treeIcons: {},
 };
 
 export default theme;
--- a/lib/ui/src/components/sidebar/Sidebar.js
+++ b/lib/ui/src/components/sidebar/Sidebar.js
@@ -7,13 +7,14 @@ import { ScrollArea } from '@storybook/components';
 import SidebarHeading from './SidebarHeading';
 import SidebarStories from './SidebarStories';
 
-const Heading = styled(SidebarHeading)({
+const Heading = styled(SidebarHeading)(props => ({
   padding: '20px 20px 12px',
-});
+  ...props.theme.navHeading,
+}));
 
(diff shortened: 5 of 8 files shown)


ndelangen (commented):
I like the idea of making all this theme-able. I'm on the fence if we should add all these variables to ThemeVars though.

WDYT @shilman, @domyen ?

shilman:
@ndelangen Yeah this seems way verbose to me. @ManBearTM what's the minimum you'd want to be able to style to achieve your desired output?

Zarabadoo (approved):

MansoorBashaBellary:
> @ndelangen Yeah this seems way verbose to me. @MansoorBashaBellary what's the minimum you'd want to be able to style to achieve your desired output? I think we should add 1-2 vars at most.

@shilman - These are the minimum things that i could think off to get the desired output. What i can do instead is to wrap all the variables in `storiesNav` like:

```
storiesNav: {
  nav: {},
  navHeading: {},
  treeSubHeading: {},
  treeItemHeader: {},
  activeMenuLink: {},
  treeExpander: {},
  treeIcons: {}
}
```

But the only thing is in `js` files, i have to do `theme.storiesNav.*` to get the desired output. Please let me know what best thing to go ahead.

ndelangen:
@MansoorBashaBellary I totally understand the request for more customization.

I was planning to enable this using addons.

I'm still unsure how much theming to allow.

On 1 hand I want users to be able to create the storybook of their dreams, but on the other hand, we don't want to have hundreds of theme variables, that we'll have to keep sane and compatible.

domyen:
@ndelangen I think this adds more complexity and maintenance surface area for the UI code so I'd prefer to leave it mostly the same. 

Storybook helps you build UIs, theming is an aesthetic choice that doesn't really contribute or detract from that value prop. I think we should keep the theming API as minimal as possible, that way maintainers can focus on building new functionality instead of maintaining support for theming edge cases into the future. 

(Background: I worked on theming support with Norbert)

MansoorBashaBellary:
> @MansoorBashaBellary I totally understand the request for more customization.
> 
> I was planning to enable this using addons.
> 
> I'm still unsure how much theming to allow.
> 
> On 1 hand I want users to be able to create the storybook of their dreams, but on the other hand, we don't want to have hundreds of theme variables, that we'll have to keep sane and compatible.

Thanks @ndelangen and @domyen. We are maintaining our storybooks with stories nav panel themed from Storybooks 4.0. In order to use SB 5.0, we needed these changes.

I don't think these changes are going to make storybooks complex. But i totally agree that theme variables are to be minimal. Let me know if we can use any other approach that i am unaware of, to get this thing done.

adamlohner:
> Storybook helps you build UIs, theming is an aesthetic choice that doesn't really contribute or detract from that value prop. I think we should keep the theming API as minimal as possible, that way maintainers can focus on building new functionality instead of maintaining support for theming edge cases into the future.

If that's the long term stance on this topic (_which is completely acceptable IMO_), can additional attrs/classes/id's be added to specific elements so that we can still override them by injecting styles using the `manager-head.html` workaround? I think if the following elements outlined in red had some sort of consistent/obvious attribute users could leverage to style, it would be sufficient for _most_ theming use cases.

nav-theming

domyen:
@adamlohner good point. Perhaps adding `classNames` so folks could style via CSS is a reasonable middle ground here?

jonmilner:
Not specifically related to the Storybook nav (I'm thread-jacking a bit here), but this seems relevant to theming within Storybook in general...

This has been one of my biggest pain points with open-source tools that use CSS-in-JS -- I certainly understand the appeal in terms of maintenance, but as an end-user, not having class names available in the DOM to make custom style overrides easy (or in some cases possible at all) can be incredibly frustrating. Even if it's not technically _supported_, simply making it _possible_ to override default styling would go a long way -- and I'd happily accept the risk of minor upgrades breaking my custom styles.

> Storybook helps you build UIs, theming is an aesthetic choice that doesn't really contribute or detract from that value prop. I think we should keep the theming API as minimal as possible, that way maintainers can focus on building new functionality instead of maintaining support for theming edge cases into the future.

Agree with @adamlohner -- while not ideal, it's certainly understandable. Storybook is an incredible tool as-is (the updates introduced in v5 are awesome). And taking theme customization too far could absolutely detract from the amazing work the Storybook team is doing.

I do, however, feel like it's not always simply about meaningless aesthetics, but rather the UX -- which _is_ consistent with that value prop if an improved UX helps people build UIs. 

Take the following example:

I want to help my team build UIs. Having the props documented in an intuitive, easy-to-read way will help them do that. 

Unfortunately, it's hidden behind this little blue button:

Screen Shot 2019-03-18 at 11 46 18 PM

I see there's an option to display it all on one screen, but it looks a bit clunky...

Screen Shot 2019-03-18 at 11 55 00 PM

What I really want is for it to show up in the addons panel. I see that's not actually supported (#1147) but there's a [workaround](https://www.npmjs.com/package/storybook-addon-react-docgen).

Screen Shot 2019-03-18 at 11 04 33 PM

This is a lot closer to what I'd actually consider a good experience, but it could use some tweaking. `addon-info` seems to allow _some_ [custom styling](https://github.com/storybooks/storybook/blob/master/addons/info/src/components/Story.js#L19):

Screen Shot 2019-03-18 at 11 05 56 PM

But there's not much else I can do here (that I'm aware of, at least). And it's still a far cry from where I want it to be:

Screen Shot 2019-03-18 at 11 17 19 PM

(^ These styles were achieved via editing styles in Chrome's dev tools. It doesn't seem this level of customization is actually possible.)

This is obviously just one example, but there have been more than a couple of areas that I've fought with only to realize overriding styles that would otherwise be very simple to do is either going to be way too much work, or is just not possible. 

I think adding class names to the Storybook UI could be *huge* -- even just a few sprinkled throughout so that I could write selectors like `#section .area h2 { ... }` rather than something like `#section > div:nth-child(2n) > div > div > div > h2 { ... }`.

adamlohner:
@jonmilner you could always write a custom addon to achieve the desired functionality rather than working around the existing notes implementation.  I actually just did this and found the Addon API to be super flexible and easy to use.

Screen Shot 2019-03-19 at 5 47 17 PM

domyen:
Thanks everyone, I definitely appreciate the good thinking behind this PR. I'm closing this because there isn't clear consensus on theme API expansion and the impact on future maintainability. 

@jonmilner That use case probably isn't in the scope of SB theming. Rather, it's the visual implementation and functionality of addons. I reckon you'll be pleased with the additional documentation features in Storybook Docs [coming soon](https://docs.google.com/document/d/1D4Ov4UvrB80Uy1CLGshbFEWQKWEpr8dL4FmOqRDwjwI/edit). Specifically the new `docs` addon that consolidates info/readme/notes addons into one. 

If folks still want extra visual customization, I'd propose an alternate PR that assigns static classNames to DOM elements for styling.

PR #6106, opened 2019-03-15
UI: open brandUrl in new window

toshi1127:
Issue: #6070

## What I did
Opened brandUrl in new window when the brand image was clicked
## How to test

- Is this testable with Jest or Chromatic screenshots? Yes. I updated the snapshot.
- Does this need a new example in the kitchen sink apps?  No.
- Does this need an update to the documentation?  No.

### DIFF
--- a/examples/official-storybook/tests/__snapshots__/storyshots.test.js.snap
+++ b/examples/official-storybook/tests/__snapshots__/storyshots.test.js.snap
@@ -8388,7 +8388,7 @@ exports[`Storyshots UI|Sidebar/Sidebar loading 1`] = `
                 <a
                   class="emotion-1"
                   href="./"
-                  target="_blank"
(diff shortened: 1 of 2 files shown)


ndelangen (changes requested):
I like it! one feature-request and then this will get merged:

Please only add the target when the url is an external link.

toshi1127:
@ndelangen 
only added the target when the url is an external link.
Please confirm!!

ndelangen:
If target is an empty string, `target=""` I guess nothing special happens.

Not having the prop at all seems a bit cleaner to me, but this is good to merge IMHO!

Good addition @toshi1127 👍

ndelangen (approved):

aganglada:
This can be closed right?

PR #6118, opened 2019-03-15
Addon-a11y: added 'name' param for generate unique key for item

Armanio:
Issue: #6107

## What I did
I added `name` prop `Report` component for pass to all child. React (and angular? and vue? and etc) in vdom  elements list by `key`. Sometimes (read #6107)  in any tabs has components with same key. It confuse react and he reuse state from other `report` component.

Im not sure that someone can understand my english, but im trying. :(
## How to test

yes
no
no


shilman (approved):
Good 👍

PR #6124, opened 2019-03-16
Add missing button titles

aamct2:
Issue: #6123

## What I did

- Added titles to some of the image buttons that were missing them. This helps to meet WCAG 2.1 (Level A) for accessibility.
  - WCAG 2.1 [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
  - Deque University's guidance on [buttons having discernible text](https://dequeuniversity.com/rules/axe/3.1/button-name)

## How to test

Run any of the examples and inspect the adjusted buttons to see that they now have a `title` attribute specified. These issues can also be picked up by [Lighthouse](https://developers.google.com/web/tools/lighthouse/) in the Accessibility audit in the Chrome Developer Tools.

### DIFF
--- a/examples/official-storybook/tests/__snapshots__/storyshots.test.js.snap
+++ b/examples/official-storybook/tests/__snapshots__/storyshots.test.js.snap
@@ -8446,6 +8446,7 @@ exports[`Storyshots UI|Sidebar/Sidebar loading 1`] = `
               >
                 <button
                   class="emotion-5"
+                  title="Shortcuts"
                 >
                   <svg
                     class="emotion-4"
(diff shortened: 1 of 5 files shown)


Armanio:
Good job @aamct2
One notice: when you update UI you must regenerate snapshots for test passes.
More information you can read [here](https://github.com/storybooks/storybook/blob/next/CONTRIBUTING.md#2a-run-unit-tests).
When the snapshots will be updated i will merge PR.
Could you update the snapshots?

Armanio:
Thanks!

PR #6137, opened 2019-03-17
Only check versions for dev/local builds

maraisr:
Might be a tad controversial, but with this PR I aim to only enable the version checking things on local/dev builds. IMO production, deployed versions shouldn't really echo their outdated.

_Or_ am I completely wrong, and the intent here is to tell them to upgrade wherever they are?

### DIFF
--- a/lib/ui/src/core/context.js
+++ b/lib/ui/src/core/context.js
@@ -71,8 +71,10 @@ export class Provider extends Component {
       initShortcuts,
       initStories,
       initURL,
-      initVersions,
-    ].map(initModule => initModule(apiData));
+      process.env.NODE_ENV !== 'production' && initVersions,
+    ]
+      .filter(Boolean)
+      .map(initModule => initModule(apiData));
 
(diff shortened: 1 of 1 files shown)


ndelangen:
Hey @maraisr thanks for this PR.

I think completely removing all functionality for the version module isn't too great.

The version notification is annoying, but the about page should still show the current and next version IMHO.

Also the module and it's api's could one day be used by addons, so I think not initializing it could lead to bugs later. Which would probably be a pain to trace, since it would only manifest in the production version.

I discussed with @shilman and we came up with a version that just hides the notification.

Do you agree this is better solution?

Again, thanks for your input and time 🙇

maraisr:
@ndelangen - oh yeah completely agree! Still learning my way around the project to have known about the other use cases 😅.

Will update this PR to only remove the _toast_ in a PRO build.

maraisr:
Sorry; AFAIK the latest `next` branch contains a `mode !== 'production'` that wraps that update notification, rendering this PR defunct. Someone else beat me to it 😎 	

Thanks again for your time 💯

PR #6154, opened 2019-03-18
📜[storysource]  📦 [sandpack] The clientApi must be reused when possible

libetl:
Issue:

## What I did
I am working on a topic : ability to live edit the story when on the screen, and ability to export the component as a project in codesandbox.
This will be done by upgrading the storysource addon, which could become an Integrated Development Environment

Sandpack will re-download storybook. 
Then we must have an idempotent behavior, which means keeping control over code run several times.
In this case, we must prevent the clientApi from being created twice, instead we must reuse the currently existing one.

## How to test

I added one test to verify that it works.

I don't think so.

I don't think so.

### DIFF
--- a/lib/core/src/client/preview/start.js
+++ b/lib/core/src/client/preview/start.js
@@ -77,9 +77,16 @@ export const getContext = (() => {
         addons.setChannel(channel);
       }
     }
-
-    const storyStore = new StoryStore({ channel });
-    const clientApi = new ClientApi({ storyStore, decorateStory });
+    let storyStore;
+    let clientApi;
+    if (typeof window !== 'undefined' && window.__STORYBOOK_CLIENT_API__) {
+      clientApi = window.__STORYBOOK_CLIENT_API__;
+      // eslint-disable-next-line no-underscore-dangle
+      storyStore = clientApi._storyStore;
+    } else {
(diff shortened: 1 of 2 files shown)


libetl:
Hi @shilman @tmeasday and @ndelangen.

I think I speaked about it with you, but don't hesitate to discuss about the intent if you don't agree.
Any better answer is welcome.

lib/core/src/client/preview/start.js
```diff
@@ -77,9 +77,16 @@ export const getContext = (() => {
         addons.setChannel(channel);
       }
     }
-
-    const storyStore = new StoryStore({ channel });
-    const clientApi = new ClientApi({ storyStore, decorateStory });
+    let storyStore;
+    let clientApi;
```
libetl: any way to make that const again ?
I know that Kotlin can do it. (free ad)

tmeasday (approved):
LGTM

libetl:
@shilman, @ndelangen ?
image

lib/core/src/client/preview/start.js
ndelangen: I'd wrap it into a function, that way the result of the function can be a const

libetl:
thanks

PR #6181, opened 2019-03-19
fix(MIGRATION): add missing commas

samouss:
Issue: The migration guide print non valid JavaScript for the full control mode.

## What I did

I've added the two missing commas.


shilman (approved):
🙇 good catch!

PR #6182, opened 2019-03-19
fix: Add the raw method to the react-native implementation

rmevans9:
Issue: The new client for `react-native` did not export the `raw` method from `clientApi` as the storyshots addon was expecting

## What I did

Added the `raw` method as an export from the `react-native` client

## How to test

- Run the storyshots addon with the latest alpha version of storybook on a `react-native` project


shilman (approved):
👍 thanks @rmevans9!!!

PR #6186, opened 2019-03-19
Add fonts and color to styling in info addon

SteffenGorenflo:
Issue: https://github.com/storybooks/storybook/issues/5723
## What I did
Add font family and color to the styling of the info addon. 

https://github.com/storybooks/storybook/commit/c7f624e54afa50622b4b45313ede89f9571f65b2#diff-36f288162c3cc9bd518d680f366bbe54 removed the font with the color and https://github.com/storybooks/storybook/pull/5759 added the font again, but not with the color and at the (imho) wrong place. 

I think with my solution the info addon has the correct fonts and color when rendering "behind" the story as well as inline.

## How to test
When rendering the Story `Button with new info` in the cra-kitchen-sink example, add `info: { inline: true }` to the options. The rendered font of the addon-info should look the same as if the addon was rendered with `info: { inline: false}` 

Yes
Maybe we can render a story with the inline options to true
No


ndelangen (approved):

ndelangen:
@SteffenGorenflo 👍

PR #6190, opened 2019-03-20
Allow forced renders in @storybook/html  

dwightjack:
Issue: #5017

## What I did

This PR adds the ability to setup an `html` object parameter to control whether force re-renders are allowed or not like this:

```js
// .storybook/config.js
import { configure, addParameters } from '@storybook/html';

addParameters({
  html: {
    preventForcedRender: false, // or true
  }
});

// ...
```

This addresses a change introduced in #4822. I understand the rationale around it, but right now it is just preventing @storybook/html to re-render stories returning a DOM node, which looks much more like a bug than a feature on  _normal_ use cases.

## How to test

I added a test story in `examples/html-kitchen-sink/stories/addon-knobs.stories.js` (`Addons > Knobs > DOM`) and the relevant configuration to `examples/html-kitchen-sink/.storybook/config.js`

### DIFF
--- a/app/html/src/client/preview/render.js
+++ b/app/html/src/client/preview/render.js
@@ -4,6 +4,7 @@ import { stripIndents } from 'common-tags';
 const rootElement = document.getElementById('root');
 
 export default function renderMain({
+  parameters = {},
   storyFn,
   selectedKind,
   selectedStory,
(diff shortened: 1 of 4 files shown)


ndelangen:
@Atekon I think this is good to merge, but after releasing this you'll likely have to update your stories.

Technically a breaking change. WDYT, about releasing this in 5.1 though?

ndelangen (approved):

PR #6197, opened 2019-03-20
[bug] Fix "lib/addons" default export for react-native.

zewish:
Issue:
Broken commonjs import on react-native (metro bundler)

## What I did
Changed require('.') to require('./index')

## How to test
Run @storybook/react-native with the @storybook/addons.

No. I can make a screenshot of the react-native error we get if required.

No - it's an internal API to public API export.

No - it's an internal change.

### DIFF
--- a/lib/addons/src/public_api.ts
+++ b/lib/addons/src/public_api.ts
@@ -1,5 +1,5 @@
 export * from './make-decorator';
-export * from '.';
+export * from './index';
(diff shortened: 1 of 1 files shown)


ndelangen (approved):

ndelangen:
@shilman just pinging you to show you this 'convenience' is more trouble than it's worth.

PR #6209, opened 2019-03-21
fix(client-api): add core-js as a dep

eddiemonge:
client-api requires core-js in the dist files but it is not listed as a dependency. This causes it to use whatever package npm/yarn/whatev thinks it should use, in my case, ver3 since that's the top most version.

See https://github.com/storybooks/storybook/issues/6204#issuecomment-475073128 for more info on the problem


ndelangen (approved):

ndelangen:
Thank you @eddiemonge

PR #6215, opened 2019-03-21
Angular: Disable prod mode for debugging components

konstantindenerz:
Issue:

## What I did
- I disabled prod mode to enable dev tools.
Now its possible to debug components with ng.probe:
image

## How to test

  - No
  - No
  - No


shilman (changes requested):
I don't know Angular but this seems like a breaking change. Shouldn't this be conditional on `NODE_ENV` or something like that?

konstantindenerz:
Now an optional option in the parameters definition is required to enable dev mode.

*prod* mode:
https://github.com/storybooks/storybook/blob/next/examples/angular-cli/.storybook/config.ts#L6-L10
```
addParameters({
  options: {
    hierarchyRootSeparator: /\|/,
  },
});
```

*dev* mode:
https://github.com/storybooks/storybook/blob/next/examples/angular-cli/.storybook/config.ts#L6-L10
```diff
addParameters({
  options: {
    hierarchyRootSeparator: /\|/,
+   ngDevMode: true.
  },
});
```

ndelangen:
@konstantindenerz I'd like to merge this, though there's 1 thing kinda troubling me, which is that it adds a property to the storybook options object.

If this would use some other property on the parameters object instead I'd merge this easily.

Would you agree that moving the config to switch this behavior from `parameters.option.ngDevMode` to possibly `parameters.ng.mode = 'dev`?

ndelangen:
Sorry for the late reply, I had one typed earlier but got distracted ( I blame @shilman ).

shilman (changes requested):
Can we somehow use NODE_ENV and just do away with the new option entirely?

konstantindenerz:
@shilman @ndelangen I found the interpolation of header template and pushed the new implementation. Did you imagine it that way?

shilman (approved):
LGTM! Does it work? 😄

konstantindenerz:
Looks like it will work:
image

I tested it with storybook/examples/angular-cli project.

And the static version (production build):
image

app/angular/src/client/preview/angular/helpers.ts
```diff
@@ -75,7 +77,9 @@ const draw = (newModule: DynamicComponentType): void => {
   if (!platform) {
     insertDynamicRoot();
     try {
-      enableProdMode();
+      if (NODE_ENV !== 'development') {
```
ndelangen: Won't this possibly throw a runtime error if NODE_ENV isn't set?

app/angular/src/client/preview/angular/helpers.ts
```diff
@@ -17,6 +17,8 @@ type DynamicComponentType = typeof componentClass;
 
 const storyData = new ReplaySubject(1);
 
+declare var NODE_ENV: string | 'development';
```
shilman: Is this supposed to set the default value to `'development'` if NODE_ENV isn't defined? I don't think it does

app/angular/src/client/preview/angular/helpers.ts
shilman: I think it will

konstantindenerz: If not defined, should it be handled as a production or development mode?

app/angular/src/client/preview/angular/helpers.ts
konstantindenerz: This specifies the possible value.

app/angular/src/client/preview/angular/helpers.ts
konstantindenerz: New suggestion: https://github.com/storybooks/storybook/pull/6215/commits/5d76d4738069ffa56bcba9d4ec534b5697dc4a46#diff-65b1b66b4dca31f62fc0028943d19079R84

mtuzinskiy:
Hi! Is there a possibility to workaround this right now, before the new version comes out? Thanks

mtuzinskiy:
We really lack the possibility to use the `ng.probe` with Storybook... When can we expect this fix to be merged in? @shilman  Thank you!

app/angular/src/client/preview/angular/helpers.ts
```diff
@@ -7,6 +7,12 @@ import { STORY } from './app.token';
 import { NgModuleMetadata, IStoryFn, NgStory } from './types';
 import { ReplaySubject } from 'rxjs';
 
+declare global {
```
shilman: @ndelangen @kroeder @gaetanmaisse @leoyli is this correct? want to merge & release this but concerned about this bit

shilman: Isn't 'string' the value "string" and not a general string type?

gaetanmaisse: I think you're right @shilman it should be `NODE_ENV: string | 'development' | undefined;`

gaetanmaisse: And this `declare` block can be moved to a `typings.d.ts` as soon as https://github.com/storybooks/storybook/pull/6570 will be merged to avoid to have it in the middle of code file.

gaetanmaisse (approved):

ndelangen:
I'll pick this up

ndelangen:
@konstantindenerz I made a tweak, could you cast your review on that?

konstantindenerz:
@ndelangen Thanks for the commit. I'm not sure I'm doing something wrong while building, but I don't get NODE_ENV on window object. `typeof NODE_ENV === 'string'` results in false. Is it the expected behavior that if NODE_ENV is undefined the debugging tools should be enabled?
Steps to reproduce:
* `yarn bootstrap --core`
* `cd app/angular && yarn link`
* `cd examples/angular-cli && yarn link @storybook/angular`
* `npm run storybook`

ndelangen:
@konstantindenerz I'll take a look

ndelangen:
@konstantindenerz 
```
cd app/angular && yarn link
```

The yarn link commands are not necessary, and might actually cause issues, I'm unsure. FYI

ndelangen:
Tested, this works flawlessly:

code:
Screenshot 2019-05-06 at 20 12 07

invocation from terminal:
Screenshot 2019-05-06 at 20 11 51

When the NODE_ENV is anything other then `production`:
Screenshot 2019-05-06 at 20 12 00

When is is `production` the entire codeblock removed from the bundle.

```sh
yarn bootstrap --core
cd examples/angular-cli
NODE_ENV=production yarn storybook  # for no debugging
NODE_ENV=development yarn storybook # for debugging on
```

ndelangen (approved):

ndelangen:
> Is it the expected behavior that if NODE_ENV is undefined the debugging tools should be enabled?

If NODE_ENV does not exist of is not `development` we enable prod mode, sounds reasonable to me?

konstantindenerz:
@ndelangen 
image
In the current implementation the dev mode is enabled if NODE_ENV does not exists. jfyi

Issue #6001, opened 2019-03-10
[Addon: A11y] - Feature - Highlight the violation on the rendered component


CodeByAlex:
**Is your feature request related to a problem? Please describe.**
There isn't a way for a non-technical user to determine where the accessibility issue is located on the component. I think it would be useful for there to be a button/checkbox/link that, when clicked, would highlight where the specific issue is located on the component.

**Describe the solution you'd like**
I would like there to be a way for a user to go to the a11y tab, look at a violation and be able to pinpoint where that issue is on the component using some form of highlighting.
 ![image](https://user-images.githubusercontent.com/29763590/54087527-10bcb200-432a-11e9-8d5c-5bc6dc4bc6c6.png)

**Implementation idea**
Run querySelector from the root of the iframe using the item -> nodes -> target (which provides a path to the element) once the elment is found, it can be highlighted. This code will be triggered on some type of action depending upon design

**Are you able to assist bring the feature to reality?**
Yes

**Additional context**
@domyen would you want to collaborate on this so that we can maintain a consistent design

shilman:
That's a cool idea! I wonder whether we can do this without modifying the component's DOM? @ndelangen maybe a new addon API?

CodeByAlex:
@shilman I like the idea of keeping the changes out of the component’s DOM if possible. Any idea  how that could be done whether it’s a new api or not?

shilman:
AFAIK it's not possible in the current setup. My suggestion would be prototyping the feature by modifying the DOM directly and then potentially rebuild it with a new API if we can figure out a good way to do that.

CodeByAlex:
Nice @Armanio! So are you saying that you are already working on adding this feature or are you saying that you would like this feature to be added?

Armanio:
Yep, i already started work on this feature. But dont worry, if you want you can implement it instead me.

CodeByAlex:
I started working on this last night so I'm not too far along yet. How much progress have you made?

In any case, at some point, I think we should touch base with @ndelangen to see if he has any ideas for the design. Were you planning on making the path a link, adding a button, or something else to trigger the highlighting

domyen:
I can take a look at the design 👍 Socialize the prototype when ready, also
lmk the outcome of the api discussion :)

Armanio:
@CodeByAlex i write prototype but code is too bad. 😖
I am ashamed to show it.

I recorded mini screencast.
[...]

PR #6218, opened 2019-03-21
Feature/highlight a11y violation on component

CodeByAlex:
Issue: #6001

## What I did

I added highlight toggles for each accessibility rule and category at every level. A user can now click on the highlight checkbox and see their violation, pass, or incomplete rule highlighted in the iframe. Each type (violation, pass, or incomplete) has a corresponding color to differentiate the rule when highlighting.

Design:

Highlight-Toggle component - can take in one to many elements that it can highlight. This means that the toggle component can be placed at any level and highlight the element or elements passed in based on a switch.

Redux is being used to keep track of a global map where the key is the element itself and the value is an object containing the toggle state of the element and the original element outline so that the component isn't changed when the highlight is removed.

The highlight toggle uses the redux map to keep track of whether or not the element or elements within it are highlighted and whether or not the toggle state is true.

I worked with @domyen on the design elements for this feature to keep consistency across the app.

## How to test

- Does this need an update to the documentation? Maybe to describe the feature?

a11y-highlight-feature


domyen:
Great work @CodeByAlex! QAing this. 

After playing with this in real life, I'd suggest these layout changes:
Screen Shot 2019-03-24 at 10 29 37 AM
I think if we align the checkboxes to the right and underneath the "highlight results" it becomes obvious what they're about (similar to column headers in a table). That means we can get rid of the repeating "highlight" label on each line item.

I can take a stab at it if you like. What do you think?

CodeByAlex:
Thanks @domyen! Thanks for the feedback. I'll make sure that the "Highlight results" and checkbox are vertically centered and uncapitalize "results". My only concern about aligning the checkboxes to the right would be in scenarios where people are using wide screens. It may be hard to determine which checkbox is associated with which rule (one above or below in most cases). What do you think?

domyen:
👍 We can solve this by showing the boundaries of the row:  
image

The right hand alignment makes the highlight checkbox a bit less distracting. After all the goal is to communicate the rules not that there's a checkbox. :)

And while we're here:
- The button should extend to the full length of the row
- The button should have a `cursor:pointer`
- button:hover should have bg color `rgba(0,0,0,.05)`

You're a legend @CodeByAlex 🙏 🙇

CodeByAlex:
I like that a lot @domyen , Ill make that design change

CodeByAlex:
@domyen, Ive made the changes. Let me know what you think
a11y-highlight-feature-mod

CodeByAlex:
I think one other touch-up that would make this look a little sleeker is putting the "Highlight results" text on the left side of the checkbox. This keeps a uniform line down the right side. Would you agree @domyen?

image

domyen (approved):
LGTM from a UI perspective

shilman (approved):
This looks amazing! Great work @CodeByAlex and @domyen! Esp love the right aligned checkbox on top -- details ❤️❤️❤️ 👌 !

Since the UI has gotten complex, and it's been awhile since I used redux for state management (cc @alexandrebodin @ndelangen), here's what I'd like to see on the code side:
- [ ] Stories for the major components
- [ ] Hooked into `official-storybook` or another top-level storybook to simplify testing

This might be outside the scope of this PR, but we should probably do this for all the core addons. It bothers me to see Jest tests with snapshots when we've already built a much nicer way to do this inside Storybook.

ndelangen:
@CodeByAlex can you please allow maintainers to push to your branch? I'd like to fix the merge conflicts and merge

CodeByAlex:
Sure thing @ndelangen

ndelangen:
Thank you! Let's get this merged, amazing feature!

CodeByAlex:
Thanks, @ndelangen!

addons/notes/src/Panel.tsx
```diff
@@ -56,7 +56,9 @@ export const SyntaxHighlighter = ({ className, children, ...props }: SyntaxHighl
   // className: "lang-jsx"
   const language = className.split('-');
   return (
-    <SyntaxHighlighterBase language={language[1] || 'plaintext'} bordered copyable {...props}>{children}</SyntaxHighlighterBase>
```
Armanio: Why it is here? 🤔

lib/api/src/version.ts
```diff
@@ -1 +1 @@
-export const version = '5.1.0-alpha.13';
+export const version = '5.1.0-alpha.14';
```
Armanio: And what is it? Does it need in this PR?

addons/notes/src/Panel.tsx
CodeByAlex: Not Exactly sure why and I’d have to look at it to figure it out but it looks like a format change

Armanio:
Looking great! 
I wrote a few comments. Check it before you merge it, please. 
Thank you for your effort.

lib/api/src/version.ts
CodeByAlex: @ndelangen bumped the version here during the merge

addons/notes/src/Panel.tsx
ndelangen: Just a formatting change, I guess it was a warning as didn't cause the build to fail before, but got fixed by accident/automatically

PR #6238, opened 2019-03-22
fix(tooltip): event bindings, logger import

Armanio:
Issue: #6211

Actually, the problem was in incorrect logger import because as any sensitive code wrapped in try/catch block. I fixed import and change catch block for more clear.

## What I did
Fix logger import
Catch error bindings

## How to test
Install https://github.com/wolfmanstout/vimium in FF
Open any sample
...
Profit!

no
no
no


shilman (approved):
Nice one 🤕

Issue #6004, opened 2019-03-10
Action logger not working


kaluabentes:
**Describe the bug**
The action logger is not working

**To Reproduce**
1. Create a story with action
2. Click on the component

**Expected behavior**
See action log

**Code snippets**
```javascript
import React from 'react'
import { storiesOf } from '@storybook/react'
import { action } from '@storybook/addon-actions'

import Button from '.'

storiesOf('Atoms/Button', module).add('default', () => (
  <Button onClick={action('clicked')}>Default</Button>
))
```

**System:**
 - OS: MacOS
 - Device: Macbook Pro
 - Browser: Chrome
 - Framework: React
 - Addons: addon-actions
 - Version: 5.0.1

shilman:
I'm not able to reproduce. Can you please provide an example?

mech:
I am able to produce it using stock CRA + SB 5.0.1

```
▶ npx create-react-app storybook-latest
▶ cd storybook-latest/
▶ npx -p @storybook/cli sb init
```

When viewing the "Hello Button", click in rapidly on it to see error. You have to open the console and click the action multiple times quickly. After this you can move to Button with emoji and it works and then move back to "Hello Button" and it works again.

This happen only the first time.

```js
RangeError: Maximum call stack size exceeded
    at Object.toString (<anonymous>)
    at isArguments (vendors~main.b2bcbb4f8de247f2094b.bundle.js:33913)
    at keys (vendors~main.b2bcbb4f8de247f2094b.bundle.js:33976)
    at equal (vendors~main.b2bcbb4f8de247f2094b.bundle.js:38890)
    at equal (vendors~main.b2bcbb4f8de247f2094b.bundle.js:38901)
    at equal (vendors~main.b2bcbb4f8de247f2094b.bundle.js:38901)
    at equal (vendors~main.b2bcbb4f8de247f2094b.bundle.js:38901)
    at equal (vendors~main.b2bcbb4f8de247f2094b.bundle.js:38901)
    at equal (vendors~main.b2bcbb4f8de247f2094b.bundle.js:38901)
    at equal (vendors~main.b2bcbb4f8de247f2094b.bundle.js:38901)
```

Dependencies as follow:

```js
"dependencies": {
    "react": "^16.8.4",
    "react-dom": "^16.8.4",
    "react-scripts": "2.1.8"
  },
"devDependencies": {
    "@storybook/react": "^5.0.1",
    "@storybook/addon-actions": "^5.0.1",
[...]

PR #6240, opened 2019-03-22
Addon-actions: safe render for cyclic obj

Armanio:
Issue: #6004

## What I did
Wrap deepEqual function call in try/catch.
I spend much time finding an elegant solution, but...
I analyzed comparing objects and they have cyclic react instance ('__reactInternalInstance...' has an element, an element has __reactblabla, and etc ). I don't know its bug or feature.

More information: 'fast-deep-equal' issue and author reply [here](https://github.com/epoberezkin/fast-deep-equal/issues/8)

And I rewrite 'addAction' method: if new state based on the previous state you must use a callback in 'setState' function. React [docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) for more information.

## How to test
Start cra-kitchen-app
Open button sample
Double click to button

no
no
no

### DIFF
--- a/addons/actions/src/containers/ActionLogger/index.tsx
+++ b/addons/actions/src/containers/ActionLogger/index.tsx
@@ -17,6 +17,14 @@ interface ActionLoggerState {
   actions: ActionDisplay[];
 }
 
+const safeDeepEqual = (a: any, b: any): boolean => {
+  try {
+    return deepEqual(a, b);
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM? Will wait for another opinion before merging

ndelangen (approved):

PR #6244, opened 2019-03-23
use Fuse.js package instead of fuzzy-search for customizable search threshold

mohsinulhaq:
Issue: https://github.com/storybooks/storybook/issues/5940

## What I did
Use Fuse.js for customizing fuzzy search threshold


shilman (approved):
Results are looking really good and this is a nice small change. Great work! 👏

ndelangen (approved):

PR #6245, opened 2019-03-23
fix(cra-config): use node-logger instead of console

Armanio:
[Hmmmm...](https://teamcity.jetbrains.com/viewLog.html?buildId=2084504&tab=Inspection&buildTypeId=OpenSourceProjects_Storybook_Lint_Warnings)


shilman (approved):
👍 👍 👍

PR #6252, opened 2019-03-24
Add custom grid size

aamct2:
Issue: #5853 `Make grid cell size customizeable`

## What I did

- Made grid background SVG dynamically generated based on custom cell size
- Update grid SVG to be more human readable (and editable)
- Grid cell size defined by `Theme`
- Pick up `gridCellSize` from Theme configuration options

## How to test

- Added Jest tests

### DIFF
--- a/examples/cra-kitchen-sink/.storybook/config.js
+++ b/examples/cra-kitchen-sink/.storybook/config.js
@@ -1,12 +1,11 @@
 import { configure, addParameters, addDecorator } from '@storybook/react';
+import { create } from '@storybook/theming';
 
 import { withA11y } from '@storybook/addon-a11y';
 
 addDecorator(withA11y);
 addParameters({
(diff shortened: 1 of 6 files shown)


tmeasday:
Thanks @aamct2, looks great!

I wonder if this should be a theme variable or an option. What do you think @ndelangen @domyen?

domyen:
An option - it's more of a functional change rather than a styling change to me.

ndelangen (approved):

ndelangen:
I'll accept this via a ThemeParam for right now, I think this feature makes sense.

Long term I'd like to move the grid into an addon and configure it via https://github.com/storybooks/storybook/issues/4169

Hypnosphi:
@tmeasday well we have things like brandURL in theme which means that we are not too picky anyway

PR #6259, opened 2019-03-25
FIX storybook having a incompatible version of webpack with vue-cli

ndelangen:
Issue: https://stackoverflow.com/a/53630702

## What I did
CHANGE the required version of webpack so it's more likely to be compatible with the users installed version.


backbone87 (approved):
lgtm, but why do the apps also require webpack and do not rely on cores webpack constraint?

ndelangen:
> lgtm, but why do the apps also require webpack and do not rely on cores webpack constraint?

Yeah we should probably use the one from core

backbone87:
@ndelangen does this also go into 5.0.x?

PR #6315, opened 2019-03-27
Added missing background color for left panel

EskelCz:
This is a problem when working on an app with darker background. It might only be a problem for react-native-navigation apps, which uses different rootView structure.

Issue: Discord conversation

## What I did
Added white background to the left panel

## How to test
Open DeviceUI on React Native, work a view with black background, other panels should be white.

Maybe with screenshots, not sure. Just not necessary in my opinion.
No.
No.


Keraito (approved):

ndelangen:
Great to have this fixed @EskelCz !

PR #6323, opened 2019-03-27
CLI: add --preview-url cli argument for custom preview

RPDeshaies:
Issue: https://github.com/storybooks/storybook/issues/5975

## What I did

- I added a new env variable called `PREVIEW_URL` that the manager can use to override the url of the iframe to laod
- I added a new CLI option called `--ignore-preview` that triggers an already available flag called `ignorePreview`
- I updated the `preview.js` component to use the `process.env.PREVIEW_URL` variable if available, otherwise falling back to `iframe.html`

## How to test

We could create a new repo I guess to test this feature ?
To test locally, I've been updating (temporarly) the `package.json` script called `storybook` from `official-storybook`

```
 "storybook": "cross-env PREVIEW_URL=http://localhost:1337 STORYBOOK_DISPLAY_WARNING=true DISPLAY_WARNING=true start-storybook -p 9011 -c ./ -s built-storybooks --no-dll --ignore-preview"
```

Then I execute a little parcel example I run locally as well

```tsx
//storybook.tsx
import { configure, storiesOf } from "@storybook/react";
import React from "react";

configure(() => {
  storiesOf("Category 1").add("Test 1", () => <div>Category 1 - Test 1</div>);
  storiesOf("Category 2")
    .add("Test 1", () => <div>Category 2 - Test 1</div>)
    .add("Test 2", () => <div>Category 2 - Test 2</div>);
}, {});
```

```html
<!-- storybook.html -->
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
    <title></title>
  </head>
  <body>
    <div id="root"></div>
    <script src="./storybook.tsx"></script>
  </body>
</html>
```

```bash
# parcel command
parcel ./lib/storybook.html --port 1337
```

Maybe ? I'm not sure @tmeasday ?

Yes, but where @tmeasday?

If you can't run it locally, here's my entire `package.json` for my parcel test

```json
{
  "name": "",
  "version": "0.0.0",
  "description": "",
  "license": "UNLICENSED",
  "main": "dist/storybook.html",
  "scripts": {
    "watch": "parcel ./lib/storybook.html --port 1337"
  },
  "dependencies": {
    "@storybook/react": "^5.0.5",
    "react": "^16.8.5",
    "react-dom": "^16.8.5"
  },
  "devDependencies": {
    "@babel/core": "^7.4.0"
  }
}
```

### DIFF
--- a/lib/core/src/server/build-static.js
+++ b/lib/core/src/server/build-static.js
@@ -193,6 +193,7 @@ export async function buildStaticStandalone(options) {
 
 export function buildStatic({ packageJson, ...loadOptions }) {
   const cliOptions = getProdCli(packageJson);
+  process.env.PREVIEW_URL = cliOptions.previewUrl;
 
   return buildStaticStandalone({
     ...cliOptions,
@@ -200,6 +201,6 @@ export function buildStatic({ packageJson, ...loadOptions }) {
     packageJson,
     configDir: loadOptions.configDir || cliOptions.configDir || './.storybook',
     outputDir: loadOptions.outputDir || cliOptions.outputDir || './storybook-static',
-    ignorePreview: loadOptions.ignorePreview || cliOptions.ignorePreview,
+    ignorePreview: !!cliOptions.previewUrl,
   });
 }
--- a/lib/core/src/server/build-static.js
+++ b/lib/core/src/server/build-static.js
@@ -200,5 +200,6 @@ export function buildStatic({ packageJson, ...loadOptions }) {
     packageJson,
     configDir: loadOptions.configDir || cliOptions.configDir || './.storybook',
     outputDir: loadOptions.outputDir || cliOptions.outputDir || './storybook-static',
+    ignorePreview: loadOptions.ignorePreview || cliOptions.ignorePreview,
   });
 }
--- a/lib/core/src/server/config/utils.js
+++ b/lib/core/src/server/config/utils.js
@@ -23,6 +23,7 @@ export function loadEnv(options = {}) {
     // even when deployed inside a subpath. (like in GitHub pages)
     // In development this is just empty as we always serves from the root.
     PUBLIC_URL: options.production ? '.' : '',
+    PREVIEW_URL: process.env.PREVIEW_URL,
   };
 
   Object.keys(process.env)
--- a/lib/core/src/server/config/utils.js
+++ b/lib/core/src/server/config/utils.js
@@ -23,7 +23,7 @@ export function loadEnv(options = {}) {
     // even when deployed inside a subpath. (like in GitHub pages)
     // In development this is just empty as we always serves from the root.
     PUBLIC_URL: options.production ? '.' : '',
-    PREVIEW_URL: process.env.PREVIEW_URL,
+    PREVIEW_URL: process.env.PREVIEW_URL || '',
   };
 
   Object.keys(process.env)
--- a/lib/core/src/server/build-dev.js
+++ b/lib/core/src/server/build-dev.js
@@ -341,5 +341,6 @@ export async function buildDev({ packageJson, ...loadOptions }) {
     ...loadOptions,
     packageJson,
     configDir: loadOptions.configDir || cliOptions.configDir || './.storybook',
+    ignorePreview: loadOptions.ignorePreview || cliOptions.ignorePreview,
   });
 }
--- a/lib/core/src/server/build-dev.js
+++ b/lib/core/src/server/build-dev.js
@@ -335,12 +335,13 @@ export async function buildDevStandalone(options) {
 
 export async function buildDev({ packageJson, ...loadOptions }) {
   const cliOptions = await getDevCli(packageJson);
+  process.env.PREVIEW_URL = cliOptions.previewUrl;
 
   await buildDevStandalone({
     ...cliOptions,
     ...loadOptions,
     packageJson,
     configDir: loadOptions.configDir || cliOptions.configDir || './.storybook',
-    ignorePreview: loadOptions.ignorePreview || cliOptions.ignorePreview,
+    ignorePreview: !!cliOptions.previewUrl,
   });
 }
--- a/lib/core/src/server/cli/dev.js
+++ b/lib/core/src/server/cli/dev.js
@@ -36,6 +36,10 @@ async function getCLI(packageJson) {
     .option('--quiet', 'Suppress verbose build output')
     .option('--no-dll', 'Do not use dll reference')
     .option('--debug-webpack', 'Display final webpack configurations for debugging purposes')
+    .option(
+      '--ignore-preview',
+      'Ignore the storybook default preview implementation so that you can rollout your own'
+    )
     .parse(process.argv);
 
   // Workaround the `-h` shorthand conflict.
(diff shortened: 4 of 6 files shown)


BudgieInWA:
It might be nicer to provide a single argument `--preview-url localhost:8000` which enables `ignorePreview` and sets the preview url.

tmeasday:
This is great @RPDeshaies! I agree though, I would have thought it makes more sense to pass the preview url in on the same flag we disable it? 

I don't really know what the purpose of `--ignore-preview` is without having an alternate URL (I appreciate the code path was already there). Is there a use case where that is helpful?

RPDeshaies:
> It might be nicer to provide a single argument --preview-url localhost:8000 which enables ignorePreview and sets the preview url.

I agree. I didn't find an easy way to do it though. I'll investigate...

> Is there a use case where that is helpful?
Didnt find anything in the code so I don't know :/

RPDeshaies:
Just refactored the code to use `--preview-url` instead of `--ignore-preview`

Ready to review again

e.g. `--preview-url=http://localhost:1234`

tmeasday:
Awesome! @igor-dv I wonder if there is a way to set the manager's environment without having to change `process.env`?

tmeasday:
PS @shilman and I were working on refactoring the URL handling on the other side for some docs-mode changes he was making and I wonder if we can sneak both sides of this into 5.1?

shilman:
@ndelangen @tmeasday one man's april fool's joke is another man's PR 😂 

Slack_-_Storybook

ndelangen (approved):

ndelangen:
Nice, I think this is a way for some users to incorporate the preview into their own apps if desired, or at least gain the control they need.

ndelangen:
The integration chromatic tests are failing though:
https://www.chromaticqa.com/snapshot?appId=5a375b97f4b14f0020b0cda3&id=5ca219ed22ee8b002095dd99

Can you take a look at why this happens @RPDeshaies ?

RPDeshaies:
@ndelangen any idea on how I could reproduce the error locally? I just run `yarn dev` and open the `cra-kitchen-sink` app and run `storybook` in it ?

shilman:
@RPDeshaies try `yarn bootstrap --reset --core && yarn build-storybooks`

After it's done, try looking at the integration examples in `examples/official-storybook`

RPDeshaies:
@shilman Thanks !

Found the issue.
For some reason `--preview-url` was `"undefined"` (the `"` are important)
I wrapped the code to get the baseUrl inside a function and now it works.


```tsx
function getBaseUrl() {
  if (!process.env.PREVIEW_URL || process.env.PREVIEW_URL === 'undefined') {
    return 'iframe.html';
  }
  return process.env.PREVIEW_URL;
}

Preview.defaultProps = {
  viewMode: undefined,
  storyId: undefined,
  path: undefined,
  description: undefined,
  baseUrl: getBaseUrl(),
};
```

RPDeshaies:
Can someone tell me why the build is failing ? I don't have access

shilman:
@RPDeshaies I think the build is fine. There are a bunch of broken build items that need maintenance, but those are generally a problem and not specific to yours. The one that is important--`build_test_deploy`--is green... 😖

RPDeshaies:
@shilman alright good to know. Well you guys can merge it when you feel like it then :)

shilman:
Thanks @RPDeshaies ! Waiting for @tmeasday 's review 👀

lib/core/src/server/config/utils.js
```diff
@@ -23,6 +23,7 @@ export function loadEnv(options = {}) {
     // even when deployed inside a subpath. (like in GitHub pages)
     // In development this is just empty as we always serves from the root.
     PUBLIC_URL: options.production ? '.' : '',
+    PREVIEW_URL: process.env.PREVIEW_URL,
```
tmeasday: maybe this should be `|| ''` as in the other options above

lib/core/src/server/build-static.js
```diff
@@ -193,12 +193,14 @@ export async function buildStaticStandalone(options) {
...
   const cliOptions = getProdCli(packageJson);
+  process.env.PREVIEW_URL = cliOptions.previewUrl;
 
   return buildStaticStandalone({
     ...cliOptions,
     ...loadOptions,
     packageJson,
     configDir: loadOptions.configDir || cliOptions.configDir || './.storybook',
     outputDir: loadOptions.outputDir || cliOptions.outputDir || './storybook-static',
+    ignorePreview: !!cliOptions.previewUrl,
```
tmeasday: can we add `previewUrl` as another option here and pass it through to `loadEnv` below? I don't love messing with `process.env`

tmeasday (changes requested):
If @ndelangen and @shilman are happy as is, I'm OK, but I think it would be better to pass through the option rather than setting `process.env` ourselves

tmeasday:
@RPDeshaies any chance you could look at the final couple of changes I was asking for?

lib/core/src/server/config/utils.js
RPDeshaies: will do

lib/core/src/server/build-static.js
RPDeshaies: I don't mind but I have no idea how the `Preview.js` component would be able to read form this "config" ? Do you have any example of a component reading from there ?

I initially thought that only the CLI /config tasks could read from the result of this function

lib/core/src/server/config/utils.js
RPDeshaies: pushed

RPDeshaies:
>@RPDeshaies any chance you could look at the final couple of changes I was asking for?

updated one comment, added quesiton for the other

lib/core/src/server/build-static.js
tmeasday: Oh, I think you misunderstand - I wasn't clear; I didn't mean don't set `process.env` for the webpack process, just don't mess with node's `process.env` on line 196 (☝️).

I'm pretty sure the `ignorePreview` option you set here will get passed through in the `options` argument of `loadEnv` below (in `utils.js` 👇) -- so if you also added `previewUrl` there, you could set the `PREVIEW_URL` env var there without needing to read it out of node's `process.env`.

Basically passing the option through rather than indirectly sending it via a global var (on `process.env`)

tmeasday (changes requested):

dgreensp:
I'm excited to see this land.

tmeasday:
👋 @dgreensp

dgreensp:
Hi Tom! :)

RPDeshaies:
@tmeasday my bad, I forgot to press the "submit review" button for like 2 weeks.

(I'm used to bitbucket reviews and there's no "submit" comments are just added automatically 😅 )

RPDeshaies:
@shilman Hey ! I'm super happy this got merged and cleaned up. 🚀 

I was wondering, I'm trying the `5.2.0-alpha.33` version right now and I can make the `start-storybook` command work with the new flag which is awesome.

But I'm not able to trigger it for `build-storybook`

Does it work on your end?
Did I miss something?

shilman:
@RPDeshaies Yep that's a bug. Fixed here & will release shortly: https://github.com/storybookjs/storybook/pull/7245

PR #6325, opened 2019-03-27
Remove styling of elements after re-running the the accessibility rules

CodeByAlex:
Issue: (No formal issue - If we would like to track it with an issue, I can create one) When tests are re-run after a11y rules are fixed, the elements that are highlighted remain highlighted until the next time the redux state is changed. This is not expected behavior.

## What I did
I added a line of code to remove the elements from the redux store(In the same way I did on tab switch) when the tests are re-run.

## How to test


Armanio (approved):

ndelangen:
Go ahead and merge @CodeByAlex 👍

ndelangen (approved):

PR #6346, opened 2019-03-30
Allow optional component declaration

MaximSagan:
Issue: #5542

Typically, when developing features in Angular, it is advantageous to use "feature modules" which provide a "context" for declared components including required imports, declarations, and providers.

To simulate this context in Storybook, we may want to use the component that has already been declared in its FeatureModule, rather than having to recreate this context manually (with imports, providers, and additional declarations) just to get the component to perform its basic functionality.

However, currently, the only way to display a component declared in its feature module is to refer to it indirectly in the template  property (e.g. `template: '<my-component [prop1]="prop1"></my-component>'`) which is used by Storybook to dynamically create a component called DynamicComponent. While this works, we lose the following benefits:
- Direct mapping between props and the @Inputs and @Outputs of the component we want to display.
- Any potential benefits of type-checking for property names.

If we want to get the above benefits, we need to specify the component by its class (e.g. `component: MyComponent`). Unfortunately, the default behavior of Storybook for Angular is to declare specified components in a dynamic module, which prevents us from using the version of the component declared in our feature module (with its appropriate context), due to the fact that a component can only be declared in one module (else Angular will error out).

## What I did

I added an optional flag called `requiresComponentDeclartion` to the NgStory object that the user specifies in the ".add()" method. When this flag is set to false, Storybook will not declare the specified component in the DynamicModule. This flag is set to true by default, which leaves the existing behavior (hence this is not a breaking change). 

## How to test

  Maybe? Please advise.
  Yes, added.
  The Storybook for Angular documentation does not seem to go very deep into specific configuration options.


MaximSagan:
Sorry, as with my other PR, snapshots were missing. Should be good now.

ndelangen:
@kroeder @gaetanmaisse Would you be able to review this possibly? Sounds useful.

gaetanmaisse (approved):

ndelangen (approved):

PR #6348, opened 2019-03-30
Fix a11y build by adding hoist-non-react-statics

kroeder:
## What I did
I encountered a build error in addon-a11y: https://circleci.com/gh/storybooks/storybook/114635?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

After reading https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33919 I thought raising `@types/react-redux` would fix it. It did not and the suggested way to fix this is to add `"hoist-non-react-statics": "^3.3.0",`

We should remove this as soon as it is fixed - I will create a separate issue for that task.

### DIFF
--- a/addons/a11y/package.json
+++ b/addons/a11y/package.json
@@ -39,12 +39,13 @@
     "memoizerific": "^1.11.3",
     "react": "^16.8.4",
     "react-redux": "^6.0.1",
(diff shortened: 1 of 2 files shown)


shilman (approved):
OK! Don't quite understand this, but merging

PR #6384, opened 2019-04-02
Update CLI options usage

Fonger:
Issue:

## What I did
Update CLI options docs
I'm curios why build has no `--quiet` mode to run on CI with no verbose output and it turns out that this options work. It's just missing from the docs.


shilman (approved):
Thanks @Fonger! 谢谢!

PR #6388, opened 2019-04-02
chore: bump react-syntax-highlighter to ^10.1.2

nadiia:
Issue:

## What I did
Upgraded react-syntax-highlighter to the latest.


ndelangen:
Hi @nadiia, thank you for taking the time to contribute, it's very appreciated.

I've delayed upgrading this particular package for some time, because we're using api within the package to make lines clickable/navigable.

You can find this code here:
https://github.com/storybooks/storybook/blob/cedc39f1a2e24f892483f43387e880178c008747/addons/storysource/src/StoryPanel.js#L7

Upgrading breaks this feature (at least last time I checked).

Could you verify this feature works after upgrading and if not perhaps look for a fix?

Thanks, let me know if you need something 👍

PR #6395, opened 2019-04-03
fix: Set custom options on first load

mortal94:
Issue:
`V5.1.0-alpha.20`: Custom options were not loaded on storybook first load, only when selecting a story (changing story)

## What I did
Saw that in v5.0.6 the above issue does not happen, so i've check what have changed and saw when moving the "Provider" from `v5.0.x` to `typescript` in `v5.1.x` in SET_STORIES event there was missing setting the first options, so i've added it back.

## How to test
1. change some options and load them with addParameters.
2. clear localStorage and referesh storybook and see that on first load it'll apply the custom options.

### DIFF
--- a/lib/api/src/index.tsx
+++ b/lib/api/src/index.tsx
@@ -142,6 +142,10 @@ class ManagerProvider extends Component<Props, State> {
 
     api.on(SET_STORIES, (data: { stories: StoriesRaw }) => {
       api.setStories(data.stories);
+      const options = storyId
+        ? api.getParameters(storyId, 'options')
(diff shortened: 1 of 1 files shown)


shilman (approved):
Thanks @mortal94!

@mortal94 @ndelangen are there tests we can add to help reduce these kinds of regressions?

mortal94:
> Thanks @mortal94!
> 
> @mortal94 @ndelangen are there tests we can add to help reduce these kinds of regressions?

NP @shilman !
About tests, sorry i don't have a good idea right now

ndelangen (approved):

ndelangen:
@mortal94 great work!

PR #6397, opened 2019-04-03
Fix docs links to custom webpack config page

jamisuomalainen:
Issue: Links to Custom Webpack Config page were pointing to a nonexisting page (404).

## What I did

Replaced all references to the non-existing configurations/webpack page with the correct found in other references to the Custom Webpack Config page.


shilman (approved):
Thanks @jamisuomalainen 🙏

PR #6398, opened 2019-04-03
feat(addon-notes): use @storybook/router <Link> to render links in notes

sto3psl:
Issue: https://github.com/storybooks/storybook/issues/6394

## What I did

Links in the notes addon cause storybook to reload. This is not the best user experience when the link points to a different story. Since storybook uses a client side router already, I used the `overrides` option in `markdown-to-jsx` to use a custom `<NotesLink />` component for all markdown links. It basically uses normal `<a>` tags for absolute links (external) and the `@storybook/router` `<Link />`
for internal links (relative). That way, nothing breaks and storybook uses the `@reach/router` in addon notes when it can.
 
## How to test

Yes, I added tests and an example in the svelte kitchen sink app.
I added a relative link in the svelte kitchen sink app but this is nothing where users have to do anything different.
No.


ndelangen (approved):
Awesome!

PR #6403, opened 2019-04-03
Migrate @storybook/addon-jest to TS

sairus2k:
Issue: #5030

## What I did

- Migrate @storybook/addon-jest to TypeScript
- Rewrite `createSubgroup` function to more types-friendly
- Extract `Message` component into separate file


addons/jest/package.json
```diff
@@ -34,11 +34,14 @@
     "@storybook/theming": "5.1.0-alpha.20",
     "core-js": "^2.6.5",
     "global": "^4.3.2",
-    "prop-types": "^15.7.2",
     "react": "^16.8.4",
     "upath": "^1.1.0",
     "util-deprecate": "^1.0.2"
   },
+  "devDependencies": {
+    "@storybook/api": "^5.1.0-alpha.20",
```
ndelangen: Are there devDependencies though? I think they are regular dependencies.

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,212 @@
+/* tslint:disable:object-literal-sort-keys */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import { styled } from '@storybook/theming';
+import colors from '../colors';
```
ndelangen: I think these colours should come from the theme if possible

addons/jest/src/register.tsx
```diff
@@ -2,15 +2,13 @@ import React from 'react';
 import addons from '@storybook/addons';
 import { ADDON_ID, PANEL_ID } from './shared';
 
-// import PanelTitle from './components/PanelTitle';
 import Panel from './components/Panel';
 
 addons.register(ADDON_ID, api => {
   const channel = addons.getChannel();
```
ndelangen: We should have no need for the channel anymore, calling api.on() === channel.on()

All relavent channel APIs are available on api

ndelangen:
addons/jest/src/styles.js → addons/jest/src/styles.ts

I'm unsure if this is still used, I can't find anything tht would use this.

I think we should delete this, or at the very least add a deprecation so we can remove it in a next release

ndelangen:
This is great! thanks for taking this on @sairus2k !

I think it works well and is merge-able.

We could do a bit more cleanup of this addon, how do you feel about doing that in this PR?

addons/jest/package.json
sairus2k: I thought they need only for type info, they don`t need in regular dependencies. But they also imported in dist/**.d.ts files. Ok, I`ll move them to regular dependencies.

addons/jest/src/components/Message.tsx
sairus2k: Hmm, I don't find similar colors in `lib/theming/src/base.ts`

addons/jest/src/register.tsx
sairus2k: OK

ndelangen:
Looks like you also need to remove `addons/jest/styles.js`
```
/tmp/storybook/addons/jest/styles.js
  1:9  error  Unable to resolve path to module './dist/styles'  import/no-unresolved
```

addons/jest/src/components/Result.tsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+
+import { styled } from '@storybook/theming';
+
+// @ts-ignore TODO: migrate Message to typescript
```
gaetanmaisse: `// @ts-ignore TODO: migrate Message to typescript` it looks like you did it 😉

addons/jest/src/components/Indicator.ts
```diff
@@ -1,8 +1,17 @@
...
 import { styled } from '@storybook/theming';
 
-const Indicator = styled.div(
+interface IndicatorProps {
+  color: string;
+  size: number;
+  children?: React.ReactNode;
+  right?: boolean;
+  overrides?: any;
+  styles?: any;
```
hipstersmoothie: could you do something like `React.CSSProperties` here?

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,208 @@
...
+  if (i === list.length - 1) {
+    // Provide a "safety net" when Jest returns a partially recognized "group"
+    // (recognized by acc.startTrigger but acc.endTrigger was never found) and
+    // it's the only group in output for a test result. In that case, accList
+    // will be empty, so return whatever was found, even if it will be unstyled
+    // and prevent next createSubgroup calls from throwing due to empty lists.
+    accList.push(null);
+
+    return accList.reduce<MsgElement[]>((eacc, el, ei) => {
+      switch (true) {
```
hipstersmoothie: this should really just be some if/else. a switch doesn't really make sense and makes the code more complex

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,208 @@
...
+  msg: MsgElement[];
+  className?: string;
+}
+
+const Sub = styled(({ msg, className }: SubProps) => (
+  <section className={className}>
+    {msg
+      .filter(item => typeof item !== 'string' || item.trim() !== '')
+      .map((item, index, list) => {
+        if (typeof item === 'string') {
```
hipstersmoothie: seems odd you filter for anything that isn't a string, then do something for only strings. seems like this code could be clearer

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+type MsgElement = string | JSX.Element;
+
+const Message = ({ msg }: MessageProps) => {
+  const data = patterns
+    .reduce((acc, regex) => acc.replace(regex, ''), msg)
+    .split(/\[2m/)
+    .join('')
+    .split(/\[22m/)
+    .reduce<string[]>((acc, item) => acc.concat(item), [])
```
hipstersmoothie:
```suggestion
    .reduce((acc, item) => acc.concat(item), [] as string[])
```

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,208 @@
...
+    .map((item, li) =>
+      typeof item === 'string'
```
hipstersmoothie: are these type checks necassary? given the above reduce you should be getting an array of strings

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,208 @@
...
+    )
+    .reduce((acc, item) => acc.concat(item))
+    .map((item, li) =>
+      typeof item === 'string'
+        ? item
+            .split(/\[31m(.*?)\[39m/)
+            .map((i, index) => (index % 2 ? <Negative key={`n_${li}_${i}`}>{i}</Negative> : i))
+        : item
+    )
+    .reduce<MsgElement[]>((acc, item) => acc.concat(item), [])
```
hipstersmoothie:
```suggestion
    .reduce((acc, item) => acc.concat(item), [] as MsgElement[])
```

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,208 @@
...
+    .reduce(
+      createSubgroup({
+        startTrigger: e => typeof e === 'string' && e.indexOf('Error: ') === 0,
+        endTrigger: e => typeof e === 'string' && Boolean(e.match('Expected ')),
+        grouper: (list, key) => <Main key={key} msg={list} />,
+      }),
+      []
+    )
+    .reduce<MsgElement[]>(
```
hipstersmoothie: probably should use `as` here instead. like my above suggestions

addons/jest/src/components/Message.tsx
```diff
@@ -0,0 +1,208 @@
...
+      (acc, it) =>
+        typeof it === 'string' ? acc.concat(it.split(/(at(.|\n)+\d+:\d+\))/)) : acc.concat(it),
+      []
+    )
+    .reduce<MsgElement[]>((acc, item) => acc.concat(item), [])
```
hipstersmoothie:
```suggestion
    .reduce((acc, item) => acc.concat(item), [] as MsgElement[])
```

addons/jest/src/components/Result.tsx
hipstersmoothie: 🎉

addons/jest/src/components/Message.tsx
sairus2k: Could you please describe the difference?

addons/jest/src/components/Message.tsx
sairus2k: In this case TS throws error `Property 'concat' does not exist on type Element`

addons/jest/src/components/Message.tsx
sairus2k: It filters for not empty strings, then removes line breaks for those strings

ndelangen (approved):

PR #6425, opened 2019-04-04
REMOVE the testing & building of examples in Teamcity

ndelangen:
Issue: Some tests were run in multiple CIs

With the introduction of now.sh we have easily accessible examples, and so we need the examples in teamcity less. We also had the running of tests duplicated in CircleCI.

Since we're taking a lot of resources from teamcity we'll lean more on CircleCI and now.sh

## What I did
I removed the CI teamcity config for tests & examples


ndelangen:
@Hypnosphi I could use some help, cleanly removing the proper things 😅 

@shilman will be moving Chromatic too

Hypnosphi:
So, basically, we're just using TeamCity for ESLint inspections? Then I can simplify the config a bit more and remove "Bootstrap" and "Build" configurations

> @shilman will be moving Chromatic too

This probably should be done first, because "Chromatic" depends on "Examples"

PR #6428, opened 2019-04-04
Migrate addon-options to typescript

kroeder:

## What I did

Migrate addon-options to TS


kroeder:
This package is deprecated - should I bother about typings for this package? I just did a quick renaming and "any" everything that turned red

ndelangen:
I'm fine with typing it for now

Just don't put large amounts of time in it.

addons/options/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "license": "MIT",
   "main": "dist/index.js",
+  "types": "dist/public_api.d.ts",
```
ndelangen: Why is this not `dist/index.d.ts`?

ndelangen (approved):

addons/options/package.json
kroeder: Because this package has multiple files with exports - I then use public_api.d.ts as entry point and export everything from one file

PR #6435, opened 2019-04-05
[bug] Move '@emotion/*' path-resolutions to the dependent package

devrelm:
Issue: #5817, #6255 

In certain cases, `yarn install` would create a directory structure where `@emotion/core` lived inside the `./node_modules/@storybook/theming/node_modules` directory. This caused `require.resolve('@emotion/core/package.json')` within `./node_modules/@storybook/ui/paths.js` to fail.

## What I did

I moved the `require.resolve` stuff for the `@emotion` packages into `@storybook/theming`, which actually lists them as dependencies. I then set up `manager-webpack.config.js` to import those paths from `@storybook/theming`.

## How to test

- Is this testable with Jest or Chromatic screenshots? _No_
- Does this need a new example in the kitchen sink apps? _No_
- Does this need an update to the documentation? _No_

### DIFF
--- a/lib/core/src/server/manager/manager-webpack.config.js
+++ b/lib/core/src/server/manager/manager-webpack.config.js
@@ -3,6 +3,7 @@ import webpack from 'webpack';
 import Dotenv from 'dotenv-webpack';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import CaseSensitivePathsPlugin from 'case-sensitive-paths-webpack-plugin';
+import themingPaths from '@storybook/theming/paths';
 import uiPaths from '@storybook/ui/paths';
 
 import findCacheDir from 'find-cache-dir';
@@ -65,6 +66,7 @@ export default ({ configDir, configType, entries, dll, outputDir, cache, babelOp
       extensions: ['.mjs', '.js', '.jsx', '.json'],
       modules: ['node_modules'].concat(raw.NODE_PATH || []),
       alias: {
+        ...themingPaths,
         ...uiPaths,
       },
     },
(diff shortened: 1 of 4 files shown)


lib/ui/paths.js
```diff
@@ -3,8 +3,6 @@ const { dirname } = require('path');
 // These paths need to be aliased in the manager webpack config to ensure that all
 // code running inside the manager uses the *same* version of react[-dom] that we use.
 module.exports = {
-  '@emotion/core': dirname(require.resolve('@emotion/core/package.json')),
-  '@emotion/styled': dirname(require.resolve('@emotion/styled/package.json')),
```
benoitdion: there's also `'emotion-theming': dirname(require.resolve('emotion-theming/package.json')),` below. Does that need to be moved too?

lib/core/src/server/manager/manager-webpack.config.js
```diff
@@ -65,6 +66,7 @@ export default ({ configDir, configType, entries, dll, outputDir, cache, babelOp
       extensions: ['.mjs', '.js', '.jsx', '.json'],
       modules: ['node_modules'].concat(raw.NODE_PATH || []),
       alias: {
+        ...themingPaths,
```
benoitdion: nit: should we spread `uiPaths` first since it depends on theming?

lib/ui/paths.js
devrelm: You're likely right. I didn't really take time to go through anything outside of `@emotion/core`, but I should check the others as well.

lib/core/src/server/manager/manager-webpack.config.js
devrelm: I think that it should go in order of leaves-to-root. Whatever spreads last will overwrite what came before, so if the idea is that the package closest to the root should get the last say in how the dependencies are resolved, then `@storybook/ui` should come last.

ndelangen (approved):
Couldn't have done it better myself

marco-silva0000:
I think this error still persists, when using yarn with pnp. I've created #7206 with some more context

PR #6438, opened 2019-04-05
Fix for issue 5932, adding default options to action(s)

LFDanLu:
Issue: https://github.com/storybooks/storybook/issues/5932

Apologies for any missteps, first time doing a pull for a open source project :).

## What I did
I noticed that both `action` and `actions` defaulted their options to `{}`, hence the reported issue. To remedy this, I imported `config` from `configureActions.ts` and set it as the new default for ActionOptions.

## How to test
To test, you can play around with the “with text” and “with some emoji” stories in the cra-kitchen-sink example story (https://github.com/storybooks/storybook/tree/next/examples/cra-kitchen-sink). 

Sample test flow:
1. Navigate to “Button” -> “with text” in the cra-kitchen-sink storybook
2. Click the “Hello Button” a couple times to generate some action logs.
3. Switch to the “with some emoji” story. Note that the action logs aren’t wiped since that action doesn’t have `clearOnStoryChange: true`
4. Click the emoji button in the “with some emoji” story. Note that a new action is logged
5. Change to a different story. Note that the action log is wiped clean since the action associated with that button doesn’t have any options, thus using the default config from `configureActions`.

Additional testing scenarios can be done by adding `configureActions` with various settings in the `config.js` of the cra-kitchen-sink storybook.


addons/actions/src/preview/__tests__/action.test.js
```diff
@@ -29,67 +28,66 @@ describe('Action', () => {
   });
 });
 
-// TODO: This functionality is removed, unsure if to add back or remove
```
LFDanLu: I saw that this test was commented out and covers the exact functionality that I was re-adding via this pull. Couldn't find the old implementation nor the reasoning for it's removal so if this was removed for a specific reason I would appreciate a pointer.

LFDanLu: Scratch that found the pull (https://github.com/storybooks/storybook/pull/4086/), but it is quite a large one so I'm still unable to glean the reason for the initial removal. Gonna change my implementation to mirror how user provided config options were merged with the config supplied by `configureActions`)

LFDanLu:
@ndelangen I not certain how to assign reviewers, is it supposed to be an automatic process?

ndelangen (approved):

PR #6448, opened 2019-04-07
Migrate app/react-native to typescript

benoitdion:
Issue: #5030

## What I did

Migrate app/react-native to typescript. Most types have been annotated properly except for the data we get from client-api.


app/react-native/src/preview/components/OnDeviceUI/navigation/index.tsx
```diff
@@ -20,13 +19,19 @@ const style = StyleSheet.create({
...
 
-export default class Navigation extends PureComponent {
-  constructor(props) {
-    super(props);
-    this.state = {
-      isUIVisible: props.initialUiVisible,
-    };
-  }
+interface Props {
+  initialUiVisible: boolean;
```
gaetanmaisse: According to https://github.com/storybooks/storybook/pull/6448/commits/ea0c8a0f4a997791ed4999c0f8d8b5cec487089a#diff-46b88ac54083f1f8e54a54e6f4a3088fL77
`initialUiVisible` is optional so maybe `initialUiVisible?: boolean;` is more accurate

app/react-native/src/preview/components/OnDeviceUI/navigation/index.tsx
```diff
@@ -1,8 +1,7 @@
 import React, { PureComponent } from 'react';
 import { View, SafeAreaView, StyleSheet } from 'react-native';
+// @ts-ignore TODO: add types for react-native-swipe-gestures
```
gaetanmaisse: or add a typings.d.ts with `declare module 'react-native-swipe-gestures'` if types are not available for this lib 😉

gaetanmaisse (approved):

app/react-native/src/preview/components/OnDeviceUI/navigation/index.tsx
benoitdion: good catch

app/react-native/src/preview/components/OnDeviceUI/navigation/index.tsx
benoitdion: added in 29ddf5aafc483b9571cac5d4378ff53d6c4ae185

PR #6484, opened 2019-04-10
Angular: Support default `storybook` project configuration

dankrieger:
Issue: Storybook Angular assumes use of `defaultProject` defined in angular.json as its source for project configuration.  In larger projects it would be beneficial to define a separate project for storybook that can define configuration settings separate from the default project.

## What I did

When resolving the angular project to use for configuration, default to look for a project named `storybook`.  If no project named `storybook` exists, use the existing resolution process.

## How to test

Added test to angular CLI config to identify that storybook project is used when available in angular.json file.

NOTE: I recognize this could be a breaking change if an existing project in angular.json is named storybook but is not intended to be the project configured used when running storybook.  However I suspect this potential conflict is very rare in existing applications.

### DIFF
--- a/app/angular/src/server/angular-cli_config.js
+++ b/app/angular/src/server/angular-cli_config.js
@@ -46,10 +46,15 @@ export function getAngularCliWebpackConfigOptions(dirToSearch) {
     throw new Error('angular.json must have projects entry.');
   }
 
-  let project = projects[Object.keys(projects)[0]];
+  // By default, look for a `storybook` project in angular.json to use as configuration default
+  let project = projects.storybook;
 
-  if (defaultProject) {
-    project = projects[defaultProject];
+  if (!project) {
+    project = projects[Object.keys(projects)[0]];
+
+    if (defaultProject) {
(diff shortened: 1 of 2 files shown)


app/angular/src/server/angular-cli_config.js
```diff
@@ -46,10 +46,15 @@ export function getAngularCliWebpackConfigOptions(dirToSearch) {
     throw new Error('angular.json must have projects entry.');
   }
 
-  let project = projects[Object.keys(projects)[0]];
+  // By default, look for a `storybook` project in angular.json to use as configuration default
+  let project = projects.storybook;
```
ThomasBurleson:
```ts
let fallbackProject = defaultProject && projects[defaultProject];
let firstProject = projects[Object.keys(projects)[0]];

let project = projects.storybook || fallbackProject || firstProject;
```

ThomasBurleson (changes requested):

ndelangen (approved):

PR #6486, opened 2019-04-10
Fix/custom parameters on preview url

ndelangen:
Issue: https://github.com/storybooks/storybook/issues/5533

## What I did
- I allowed customQueryParameters to be set the preview URL
- I allowed customQueryParameters to be set on the URL of the new window
- I changed knobs so it would read the customQueryParameters for knobs

## How to test

Here is it in action:
https://www.youtube.com/watch?v=NQsktVyZIsA

### DIFF
--- a/addons/knobs/src/components/Panel.js
+++ b/addons/knobs/src/components/Panel.js
@@ -85,7 +85,7 @@ export default class KnobPanel extends PureComponent {
           if (urlValue !== undefined) {
             const value = Types[knob.type].deserialize(urlValue);
             knob.value = value;
-            queryParams[`knob-${name}`] = value;
+            queryParams[`knob-${name}`] = Types[knob.type].serialize(value);
           }
         }
       });
--- a/addons/knobs/src/components/Panel.js
+++ b/addons/knobs/src/components/Panel.js
@@ -87,8 +87,8 @@ export default class KnobPanel extends PureComponent {
           }
         }
 
-        // set all knobsquery params to be deleted from URL
-        queryParams[`knob-${name}`] = null;
+        // set all knobsquery params to serialized value
+        queryParams[`knob-${name}`] = Types[knob.type].serialize(knobs[name].value);
       });
 
       api.setQueryParams(queryParams);
--- a/addons/knobs/src/components/Panel.js
+++ b/addons/knobs/src/components/Panel.js
@@ -72,7 +72,7 @@ export default class KnobPanel extends PureComponent {
 
   setKnobs = ({ knobs, timestamp }) => {
     const queryParams = {};
-    const { api, channel } = this.props;
+    const { api } = this.props;
 
     if (!this.options.timestamps || !timestamp || this.lastEdit <= timestamp) {
       Object.keys(knobs).forEach(name => {
@@ -80,15 +80,14 @@ export default class KnobPanel extends PureComponent {
         // For the first time, get values from the URL and set them.
         if (!this.loadedFromUrl) {
           const urlValue = api.getQueryParam(`knob-${name}`);
+
+          // If the knob value present in url
           if (urlValue !== undefined) {
-            // If the knob value present in url
-            knob.value = Types[knob.type].deserialize(urlValue);
-            channel.emit(CHANGE, knob);
+            const value = Types[knob.type].deserialize(urlValue);
+            knob.value = value;
+            queryParams[`knob-${name}`] = value;
           }
         }
-
-        // set all knobsquery params to serialized value
-        queryParams[`knob-${name}`] = Types[knob.type].serialize(knobs[name].value);
       });
 
       api.setQueryParams(queryParams);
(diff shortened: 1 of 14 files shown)


addons/knobs/src/components/Panel.js
```diff
@@ -72,23 +72,22 @@ export default class KnobPanel extends PureComponent {
...
         const knob = knobs[name];
         // For the first time, get values from the URL and set them.
         if (!this.loadedFromUrl) {
           const urlValue = api.getQueryParam(`knob-${name}`);
+
+          // If the knob value present in url
           if (urlValue !== undefined) {
-            // If the knob value present in url
-            knob.value = Types[knob.type].deserialize(urlValue);
-            channel.emit(CHANGE, knob);
```
leoyli: Should we still fire an event?! I.e. `api.emit(CHANGE, knob)`; otherwise, this may be a minor breaking change for some people (-> can not cherry picked this one before v5.1).

lib/ui/src/containers/preview.js
```diff
@@ -20,6 +20,10 @@ const mapper = ({ api, state: { layout, location, path, storyId, viewMode, selec
     : {
         api,
         getElements: api.getElements,
+        getQueryParams: () =>
+          Object.entries(api.getUrlState().queryParams).reduce((acc, [k, v]) => {
+            return `${acc}&${k}=${v}`;
+          }, ''),
```
leoyli: I was wondering for a while until I see the leading `&` was added here. Not so intuitive but I think that is okay since it become extendable by default

leoyli (approved):
Thank you @ndelangen!  It's a great work indeed 👍.  Minor concerns on RN side but I have no experience on RN world so will let people comment about that. 😄

lib/ui/src/containers/preview.js
tmeasday: I think it would make more sense to pass the raw API (`getUrlState()`) down the the preview and have it do the putting it together into a string (probably using `import { URL } from 'url'` too!)

lib/api/src/modules/url.ts
```diff
@@ -76,7 +76,7 @@ const initialUrlSupport = ({ navigate, location, path }: Module) => {
     setTimeout(() => navigate(`/story/*`, { replace: true }), 1);
   } else if (Object.keys(query).length > 1) {
     // remove other queries
-    setTimeout(() => navigate(`/${queryPath}`, { replace: true }), 1);
```
tmeasday: What's up with this change?

lib/client-api/src/index.js
```diff
@@ -13,3 +15,12 @@ export {
   pathToId,
   splitPath,
 };
+
+export const getQueryParams = () => {
+  return qs.parse(document.location.search, { ignoreQueryPrefix: true });
+};
+
+export const getQueryParam = key => {
```
tmeasday: Should we also use this API in the story store where we read the query?

tmeasday (changes requested):

lib/api/src/modules/url.ts
ndelangen: Noticed while testing the copy feature, a `/` would get prepended every time..

(you can actually see this in the video)

This fixes it

lib/client-api/src/index.js
ndelangen: Yeah sounds good to me

addons/knobs/src/components/Panel.js
ndelangen: The event will be redundant, the knob manager will already be in that exact state.

I can fire the event, to be sure, but I don't see the point? What usage is this breaking?

ndelangen: Ow.. 
> Minor concerns on RN

that probably?

lib/ui/src/containers/preview.js
ndelangen: ok

ndelangen:
Uhm, damn this seems unnecessary:

```
info Publishing to channel fix-customParameters-on-preview-url 
[09:36:05] Release channel name can only contain lowercase letters, numbers and special characters . _ and -
```

This fails the build.. a single capital letter in my branch name..

shilman (approved):
awesome!

addons/knobs/src/components/Panel.js
leoyli: I think SB using events heavily... although in theory if the event is name spaced, it should be no problem at all because it is meant to be used internally.  But since SB is also highly customizable, so standing the user point, I will take more conservative approach. At least in the doc level... Don't won't freak out people if they have customized code working well for long time but fail without any notice... (so I say "may" be a breaking change for some users :p)

ndelangen: Well not anymore, I added it back, be it with the `api.on` method, but that's all the same.

PR #6502, opened 2019-04-12
fix(background): move 'iframe' styles to background wrapper

Armanio:
Issue: https://github.com/storybooks/storybook/issues/6392

## What I did
Fix iframe styles: remove iframe inline styles, set styles from bg wrapper

## How to test
Run your storybook
Check bg
Check zoom
Check iframe behavior - all must work as expected

yes
no
no


shilman (approved):
Thanks for taking care of this. Fix looks good!

PR #6535, opened 2019-04-16
Fix for issue #6338: Changed default value for sourceMap property of angulars webpack config

aspidov:
Issue: #6338

## What I did

changed value of sourceMap from {} to false

## How to test

Create a project via latest angular-cli and add storybook, it will fail, because sourceMap property should be boolean

### DIFF
--- a/app/angular/src/server/angular-cli_config.js
+++ b/app/angular/src/server/angular-cli_config.js
@@ -71,7 +71,7 @@ export function getAngularCliWebpackConfigOptions(dirToSearch) {
     tsConfig,
     supportES2015: false,
     buildOptions: {
-      sourceMap: {},
+      sourceMap: false,
       optimization: {},
       ...projectOptions,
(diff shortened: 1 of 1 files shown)


shilman (commented):
@aspidov Is this backwards compatible? If so, how far back?

aspidov:
@shilman , this property is itroduced in this commit https://github.com/angular/angular-cli/commit/cde624ad76a3f713768a44dee43e14be72e9fdaf#diff-dffa709a67c66e298beea219faae843fR595 , and always was boolean.

aspidov:
@igor-dv @kroeder @shilman please let me know if you need any additional information.

ndelangen:
@yjcxy12 According to git you added this line, would you be able to pitch in here and see if the proposed change is correct too?

ndelangen (approved):

PR #6539, opened 2019-04-16
Fix: addon-jest display wrong result

sairus2k:
Issue: #5809

## What I did
- Fix fileName regex to capturing filenames with slash
- Add supports for Typescript test files
- Improve typings


addons/jest/src/index.ts
```diff
@@ -3,13 +3,17 @@ import deprecate from 'util-deprecate';
...
+}
+
 const findTestResults = (
   testFiles: string[],
   jestTestResults: { testResults: Array<{ name: string }> },
   jestTestFilesExt: string
 ) =>
   Object.values(testFiles).map(name => {
-    const fileName = `${name}${jestTestFilesExt}`;
+    const fileName = `[\\/\\\\]${name}${jestTestFilesExt}`;
```
ndelangen: Can you give some context around this funny business?

ndelangen:
Thanks for this @sairus2k!

Looks good, if you could add a comment so we understand this code later we can merge this I think.

addons/jest/src/index.ts
sairus2k: I have replaced regex with upath.sep.
I think this is more obvious and don't require code-comment.

ndelangen (approved):

PR #6544, opened 2019-04-17
FIX null brandImage

kennethtruong:
Issue: #5866

## What I did
Make it so that `brandImage` is used if defined instead of `null`

(Addresses the issue introduced in https://github.com/storybooks/storybook/pull/6120)


shilman (approved):
Thanks @kennethtruong ! Merging and will release

PR #6550, opened 2019-04-17
MIGRATE addon-storysource from using channel to api

ndelangen:
Issue:
We want to promote the usage of api over channel

## What I did
CLEANUP channel should not be used in manager, just use api instead


libetl:
Ok.
I will have to revamp both the editor and the addon in the editor branch.

libetl (approved):
:ok:

PR #6551, opened 2019-04-17
FIX viewmode being hard-coded making custom addons of type TAB impossible

ndelangen:
Issue:
I was documenting the TAB addon type, and discovered we had the `info` and `story` viewmodes hardcoded, this meant it was pretty much impossible to create custom viewmodes & thus addons of type TAB.

This PR changes the way we detect whether we're in `/settings/xyz`.

Now addons can easily create their own TAB! no hassle.

## What I did
FIX router & desktop layout, which made it impossible to add custom viewmodes

### DIFF
--- a/lib/router/src/utils.ts
+++ b/lib/router/src/utils.ts
@@ -6,7 +6,7 @@ interface StoryData {
   storyId?: string;
 }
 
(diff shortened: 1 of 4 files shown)


shilman (approved):
LGTM!

PR #6561, opened 2019-04-18
Emit event on updating background

pajter:
Issue: n/a

## What I did

I made the backgrounds addon emit an event when a user changes the background. This allows developers to listen for this event and implement any custom logic they need in their own Storybook (plugin), eg. setting a class name for theming on the preview body element based on the background color (light/dark).

I'm not sure if there's a specific convention I should follow for event names. I looked at and copied what I found in other addons.

## How to test

1. Fire up Storybook with backgrounds configured
2. Change the background through the background selector menu
3. Observe the `storybook/background/update` event message in the console

### DIFF
--- a/addons/backgrounds/src/containers/BackgroundSelector.tsx
+++ b/addons/backgrounds/src/containers/BackgroundSelector.tsx
@@ -1,12 +1,12 @@
 import React, { Component, Fragment } from 'react';
 import memoize from 'memoizerific';
 
-import { Combo, Consumer } from '@storybook/api';
+import { Combo, Consumer, API } from '@storybook/api';
 import { Global, Theme } from '@storybook/theming';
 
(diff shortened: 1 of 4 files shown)


addons/backgrounds/src/containers/BackgroundSelector.tsx
```diff
@@ -96,13 +96,16 @@ interface State {
...
+export class BackgroundSelector extends Component<{ api: API }, State> {
   state: State = {
     selected: null,
     expanded: false,
   };
 
-  change = (args: State) => this.setState(args);
+  change = ({ selected, name }: { selected: string; name: string }) => {
+    this.props.api.emit(EVENTS.UPDATE, { selected, name });
+    this.setState({ selected, expanded: false });
```
pajter: The menu always closes when selecting a color, so I moved `expanded: false` into here.

ndelangen (approved):

PR #6563, opened 2019-04-18
A11y design enhancements

CodeByAlex:
## What I did

I have modified the X and check marks to be more accessible because the color of the X alone is difficult extract meaning from if you are visually impaired. They now are colored and list out the severity/urgency of the issue (which really provides better information anyway).
OLD:
image

NEW:
image

I also updated the text that appears when there aren't any rules within a tab. It used to use the word "A11y" and now it uses the word "accessibility". This helps people that don't understand what a11y means.
OLD:
image

NEW:
image

The global checkbox and all of the badges now respond nicely (when the addons panel is mobile, vertical, or horizontal).
OLD:
image

NEW:
image

Capitalized the first character in colorblindness emulation text
image

## How to test

- Is this testable with Jest or Chromatic screenshots? Sure


Armanio:
Wow! Looks good!

Firstly, I think need to fix conflicts.

addons/a11y/src/components/Report/Rules.tsx
```diff
@@ -40,30 +53,63 @@ const Status = styled.div(({ passes, impact }: { passes: boolean; impact: string
...
+      break;
+    case ImpactValue.MODERATE:
+      badgeType = 'warning';
+      break;
+    case ImpactValue.MINOR:
+      badgeType = 'neutral';
+      break;
+    default:
+      break;
+  }
```
Armanio:
```suggestion
const badgeType = rule.impact in ImpactValue ? rule.impact : null;
```
can simplify?

addons/a11y/src/components/Tabs.tsx
```diff
@@ -18,26 +18,34 @@ const Container = styled.div({
...
+    fontSize: '14px',
+    padding: elementWidth > maxElementWidth ? '12px 15px 10px 0' : '12px 0px 3px 12px',
+    height: '40px',
+    border: 'none',
+    marginTop: elementWidth > maxElementWidth ? '-40px' : '0px',
+    float: elementWidth > maxElementWidth ? 'right' : 'left',
+    display: elementWidth > maxElementWidth ? 'flex' : 'block',
+    alignItems: 'center',
+    width: elementWidth > maxElementWidth ? 'auto' : '100%',
+    borderBottom: elementWidth > maxElementWidth ? 'none' : '1px solid rgba(0,0,0,.1)',
```
Armanio: why you didn't use media queries?

Armanio (changes requested):

CodeByAlex:
Shoot I removed that magic number constant when I was testing something.
Nice catch!  I’ll make sure to add that back in. I originally had media
queries but media queries aren’t responsive when the addon panel is
shortened in width in the vertical view.

CodeByAlex:
I did try to find a css solution before going with a JS solution but
nothing came up for what I needed for this use case.

domyen:
The width logic sounds good to me. There's no way to tell container dimensions with media queries so when the addon panel's orientation changes it may end up using a suboptimal layout.

addons/a11y/src/components/Report/Rules.tsx
```diff
@@ -15,18 +16,30 @@ const impactColors = {
...
   fontWeight: '400',
 } as any);
 
-const Item = styled.div({
-  display: 'flex',
-  flexDirection: 'row',
-  marginBottom: '6px',
-});
+const Item = styled.div(({ elementWidth }: { elementWidth: number }) => ({
+  flexDirection: elementWidth > 407 ? 'row' : 'inherit',
```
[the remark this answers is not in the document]
shilman: Yeah please make this a named constant at least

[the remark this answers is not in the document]
CodeByAlex: Ill definitely do that, shouldnt have been in there as a magic string

addons/a11y/src/components/Report/Rules.tsx
CodeByAlex: I think that would be a good solution if the badge names were the same as the severity names but unfortunately, they are not

addons/a11y/src/components/Report/Rules.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { FunctionComponent } from 'react';
 import { styled } from '@storybook/theming';
-
-import { Icons } from '@storybook/components';
+import { Badge, Icons } from '@storybook/components';
 import { CheckResult } from 'axe-core';
+import { RuleType } from '../A11YPanel';
+import { SizeMe } from 'react-sizeme';
```
ndelangen: We could cut back on the number of dependencies by using https://github.com/storybooks/storybook/blob/tech/loader-scope-changes/lib/ui/src/app.js#L4

```js
import ResizeDetector from 'react-resize-detector';
```

CodeByAlex: That's a good point. I hadn't spotted that in the repo. It looks like it also has the "refreshMode" feature so it fits the needs. Ill convert what I have to use react-resize-detector. Thanks!

CodeByAlex: So I tried the react-resize-detector and there is a delay between when you start an action and the resizing occurs. Think We should switch back to ResizeMe where this did not occur. react-sizeme does an initial, invisible render of the component to measure its size. Then, when the component renders, the size is known and the component can render properly from the beginning.
glitchy-response

shilman:
@CodeByAlex build fails:

```
yarn run v1.12.3
$ node ./scripts/repo-dirty-check
 M yarn.lock
Git repo is dirty, please consider updating lockfiles by running `yarn bootstrap --reset --core`
```

CodeByAlex:
@Shilman Done!

addons/a11y/src/components/Report/Rules.tsx
CodeByAlex: I tried switching out the react-resize-detector for resize me but for that use case, react-resize-detector was the better library

CodeByAlex: @ndelangen what do you think about keeping both?

addons/a11y/src/components/Report/Rules.tsx
CodeByAlex: I’ve looked into this a bit more and the best option as far as performance goes is to use a switch statement. So I’d like to leave it as is.

This article goes into some of the pros and cons which I found interesting: https://medium.com/front-end-weekly/switch-case-if-else-or-a-lookup-map-a-study-case-de1c801d944

Armanio (approved):

addons/a11y/src/register.tsx
```diff
@@ -20,56 +20,56 @@ const PreviewWrapper: FunctionComponent<{}> = p => (
...
-          <filter id="protanopia">
+          <filter id="Protanopia">
             <feColorMatrix
               in="SourceGraphic"
               type="matrix"
               values="0.567, 0.433, 0, 0, 0 0.558, 0.442, 0, 0, 0 0, 0.242, 0.758, 0, 0 0, 0, 0, 1, 0"
             />
           </filter>
-          <filter id="protanomaly">
+          <filter id="Protanomaly">
```
shilman: why capitalize?

addons/a11y/src/components/ColorBlindness.tsx
```diff
@@ -72,15 +72,15 @@ export class ColorBlindness extends Component<ColorBlindnessProps, ColorBlindnes
...
-      'mono',
+      'Protanopia',
+      'Protanomaly',
+      'Deuteranopia',
+      'Deuteranomaly',
+      'Tritanopia',
+      'Tritanomaly',
+      'Achromatopsia',
+      'Achromatomaly',
+      'Mono',
```
shilman: Just upcase the first letter instead?

shilman (approved):
LGTM!

addons/a11y/src/register.tsx
CodeByAlex: You have a good point, ids are normally lowercase. The id is used to check against the list of the names of the filters which are now capitalized so I switched it. However,  I could just as easily do to lowercase and check against the id

addons/a11y/src/components/ColorBlindness.tsx
CodeByAlex: Actually, like this idea better because it has the least impact so I made that change instead.

addons/a11y/src/components/ColorBlindness.tsx
```diff
@@ -85,9 +85,9 @@ export class ColorBlindness extends Component<ColorBlindnessProps, ColorBlindnes
       id: i,
```
shilman: is this `id` used anywhere?

shilman: suggest:
```js
id: i.toLowerCase()
...
  this.setFilter(this.id)
```

CodeByAlex: after I modified the array of types, its still lowercase and only the first char is capitalized in the title. Not sure if the id is being used but I can check.

ndelangen:
Wow, this takes the a11y addon from 'useful' to 'amazing', this is fantastic @CodeByAlex !

PR #6570, opened 2019-04-20
Migrate @storybook/angular to typescript

kroeder:


app/angular/src/server/__tests__/angular-cli_config.test.ts
```diff
@@ -1,5 +1,5 @@
-import { getAngularCliWebpackConfigOptions } from './angular-cli_config';
-import angularJson from '../../../../examples/angular-cli/angular.json';
+import { getAngularCliWebpackConfigOptions } from '../angular-cli_config';
```
igor-dv: just checking what is happening around, do we have some new convention to put tests in the `__tests` dir ?

kroeder: We don't. I just wanted to bring some order in the server directory. 

image

Is it worth to caring about this? I don't know. I can leave it as is 🙂

app/angular/package.json
```diff
@@ -46,9 +49,12 @@
     "@angular/forms": ">=6.0.0",
     "@angular/platform-browser": ">=6.0.0",
     "@angular/platform-browser-dynamic": ">=6.0.0",
+    "@angular-devkit/build-angular": ">=6.0.0",
     "autoprefixer": "^8.1.0",
     "babel-loader": "^7.0.0 || ^8.0.0",
     "rxjs": "^6.0.0",
+    "typescript": "^3.4.4",
```
ndelangen: Do we REQUIRE the .4 patch version? if not, try to keep dependencies to `.0` patch version for maximum compatibility.

Our lockfile will make sure we'll actually use the latest patch.

If we keep these version rounded down to the lowest patch, it's less likely users will end up with multiple versions installed.

ndelangen (approved):

app/angular/package.json
kroeder: Done

app/angular/src/server/__tests__/angular-cli_config.test.ts
ndelangen: I like colocation of tests, until the folder gets like 6+ files, then having a `tests` folder is kinda nice. I don't care enough about it either way really. 👍

PR #6571, opened 2019-04-20
feat(storiesNav): adding classNames to stories nav elements

mbellary-chwy:
Feature:

adding classNames to stories nav elements

## What I did

- adding classNames to stories nav elements
- updating LogoLink to support html elements

This PR is a replacement for the deep theming PR:

https://github.com/storybooks/storybook/pull/6098

### DIFF
--- a/lib/ui/src/components/sidebar/SidebarItem.js
+++ b/lib/ui/src/components/sidebar/SidebarItem.js
@@ -111,9 +111,19 @@ export default function SidebarItem({
   }
 
   return (
-    <Item isSelected={isSelected} depth={depth} loading={loading} {...props}>
-      <Expander isExpandable={!isLeaf} isExpanded={isExpanded ? true : undefined} />
-      <Icon icon={iconName} isSelected={isSelected} />
+    <Item
+      isSelected={isSelected}
+      depth={depth}
+      loading={loading}
+      {...props}
+      className={isSelected ? 'sidebar-item selected' : 'sidebar-item'}
+    >
+      <Expander
+        className="sidebar-expander"
+        isExpandable={!isLeaf}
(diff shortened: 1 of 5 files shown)


lib/ui/src/components/sidebar/SidebarItem.js
```diff
@@ -111,9 +111,19 @@ export default function SidebarItem({
...
   return (
-    <Item isSelected={isSelected} depth={depth} loading={loading} {...props}>
-      <Expander isExpandable={!isLeaf} isExpanded={isExpanded ? true : undefined} />
-      <Icon icon={iconName} isSelected={isSelected} />
+    <Item
+      isSelected={isSelected}
+      depth={depth}
+      loading={loading}
+      {...props}
+      className={isSelected ? 'sidebar-item selected' : 'sidebar-item'}
```
ndelangen: I'm ok, with this, though I think it'd be safer to encapsulate the adding of these classNames to the components. That way we can just use them, and we don't have to worry about adding these classnames on the jsx.

WDYT?

ndelangen (approved):
Please consider my comment above, if you think it's worth it to encapsulate the adding of classnames.

ndelangen:
Looks like the CI doesn't like the branchname, but all else looks ok:
https://circleci.com/gh/storybooks/storybook/119943

lib/ui/src/components/sidebar/SidebarItem.js
MansoorBashaBellary: I tried on adding classNames to components when i started working on this change, but i overcame some issues of initial styles not getting applied. So, i went ahead and applied classNames this way, which was not breaking in any of the cases.

PR #6594, opened 2019-04-23
Ondevice actions

ForbesLindesay:
Issue:

## How to test

- Is this testable with Jest or Chromatic screenshots? - maybe
- Does this need a new example in the kitchen sink apps? - I've added it to crna
- Does this need an update to the documentation? - I've updated ADDONS_SUPPORT

It looks like this:

Screenshot 2019-04-23 at 14 10 59
Screenshot 2019-04-23 at 14 11 29


shilman:
@ForbesLindesay awesome awesome awesome 💯 this is really exciting!

addons/ondevice-actions/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "types": ["webpack-env"]
```
benoitdion: this may not be needed

addons/ondevice-actions/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "bugs": {
+    "url": "https://github.com/storybooks/storybook/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/storybooks/storybook.git"
+  },
+  "license": "MIT",
+  "main": "dist/index.js",
+  "jsnext:main": "src/index.js",
```
benoitdion: since this package is typescript only you shouldn't need this.

addons/ondevice-actions/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "scripts": {
+    "prepare": "node ../../scripts/prepare.js"
+  },
+  "dependencies": {
+    "@storybook/addons": "5.1.0-alpha.29",
+    "@storybook/core-events": "5.1.0-alpha.29",
+    "core-js": "^2.5.7",
+    "fast-deep-equal": "^2.0.1",
+    "global": "^4.3.2",
+    "lodash": "^4.17.11",
```
benoitdion: are all these dependencies used by your addon?

addons/ondevice-actions/package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "peerDependencies": {
+    "@storybook/addon-actions": "*",
+    "react": "*",
+    "react-native": "*"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "gitHead": "fbd7ee4c80df437fed4bdc6e11140733fd450080"
```
benoitdion: this shouldn't be required

benoitdion:
hey @ForbesLindesay, this is awesome! A few minor comments but looks really good overall. 

Do you mind adding a story in the crna-kitchen-sink app to showcase usages of this new addon?

ForbesLindesay:
@benoitdion Thanks for the feedback. I've hopefully resolved all of it now.  I had committed but forgotten to push the update to crna-kitchen-sink. It doesn't actually need a new example as the existing examples already have actions, so simply adding the new ondevice-actions addon does the trick.

benoitdion:
Thanks for taking the time to address the feedback @ForbesLindesay! It looks like your build is failing, you may need to run yarn. 

I took it out for a spin and it's working great but the ui looks a bit different for me:
Screenshot_1556106111

do you know what might be going on with the expand/collapse buttons? Also, after expanding a coule actions the clear button disappears. It may need to be part of the ScrollView or sticky at the bottom.

What do you think?

ndelangen:
WOW awesome contribution @ForbesLindesay

@benoitdion looks like RN is really getting the attention it deserves, are you able to handle this as well?

ndelangen:
I'll fix the yarn lock situation

ForbesLindesay:
> After expanding a couple actions the clear button disappears.

It's in a scroll view, so this works fine for me. I only tested on iOS though

> do you know what might be going on with the expand/collapse buttons?

I only tested on iOS, I think probably the `Button`s are styled differently on the different platforms, probably we should be using something like `TouchableHighlight` instead.

PR #6598, opened 2019-04-23
Check for color knob value existence before applying uppercase

### LINKED ISSUE #6597: Having no default value for color knob breaks
**Describe the bug**
When there is no default value for a color knob, there is a fatal error of:
`TypeError: Cannot read property 'toUpperCase' of undefined`

Here is the source of this error:
https://github.com/storybooks/storybook/blob/c6d704225eec98c20a729242f14987852071ddb1/addons/knobs/src/components/types/Color.js#L94

**To Reproduce**
1. In a story, create a color knob with no default value:
```
.add('Overview', () => {
    const iconColor = color('Icon Color')
    return <Component iconColor={iconColor} />
  })
```
2. Navigate to this story

**Expected behavior**
The color knob should not require a default value. (This didn't break prior to be822727e10893d4853639b2e0bdb4abaef815f2)

This is important because some components may have a color prop that is only used when non-falsy.

**Possible Solution**
Check if `knob.value` exists before using `toUpperCase()`.

binomialstew:
Issue: https://github.com/storybooks/storybook/issues/6597

## What I did

Fixes https://github.com/storybooks/storybook/issues/6597

## How to test

- Remove the default color value:
https://github.com/storybooks/storybook/blob/c6d704225eec98c20a729242f14987852071ddb1/examples/official-storybook/stories/addon-knobs.stories.js#L65
- Storybook should not break

`["bug"]`

### DIFF
--- a/addons/knobs/src/components/types/Color.js
+++ b/addons/knobs/src/components/types/Color.js
@@ -82,7 +82,6 @@ class ColorType extends React.Component {
     const colorStyle = {
       background: knob.value,
     };
-
     return (
       <ColorButton
         active={displayColorPicker}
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM! Thanks for the fix! 🥇

PR #6634, opened 2019-04-25
Handle loading `.storybook/babel.config.js` (#6633)

lukeaskew:
#6633


shilman (approved):
Great fix! Did you test it?

@ndelangen are there any good places to drop in tests for this?

ndelangen:
@LukeAskew Would you be able to see what we can do to fix the failing cli test?
https://circleci.com/gh/storybooks/storybook/121204

LukeAskew:
Sure, but I could use a little guidance.

1. These same cli tests fail locally when I run at the HEAD of `next`. I think this is because the babel configs in these fixtures don't include `preset-react`. I am unsure of why these pass on Circle.
1b. What is the intended behavior of test cases like [this](https://github.com/storybooks/storybook/blob/d0564a2d658cd5e65dd2dcdaa8ead658bacd15b3/lib/cli/test/fixtures/react_babel_config_js/babel.config.js)? I would expect this config to be used instead of the default and I would expect this to fail because it lacks the necessary presets.
2. How can I link my changes in `core` to the cli test? It appears that the fixtures pull directly from npm. It is unclear how I can test my changes across the repo.

igor-dv:
AFAIR, The similar fix was already considered some time ago. One problem is that it might be a breaking change. SB could start requiring babel configs from root, that are existing in the project and where not required before. 

Also some discussion was here - #2582

LukeAskew:
Updated the code to just handle `.storybook/babel.config.js` as well. This allows folks to piggyback on the root config via:

```js
module.exports = require('../babel.config');
```

ndelangen (approved):

PR #6669, opened 2019-04-28
Bugfix: Have no cancel option in UI if the context have no param

leoyli:
Issue: #6668 

## What I did
- [bugfix] Have no cancel option in UI if the context have no param.
- [type] enable more strict typing with `--strictNullChecks` (@`addon-contexts` level).
- [chore] simplify to make code more readable.

## How to test
N/A


shilman (approved):
🔥

PR #6688, opened 2019-04-30
Fix regression with deep linking

tmeasday:
Issue: #5925

The issue was the `iframe` did not update when its `src` was changed as we navigated to the story id.

The reason it sometimes worked depended on whether the preview had set up its channel in time (the event *was* emitted).

## What I did

Ensured the preview's URL changed.

## How to test

http://localhost:9011/?selectedKind=UI%7CPanel&selectedStory=no+panels

### DIFF
--- a/lib/ui/src/components/preview/iframe.js
+++ b/lib/ui/src/components/preview/iframe.js
@@ -12,8 +12,8 @@ export class IFrame extends Component {
   }
 
   shouldComponentUpdate(nextProps) {
-    const { scale } = this.props;
-    return scale !== nextProps.scale;
+    const { scale, src } = this.props;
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM - fixed the same bug in the Docs work 😭

ndelangen (commented):
This will cause refreshes in the preview, as any change the an iframe's src will cause it to refresh

tmeasday:
We should probably revert this PR and figure out what is wrong with the existing buffer in the postmessage channel

tmeasday:
Also we should make it more explicit in the `IFrame` component that `src` should not change

PR #6692, opened 2019-04-30
Use `http-server` instead of `serve` in the example

Hypnosphi:
Unfortunately, current version of `serve` doesn't work without [configuration](https://github.com/zeit/serve/issues/393#issuecomment-395978195), as it removes the query string from iframe URL


shilman (approved):
WOW well that explains that. Thanks @Hypnosphi !! 🎖

PR #6695, opened 2019-04-30
Update how we install dev dependencies in cli

MikaelCarpenter:
Issue: #4805 

## What I did
_Scoped to `lib/cli/`_
- add a new option, `installAsDevDependencies`, to `npmOptions` in `initiate.js`
- update `installDependencies` helper function to use that new option
  - update usage of this function to not pass `'-D'` in list of dependencies anymore

Relates to a comment from a previous PR: [comment](https://github.com/storybooks/storybook/pull/6629#discussion_r278770095)

## How to test
`yarn test --cli`

- While this is testing, it might be good to look into `test/run/` and look at the `package.json`s to ensure dependencies are being installed as expected
  - a good candidate to look at is the meteor fixture since it tries to install normal dependencies and dev dependencies
- If it helps you can comment out `trap cleanup EXIT` in `run_tests.sh` to avoid deleting the `run` directory after testing

### DIFF
--- a/lib/cli/generators/ANGULAR/index.js
+++ b/lib/cli/generators/ANGULAR/index.js
@@ -46,7 +46,6 @@ async function addDependencies(npmOptions) {
     `@storybook/addon-links@${linksVersion}`,
     `@storybook/addons@${addonsVersion}`,
     ...babelDependencies,
-    '-D',
   ]);
 }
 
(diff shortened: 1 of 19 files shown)


shilman (approved):
Thanks for the follow-up @MikaelCarpenter. Looks great!

PR #6696, opened 2019-05-01
fix: Preserve dimensions on resizing for panel

kohakukun:
Issue: https://github.com/storybooks/storybook/issues/6343

## What I did

- Update resizer position when changing the boundaries of the container
for the addon panel and preview

## How to test

storybookPR

### DIFF
--- a/lib/ui/src/components/layout/container.js
+++ b/lib/ui/src/components/layout/container.js
@@ -1,3 +1,5 @@
+/* eslint-disable react/no-did-update-set-state */
+
 import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { styled, withTheme } from '@storybook/theming';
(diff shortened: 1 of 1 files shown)


kohakukun:
@shilman PTAL. 

P.S.: Not sure what to do about the eslint rule. Should we use another solution?

lib/ui/src/components/layout/container.js
```diff
@@ -285,13 +287,35 @@ class Layout extends Component {
...
     const { resizerPanel, resizerNav } = this.state;
 
     persistance.set({
       resizerPanel,
       resizerNav,
     });
+    const { width: prevWidth, height: prevHeight } = prevProps.bounds;
+    const { bounds, options } = this.props;
+    const { width, height } = bounds;
+    if (width !== prevWidth || height !== prevHeight) {
```
igor-dv: Why can't this block be done in the `getDerivedStateFromProps` ?

kohakukun: I want to move the resizer by the delta of boundaries. I could move it to getDeriverStateFromProps by changing the resizerPanel state to store width and height instead of a position. What do you think?

igor-dv: yeah, It may work I think.

ndelangen (approved):

ndelangen:
Thank you @kohakukun !

Aaron-Pool:
@kohakukun  OH MY GOSH THANK YOU. This was driving me _crazy_.

PR #6707, opened 2019-05-02
Remove teamcity CLI tests

benoitdion:
The teamcity CLI tests have been failing on Teamcity but seem just fine on circleci. Since we have duplicate coverage and the failure seem to be environmental, remove the teamcity job.

Example failure: https://teamcity.jetbrains.com/viewLog.html?buildId=2235626&buildTypeId=OpenSourceProjects_Storybook_CliTest


shilman (approved):
Not sure how all this works, but will merge!

PR #6745, opened 2019-05-08
Typo in svelte stories example (non-valid JS)

thollander:
## What I did
Fix non-valid JS example in svlete stories example.

## How to test
The documentation example should now be correct.


shilman (approved):
Great 💯

PR #6750, opened 2019-05-09
Fix duplicate key in svelte guide

thollander:
Issue:
- In storybook, the emojies example erases the first example (same key)in svelte docs
- Implies that the example contains only one example instead of two

## What I did
Change the key of the emojies example to be compliant with the other examples

## How to test
When running the examples in Storybook, we should get two stories instead of one.


shilman (approved):
🥇

PR #6755, opened 2019-05-09
Add documentation on searching and "notes"

### LINKED ISSUE #6483: Feature request for greater discoverability in large storybooks

I work for an organization that has hundreds of apps and websites. We have helped migrate many of them to use storybook and will continue to do so going forward. Storybook is fantastic and meets our needs well for the most part. But one area it struggles in is in projects that have many components (some of them hundreds) it is hard to find specific components through folders and titles + searching alone.

It would be very helpful to be able to be able to include keywords that pull up specific stories from searching; so keywords that are not in the story title could be included in search.

We've tried to organize stories the best we can using the nested folder structure. This works well for components that only have one function. But many components have multiple functions, so a 1 to 1 ratio of discoverability doesn't work well.

Possibly, if I could get a bit of direction from a storybook admin :) But with my current workload, having the storybook team implement would be helpful.

trevordmiller:
Resolves https://github.com/storybooks/storybook/issues/6483

### DIFF
--- a/docs/src/pages/basics/writing-stories/index.md
+++ b/docs/src/pages/basics/writing-stories/index.md
@@ -157,6 +157,22 @@ storiesOf('Component', module).add('With Markdown', () => <MyComponent />, {
 });
 ```
 
+## Searching
+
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM. Thanks @trevordmiller

PR #6761, opened 2019-05-10
Corrected failing example in Vue guide

ffxsam:
This example failed out of the box. Registering every time is necessary, apparently

Issue:
Example in Vue guide threw errors on fresh install

## What I did
Added component registration in every `.add()`

### DIFF
--- a/docs/src/pages/guides/guide-vue/index.md
+++ b/docs/src/pages/guides/guide-vue/index.md
@@ -132,8 +132,14 @@ import { storiesOf } from '@storybook/vue';
 import MyButton from './Button.vue';
 
 storiesOf('Button', module)
-  .add('with text', () => '<my-button>with text</my-button>')
-  .add('with emoji', () => '<my-button>😀 😎 👍 💯</my-button>')
+  .add('with text', () => ({
+    components: { MyButton },
+    template: '<my-button>with text</my-button>'
+  }))
+  .add('with emoji', () => ({
+    components: { MyButton },
+    template: '<my-button>😀 😎 👍 💯</my-button>'
+  }))
   .add('as a component', () => ({
     components: { MyButton },
     template: '<my-button :rounded="true">rounded</my-button>'


shilman:
@backbone87 @elevatebart can you confirm? seems like a serious regression if this is correct...

ndelangen (approved):

ndelangen:
@shilman No the option for a single string was removed at some point in the past.

These are the option currently available for vue users.
https://github.com/storybooks/storybook/blob/next/examples/vue-kitchen-sink/src/stories/custom-rendering.stories.js#L9-L92

AFAIK, this should work with globally register components:
```js
  .add('template', () => ({
    template: `
      <div>
        <h1>A template</h1>
        <p>rendered in vue in storybook</p>
      </div>`,
  }))
```

elevatebart:
I believe the component MyButton is registered globally which makes these changes normally unnecessary. 

I will check this immediately

elevatebart:
If it is the case, the documentation needs urgent update. I need to check.

Bart Ledoux
Twitter: @bartledoux
Github, Gitter, Discord: @elevatebart

ndelangen:
Did you find if this is necessary @elevatebart?

elevatebart:
Not yet, I will check this in the coming hours.

Bart Ledoux
Twitter: @bartledoux
Github, Gitter, Discord: @elevatebart

elevatebart:
I just checked on `next` branch and all seems fine. 
No need to update the docs. 
The pure string story is still working fine in 5.1.0-rc.0.

@ffxsam you seem to have missed - so had I for the record - is that [in the guide]https://github.com/storybooks/storybook/blob/next/docs/src/pages/guides/guide-vue/index.md#step-3-create-the-config-file(), vue components are meant to be registered globally in the `config.js` file. It is only mentioned in the details sections. Maybe we should clarify this. 

Does it make more sense now?

Capture d’écran 2019-05-23 à 07 32 02

ndelangen:
Thanks for all your detective work 🕵 @elevatebart 👍

ndelangen:
@ffxsam thank you for helping! Super appreciated

ffxsam:
> I just checked on `next` branch and all seems fine.
> No need to update the docs.
> The pure string story is still working fine in 5.1.0-rc.0.
> 
> @ffxsam you seem to have missed - so had I for the record - that [in the guide](https://github.com/storybooks/storybook/blob/next/docs/src/pages/guides/guide-vue/index.md#step-3-create-the-config-file), vue components are meant to be registered globally in the `config.js` file. It is only mentioned in the details sections. Maybe we should clarify this.

Ah, I did see this, but this intention was not clear. It states:

>You might be using global components or vue plugins such as vuex, in that case you'll need to register them in this config.js file.

I would consider global components to be from plugins or other third party libraries such as Vuetify and Buefy. I don't consider my own internal components to be global at all (as I `import` them when necessary).

backbone87:
we should update the storybook vue docs and explain what i wrote here: https://github.com/storybookjs/storybook/issues/6357#issuecomment-478514130

storybook adds no magic regarding component registration

Issue #6623, opened 2019-04-25
Force rerender with storybook/html


hsablonniere:
Hey wonderful team of storybook :smile:

I'm trying to force rerender my stories when a knob is changed with storybook/html but it does nothing when story is an HTML element.

I looked at the code, knob changes trigger a rerender so it should be OK but when the code execution arrives in `renderMain` if it sees `forceRender === true` it does nothing :cry:

I'm probably missing the point here but reading this, I expected a force render to clear `rootElement` and reattach the node. The opposite of what it does right know.

https://github.com/storybooks/storybook/blob/master/app/html/src/client/preview/render.js#L20-L25

* Can someone tell me more about this? @Atekon 
* Maybe I'm the only one expecting this since I use custom elements and a detach/reattach would trigger a rerender of my component.

JohnMunsch:
I think I'm seeing the same thing when my object knob supplies new data to create a Redux model and assigns that to my LitElement component. I can see it go through the story, create a new model, create a new component, assign the model to the component, and then return the component. However, the rendered component shown in Storybook is still the previous one.

Hypnosphi:
@igor-dv @Atekon looks like #4822 broke knobs for "HTML element" stories

hsablonniere:
Thanks very much!!! :tada:

PR #6783, opened 2019-05-14
HTML: support knobs for both cached and uncached nodes

### LINKED ISSUE #5017: Stories that return a DOM element don't update when knobs change

Hypnosphi:
Issue: #6190 undid #4822

## What I did
Used reference comparison to fix #5017 and #6623 while keeping the effect of #4822

## How to test
– Open Knobs/DOM story, check that the knob still works
– Open Knobs/CSS transitions story, observe transitions when changing color

### DIFF
--- a/app/html/src/client/preview/render.js
+++ b/app/html/src/client/preview/render.js
@@ -12,14 +12,14 @@ export default function renderMain({
   showError,
   forceRender,
 }) {
-  const { html = {} } = parameters;
   const element = storyFn();
 
   showMain();
   if (typeof element === 'string') {
     rootElement.innerHTML = element;
   } else if (element instanceof Node) {
-    if (html.preventForcedRender === true && forceRender === true) {
+    // Don't re-mount the element if it didn't change and neither did the story
+    if (rootElement.firstChild === element && forceRender === true) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


shilman (approved):
LGTM. 👍  Will wait for another review to merge!

ndelangen (approved):

PR #6794, opened 2019-05-16
Scrollbar supports theming again

domyen:
Issue: #6753

Core issue is #6753. @ndelangen's #6785 reimplements simplebar styles. This PR adds theming support back in.

## What I did
Add theming support back in

## How to test
- Change to dark theme and see white scroll bar
- Change to light theme and see dark scroll bar

Screen Recording 2019-05-16 at 09 21 AM
Screen Recording 2019-05-16 at 09 20 AM


shilman (approved):
@domyen we should probably fix theming so we don't need to hardcode it like this. Merging for now so I can release.

PR #6795, opened 2019-05-16
typo fix in ember setup guide

gokatz:
This seems to be a typo in the setup guide of the Ember storybook.

 Also, in the particular line, it is mentioned that the `hbs` file has to be imported from the storybook babel plugin. But in the sample code, `has` was imported from `htmlbars-inline-precompile`. Is that intentional?


shilman (approved):
HAHA thanks! 🙇

PR #6815, opened 2019-05-17
ISS-6802 Fix issue with init in create-react-library projects

MikaelCarpenter:
Issue: #6802 

There's two causes to this issue: 
- `create-react-library` puts `react-scripts` in devDeps
- `semver.gtr()` fails if you pass it `undefined`

## What I did
- In the `REACT_SCRIPTS` generator, check both dependencies and devDependencies for `react-scripts`
- Also, make sure it exists before trying to use `semver`

## How to test

1. npm install -g create-react-library
2. create-react-library -s -t default test
3. cd test
4. npx -p @storybook/cli@next sb init

I tested init before and after the changes and looked good to me 👍


ndelangen (approved):

shilman (approved):
Makes sense!

PR #6818, opened 2019-05-18
CHANGE lib/theming so it no longer depends on react-inspector

ndelangen:
Issue: https://github.com/storybooks/storybook/issues/6764

## What I did
COPY the chromeDark & chromeLight variables into the module itself


ndelangen:
This adds the code for theming the syntaxhighlighter into lib/theming instead of pulling it from react-inspector.

We might want to create packages for these themes, so we can share this code again.

If anyone would be interested in creating those packages?

ndelangen:
@beauroberts could you help us review this?

beauroberts (approved):
LGTM, although I suspect the actual source of the error I encountered might reside in `addon-actions`:

[`ActionLogger`](https://github.com/storybooks/storybook/blob/next/addons/actions/src/components/ActionLogger/index.tsx#L4) imports `Inspector` from `storybooks/react-inspector`. The different inspector implementations (`DOMInspector`, `ObjectInspector`) import [`TreeView`](https://github.com/storybooks/react-inspector/blob/b542e8c1562f85fdc4b84e547586ab838c07e4ed/src/tree-view/TreeView.js#L6), which makes use of hooks and `memo`, both unavailable in React < 16.6, (or 16.8 in the case of hooks).

But this will probably sort my issue if I remove use of the Actions addon. Thank you!

shilman (approved):

PR #6825, opened 2019-05-19
(feature): Support loglevel for production/ci builds.

kaidjohnson:
Issue:

In order to better support various ci and package management workflows, it would be nice to have the ability to suppress some (or all) of build-storybook's logging output. Given Storybook uses npmLog, and npmLog has a feature to support this functionality, it's simple to surface that functionality to the cli through the addition of a `--loglevel` argument.

## What I did

Surfaced cli argument `--loglevel` for build-storybook to hook into npmlog's level setting.

## How to test

Maybe? Does Storybook show examples for all build options? It wasn't immediately clear to me.
Yes. The cli-options index.md is updated in this PR to reflect the change.

### DIFF
--- a/lib/core/src/server/cli/prod.js
+++ b/lib/core/src/server/cli/prod.js
@@ -13,10 +13,12 @@ function getCLI(packageJson) {
     .option('-c, --config-dir [dir-name]', 'Directory where to load Storybook configurations from')
     .option('-w, --watch', 'Enable watch mode')
     .option('--quiet', 'Suppress verbose build output')
+    .option('--loglevel [level]', 'Control level of logging during build')
     .option('--no-dll', 'Do not use dll reference')
     .option('--debug-webpack', 'Display final webpack configurations for debugging purposes')
     .parse(process.argv);
 
+  logger.setLevel(program.loglevel);
   logger.info(chalk.bold(`${packageJson.name} v${packageJson.version}\n`));
 
   // The key is the field created in `program` variable for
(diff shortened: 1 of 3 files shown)


lib/core/src/server/cli/prod.js
```diff
@@ -13,10 +13,12 @@ function getCLI(packageJson) {
     .option('-c, --config-dir [dir-name]', 'Directory where to load Storybook configurations from')
     .option('-w, --watch', 'Enable watch mode')
     .option('--quiet', 'Suppress verbose build output')
+    .option('--loglevel [level]', 'Control level of logging during build')
     .option('--no-dll', 'Do not use dll reference')
     .option('--debug-webpack', 'Display final webpack configurations for debugging purposes')
     .parse(process.argv);
 
+  logger.setLevel(program.loglevel);
```
ndelangen: What's the level being set to now, when the cli argument isn't given?

Would wrapping this with an if-statement make sense?

ndelangen:
Thanks @kaidjohnson!

lib/core/src/server/cli/prod.js
kaidjohnson: I set a default value in the node-logger index.ts to 'info' to avoid an if wrapper.

Previously, no default was set, but node-logger doesn't wrap beyond 'info', so it's a safe value to support current functionality while supporting long-term best-practice of ensuring debugging output is protected from users who don't explicitly ask for it.

ndelangen (approved):

PR #6828, opened 2019-05-19
Presets: Ability to combine presets into another preset

igor-dv:
Issue: We want to be able to combine presets into another custom preset file, so it will be easier to customise them.

## What I did
Currently we support preset file to be something like this:

```js
export async function webpack() { }
export async function babel() { }
```

or 

```js
function webpack() {}
function babel() {}

module.exports = {
  webpack,
  babel,
}
```

In both options an **object** is exported from the module.

Now, we can export an array, which means we are exporting a set of presets out of preset (yo dawg)


shilman (approved):
Pretty nice -- great work! 👍 

Let's merge this later for 5.2

joeycozza:
@igor-dv 
I have a question about a potential scenario with a preset of presets.

Let's say I'm making a preset-foo that uses preset-typescript and preset-knobs under the hood.

Now app Bar is using my preset-foo in their .storybook/presets.js file.

Question 1) Does the Bar app still need to do the following? If not, what is the mechanism to use addDecorator within a preset? 
```javascript
//Inside the Bar app's .storybook/config.js file

import { addParameters, addDecorator } from '@storybook/react'
import { withKnobs } from '@storybook/addon-knobs'
addDecorator(withKnobs)
```
Question 2) The Bar app will want to utilize knobs like `boolean`, `text`, etc. Does preset-foo need to do an `export * from '@storybook/addon-knobs` in order to expose them to the Bar app? Or is the Bar app expected to have @storybook/addon-knobs as a direct dependency and access/use it directly in their .stories files?

Thanks for your help and clarification!

igor-dv:
@joeycozza , sorry for the delay.

1) `addDecorator` could be used in preset if preset extends the `config` and `addons` entry point.
something like this:
```js
// this will customise the preview entry point
export function config(entry = []) {
  return [
    require.resolve('./path/to/knobs-config.js'),
    // "entry" here will contain a config.js from the .storybook dir, 
    // so you need to configure before or somehow skip it. 
    // maybe we need to have a preset for that =)
    ...entry, 
  ] 
}

// this is an alternative for "addons.js"
export function addons(entry = []) {
  return [
    ...entry, 
    require.resolve('@storybook/addon-knobs/register')
  ];
}
```

where the `knobs-config.js` will be something like this:

```js
import { addDecorator } from '@storybook/react'
import { withKnobs } from '@storybook/addon-knobs'
addDecorator(withKnobs)
```

2) Here could be 2 possibilities:
a. The addon (for example addon-knobs) will have a preset for itself, so you will need to install it anyway to use preset and to use what it exports.
b. We have a different preset that needs to have a peer dep to the `addon-knobs` in order to not mess with the dependencies hell, so you will need to install it anyway.
 
Does it answer your question ?

shilman (approved):
👍

ndelangen:
I'm working on this to be inline with #8597

ndelangen:
presets can now export presets like this:

```js
// preset.js
module.exports = ['preset1', 'preset2'];
```

but also like so:

```js
// preset.js
module.exports = {
  presets: ['preset1', 'preset2'],
  otherValue: true,
};
```

ndelangen (approved):

PR #6847, opened 2019-05-21
Feature/jest design changes

CodeByAlex:
Issue: https://github.com/storybooks/storybook/issues/6805

## What I did
I moved the name of the test class above the progress bar, moved the passing percentage below the progress bar, and increased contrast in the error message box.

Original:
image

New:
image

## How to test


ndelangen:
Feel free to change it in more radical ways, and extracting the re-usable component into `lib/components` 👍

ndelangen:
It would also be great to actually have stories for this component, so we don't have to paste screengrabs, but can approve with chromatic 👍

PR #6854, opened 2019-05-21
Fix documentation for themeing

ColinRTaylor:
Issue: The docs have the wrong path, I updated it to use the correct one.

## What I did
Fixed the docs to use the right path after I figured out the correct one.

## How to test


ndelangen:
Thank you!

I'll merge this asap

PR #6884, opened 2019-05-27
Call onChanges() only for changed props

thomashilzendegen:
Issue:
The lifecycle hook `ngOnChanges` is called for every input, regardless if it was changed or not. Also, the `firstChange` property is determined from the instance's value (if it is still `undefined`) instead of the first real value change.

## How to test


kroeder (approved):
LGTM

kroeder:
had to change the `hasOwnProperty` check. See https://eslint.org/docs/rules/no-prototype-builtins for details

PR #6907, opened 2019-05-29
make it compatible with new version of read-pkg-up

mzdunek93:
Issue: storyshots doesn't work due to incompatibility with the newest version of the `read-pkg-up` module.

## What I did

Modified properties read from the object returned by the module.


shilman (approved):
Thanks for the fix! Will get it out on the next RC 🙇

PR #6921, opened 2019-05-31
PanelTabOptions WIP

JoshRosenstein:
Issue: 
no options to customize panel tabs. similar to `showPanel`, allow user to config Panel tabs globally or per story.
## What I did
- added two options to layout api, `hiddenTabs` & `tabsOrder` (names up for discussion), in order to customize panel tabs.
- Add sorter method within panel component

***Note, having trouble building (on windows), will rely on CI for now , therefore this PR may be temporary.
 
## How to test

maybe?
yep


lib/ui/src/components/panel/panel.js
```diff
@@ -10,6 +10,20 @@ const DesktopOnlyIconButton = styled(IconButton)({
   },
 });
 
+const tabSorter = (panels, panelOrder) => {
```
tmeasday: Could we pull this out into a separate lib file and and a couple of simple tests? (Perhaps coupled with the hidden logic too?)

tmeasday (commented):
This is great! I think we'll need to solve #5857 before it can really be used however

lib/ui/src/components/panel/panel.js
JoshRosenstein: @tmeasday  Sure thing.

ndelangen:
I think this feature is the same that was recently added by @atanasster.

PR #6923, opened 2019-05-31
Addons: Disable option for addon tab

jsomsanith-tlnd:
Issue: #6849

## What I did
The issue above ask for a feature to disable a11y panel for a single story. But the idea of this PR is to provide a solution at a common level instead of at addon level.

The principle : 
* addons register their tab panel, passing the parameter key
* user can then disable a panel for a group of stories or a specific story
```javascript
storiesOf('My Component', module)
  // disable the a11y tab for all "My Component" stories
  .addParameters({
    a11y: { disabled: true },
  })
  .add('story 1', story1)
  .add('story 2', story2);
```

```
storiesOf('My Component', module)
  // disable the a11y tab for only "story 1"
  .add('story 1', story1, {  a11y: { disabled: true } })
  .add('story 2', story2);
```

## How to test

- Does this need a new example in the kitchen sink apps? Yes, a `Addon panel` example has been added
- Does this need an update to the documentation? Yes, it's a common addon behavior, but addons need to pass the param key now. Documentation has been added.


tmeasday:
This seems reasonable to me

jsomsanith:
Thanks @tmeasday, I'll work on the doc/test/kitchen sink

lib/ui/src/containers/panel.js
```diff
@@ -10,8 +10,25 @@ const createPanelActions = memoize(1)(api => ({
...
+      filteredPanels[id] = panel;
+    }
+  });
+
+  return filteredPanels;
+};
+
 const mapper = ({ state, api }) => ({
-  panels: api.getPanels(),
+  panels: filterPanel(api.getPanels(), api.getParameters(state.storyId)),
```
ndelangen: should `getFilteredPanels` be an api?
should `getPanels` just return the filtered list

We should make sure this is memoized

jsomsanith: I added a `getPanelForStory(storyParameters)` in the addon api.
Since the `api.getParameters()` is in the story api, it feels wrong to use it in the addon api. My first thought is that those are separate modules.

ndelangen: The modules addons, layout, stories are all interconnected.

It's OK to call methods on the final / complete api in methods of the SubAPI.

It's probably not a best practice or anything, but duplicating code and making work arounds so "modules" **appear** self-contained when in fact they aren't really is maybe even worse?

jsomsanith: You're right, and I haven't noticed it but the addon api already use other apis.
Changed it :)

ndelangen:
@jsomsanith there seems to be a runtime error in the build versions, could you check it out?

Please ask me for help, if you need it, happy to pair on it if you'd like.

ndelangen (approved):

ndelangen:
Awesome, does anyone else want to review before I merge?

jsomsanith:
It’s been 15 days, when this PR will be merged ? :)

shilman (approved):
SOLID 👊

Issue #6756, opened 2019-05-09
Clicking Shortcuts Button Opens Blank Screen


scottleblang:
**Describe the bug**
When clicking the Storybook 3-dot shortcuts button in the left panel, rather than open a menu of options for Storybook, the page refreshes to a blank screen . This same behavior also occurs if I press the 'Change the size of the preview' button along the top header panel.

**Expected behavior**
A menu should open next to the Shortcuts button showing selectable options -see attached screenshot of expected behavior.

**Error Message**
Uncaught Error: Couldn't parse the color string. Please provide the color as a string in hex, rgb, rgba, hsl or hsla notation.
    at parseToRgb (vendors~main.270ffce4b0377143551d.bundle.js:8469)
    at parseToHsl (vendors~main.270ffce4b0377143551d.bundle.js:8550)
    at lighten (vendors~main.270ffce4b0377143551d.bundle.js:9085)
    at fn (vendors~main.270ffce4b0377143551d.bundle.js:8796)
    at vendors~main.270ffce4b0377143551d.bundle.js:132612
    at handleInterpolation (vendors~main.270ffce4b0377143551d.bundle.js:897)
    at serializeStyles (vendors~main.270ffce4b0377143551d.bundle.js:989)
    at vendors~main.270ffce4b0377143551d.bundle.js:1294
    at updateContextConsumer (vendors~main.270ffce4b0377143551d.bundle.js:26277)
    at beginWork (vendors~main.270ffce4b0377143551d.bundle.js:26465);
The above error occurred in the <Context.Consumer> component:
    in Styled(div) (created by Tooltip)
    in Tooltip (created by Tooltip)
    in Tooltip (created by InnerPopper)
    in InnerPopper (created by Context.Consumer)
    in Popper (created by TooltipTrigger)
    in Manager (created by TooltipTrigger)
    in TooltipTrigger (created by WithTooltipPure)
    in WithTooltipPure (created by lifecycle(WithTooltipPure))
    in lifecycle(WithTooltipPure)
    in div (created by Context.Consumer)
    in Styled(div)
    in Unknown (created by withState(Component))
    in withState(Component) (created by Context.Consumer)
    in Styled(withState(Component)) (created by Sidebar)
    in nav (created by Context.Consumer)
    in Styled(nav) (created by Sidebar)
    in Sidebar (created by Context.Consumer)
    in _default (created by Layout)
    in div (created by Context.Consumer)
    in Styled(div) (created by Nav)
[...]

PR #6939, opened 2019-06-02
Fix crash when invalid args for library polished's functions

### LINKED ISSUE #6855: Gradients in theme break shortcut menu
**Describe the bug**
Whenever one uses a gradient in the theme, the shortcut button will cause the entire app to crash. 
**To Reproduce**
Theme with crash: 
```js
import { create } from '@storybook/theming';

const blue = '#0063ba';
const grey = '#e6e6e6';
const white = '#ffffff';

export default create({
  base: 'light',

  colorPrimary: blue,
  colorSecondary: blue,

  // UI
  appBg: `linear-gradient(to bottom, ${grey}, ${white})`,
  appContentBg: white,
  appBorderColor: blue,
  appBorderRadius: 0,
});
```
---
Theme without crash:
```js
import { create } from '@storybook/theming';

const blue = '#0063ba';
const grey = '#e6e6e6';
const white = '#ffffff';

export default create({
  base: 'light',

  colorPrimary: blue,
  colorSecondary: blue,

  // UI
  appBg: white,
  appContentBg: white,
  appBorderColor: blue,
  appBorderRadius: 0,
});
```
**Expected behavior**
The gradient background works, but I also expect the shortcut menu to open normally.

**Screenshots**
A screenshot would simply be a white page.

**System:**
 - OS: [MacOS]
 - Device: [Macbook Pro 2018]
 - Browser: [chrome]
 - Framework: [react]
 - Version: [5.0.6]

**Additional**
Also discussed here: #6756 
I think a better error is required to prevent confusion with the theme.

lonyele:
Issue: Fixes #6855, #6756, #6135 

#### Problem 
When using some functions from [polished](https://github.com/styled-components/polished) such as `lighten(...), darken(...)`, its arguments that can't be converted to RGB [throws the error](https://github.com/styled-components/polished/blob/master/src/color/parseToRgb.js#L123) 

Commit that possibly introduced this [bug](https://github.com/storybookjs/storybook/pull/5562/commits/9a50e803e39006ece5718fc7b896143f07e6b63e#diff-0d0c72f854c189e6d0d17a1948570d5dR39)

#### Implementation
- Check if arguments of those functions are valid(gradient, css variable, calc etc...)
- Since it is used on different places, I put it into `util.js` from `@storybook/theming` (I couldn't find the place to put some util/shared functionalities across packages. If it doesn't look ok, please let me know where to put these kinds of functions)

## How to test

### DIFF
--- a/lib/theming/src/utils.ts
+++ b/lib/theming/src/utils.ts
@@ -6,17 +6,17 @@ export const mkColor = (color: string) => ({ color });
 // to library polished's functions such as lighten or darken throws the error
 // that crashes the entire storybook. It needs to be guarded when arguments
 // of those functions are from user input.
-const isLinearGradient = (color: string) => {
-  return typeof color === 'string' && color.includes('linear-gradient');
+const isColorVarChangeable = (color: string) => {
+  return !!color.match(/(gradient|var)/);
 };
 
 const colorFactory = (type: string) => (color: string) => {
   if (type === 'darken') {
-    return isLinearGradient(color) ? color : rgba(`${darken(1, color)}`, 0.95);
+    return isColorVarChangeable(color) ? color : rgba(`${darken(1, color)}`, 0.95);
   }
 
   if (type === 'lighten') {
-    return isLinearGradient(color) ? color : rgba(`${lighten(1, color)}`, 0.95);
+    return isColorVarChangeable(color) ? color : rgba(`${lighten(1, color)}`, 0.95);
   }
 
   return color;
--- a/lib/theming/src/utils.ts
+++ b/lib/theming/src/utils.ts
@@ -2,10 +2,10 @@ import { rgba, lighten, darken } from 'polished';
 
 export const mkColor = (color: string) => ({ color });
 
-// Passing linear-gradient string to library polished's functions such as
-// lighten or darken throws the error that crashes the entire storybook.
-// It needs to be guarded when arguments of those functions are from
-// user input.
+// Passing arguments that can't be converted to RGB such as linear-gradient
+// to library polished's functions such as lighten or darken throws the error
+// that crashes the entire storybook. It needs to be guarded when arguments
+// of those functions are from user input.
 const isLinearGradient = (color: string) => {
   return typeof color === 'string' && color.includes('linear-gradient');
 };
--- a/lib/theming/src/utils.ts
+++ b/lib/theming/src/utils.ts
@@ -10,16 +10,29 @@ const isColorVarChangeable = (color: string) => {
   return !!color.match(/(gradient|var)/);
 };
 
-const colorFactory = (type: string) => (color: string) => {
+const applyPolished = (type: string, color: string) => {
   if (type === 'darken') {
-    return isColorVarChangeable(color) ? color : rgba(`${darken(1, color)}`, 0.95);
+    return rgba(`${darken(1, color)}`, 0.95);
   }
 
   if (type === 'lighten') {
-    return isColorVarChangeable(color) ? color : rgba(`${lighten(1, color)}`, 0.95);
+    return rgba(`${lighten(1, color)}`, 0.95);
   }
 
   return color;
+};
+
+const colorFactory = (type: string) => (color: string) => {
+  if (isColorVarChangeable(color)) {
+    return color;
+  }
+
+  // Guard anything that is not working with polished.
+  try {
+    return applyPolished(type, color);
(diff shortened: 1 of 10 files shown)


shilman (dismissed):
Nice!!! Can you call these `lighten` and `darken` for consistency's sake?

lonyele:
Yeap, I applied your feedbacks~

lib/theming/src/utils.ts
```diff
@@ -1 +1,26 @@
+import { rgba, lighten, darken } from 'polished';
+
 export const mkColor = (color: string) => ({ color });
+
+// Passing arguments that can't be converted to RGB such as linear-gradient
+// to library polished's functions such as lighten or darken throws the error
+// that crashes the entire storybook. It needs to be guarded when arguments
+// of those functions are from user input.
+const isLinearGradient = (color: string) => {
+  return typeof color === 'string' && color.includes('linear-gradient');
```
ndelangen: I think this function is too specific, linear-gradients aren't the only type of gradients.

How about:

```js
const isColorVarChangeable = (color: string) => {
  return !!color.match(/(gradient|var)/);
}
```

related to #6135

ndelangen (changes requested):

ndelangen:
Thanks for contributing @lonyele 🎉

lib/theming/src/utils.ts
lonyele: Oh, you are right... I only thought about the linear-gradient even though I wrote a comment `things that can't be converted to RGB`  I change to your function! Please check

ndelangen: Could you add a few test cases for the function?

lonyele:
Oh, I'm working on this one, please wait for a while

lonyele:
Hm... I'm not sure if the snapshot should be updated(https://circleci.com/gh/storybookjs/storybook/130895?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) Help is needed here...

I found from library `polished` that if anything goes wrong, it throws [`new PolishedError(...)`](https://github.com/styled-components/polished/search?q=Error&type=Code)  I feel like this is a small hot fix, but it is a not bad start to be safe from `polished`'s exception. Potentially any user inputs that are being passed to `polished` can go wrong. I'm not sure if it should be refactored since there are a lot and it is only for the few cases, but if it is needed then any custom logic can be added easily [here](https://github.com/storybookjs/storybook/pull/6939/files#diff-83aa48423f4a4a8b26b8eb4786688a9fR29)

lib/theming/src/utils.ts
lonyele: @ndelangen I wrote some tests and it worked fine, but I'm not sure about updating [the snapshot test](https://circleci.com/gh/storybookjs/storybook/130895?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link)

lonyele (commented):
[sorry mistake](https://github.com/storybookjs/storybook/pull/6939/files#diff-83aa48423f4a4a8b26b8eb4786688a9fR54)

lib/theming/src/utils.ts
```diff
@@ -1 +1,59 @@
...
+
+  if (!isValidColorForPolished(color)) {
+    return color;
+  }
+
+  // Guard anything that is not working with polished.
+  try {
+    return applyPolished(type, color);
+  } catch (error) {
+    return color;
```
lonyele: Logging a message can maybe helpful to the users or tracking bug later, but I'm not sure

ndelangen: Yeah, we'll see if an issue opens up, let's keep it not logging anything for now.

ndelangen:
@lonyele run:
```sh
yarn test --core --update
```

Then commit the changes generated, to fix the snapshots.

lonyele:
@ndelangen Hm... my changes of chromatic looks very wrong. I'll investigate this

ndelangen:
Indeed, I denied the changes that seems wrong.

lonyele:
@ndelangen  oh god... one of the most stupid mistake I've made in few months. I hung my head in shame...
Just a side note, because of [this try/catch](https://github.com/storybookjs/storybook/pull/6939/files#diff-83aa48423f4a4a8b26b8eb4786688a9fR51) tests were all passed even with this stupid mistake.

ndelangen (approved):

PR #6940, opened 2019-06-02
Improve tsconfig.app.json detection for app/angular

kroeder:
Issue: https://github.com/storybookjs/storybook/issues/6934

## What I did

I improved the `tsconfig.app.json` detection as it was broken because of a breaking change in Angular 8. They moved `tsconfig.app.json` from `src/tsconfig.app.json` to `./tsconfig.app.json` 

I changed it so that it reads the correct path from `angular.json` instead of setting a fixed path. That way every Angular version should be compatible 

## How to test

- Checkout this branch
- Link `lib/cli` 
- `sb init -f` a project and check if it can start


shilman (approved):
Thanks for fixing this!

PR #6949, opened 2019-06-03
CLI: Don't output startup information on smoke test

iliyanyotov:
Issue: https://github.com/storybookjs/storybook/issues/5164

Don't output the startup information when running a smoke test.

## What I did
Moved the function call into the condition so it's not showing when running a smoke test.

The snapshots were failing prior to my changes, so I updated them in order to run the tests successfully.

## How to test

- Run the CLI command without the smoke test flag, startup information should be available.
- Run the CLI command with the smoke test flag, you should see only the stats, but not the startup information.


libetl:
next or master ?

libetl:
This change was created on top of https://github.com/storybookjs/storybook/tree/release/5.0

libetl:
Rebasing just to compare, but requesting changes.

libetl (dismissed):
this is wrong, you should rebase on `next`

ndelangen (approved):

ndelangen:
cherry-picked & rebased

ghengeveld:
This changed introduced the unwanted side effect that `--ci` now also doesn't show the startup info anymore. This makes it impossible to know for sure on which port Storybook was actually started.

This is about to be fixed: https://github.com/storybookjs/storybook/pull/7822

PR #6967, opened 2019-06-04
add @babel/core & babel-loader

lukethacoder:
Issue:

Module build failed without @babel/core & babel-loader installed

## What I did

Append @babel/core & babel-loader to package.json

## How to test

Pull down the project files, `npm i`, `npm run storybook`

### DIFF
--- a/examples/html-kitchen-sink/package.json
+++ b/examples/html-kitchen-sink/package.json
@@ -14,6 +14,7 @@
     "storybook": "start-storybook -p 9006"
   },
   "devDependencies": {
+    "@babel/core": "^7.4.5",
     "@storybook/addon-a11y": "5.1.0-rc.4",
     "@storybook/addon-actions": "5.1.0-rc.4",
     "@storybook/addon-backgrounds": "5.1.0-rc.4",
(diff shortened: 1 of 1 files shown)


ndelangen:
Users are supposed to install these dependencies:
https://storybook.js.org/docs/guides/guide-html/#add-babelcore-and-babel-loader

If we start having them inside our package.json dependencies, the version conflicts and the problems following would be really bad, for us and users.

Yankovsky:
@ndelangen Can we add some friendly reminder if @babel/core and babel-loader are not installed as dependencies in project?
I was cleaning up the project by removing unused dependencies and I've spent whole day trying to figure out why storybook won't start. Even worse in my case after first `yarn install` babel-loader was right in root node_modules, but after second it went to react-scripts/node_modules and because of that storybook couldn't find babel-loader.
I think that we can implement something similar to create-react-app reminder about different version of required dependency  and run this check on storybook start. What do you think about it?

Yankovsky:
Should I open a separate issue?

Yankovsky:
I know that there are notifications about missing peer dependencies but they are easy to ignore. 
They should be not warnings but errors.

Yankovsky:
Do you know maybe there is a flag for yarn or npm that turns these warnings into errors?

PR #6968, opened 2019-06-04
FIX the customCanvas option

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/6953 https://github.com/storybookjs/storybook/issues/6493

## What I did
I changed the Preview component so it accepts a customCanvas function as prop, and used this over the `renderPreview` function if passed.

I also fixed the customCanvas component for react-native-server so it hides when the viewmode isn't `story`.

I also changed the toggleFullscreen api, so if the passed value isn't a boolean, it ignores it. (useful so there's no need to wrap the api in a function to strip out an event-object from an event-handler)


shilman (approved):
LGTM!

PR #6970, opened 2019-06-04
Fix links that are not working with plain left click

lonyele:
Issue: #6955 

#### Problem
`Link` from `@storybook/components/src/typography` is used without `cancel={false}` thus it is not clickable with plain left click. [go to code](https://github.com/storybookjs/storybook/blob/next/lib/components/src/typography/link/link.tsx#L13)

Places where these are happening from #6955 and not yet reported `placeholder.stories.tsx` and `about.js`

#### Implementation
I added properties `cancel={false}` and `secondary` to all the links that are expected to work with the plain left click.

#### Question
@ndelangen I don't fully understand about this cancel click functionalities but I feel like fixing is needed on `link.stories.tsx` [story from line 15 and 25 is the same. cancel is defaulted to true](https://github.com/storybookjs/storybook/blob/next/lib/components/src/typography/link/link.stories.tsx)   but I'm not super sure. Please guide me on what to do.
## How to test


ndelangen (approved):

shilman (approved):
LGTM! Thanks @lonyele!

lonyele:
@ndelangen Hi, it is a newbie question...(actually I am) Why was snapshot changed?

ndelangen:
Hey @lonyele we have stories for lots of UI bits that make up storybook, some of those UI bits/components use the component you changed.

All our stories are 'tested'/change-detected using jest snapshots & chromatic.

The jest-snapshots have to be updated as part of the code-change; chromatic is approved from a web-ui.

I hope this explains the process to you. Let me know if you want to learn more 👍

lonyele:
@ndelangen Thanks for your explanation! I wasn't sure about the changes of `hover`,`active` etc but come to think of it, I changed the `Link`. :+1:

PR #6972, opened 2019-06-04
IMPROVE HMR error reporting, no refreshes needed for error recovery

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/5822

## What I did
I fixed the HMR for error handling and improved the display of errors a bit

## How to test
- open any example
- do a code-change in a story file
- assert the codechange is applied without a refresh
- do a code-change with a syntax error
- assert there is a error message
- navigate to another story
- assert that the error message stays in view
- fix the syntax error
- assert that the error disappears and you will navigate to the correct story

also:
- throw an error in the storyFn
- assert that the error is displayed
- fix the error in the story
- assert that the story renders correctly without a refresh

Here's how the error look now:
Screenshot 2019-06-04 at 21 55 09
Screenshot 2019-06-04 at 21 54 58


ndelangen:
By it's nature of HMR only working and doing anything during development, this PR should be tested manually on multiple machines!

tmeasday:
Wow, this is cool @ndelangen. I'd like to take some time to test it out properly

lib/core/src/client/preview/start.js
```diff
@@ -118,13 +122,17 @@ export default function start(render, { decorateStory } = {}) {
...
 
   const renderMain = forceRender => {
     const revision = storyStore.getRevision();
     const selection = storyStore.getSelection();
-    const { kind, name, getDecorated, id } = selection || {};
+    const loadError = storyStore.getError();
+    const { kind, name, getDecorated, id, error } = selection || {};
 
-    if (getDecorated) {
+    if (loadError || error || (revision === previousRevision && previousError)) {
```
tmeasday: What is the situation where `previousError` is defined but `error` is not?

tmeasday: I don't even see `previousError` getting set anywhere...?

tmeasday (commented):
Code seems general good (one comment) but I'd like to try this out a bit.

lib/core/src/client/preview/start.js
ndelangen: good catch, will fix this

ndelangen:
@tmeasday were you able to give this a test locally?

tmeasday:
Oh, sorry I will try and do it today

tmeasday:
Works great @ndelangen. PS I noticed that our `errors` story in official storybook (http://localhost:9011/?path=/story/core-errors--story-errors) -- where we are supposed to return something that React cannot render no longer works. I think this is because React can now render `null`?

Should we change it to `export const errors = () => ({});` -- react doesn't like that.

ndelangen:
Exciting! possibly a headline feature @shilman ?

I'll fix up the merge conflicts & make sure that story you mentioned actually throws an error.

ndelangen:
@tmeasday if a storyFn returns something that react can't render, it doesn't seem to do anything, it just renders nothing in the place. I can't find a way to make that actually throw anymore. React just seems to accept it 🤷‍♂ 

This is because the storyFn isn't executed in the react render stack, and so no InvariantError occurs.

When a component rendered in a story tries to do the same though, it DOES throw a InvariantError.

Would be great to address this at some point, but that doesn't feel like a stopper for this PR.

tmeasday:
Sure thing @ndelangen

PR #6991, opened 2019-06-06
Don't redefine `process` variable

Hypnosphi:
Issue: by default, webpack injects [`process/browser`](https://github.com/defunctzombie/node-process/blob/master/browser.js) import whenever `process` global variable is used. Our `DefinePlugin` overrides it which breaks packages that use stuff like `process.cwd()`. One example is `unified` package used by `react-markdown` internally.

Honestly, I don't remember why I had to add [this line](https://github.com/storybookjs/storybook/blame/86a9d2a9b0ed17878475178e2463605076e6f1d5/examples/official-storybook/webpack.config.js#L39) which were later copypasted to other places. Anyway, GraphQL addon seems to work OK after my current changes.

### DIFF
--- a/lib/core/src/server/manager/manager-webpack.config.js
+++ b/lib/core/src/server/manager/manager-webpack.config.js
@@ -59,7 +59,7 @@ export default ({ configDir, configType, entries, dll, outputDir, cache, babelOp
       new Dotenv({ silent: true }),
       // graphql sources check process variable
       new DefinePlugin({
-        process: { browser: true, env: stringified },
+        'process.env': stringified,
         NODE_ENV: JSON.stringify(process.env.NODE_ENV),
       }),
       // See https://github.com/graphql/graphql-language-service/issues/111#issuecomment-306723400
--- a/lib/core/src/server/preview/iframe-webpack.config.js
+++ b/lib/core/src/server/preview/iframe-webpack.config.js
@@ -59,7 +59,7 @@ export default ({
         template: require.resolve(`../templates/index.ejs`),
       }),
       new DefinePlugin({
-        process: { browser: true, env: stringified },
+        'process.env': stringified,
         NODE_ENV: JSON.stringify(process.env.NODE_ENV),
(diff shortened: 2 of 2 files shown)


tmeasday:
There is another (recently merged?) PR that was fixing this: https://github.com/storybookjs/storybook/pull/6946? 

This solution seems better to me however

Hypnosphi:
> There is another (recently merged?) PR that was fixing this: #6946?

It only adds `browser` field, while there's a lot more in https://github.com/defunctzombie/node-process/blob/master/browser.js

tmeasday:
> This solution seems better to me however

🤷‍♂

shilman:
I was just blindly copy-pasting a fix from another PR. As long as this doesn't break things, I agree this is much cleaner. Any good way to test this?

Hypnosphi:
> Any good way to test this?

Open https://monorepo-git-dont-redefine-process.storybook.now.sh/examples/official-storybook/?path=/graphql/addons-graphql--get-pickachu and see that it works

shilman:
Ok, I'm not sure exactly what that proves, but I'll take your word for it 😘

Hypnosphi:
@shilman No, this proves that defining the whole `process` variable on our side isn't really needed. That becomes the responsibility of default webpack behavior.

Hypnosphi:
The compiled code for `process` usage looks roughly like this, you can check it yourself in develomment, in `vendors` file:

```
/* WEBPACK VAR INJECTION */(function(process) {
  <...>
/* WEBPACK VAR INJECTION */}.call(this, __webpack_require__("../../node_modules/process/browser.js")))

```

cloud-walker:
I use `react-markdown` and I've upgraded storybook to v5.1 and part of my stories are like this right now: 

Schermata 2019-06-06 alle 15 00 49

Hypnosphi:
@cloud-walker yeah, that's exactly the issue this PR fixes

PR #7016, opened 2019-06-10
fix: add missing core-js dependency

ForbesLindesay:
This should resolve the issue some people are seeing in #6204 

## What I did

Added an explicit dependency on core-js. This was mostly working before because the dependency flattening means most people have a copy of core-js in their node_modules folder, but after version 3 was released, some people had version 2 and others had version 3. This explicitly adds the dependency on version 3.


shilman (approved):
LGTM!

shilman (approved):
LGTM!

PR #7051, opened 2019-06-11
core-js 3 errors fix

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/6984 https://github.com/storybookjs/storybook/issues/6928

## What I did
CHANGE to use corejs-upgrade-webpack-plugin over local implementation
ADD this plugin to manager & preview

optional:
Upgrade webpack to v4.33.0 (not really necessary, other then that I made the webpack plugin require the latest, I could lower that version range)

code for corejs-upgrade-webpack-plugin now lives here:
https://github.com/ndelangen/corejs-upgrade-webpack-plugin

This plugin does nothing if it can find core-js v2 references, if they cannot be found it enforces all code to use core-js 3. If THAT can't be found, it throws an error.


leoyli (approved):
LGTM

shilman (approved):
LGTM! Will release on alpha! "Shenanigans" indeed! 😆

PR #7080, opened 2019-06-13
FIX warning of loading prop not being a string

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/7049

## What I did
The component for SideBarItem leaked the `loading` prop through to the html-element, which caused an warning when react was in StrictMode.

I made sure only `className` & `children` are passed on.


B3Kay (approved):
Looks good!

PR #7096, opened 2019-06-14
Updated template to not import addons

Gongreg:
These should not be imported in React Native. They are used for web and imported automatically.


shilman (approved):
LGTM!

PR #7158, opened 2019-06-20
Fix/indenting for addon-notes & addon-info on markdown code blocks

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/5574

## What I did
When code examples come from markdown, they don't have the issue that the format prop is designed to solve, and so setting `format={false}` solves the problem.


ndelangen:
@martinmcwhorter would you be able to contribute a review of this change?

martinmcwhorter (approved):

shilman (approved):
LGTM

PR #7177, opened 2019-06-24
Migrate addon viewport to TS

gaetanmaisse:

## What I did

 - Migrate `@storybook/addon-viewport` to TypeScript
 - Refactor some imports/exports
 - Add some types and interfaces and use them to type codebase


gaetanmaisse:
@ndelangen I think chromatic is ok but not sure, can you take a look at the 7 remainings changes? They are all linked to a new `viewMode` attribute added in DOM

ndelangen (approved):

PR #7190, opened 2019-06-25
Migrate @storybook/addon-events to Typescript

lonyele:
Issue: #5030 

## What I did
I migrated addon-events to Typescript.

I wrote some comments where I think more attention is needed
## How to test


addons/events/src/components/Event.tsx
```diff
@@ -5,10 +5,19 @@ import isEqual from 'lodash/isEqual';
...
 import { styled } from '@storybook/theming';
 import json from 'format-json';
-
 import Textarea from 'react-textarea-autosize';
+import { OnEmitEvent } from '../index';
+
+interface StyledTextareaProps {
+  shown: boolean;
+  failed: boolean;
+  value?: string;
```
lonyele: Probably I missed something... but putting only `...rest` didn't work. I also had to add `value`, `onChange` which seems unnecessary.

addons/events/src/components/Event.tsx
```diff
@@ -67,7 +76,7 @@ const Label = styled.label({
   textAlign: 'right',
   width: 100,
   fontWeight: '600',
-});
+} as any);
```
lonyele: This one also gives me errors and I couldn't resolve it. Error goes like this 
```
Argument of type '() => { display: string; boxSizing: "border-box"; verticalAlign: string; paddingRight: number; paddingTop: number; textAlign: "right"; width: number; fontWeight: "600"; }' is not assignable to parameter of type 'TemplateStringsArray'.
  Type '() => { display: string; boxSizing: "border-box"; verticalAlign: string; paddingRight: number; paddingTop: number; textAlign: "right"; width: number; fontWeight: "600"; }' is missing the following properties from type 'TemplateStringsArray': raw, concat, join, slice, and 16 more.ts(2345)
```
Changing code like this also didn't work
```
const Label = styled.label(() => ({
  display: 'table-cell',
  boxSizing: 'border-box',
  verticalAlign: 'top',
  paddingRight: 5,
  paddingTop: 7,
  textAlign: 'right',
  width: 100,
  fontWeight: '600',
}));
```

Similar code is already used at other parts
https://github.com/storybookjs/storybook/blob/8aae29d55234b3267cfd131d2bf94e1ef44b9578/addons/a11y/src/components/Tabs.tsx#L18-L24
https://github.com/storybookjs/storybook/blob/2887028ad221f4641d5cc7c33cb1269cc2bf1d01/lib/components/src/form/field/field.tsx#L4-L13

Here are for more information [this](https://github.com/emotion-js/emotion/issues/1298),  and [this](https://spectrum.chat/emotion/help/emotion-typescript-debugging~df93d2c1-d607-4ff5-80e0-3914219ad8ea?authed=true#)

addons/events/src/components/Event.tsx
```diff
@@ -98,12 +121,17 @@ class Item extends Component {
...
+    payload: null,
+    payloadString: '',
+    // eslint-disable-next-line react/no-unused-state,
+    prevPayload: null,
   };
 
-  onChange = ({ target: { value } }) => {
-    const newState = {
+  onChange = ({ target: { value } }: React.ChangeEvent<HTMLTextAreaElement>) => {
+    const newState: Partial<ItemState> = {
```
lonyele: Just another newbie question... What is the right way to initialize an object in typescript? I've read [this](https://basarat.gitbooks.io/typescript/docs/tips/lazyObjectLiteralInitialization.html) and articles on google in general... but still I'm not sure. I've used `Partial` because the current setting is `no-object-literal-type-assertion`. Not sure below's `this.setState(newState as ItemState)` too

addons/events/src/components/Event.tsx
```diff
@@ -98,12 +121,17 @@ class Item extends Component {
     payload: {},
   };
 
-  state = {
+  state: ItemState = {
     isTextAreaShowed: false,
+    failed: false,
+    payload: null,
+    payloadString: '',
+    // eslint-disable-next-line react/no-unused-state,
```
lonyele: I've disabled eslint for this one because of https://github.com/yannickcr/eslint-plugin-react/issues/1759  . That is a pretty old one, maybe there is a better way?

lonyele:
sorry if there are so many questions!

ndelangen:
@gaetanmaisse think you could help here?

addons/events/src/components/Event.tsx
wuweiweiwu: You can also use something like `React.HTMLProps<HTMLTextAreaElement>`

addons/events/src/components/Event.tsx
wuweiweiwu: I think partial is fine! However, you should remove `newState as ItemState`

gaetanmaisse: @wuweiweiwu removing `newState as ItemState` will cause TS error 😞 

A workaround is to do something like:
```ts
try {
    this.setState({
        payloadString: value,
        payload: JSON.parse(value.trim()),
        failed: false
    });
} catch (err) {
    this.setState({
        payloadString: value,
        failed: true
    });
}
```
But I'm not really convinced...

addons/events/src/components/Event.tsx
lonyele: Thanks, I'll try that!

addons/events/src/components/Event.tsx
lonyele: @gaetanmaisse yeah I'll try to find something more elegant(?) haha

gaetanmaisse (approved):

addons/events/src/components/Event.tsx
lonyele: I tried some other ways but this one looks better to me. In this example, `state` is `Readonly<ItemState>` and `newState` is `Partial<ItemState>` 
```
this.setState(state => ({ ...state, ...newState }));
```

addons/events/src/components/Event.tsx
lonyele: hm... It didn't work for me. Maybe I implemented in a wrong way?

ndelangen:
`/tmp/storybook/addons/events/src/components/Panel.tsx
  5:1  error  '@storybook/api' should be listed in the project's dependencies. Run 'npm i -S @storybook/api' to add it  import/no-extraneous-dependencies
`

lonyele:
@ndelangen oh... how come this happened. I `yarn bootstrap --core`, `yarn dev`, `yarn start`, `yarn test` and couldn't catch this big mistake? hm...

shilman (approved):
LGTM!

PR #7200, opened 2019-06-26
Fix crna-kitchen-sink not working

lonyele:
Issue: #7199 

## What I did
I've matched `react`, `react-native`, `expo sdk` version by bumping up
- `react-native` to latest stable version `0.59.9`  
- `react-native` to sdk `33.0.0`
- `expo` version to `33.0.7`
- `expo sdkVersion`  to `33.0.0`

#### More info
I'm not totally sure but there was a mismatch of `react`, `react-native`, `expo sdk` version.

Possibly introduced from [here](https://github.com/storybookjs/storybook/commit/a6a8dd93fd33d2c6dc9184ffe268ab05472f3892#diff-c95252b436248cebf2b6d1b022b95a22L20) that upgrading `react` version from `16.5.1` to `16.8.6`
or where `expo sdk` was not upgraded when the upgrade of `react` and `react-native` has happened

I referenced [here](https://stackoverflow.com/questions/52817118/module-schedule-tracking-does-not-exist-in-the-haste-module-map/54427170), [here](https://github.com/facebook/react-native/issues/21150)

I'm not sure about bumping up `expo-cli` or peer-dependencies of `react-native` to `>=0.59` Should I upgrade it? 

Personally, ugrading `react-native` scares me... so if this breaks something please comment on this

## How to test

### DIFF
--- a/examples-native/crna-kitchen-sink/app.json
+++ b/examples-native/crna-kitchen-sink/app.json
@@ -1,6 +1,6 @@
 {
   "expo": {
-    "sdkVersion": "32.0.0",
+    "sdkVersion": "33.0.0",
     "platforms": [
       "ios",
       "android"
--- a/examples-native/crna-kitchen-sink/package.json
+++ b/examples-native/crna-kitchen-sink/package.json
@@ -15,10 +15,10 @@
     "preset": "jest-expo"
   },
   "dependencies": {
-    "expo": "^32.0.6",
+    "expo": "^33.0.7",
     "prop-types": "^15.6.2",
     "react": "16.8.6",
-    "react-native": "https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz",
+    "react-native": "https://github.com/expo/react-native/archive/sdk-33.0.0.tar.gz",
     "whatwg-fetch": "^3.0.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -2156,13 +2156,12 @@
   resolved "https://registry.yarnpkg.com/@expo/traveling-fastlane-linux/-/traveling-fastlane-linux-1.9.9.tgz#fd05ec401e1f45052bf8681dc4bfd21e4e05d66b"
   integrity sha512-DKfeDhfeXmYhfM/CT7yiPjyvEMFyLvEDuKU3RIdieMZ3CJOobTkdaOQlQd39VYZ+q1a1qL5QvTiIlPvcSXv9PQ==
 
-"@expo/vector-icons@~9.0.0":
-  version "9.0.0"
-  resolved "https://registry.yarnpkg.com/@expo/vector-icons/-/vector-icons-9.0.0.tgz#7f18e21d3edc8b99b76d7d1b8e26b212393e08b3"
-  integrity sha512-k5ndrW3oueW5jRDLt3o8iXKmiU+CvvCZPewOvxY7eRMivi8hIr6TkW6tMCGE1vS5fwmXffIkIpKGZkSbX7TxwA==
+"@expo/vector-icons@^10.0.1":
+  version "10.0.3"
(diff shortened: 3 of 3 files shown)


Gongreg:
Hey @lonyele, thanks for fixing it! :) I never really use CRNA example, so it keeps getting outdated. We should really improve the situation there, maybe having e2e tests will help.

Gongreg (approved):

app/react-native/package.json
```diff
@@ -33,8 +33,8 @@
     "rn-host-detect": "^1.1.5"
   },
   "devDependencies": {
-    "@types/react-native": "^0.57.57",
-    "react-native": "^0.57.8"
+    "@types/react-native": "^0.57.64",
+    "react-native": "^0.59.9"
```
benoitdion: our peer dependency still says `"react-native": ">=0.57.0"` but it looks like we broke backward compatibility with older version of RN. We may need a deeper look to understand why that happened to see if it can be fixed. With our next major release, we may want to document what version of RN/expo we're going to support and for how long. We should be able to support 0.59 for longer than 0.57 since that's the first version with react hooks support.

lonyele: yeap It seems like more investigation is needed. I didn't get much from seeing why the deploy/build failed. For now, I don't know what to do next... If you have a guess, can you tell me where to look?

Gongreg: There were few lines of changes yet I didn't see this one. :/
What exactly is the reason to currently lock it to .57 ? What breaking change did we introduce?
And do we need to lock it to .59? If there is need we can do it for next major version I guess.
But I don't think that should be the reason why expo build is failing. I could try to take a look into it too.

benoitdion: the main one is some of the storybook core code started using hooks which weren't supported before 59

Gongreg: Do you know which part? Maybe we can do something about it? Because if we can't change it we have to lock the version and notify people about it.

benoitdion: I'm not really sure, this is mostly a guess 😅

lonyele: Sorry guys. CI failed because of source-loader problem. After I merged latest `next` branch, there was no test fail anymore.

lonyele:
Hm... [CI for build and test](https://github.com/storybookjs/storybook/pull/7200/checks?check_run_id=164623588) looks fine but only [now deployement](https://zeit.co/storybook/monorepo/4063tm29u) has a problem. What should I do? I don't know why now deployment has failed also I can't rerun the ci...

shilman (approved):
LGTM @benoitdion are you ok with this?

benoitdion:
I'm ok merging since the peer dependency is version is unchanged but we should still spend time understanding why the bump was required for the kitchen sink app. It could still mean `next` broke compatibility with RN 0.57

lonyele:
@benoitdion I would just list errors that I've met. Most help is got from [here](https://stackoverflow.com/questions/52817118/module-schedule-tracking-does-not-exist-in-the-haste-module-map/54427170)

- `module schedule/tracking does not exist in the Haste module map` error.
   --> I've bumped up `react-native` to the latest
- `console.error: React Native version mismatch.` error. From a search
> Sidenote for Expo Users:
> 
> If you are using expo, be aware that expo <= 32.0.0 uses a react-native version <= 0.57.2 and therefore will require the react version to be <= 16.5.1
  
 --> I've bumped up all expo related versions to `33.0.0`(sdk, expo, react-native)

benoitdion:
it's likely that we are now incompatible with pre-hook versions of react

lonyele:
@benoitdion hm.... I revert back `react-native` and only left what is in the crna-kitchen-sink and it is working! Come to think of it, `app/react-native` isn't relevant from the start?

lonyele:
Phew... no errors now. Can you guys check? @shilman @Gongreg

PR #7224, opened 2019-06-28
[bug] Fix automatic Angular module metadata extraction for forRoot imports

pascaliske:
Issues: #7106, #7157

## What I did

As described in #7106 the automatic detection for component declarations in Angular storybooks does not work with the really common pattern of `forRoot`-imports. This PR fixes the behaviour.

The `importItem` for extracting the `NgModule`-metadata will be checked for a key `ngModule` first which exists in the case of an `forRoot`-import because the method does not return the `NgModule` directly but an object containing it instead (called [`ModuleWithProviders<MyModule>`](https://github.com/angular/angular/blob/master/packages/core/src/metadata/ng_module.ts#L83-L95)).

Besides that this PR fixes the error mentioned in #7157 which will be thrown if the decorator key `__annotations__` can not be found on the target which is the case on importing AOT-compiled modules. I added another condition for checking if the decorator key is defined.

## How to test

The additional example stories introduced in #6346 will still suffice for testing.


MatissJanis:
Please don't close this. Angular is broken currently.

addons/storyshots/storyshots-core/src/frameworks/angular/helpers.ts
```diff
@@ -31,11 +31,14 @@ const createComponentFromTemplate = (template: string) => {
   })(componentClass);
 };
 const extractNgModuleMetadata = (importItem: any): NgModule => {
+  const target = importItem && importItem.ngModule ? importItem.ngModule : importItem;
```
kroeder: Instead of checking if `importItem` exists can we instead add an interface for `importItem: any`? I haven't checked but maybe angular already provide such an interface. Or we just write down what we need

```
interface StorybookAngularMetaData {
  ngModule?: AngularModule;
}
```
Though, as I am writing this I just noticed we need dynamic access to `importItem[decoratorKey]` 

@ndelangen haven't we encountered the same type conflict in one of your PRs where we need some fix types and additionally `[key: string]: object` ? 

@pascaliske what do you think?

kroeder (commented):
LGTM - just left 1 comment

Though, by clicking through the demo I found this and it throws an error
https://monorepo-git-fork-pascaliske-7106-angular-fix-auto-2d5086.storybook.now.sh/examples/angular-cli/?path=/story/custom-dependencies--inputs-and-inject-dependencies

Is this related to this PR? 

https://monorepo-git-fork-pascaliske-7106-angular-fix-auto-2d5086.storybook.now.sh/examples/angular-cli/?path=/story/addon-centered--centered-component as well throws an error but seems unrelated 🙇 

**Edit** Compared it with the current published storybook version https://monorepo.storybook.now.sh/examples/angular-cli/?path=/story/custom-dependencies--inputs-and-inject-dependencies

It's not related

kroeder (approved):

kroeder:
One thing before merging this: Can you add a story to the angular example for having at least a manual test for the moment?

addons/storyshots/storyshots-core/src/frameworks/angular/helpers.ts
pascaliske: It would definitely be good to type out `importItem` with an interface but we still need to check for the existence to gracefully cover edge cases where the item could be undefined, I think...

pascaliske:
> One thing before merging this: Can you add a story to the angular example for having at least a manual test for the moment?

Just pushed a story `Feature Module as Context with forRoot` based on the existing `ChipsModule` but with the appropriate `forRoot` usage. 🙂

kroeder (approved):

pascaliske:
I also added the missing storyshots which I forgot... 😄

PR #6993, opened 2019-06-06
Tech/improve bootstrap list

ndelangen:
Issue: running `yarn bootstrap` requires you run `yarn install` first

## What I did
I optimized the install & give the bootstrap list a few improvements

[](http://www.youtube.com/watch?v=xku1pRzIszI "")

PR #7239, opened 2019-07-01
ADD an ignore for an eslint error about a missing dependency (puppeteer)

ndelangen:
Issue: https://github.com/storybookjs/storybook/pull/6993#discussion_r298812737

## What I did
ADD a ignore rule


shilman (changes requested):
Add this in the file instead of globally?

ndelangen:
@shilman no, the issue is: sometimes the package is installed, sometimes it isn't.

In the case the package is installed, eslint will complain about an eslint ignore that's unnecessary. It's pretty meta, but your suggestion won't work.

Hypnosphi:
I think it's better to add `.eslintrc.js` to storyshots-puppeteer and specify puppeteer in `coreDependencies`. That way, it won't ignore other import errors

shilman (approved):
LGTM tho there are some deepscan errors with the unused vars

ndelangen:
I'm fine with those

addons/storyshots/storyshots-puppeteer/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+const error = 2;
+const warn = 1;
+const ignore = 0;
+
+module.exports = {
+  rules: {
+    'import/no-unresolved': ignore,
```
Hypnosphi:
```
'settings': {
  'import/core-modules': ['puppeteer']
}
```

https://github.com/benmosher/eslint-plugin-import#importcore-modules

ndelangen: thanks, that works!

PR #7240, opened 2019-07-01
FIX by wrapping the pages with a Page component

ndelangen:
Issue: the links are un-clickable in the about page & keyboard shortcuts page is not working well.

## What I did
Made sure the pages render correctly and are never overlapped by hidden boxes.

### DIFF
--- a/lib/ui/src/settings/about.js
+++ b/lib/ui/src/settings/about.js
@@ -17,6 +17,17 @@ import {
 
 import SettingsFooter from './SettingsFooter';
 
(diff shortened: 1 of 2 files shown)


shilman (approved):
LGTM

lonyele:
@ndelangen @shilman   It took more than I thought...(I am late...) This PR can resolve the issue by showing settings page on top of other components, but persisting settings page url and more settings page routes in the future needed `Page` component. I know I'm late but would you also consider this #7241?

PR #7243, opened 2019-07-01
React-Native: Fix null story check

Gongreg:
Issue: #4716

## What I did
Proper check for story since it might be not initialised at first.


shilman (approved):
LGTM

PR #7256, opened 2019-07-02
FIX addon-actions performance issue

ndelangen:
Issue: 
> Serialising functions takes a lot of time/CPU, the default setup where, actions gets transferred across the channel including methods & functions is slow.

Related:
#6287

## What I did
ADD support for allowFunction boolean option on [telejson](https://github.com/storybookjs/telejson) & set default value for action to `false`.

## How to test
By default the data passed to `action` will sanitised from functions,
I added a case


shilman (approved):
LGTM

PR #7312, opened 2019-07-04
Setup wercker

ndelangen:
Issue: testing wercker for CLI tests


ndelangen:
maybe later, opted for travis at this time

PR #7338, opened 2019-07-08
Migrate @storybook/html to Typescript

emilio-martinez:
Issue: #5030

## What I did

Migrate @storybook/html to Typescript

## How to test

Everything should function the same, except the package should now have type definitions.

_Note:_ The first commit is intentionally mostly a rename from js => ts to make the review easier via the subsequent commits.


gaetanmaisse (approved):

app/html/src/server/framework-preset-html.ts
```diff
@@ -1,4 +1,7 @@
-export function webpack(config) {
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { Configuration } from 'webpack';
```
gaetanmaisse: I don't know well how this part of SB is working but maybe `webpack` should be a `peerDependency`? 
 @ndelangen can surely help us with this ;)

ndelangen: You can create a dependency on webpack for now, in monoconfig, I'll try an consolidate these types into a coherent whole.

When you add the dependency, be sure to add a wide version range.

Hypnosphi: I think we should export the preset function types from `@storybook/core`

emilio-martinez: Just to clarify, in this case `Configuration` is being picked up through `@types/webpack` which is a devDependency at the root of the repo. Technically, there's no runtime use or type definitions coming from `webpack` here.

It would make sense to add `@types/webpack` as a devDependency here—I could match the version at the root of the repo to avoid any conflicts. Given the above, do you guys still feel I should add `webpack` as a dependency? Let me know your thoughts 😄

emilio-martinez: @gaetanmaisse @ndelangen let me know your thoughts on my last comment and I'll make the change :)

ndelangen:
I recently did some work on app/react & app/angular

Those are now typescript as well. Would be great if this PR would follow those examples.

emilio-martinez:
@ndelangen latest commit is up for you to take a look—rebased against `next` and aligned the types to app/react & app/angular.

ndelangen (approved):

PR #7356, opened 2019-07-09
Addon-knobs: Fix select options types to allow string[] and null

emilio-martinez:
Issue: #7348

## What I did

Fixed `select(...)` knob types to allow for:
- Values of type `string`, `string[]` and `null`.
- Options passed as `string[]` in addition to objects.

This better aligns with the [Knobs Addons select docs](https://github.com/storybookjs/storybook/tree/master/addons/knobs#select), i.e.:

```
const options = {
  Red: 'red',
  Blue: 'blue',
  Yellow: 'yellow',
  Rainbow: ['red', 'orange', 'etc'],
  None: null,
};

...

You can also provide options as an array like this: ['red', 'blue', 'yellow']
```

### DIFF
--- a/addons/knobs/src/components/types/Select.tsx
+++ b/addons/knobs/src/components/types/Select.tsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import { Form } from '@storybook/components';
 
-type SelectTypeKnobValue = string;
+export type SelectTypeKnobValue = string | number | null | undefined;
 
 export interface SelectTypeKnob {
   name: string;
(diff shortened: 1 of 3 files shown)


addons/knobs/src/components/types/Select.tsx
```diff
@@ -3,17 +3,15 @@ import PropTypes from 'prop-types';
 
 import { Form } from '@storybook/components';
 
-type SelectTypeKnobValue = string;
+export type SelectTypeKnobValue = string | string[] | null;
```
kroeder: Are `select()` knobs always strings? Is `select(1, 2, 3, 4)` not possible? 

Also, if you allow nulll.. is the intention to pass in a null value as knob, like that `select(null, 'string')`? 
If yes, this could be useful too: `select(null, undefined, 'string')`

emilio-martinez: Good points! You're right, technically numbers should be fine because they'd pass the strict eq. comparison. I had added `null` because it was in the example, and similarly `undefined` should work.

Updated.

shilman (approved):
LGTM

PR #7358, opened 2019-07-09
FIX prepare script to delete tests & snapshots etc from dist

ndelangen:
Issue:
- We're shipping snapshots & tests in our dist packages
- snapshots in the dist may be detected by jest and flagged for deletion. These can be impossible to resolve because they are ignored by source control.

## What I did
FIX prepare script so it deletes the files copied by babel --copy-files arg, but should have been copied at all

### DIFF
--- a/scripts/compile-babel.js
+++ b/scripts/compile-babel.js
@@ -6,24 +6,11 @@ const shell = require('shelljs');
 function getCommand(watch) {
   const babel = path.join(__dirname, '..', 'node_modules', '.bin', 'babel');
 
-  const ignore = [
-    '**/__mocks__/',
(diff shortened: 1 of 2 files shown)


shilman (approved):
AFAICT this deletes stories from the distribution. I think we should include stories so that they can be used by other storybooks, at least in lib/components.

shilman (changes requested):
Sorry previous comment was supposed to be a change request

ndelangen:
@shilman Something to discuss long term, I think, we're publishing our full-source with stories & tests in the package already.

I personally think `dist` should be what consumers need.

PR #7359, opened 2019-07-09
Set up CI with Azure Pipelines

ndelangen:


ndelangen:
maybe later, opted for travis for now

PR #7385, opened 2019-07-11
REFACTOR addon-viewport & addon-background to FIX buggy behaviour relating to iframe styling && REMOVE recompose

ndelangen:
Issue: 
- This should decrease the bundlesize a bit, uses hooks over recompose
- addon-viewport & addon-backgrounds interopt oddly, sometimes styling the iframe itself, sometimes a wrapper.
- the grid feature should be part of the addon-background

## What I did
- Remove recompose over hooks
- Make the styling added by addon-viewport & addon-background impact the iframe, not the wrapper
- Remove the wrapper div
- Improve the styling of addon-viewport
- Make viewports centered when smaller then main area
- Improve the tooltip component usage


ndelangen:
Having to tweak a bunch of tools using the tooltip in order to remove recompose, I attended some UI improvements, 
- like the grid tool belongs with the background addon now, 
- the viewports addon uses the new api
- I found a bug in the setAddonState api, which I fixed
- viewport is centered, and still scrollable, styling is better too

PR #1347, opened 2017-06-23
Use uuid for action IDs instead of Math.random (fixes #1109)

marcfallows:
Issue: You can see the issue here: https://github.com/storybooks/storybook/issues/1109

## What I did

I replaced the randomly generated ID with a uuid.

## How to test

I included a test, so view/run the unit test included. Or, run storybook for the actions package and you'll see that IDs are now uuids.

PR #7397, opened 2019-07-12
Use v4 UUID instead of v1 for action IDs

ctavan:
Issue: none

## What I did

I'm a co-author of the [uuid npm module](https://github.com/kelektiv/node-uuid) which is being used in some places within the code base of storybook.

I'm currently trying to understand real-world use cases of [time-based UUIDs ("v1 UUIDs")](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_(date-time_and_MAC_address)).

I found one occurrence where v1 UUIDs were being used instead of v4 UUIDs and I was wondering if this case really requires the semantically much more complex v1 UUIDs or whether we could live equally well with purely random v4 UUIDs?

At least the test suite still passes after my changes, so apparently this doesn't seem to introduce any known regressions.

I'd be really curious to understand the motivation for choosing v1 over v4 UUIDs in the first place, so any feedback on this would be highly appreciated!

@marcfallows, as far as I can tell from the git history you introduced this in https://github.com/storybookjs/storybook/pull/1347, I'd be really curious about your feedback!

## How to test

No noticeable change behavior expected.

### DIFF
--- a/addons/actions/src/preview/action.ts
+++ b/addons/actions/src/preview/action.ts
@@ -1,4 +1,4 @@
-import uuid from 'uuid/v1';
+import uuid from 'uuid/v4';
 import { addons } from '@storybook/addons';
 import { EVENT_ID } from '../constants';
 import { ActionDisplay, ActionOptions, HandlerFunction } from '../models';


ndelangen (approved):

ndelangen:
@ctavan imagine, no-one really invested a lot of time in understanding the difference on our end, and picked the first working solution. 🤷‍♂

ctavan:
@ndelangen thank you so much for the feedback and for the quick merge, highly appreciated!

ndelangen:
You're welcome @ctavan, and thank you for the effort, time and energy you put into maintaining uuid ♥️.

PR #7409, opened 2019-07-13
Addon-info: Add configurable component comparator

### LINKED ISSUE #3758: addon withInfo propTablesExclude does not exclude

I am trying to exclude components from being shown in Prop Tables section like mentioned herer: docs https://github.com/storybooks/storybook/tree/master/addons/info#options-and-defaults
```js
{
/**
   * Exclude Components from being shown in Prop Tables section
   * Accepts an array of component classes or functions
   * @default []
   */
  propTablesExclude: Array<React.ComponentType>,
}
```
However it does not seem to work and I am getting Prop Tables even for excluded components.

### Steps to reproduce

Here is small sample code to test `propTablesExclude`:
```js
// @flow
import React from 'react';
import { withInfo } from '@storybook/addon-info';

/* eslint-disable react/no-unused-prop-types,react/require-default-props */
type PropsType = {
  label: string,
  onClick?: Function,
  disabled?: boolean,
};

const Button = ({ label, onClick, disabled }: PropsType) => (
  <button onClick={onClick} disabled={disabled}>
    {label}
  </button>
);

Button.defaultProps = {
  disabled: false,
  onClick: () => {},
};

storiesOf('Test', module).add(
  'Exclude component from prop tables',
  withInfo({
    text: 'This can exclude extraneous components from being displayed in prop tables.',
    propTablesExclude: [Button],
  })(() => (
    <div>
      <Button label="Button" />
    </div>
  ))
);
```
Here is screenshot of Prop Tables for Button being not excluded:
screenshot

```
    "@storybook/addon-info": "4.0.0-alpha.9",
    "@storybook/react": "4.0.0-alpha.9",
```
### Where to start

My humble guess is that something goes wrong somewhere here: https://github.com/storybooks/storybook/blob/master/addons/info/src/components/Story.js#L314-L324
```js
if (
  typeof children === 'string' ||
  typeof children.type === 'string' ||
  (Array.isArray(this.props.propTablesExclude) && // also ignore excluded types
    ~this.props.propTablesExclude.indexOf(children.type)) // eslint-disable-line no-bitwise
) {
  return;
}
if (children.type && !types.has(children.type)) {
  types.set(children.type, true);
}
```
When I have changed this code a bit into something like this:
```js
if (
  typeof children === 'string' ||
  typeof children.type === 'string' ||
  ~this.props.propTablesExclude.toString().indexOf(children.type) // eslint-disable-line no-bitwise
) {
  return;
} else if (!types.has(children.type)) {
  types.set(children.type, true);
}
```
it seemed to fix the issue. 
But I am not sure if this is the correct/safe way.
It would be cool if someone could investigate this.

Thanks

crimx:
Issue: #3758

`react-hot-loader` does some magic proxies behind the scene which may [cause Component type comparing to fail](https://github.com/gaearon/react-hot-loader#checking-element-types). 

## What I did

Add a compare function that detects `react-hot-loader` global and use the `areComponentsEqual` when possible. Fallback to strict comparison.

## How to test

This fix is backward compatible I don't think additional test is needed.

I've exposed the compare function to options but didn't add it to the doc. Looking for comments.


shilman (approved):
Thanks for the fix and sorry for the delay in commenting. Lots of stuff going on. Can you update the snapshots so the tests don't fail?

crimx:
Sorry it's been a while and now I am having trouble running test locally. Merging with upstream didn't help.

PR #7411, opened 2019-07-13
Addon-knobs: improve types via generics and readonlyarray

emilio-martinez:
Issue: Needed to add more fidelity to the knobs value types to cover even more test cases, including [`ReadonlyArray` cases](https://github.com/storybookjs/storybook/issues/7348#issuecomment-510461977) (#7348) and enum cases (#7420).

## What I did

- Added more fidelity to the knobs value types, particularly where many different kinds of values are allowed.
- Added types to cover cases where a `ReadonlyArray` may be provided (Typescript seems to treat this differently as a normal array)
- Added a ts file with various use cases as a form of "tests" (would love to get feedback on approach)

## How to test

The knobs addon spans a fair number of use cases; therefore, I had the need to create a set of test cases with which I want to ensure compilation passes. To this effect, I added a file under `src/__types__/` with a series of test cases. The successful compilation of this file should indicate types are passing correctly, at least as far as that file is intending to document.

I don't believe there's currently an established process for testing type definitions specifically in the repo, so I'd love to get thoughts/feedback on the approach.

### DIFF
--- a/addons/knobs/src/components/types/Array.tsx
+++ b/addons/knobs/src/components/types/Array.tsx
@@ -2,18 +2,16 @@ import PropTypes from 'prop-types';
 import React, { ChangeEvent, Component, WeakValidationMap } from 'react';
 
 import { Form } from '@storybook/components';
+import { KnobControlConfig, KnobControlProps } from './types';
 
-type ArrayTypeKnobValue = string[];
+export type ArrayTypeKnobValue = string[] | readonly string[];
 
-export interface ArrayTypeKnob {
-  name: string;
-  value: ArrayTypeKnobValue;
+export interface ArrayTypeKnob extends KnobControlConfig<ArrayTypeKnobValue> {
   separator: string;
 }
 
-interface ArrayTypeProps {
+interface ArrayTypeProps extends KnobControlProps<ArrayTypeKnobValue> {
   knob: ArrayTypeKnob;
-  onChange: (value: ArrayTypeKnobValue) => ArrayTypeKnobValue;
 }
 
 function formatArray(value: string, separator: string) {
@@ -41,7 +39,7 @@ export default class ArrayType extends Component<ArrayTypeProps> {
 
   static serialize = (value: ArrayTypeKnobValue) => value;
 
-  static deserialize = (value: ArrayTypeKnobValue) => {
+  static deserialize = (value: string[]) => {
     if (Array.isArray(value)) return value;
 
     return Object.keys(value)
--- a/addons/knobs/src/KnobManager.ts
+++ b/addons/knobs/src/KnobManager.ts
@@ -7,7 +7,7 @@ import { getQueryParams } from '@storybook/client-api';
 import { Channel } from '@storybook/channels';
 
 import KnobStore, { KnobStoreKnob } from './KnobStore';
-import { Knob, KnobType } from './type-defs';
+import { Knob, KnobType, Mutable } from './type-defs';
 import { SET } from './shared';
 
 import { deserializers } from './converters';
@@ -72,7 +72,7 @@ export default class KnobManager {
     return this.options.escapeHTML ? escapeStrings(value) : value;
   }
 
-  knob<T extends KnobType = any>(name: string, options: Knob<T>): Knob<T>['value'] {
+  knob<T extends KnobType = any>(name: string, options: Knob<T>): Mutable<Knob<T>['value']> {
     this._mayCallChannel();
 
     const knobName = options.groupId ? `${name}_${options.groupId}` : name;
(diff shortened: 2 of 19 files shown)


CodeByAlex (approved):
LGTM, nice fix! I like your approach to testing the types 👍

CodeByAlex:
@shilman any chance we could get this one in sooner rather than later. It's a blocker to upgrade to a version higher than ~.35

CodeByAlex:
@emilio-martinez, is this working for you in your project? I'm getting the following errors in mine: 

ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Array.d.ts
ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Array.d.ts(3,62):
TS2693: 'string' only refers to a type, but is being used as a value here.

ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts
ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts(4,134):
TS2304: Cannot find name 'readonly'.

ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts
ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts(4,143):
TS2365: Operator '>' cannot be applied to types 'boolean' and 'undefined[]'.

ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts
ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts(4,143):
TS2693: 'NonNullable' only refers to a type, but is being used as a value here.

ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts
ERROR in /node_modules/@storybook/addon-knobs/dist/components/types/Options.d.ts(4,155):
TS2304: Cannot find name 'T'.

emilio-martinez:
@CodeByAlex will ping you on Discord

addons/knobs/src/components/types/Array.tsx
```diff
@@ -2,18 +2,16 @@ import PropTypes from 'prop-types';
 import React, { ChangeEvent, Component, WeakValidationMap } from 'react';
 
 import { Form } from '@storybook/components';
+import { KnobControlConfig, KnobControlProps } from './types';
 
-type ArrayTypeKnobValue = string[];
+export type ArrayTypeKnobValue = string[] | readonly string[];
```
nickofthyme: Is this breaking type change intentional? Using `number`s will cause a typescript error.

emilio-martinez: To quote the README for this add-on:

> array
> Allows you to get an array of strings from the user.

Unfortunately, the package from DefinitelyTyped is inaccurate—that field uses a text-based input of comma delimited values, so even if you were to pass a number array initially, it would subsequently return a string array on change, i.e., here's no number parsing logic for that control.

That said, I'd love to hear more about your use case and help in any way I can. Would you mind sharing?

nickofthyme: Hey @emilio-martinez Thanks for the reply. Yeah it seems like you guys are just having to bite the bullet on these type changes going from javascript to typescript, which I totally feel your pain.

This is my use case 👇. It's fixable by just parsing the values and there is probably a better way to go about it. Just got this error when upgrading from `v5.1.9` to `v5.2.0-beta.19`.
https://github.com/elastic/elastic-charts/blob/8ebe600df1ae476e25032ab1c7707de398143e43/stories/annotations.tsx#L41

I'm curious if you guys are gradually upgrading you codebase to ts or doing it in a single release (I'm too lazy to check for myself)? If it is in a single release why wouldn’t you just release ts in `v6` rather than producing all these stricter type errors?

Thanks 👍

gaetanmaisse: TS migration is an epic started at the end of 2018, we recently reached an important step as 50% of Storybook codebase in TS 🎉.
Migration progress is followed in https://github.com/storybookjs/storybook/issues/5030 😉. Avoiding breaking changes is one of our goals but we somethings found weird/wrong types/docs even in SB own codebase.

nickofthyme: @gaetanmaisse Sweet congrats on the big 5-0!! 🎉 All the best keep up the great work.

leoyli: Actually, there are many work around and we don't buy in strict type and `noImplicitAny` at day one.  For that reason, there are sheer amount of `any` flood in our typing. That somehow explains why we inevitably will have breaking changes in type system, we need to improve it gradually due to the fact that the codebase is pretty huge.

PR #7433, opened 2019-07-16
Add dev flag for install command

RWOverdijk:
Add -D for `yarn install`. I'm from the copy/paste generation. 😄


ndelangen:
Do we have the flag in other places?

RWOverdijk:
@ndelangen Afaik: yes. Makes sense since it generally runs in dev right?

https://www.npmjs.com/package/@storybook/addon-ondevice-knobs it's specified here, too, for example.

shilman (approved):
nice catch!

PR #7448, opened 2019-07-17
Addon-viewports: Add back default viewports

ndelangen:
Issue:
- viewports no longer had any defaults
- there was an issue where if the viewport width was > main area, the viewport wouldn't render at full width.

## What I did
- added a minimal default of 3 viewports as a sane list of viewports, hopefully most people will find useful.
- fixed the layout issue, by using css-grid. flexbox fails for this usecase. This should fail gracefully on browsers that don't support it: they'd just not render the iframe centered.


shilman (changes requested):
@ndelangen Why does this need to be a breaking change? Let's not remove `withViewport` until 6.0? And what's the problem with just using `INITIAL_VIEWPORTS`?

ndelangen:
withViewport is already not exported... Might have happened in the TS migration?

addons/viewport/src/Tool.tsx
```diff
@@ -195,12 +196,15 @@ export const ViewportTool: FunctionComponent<{}> = React.memo(
                 },
                 [`#${wrapperId}`]: {
                   padding: theme.layoutMargin,
-                  display: 'flex',
                   alignContent: 'center',
                   alignItems: 'center',
                   justifyContent: 'center',
```
wuweiweiwu: if its not a flex container do we still need justify content and align items?

ndelangen: `display: grid`

PR #7449, opened 2019-07-17
React native: Emit event when story is rendered

Gongreg:
Issue: N/A

I would like to get an event to know when the story has finished rendering.

## What I did
Added event when story is rendered


shilman (approved):
LGTM

PR #7459, opened 2019-07-17
Marko: Upgrade loader & config

### LINKED ISSUE #7404: @storybook/marko: Migrate from 'marko-loader' to '@marko/webpack/loader'
Migrate from 'marko-loader' to '@marko/webpack/loader' which provides some bug fixes and enhanced functionality.

https://github.com/marko-js/webpack

DylanPiercey:
Issue: #7404

## What I did

Upgraded the Marko Webpack loader to use the [new package](https://github.com/marko-js/webpack#loader-markowebpackloader).

I also removed the dependency on `raw-loader` since it was extraneous and the same for the peerDependency of `marko-widgets` and `babel-loader`.

## How to test

Things should continue to work as before, no API change.


DylanPiercey:
Looking through, it looks like other frameworks also have a peerDep of `babel-loader`, is this something that is needed? Can add that back if so.

shilman (approved):
LGTM

ndelangen (approved):

PR #7477, opened 2019-07-18
Addon-info: only render Components in propTables

steviewonders3:
Issue: https://github.com/storybookjs/storybook/issues/7254

The [`propTables`](https://github.com/storybookjs/storybook/blob/next/addons/info/README.md) config property doesn't behave as expected.

From the README:
```
/**
   * Components used in story
   * Displays Prop Tables with these components
   * @default []
   */
  propTables: Array<React.ComponentType>,
```

Most of my stories use a `<ThemeProvider/>` and `<Box/>` as a story wrapper, and they always show up in the prop tables, which is not desired. This property seems like it is supposed to be an inclusive list of components to show in propTables, but it currently doesn't work. 

## What I did

Added a check to see if `propTables` is an array with at least one item in it, and if it is, only include story children in rendered prop tables that are in the `propTables` array.

> Note - There are no existing tests that render prop tables based on `propTables` having been set in the info config, with multiple different components in the test story.

## How to test

No
No
No

### DIFF
--- a/addons/info/src/components/Story.js
+++ b/addons/info/src/components/Story.js
@@ -319,7 +319,7 @@ class Story extends Component {
       return null;
     }
 
-    if (Array.isArray(propTables)) {
+    if (propTables) {
       propTables.forEach(type => {
         types.set(type, true);
       });
--- a/addons/info/src/components/Story.js
+++ b/addons/info/src/components/Story.js
@@ -319,11 +319,9 @@ class Story extends Component {
       return null;
     }
 
-    if (propTables) {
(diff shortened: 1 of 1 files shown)


smasontst:
This deepscan thing is kind of annoying 😠

smasontst:
I'm going to disregard deepscan, I changed one file and it keeps telling me I introduced new issues on a bunch of other unrelated files? I don't understand what I'm supposed to do about that.

ndelangen (approved):

PR #7484, opened 2019-07-19
chore: log loading angular cli config error

JounQin:
It would be mush more helpful for users to check why the config is not loaded successfully

Issue:

## What I did
Just add log info for loading angular cli config error

## How to test

Not needed

Not needed

Not needed


kroeder (approved):
LGTM

PR #7510, opened 2019-07-20
source-loader: separate server and client code

Hypnosphi:
Issue: 5.2.0-beta doesn't work in IE11. The reason is that client code is transpiled to node target, see https://github.com/storybookjs/storybook/pull/7117/files#r305589980

## What I did
Separated server and client code so that they are transpiled to respective targets


shilman (approved):
LGTM

PR #7515, opened 2019-07-21
Addon-knobs: enable Typescript `strict` flag

emilio-martinez:
Issue: People using Typescript with the `strict` flag on may see some type mismatches due to the added type sensitivity, e.g., https://github.com/storybookjs/storybook/issues/7348#issuecomment-512846401.

## What I did

- Enables TS's `strict` flag in the knobs add-on (Enabling for the entire storybook repo is a tall order and can be done incrementally over time).
- Leverages the types from `@storybook/api` to align function definitions (see `Panel.tsx`).
- Removes former uses of `any` around PropTypes.
- Generally reconciles types to meet strict check requirements

## How to test

There should be no noticeable runtime changes, only a better experience for users using Typescript in `strict` mode. I've updated `__types__/knob-test-cases.ts` to more explicitly reflect some of the nullable scenarios.

### DIFF
--- a/addons/knobs/package.json
+++ b/addons/knobs/package.json
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@storybook/addons": "5.2.0-beta.5",
+    "@storybook/api": "5.2.0-beta.5",
     "@storybook/client-api": "5.2.0-beta.5",
     "@storybook/components": "5.2.0-beta.5",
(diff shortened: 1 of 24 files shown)


leoyli (approved):
Nice job! LGTM!

gaetanmaisse (approved):

PR #7568, opened 2019-07-26
Fix select knob default selection with array

elliotlarson:
Issue: #7569

## What I did

When using a select knob with arrays as option values, the default value was not getting set properly on the select list.  The code was comparing two arrays with ===, which is always false.  

I updated the code to stringify the arrays for doing the comparison.

## How to test

There was not an existing test for the select knob that I could augment/add onto for this edge case.  However, this is a pretty small update.  I did try it out locally and it worked (manual test).

### DIFF
--- a/addons/knobs/src/components/__tests__/Select.js
+++ b/addons/knobs/src/components/__tests__/Select.js
@@ -24,6 +24,21 @@ describe('Select', () => {
       expect(green.text()).toEqual('Green');
       expect(green.prop('value')).toEqual('Green');
     });
+
+    it('should set the default value for array-values correctly', () => {
+      knob = {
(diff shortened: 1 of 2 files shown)


kroeder (approved):
I found a good place for a test 🙂 

Although, I'm not sure I test this correctly 🤔 
Even though the component itself has the correct value does this ensure that it is actually selected?

@ndelangen can you help me out here? I don't know much about enzyme nor react components

PR LGTM

elliotlarson:
@kroeder Thank you for adding a spec!  I usually put my specs side by side with the implementation files.  Somehow I totally missed the `__tests__` directory 🤦‍♂️.

PR #7578, opened 2019-07-26
Ts migration/vue

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/5030

## What I did
I migrated app/vue to typescript

It still has a couple of `any`s in there. So maybe someone with more Vue knowledge could help me out here?

## How to test
- bootstrap the repo
- run the vue example


ndelangen:
I improved the typings for DecoratorFn a bit, had to fix an broken type on the jest addon to get it to work..

app/vue/package.json
```diff
@@ -26,13 +26,18 @@
...
   },
   "dependencies": {
+    "@storybook/addons": "5.2.0-beta.10",
     "@storybook/core": "5.2.0-beta.10",
     "common-tags": "^1.8.0",
     "core-js": "^3.0.1",
     "global": "^4.3.2",
-    "regenerator-runtime": "^0.12.1"
+    "regenerator-runtime": "^0.12.1",
+    "webpack": "^4.33.0"
```
leoyli: Why they now being part of the dependency specially to this package? And `webpack` are usually belongs to the devDep I believe... nope?

app/vue/src/client/preview/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  showMain: () => void;
+  showError: (args: ShowErrorArgs) => void;
+  showException: (...args: any[]) => void;
+  forceRender: boolean;
+}
+
+// TODO: some vue expert needs to look at this
+export type StoryFnVueReturnType = any;
+
+export interface ICollection {
```
leoyli: why sometimes prefix with `I` but sometimes not?  Can we have some convention here (I know there is a ESLint-TypeScript rule for that out of box...)?

app/vue/src/client/preview/util.ts
```diff
@@ -1,19 +1,19 @@
...
-    // known limitation: we dont have the component instance to pass
+    // known limitation: we don't have the component instance to pass
     return def.call();
   }
 
   return def;
 }
 
-export function extractProps(component) {
+export function extractProps(component: any) {
```
leoyli: potentially can use `VueComponent` type from Vue types?

app/vue/package.json
ndelangen: app/vue is a devDependency for users, which has webpack as a normal dependency, because it's actually used in the runtime of the user.

most apps have this, that need something tweaked to the webpack config.

app/vue/src/client/preview/types.ts
ndelangen: I personally prefer the naming convention without the I.

app/vue/src/client/preview/util.ts
ndelangen: This is where I could use your help @leoyli, would you like to give that a try and add to this PR?

ndelangen: In storybook we're doing something custom with this `.options` I think it's not vue behavior. I don't think the value actually is a Vue.Component here..

app/vue/src/client/preview/types.ts
ndelangen: Having some more consistency wouldn't hurt, but it's not a really big deal IMHO.

app/vue/src/client/preview/util.ts
leoyli: oh I see... so it's a bit hard to tell what really is based on that snippets of code

app/vue/src/client/preview/types.ts
leoyli: I guess due to the fact that type scoped variables can be imported/exported together with the normal lexical scoped variables, so it is a bit hard to read/review without using IDE. So it makes sense to have prefix to give hints in such cases.  I'm also preferring without `I` though.

While it is true that it just a matter of code cosmetic and bike-shading, but it is actually quite important to maintain some level of consistency so more people can involve and make contribution easier, IMHO.

leoyli (approved):

app/vue/src/client/preview/render.ts
```diff
@@ -48,6 +49,7 @@ export default function render({
     root[COMPONENT] = element;
   }
 
+  // @ts-ignore I'm not sure why we need this, but types are incompatible
   root[VALUES] = element.options[VALUES];
```
backbone87: the whole `VALUES` stuff is a workaround to make knobs work. each time a knob or story changes, this render function is called and also an entirely new component is created, when `const element = storyFn();` is called above. the thing is, that we dont want to replace the component, when a knob value changes, we just want to change some props based on the new knob values. only when the story changes we want to replace the component (see above `root[COMPONENT] = element;`). as it happens to be, `forceRender` is `true`, if just the knob values change (which is quite counterintuitive).

the problem is that we have to do `const element = storyFn();` in order to update knobs and get their values. thats where the `prepare` function of the story decoration kicks in, which is executed on all components returned from the story or story decorator functions and collects the default values of their props (which contain the new knob value, when we declare our knob like `{ props: { myProp: { default: TEXT_KNOB(...) } } }`). this way we can access all of these values in the component returned from the render story function: `storyFn().options[VALUES]`. in addition to that the `prepare` function also decorates all components returned from the story or story decorator functions with a functional component that passes (and therefore binds) the data from the root component's `[VALUES]` data (which is reactive) to the decorated component via props (https://github.com/storybookjs/storybook/pull/7578/files#diff-cf1aa09935b730d1855cb4c973861de0R46). so each time we change `root[VALUES]`, vue's reactivity kicks in and rerenders all components that use this data, which is basically all components in our case, because the functional component decorators of the `prepare` function access the whole data object by doing `...parent.$root[VALUES]`, which is still cheap in vue, because we dont create new component instances (that arent necessary), like when we would do `root[COMPONENT] = storyFn()` (which would also work).

app/vue/src/client/preview/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    [WRAPS]: story,
+    [VALUES]: { ...(innerStory ? innerStory.options[VALUES] : {}), ...extractProps(story) },
+    functional: true,
+    render(h, { data, parent, children }) {
+      return h(
+        story,
+        {
+          ...data,
+          // @ts-ignore // some Vue expert needs to look at this
+          props: { ...(data.props || {}), ...parent.$root[VALUES] },
```
backbone87: `VALUES` is a reactive data property on the $root vue instance. see the comment in `render.ts` for an explanation. (since this is a functional component, we cant do `this.$root`, but have to use the parents `$root`)

app/vue/src/client/preview/index.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import './globals';
+import { IStorybookSection, StoryFnVueReturnType } from './types';
+
+import render, { VALUES } from './render';
+import { extractProps } from './util';
+
+export const WRAPS = 'STORYBOOK_WRAPS';
+
+function prepare(rawStory: any, innerStory?: any): StoryFnVueReturnType {
```
backbone87: rawStory is `string | Component`
innerStory is `VueConstructor`

app/vue/src/client/preview/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  name: 'unspecified',
+  kind: 'unspecified',
+  parameters: {},
+};
+
+// (storyFn: StoryFn, decorators: DecoratorFunction[])
+function decorateStory(
+  storyFn: StoryFn<StoryFnVueReturnType>,
+  decorators: DecoratorFunction<StoryFnVueReturnType>[]
+): StoryFn<StoryFnVueReturnType> {
```
backbone87: The return type is normalized here to `StoryFn<VueConstructor>`

app/vue/src/client/preview/index.ts
```diff
@@ -0,0 +1,127 @@
...
+        story = decorated(context);
+      }
+
+      if (decoratedStory === story) {
+        return story;
+      }
+
+      return prepare(decoratedStory, story);
+    },
+    (context => prepare(storyFn(context))) as StoryFn<StoryFnVueReturnType>
```
backbone87: this is `StoryFn<VueConstructor>`

app/vue/src/client/preview/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  parameters: {},
+};
+
+// (storyFn: StoryFn, decorators: DecoratorFunction[])
+function decorateStory(
+  storyFn: StoryFn<StoryFnVueReturnType>,
+  decorators: DecoratorFunction<StoryFnVueReturnType>[]
+): StoryFn<StoryFnVueReturnType> {
+  return decorators.reduce(
+    (decorated, decorator) => (context: StoryContext = defaultContext) => {
```
backbone87: `(decorated: StoryFn<VueConstructor>, decorator: StoryFn<StoryFnVueReturnType>): StoryFn<VueConstructor> => ...`

app/vue/src/client/preview/types.ts
```diff
@@ -0,0 +1,35 @@
...
+import { StoryFn } from '@storybook/addons';
+// TODO, 'any' should be what is actually expected from a storyFn
+
+export interface ShowErrorArgs {
+  title: string;
+  description: string;
+}
+
+export interface RenderMainArgs {
+  storyFn: StoryFn<StoryFnVueReturnType>;
```
backbone87: this is a normalized story function: `StoryFn<VueConstructor>`

backbone87 (changes requested):

ndelangen:
@backbone87 I think I've addressed all of your comments!

PR #7622, opened 2019-07-30
CHANGE define plugin variables to globals instead

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/7024

## What I did
I moved the 2 variables from being code-injected by webpack to being injected into the global app runtime via the template

@shilman


shilman (approved):
Awesome. 🚀 Thanks for taking care of this!!

PR #7640, opened 2019-07-31
Addon-centered: Fix disappearing when zoomed

### LINKED ISSUE #7167: Addon Centered causes component to disappear when zooming

ndom91:
This pull request is based on the feedback from everyone in #7167 

This now takes into account the margin, as well as updates this width calculation on each zoom step.

### DIFF
--- a/lib/core/src/server/templates/base-preview-head.html
+++ b/lib/core/src/server/templates/base-preview-head.html
@@ -5,6 +5,10 @@
     width: 100%;
     height: 100%;
   }
+  
+  body {
+    width: calc(100% - 16px);
+  }
 
   :not(.sb-show-main) > .sb-main,
   :not(.sb-show-nopreview) > .sb-nopreview,
--- a/lib/ui/src/components/preview/iframe.js
+++ b/lib/ui/src/components/preview/iframe.js
@@ -29,7 +29,7 @@ export class IFrame extends Component {
     if (scale !== nextProps.scale) {
       this.setIframeBodyStyle({
         width: `calc(${nextProps.scale * 100}% - 16px)`,
-        height: `calc(${nextProps.scale * 100}% - 16px)`,
+        height: `${nextProps.scale * 100}%`,
         transform: `scale(${1 / nextProps.scale})`,
         transformOrigin: 'top left',
       });
--- a/lib/ui/src/components/preview/iframe.js
+++ b/lib/ui/src/components/preview/iframe.js
@@ -28,8 +28,8 @@ export class IFrame extends Component {
 
     if (scale !== nextProps.scale) {
       this.setIframeBodyStyle({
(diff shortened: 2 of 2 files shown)


ndom91:
Fixes issue #7167

lib/core/src/server/templates/base-preview-head.html
```diff
@@ -5,6 +5,10 @@
     width: 100%;
     height: 100%;
   }
+  
+  body {
+    width: calc(100% - 16px);
```
ndelangen: what is this magic 16px number?

ndom91: 2x 8px margin

shilman (approved):
LGTM!

lib/core/src/server/templates/base-preview-head.html
Hypnosphi: But people may have other margin values

Hypnosphi:
What about height?

lib/core/src/server/templates/base-preview-head.html
ndom91: this is the margin hardcoded by storybook in the iframe, not from the user.

ndom91:
> What about height?

afaik height does not have this overflow problem on zoom

lib/core/src/server/templates/base-preview-head.html
Hypnosphi: Not by storybook, by browsers. But users can override it

Hypnosphi:
Right now, `height: 100%` on body adds vertical scroll even if it's not needed

Generally, I don't really like the fact that a fix for some particular addon affects layout even if you don't use the addon at all. Let's add those styles only for stories that use addon-centered decorator?

ndom91:
This does make more sense.. The centered addon is part of the primary storybook repo. Right?

Hypnosphi:
Yes: https://github.com/storybookjs/storybook/tree/next/addons/centered

Unfortunately, there's separate implementations for each framework. You may want to create some common helper and reuse it in all implementations.

BTW maybe it would be enough to set `body {transformOrigin: center center;}` for centered stories

lib/core/src/server/templates/base-preview-head.html
ndelangen: Yeah this seems fragile to me @shilman

PR #7651, opened 2019-08-01
Add .pdf support to file-loader

testerez:
Issue: the current webpack config does not support PDF files

## What I did

Just added `.pdf` to file-loader supported file extentions

## How to test

Import a PDF file from a component. You should get a link to download this PDF file.


shilman (approved):
seems reasonable to me

wuweiweiwu (approved):
nice!

PR #7705, opened 2019-08-07
Addon-jest: fix console warning

XYShaoKang:
Key should be in the outermost component

Issue:

## What I did

Move the key to the outermost component


CodeByAlex (approved):
LGTM

PR #7731, opened 2019-08-10
[Try] Add support for LitElement

lonyele:
I'm pretty sure I'm not the best person doing this one, but I gave a shot for the super basic settings. I've read #4958, #6208, [Polymer Issue](https://github.com/Polymer/polymer/issues/5240) etc... and came to a conclusion... that it is kind of complicated than I thought about this `Web Components` thing(first time looking at it...) 

From the Polymer side there are `polymer1/2`, `polymer3`, `lit-html`, `lit-element` or any frameworks/compiler that converts things to web components such as `stencil`, `svelte` etc.

For the users, installing `@storybook/web-components` and it works for every case I wrote above is the best thing since It will be ended up just `'Web Components'`, but I'm not sure if it's possible(It looks easy on theory, but I'm not sure in practice) So I made the working example only for `lit-element` for a starter(@storybook/lit-element) I also have `lit-html` example on my local(possibly `@storybook/lit-html` and it's almost same as this PR) 

I found that there are a lot discussions and opinions on this matter, I just want to hear some thoughts on this. For me, I don't really have much opinions(almost no knowledge and experience), just wanted to make things working(please ignore some copy/paste texts, I'll fix it once it looks ok)

## What I did
- Add `@storybook/lit-element`
- Add `lit-element-kitchen-sink` example

## Need help
- Add [any polyfills](https://github.com/webcomponents/polyfills) and more advanced webpack settings for different environments
- `@storybook/addon-link` support. Everything is in shadow-dom and I couldn't make it working...
- preset?
- A lot... probably


app/lit-element/src/client/preview/render.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!forceRender || !rootElement.querySelector('[id="root-inner"]')) {
+      rootElement.innerHTML = '<div id="root-inner"></div>';
+    }
+    const renderTo = rootElement.querySelector('[id="root-inner"]');
+
+    // Use lit-html's render function that is being used at lit-element
+    // internally to cover the case using only html`...` instead of
+    // lit-element which makes using lit-html and lit-element at the same
+    // time.
+    LitElement.render(element, renderTo, { scopeName: '' });
```
lonyele: I think this is a hack I've found...

examples/lit-element-kitchen-sink/.storybook/webpack.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = async ({ config }) => {
```
lonyele: This is the place where help is needed. Please [see](https://github.com/Polymer/lit-element/issues/54#issuecomment-439901575)

lonyele:
Oh... wait...

app/lit-element/README.md
```diff
@@ -0,0 +1,31 @@
+# Storybook for Polymer
```
fernandopasik: probably this title needs to be adjusted to `Lit-Element`

fernandopasik: Well, there are other mentions in the readme as well, work in progress I guess

fernandopasik: great work btw!

steel:
This is neat. I'm also using this for custom elements with `lit-element`. We are using a custom decorator to detect the output type of the story and wrap it in a div if it's a lit template. Everything just works.

lonyele:
Hm... I'm not really sure, it is too small for a so called framework support... putting `if/else` at [render](https://github.com/storybookjs/storybook/pull/7731/files#diff-fe029887bd97a69d1e001762667f6d7fR32) is the only useful code...

ndelangen:
@fernandopasik @emilio-martinez Could we get a seal of approval from you or some assistance maybe getting this in a state it's merge-able?

lonyele:
It looks like [much better implementation](#8400) is merged. Check that out!

PR #7736, opened 2019-08-10
Migrate @storybook/addon-ondevice-backgrounds to Typescript

lonyele:
## What I did
I migrated addon-ondevice-backgrounds to Typescript

## How to test

Tracking Issue: #5030


addons/ondevice-backgrounds/src/container.tsx
```diff
@@ -32,18 +41,3 @@ export default class Container extends React.Component {
     );
   }
 }
-
```
gaetanmaisse: W

addons/ondevice-backgrounds/src/container.tsx
```diff
@@ -32,18 +41,3 @@ export default class Container extends React.Component {
...
-Container.propTypes = {
```
gaetanmaisse: Why do you remove these `propTypes` and not the one of `Swatch` component?

gaetanmaisse (approved):

addons/ondevice-backgrounds/src/container.tsx
lonyele: As I know of, React class component can't be used with PropTypes? I get this error `Property 'propTypes' does not exist on type 'typeof Container`  So it seems like only working with functional component(`Swatch` is a functional component) I wanted to leave `PropTypes` for runtime check but I couldn't do it. Is there any other way to do it?

gaetanmaisse: True! I already faced that problem before... and finally, I did not found a proper solution. Either cast `Container` as any to define propTypes or remove them, I picked the 2nd one just like you did.

lonyele: Yeah I feel like it's weird but it seems like it is working as intended

PR #7774, opened 2019-08-14
React: Fix missing props in the Welcome.tsx demo

joeycozza:
Issue: 
When the typescript conversion occurred, a few props were destructured out of the props of Link and NavButton (in order to be able to define types for them I assume), but they weren't put into the rendered jsx for their component. (before the typescript conversion they were "automatically" placed in the jsx because of using rest and spread operators with props)

## What I did
I explicitly attach those props to the element that needed them.


shilman (approved):
Good catch! Thanks for fixing 🙏

PR #7808, opened 2019-08-19
Addon-docs: Fixed some sample code to be more valid

joeycozza:
Issue: I copied some sample code in the recipes.md file, and there was a pathing problem to the blocks, and a syntax issue with jsx in the mdx file.

## What I did
Fixed the pathing and the syntax issue in the sample code.


shilman (approved):
Thanks! 🙏

PR #7809, opened 2019-08-19
React-Native: Fix tabs layout issue on simulator

bqrichards:
Issue: #6917 

## What I did

I added a SafeAreaView to wrap the OnDeviceUI addon panel.


BatDroid:
that would be great if we could have it

Gongreg (approved):

PR #7824, opened 2019-08-20
Add tests and src to npmignore

Diokuz:
Hey, guys!

I just added some non-production folders to npmignore.

There are many files, which slow down npm install, docker build and other fs/https-related operations.

I could add more ignores to other packages, if that is okay.

What do you think?


ndelangen:
@Hypnosphi what do you think about this?

Hypnosphi:
It's a good idea, but we also need to remove `jsnext:main` fields:
https://github.com/storybookjs/storybook/search?q=jsnext%3Amain&unscoped_q=jsnext%3Amain

Hypnosphi:
I wonder whether a single `.npmignore` on the root level would work. @shilman can you experiment with it maybe?

ndelangen:
@Hypnosphi doesn't seem to be an option:
https://github.com/lerna/lerna/issues/296

ndelangen:
@Hypnosphi I've removed all `.npmignore` files, and instead have added an allow-list to all `package.json`.

PR #7865, opened 2019-08-26
UI: Fix TooltipLinkList not scrollable

lucasterra:
Issue: #6150

TooltipLinkList was not scrollable. This was a problem for plugins with many many options. i.e.: a i18n plugin. See screenshots below:

Before:
Screenshot 2019-08-26 at 16 48 59

After:
Screenshot 2019-08-26 at 16 49 30

## What I did

I added a maxHeight of 10.5 items (the half is to make it obvious that there are more items if you scroll down), and made overflow-y: auto.

## How to test

- Try making a plugin with many many options.


shilman (approved):
LGTM and tested!

PR #7939, opened 2019-08-31
Addon API: Preview wrapper devkit example

atanasster:
Issue: 
allow hooks for the UI around the iFrame
create addon that will use the hooks and will display edit this page links

## What I did
1. added two hook types to lib/addons/types.ts
```
  IFRAME_START = 'iframe_start',
  IFRAME_END = 'iframe_end',
```

2. Updated the lib/ui/preview component to display any add ons registered for those hooks
```
const iFrame = (
    <Fragment>
      {iframeStart.map(({ render }) => render())}
      {base}
      {iframeEnd.map(({ render }) => render())}
    </Fragment>
  );
```

## How to test

Please review the functionality as well as the appearance of the default edit page panel - it uses the background color of toolbar


atanasster:
@domyen ,I am attaching screenshots of how the addon looks in Preview and Docs pages both in light and dark theme, if you can please review and let me know your input.

```
const StyledContainer = styled.div<{}>(({ theme }) => ({
  padding: '10px',
  background: theme.background.bar,
  borderBottom: `1px solid ${theme.color.border}`,
}));
```
screenshot808

screenshot809


screenshot810


screenshot811

domyen:
Firstly, this is awesome Atanas! 😍👏

There are a few modalities here that are kind of nuanced.

**Canvas**
This is a component dev environment primarily so having the edit button in the iframe would be pretty distracting and subject to the users global styles. 

image


**Docs**
This is where you can view docs in a new tab while developing your components. I think the above solution would work well too:

image


**Docs build**
This is the static site intended for read-only use that SB docs generates. It has no toolbar. So the inline solution would be the best choice. I'd suggest placing it at the end similar to what Shopify Polaris does [here](https://polaris.shopify.com/components/images-and-icons/avatar#navigation). @shilman do you think this could be injected into a DocsPage slot somehow?

image

shilman:
@domyen looks great. i'm not sure the best way to fit that into DocsPage -- I'll have to chew on it. There's currently no "addons API" for DocsPage, you're getting into uncharted territory here. 😄

atanasster:
Thanks for the feedback @domyen.
@shilman - so should I remove completely the API to insert elements before/after the IFrame? And add an API to docs page to insert elements before/after the docs content?

shilman:
@atanasster I propose a slightly different approach, which is an `IFRAME_DECORATOR` slot, that takes a decorator function, which renders **AROUND** the iframe. That way the user can:
- insert something before or after or both
- insert something around (e.g. to change size, a la viewports addon)

lib/addons/src/types.ts
```diff
@@ -7,6 +7,8 @@ export enum types {
   TOOL = 'tool',
   PREVIEW = 'preview',
   NOTES_ELEMENT = 'notes-element',
+  IFRAME_START = 'iframe_start',
```
ndelangen: Sorry, I don't think this API is a good idea.

I want to be super careful with adding addon types, as we'll be stuck with supporting them forever.

I'd rather have a single IframeWrapper type addon, which I might have a usecase for when i want to extract the zoom functionality into an addon.

I think you can do everything in a wrapper you could do with a pre & post

ndelangen (changes requested):

atanasster:
@ndelangen @shilman - it turns out there was already a `types.PREVIEW` wrapper used by the a11y addon. 

I ended up just adding an example for the preview_wrapper in the dev-kits example.

shilman (approved):
LGTM! @ndelangen ?

ndelangen (approved):

PR #7945, opened 2019-08-31
Story render react hooks

atanasster:
Issue:
`<Story />` issue #7924 rendering stories with React hooks. 
## What I did
Replaced the storyFn instead of direct call, to be rendered as a react element

## How to test

added `reactHook` story to `Stories.stories.tsx` and can be tested in official-storybook

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,6 +5,7 @@
 /addons/actions/ @rhalff
 /addons/backgrounds/ @ndelangen
 /addons/centered/ @kazupon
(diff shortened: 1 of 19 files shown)


ndelangen:
Thanks for this @atanasster !

This also seems to add a new addon to the monorepo,
Would you mind splitting that into a separate PR?

atanasster:
oops, sorry - somehow my other PR is also linked here. Will read up on splitting PRs- do you have a good tutorial to point me to?

atanasster:
splitting this and will send a new PR

PR #7948, opened 2019-09-01
UI: Add shortcuts, events for collapse/expand all

atanasster:
Issue:
ability to collapse and expand all #6157  

note: implemented using events, to avoid moving around the treeviw estate into global state.

## What I did
- added shortcuts:
```
  collapseAll: ['ctrl', 'shift', 'ArrowUp'],
  expandAll: ['ctrl', 'shift', 'ArrowDown'],
```
- the collapse all was added to the nav menu, as this is usually the used one
- collapse all and expand all shortcuts added to the shortcuts page

- added two new events to the list of channel events
`STORIES_COLLAPSE_ALL` and `STORIES_EXPAND_ALL`

- added collapseAll() and expandAll() to the channels API , which emit the above events

- added event receiver to treeview.js and updating the local state for `unfilteredExpanded`. On collapsing all - all elements, except the current story are collapsed.

hope you like it, I used this to learn more of the code, so if its not a priority and you guys don't want to integrate it, that's fine :)


lib/api/src/modules/shortcuts.ts
```diff
@@ -251,7 +254,14 @@ export default function initShortcuts({ store }: Module) {
           fullApi.navigate('/settings/shortcuts');
           break;
         }
-
+        case 'collapseAll': {
```
shilman: Can we use the enums for these?

shilman (approved):
This looks great. You're on 🔥

shilman (approved):
This looks great. You're on 🔥

lib/api/src/modules/shortcuts.ts
atanasster: yes, it would be cleaner, but didn't want to unsettle the rest of the code. I can add it to this PR, or maybe create a new one?

atanasster:
@domyen adding screenshots , would love to get your input.
- The shortcut keys used: Cmd+Shift+Up (collapse) and Cmd+Shift+Down (expand) - chosen not to conflict with default Mac and Win system shortcuts.
- Only collapse all is added (at the very end) to the sidebar menu (expand all is probably not very used in large projects)
- the key shortcuts are updated on the shortcuts page

screenshot812
screenshot813

PR #7957, opened 2019-09-02
Addon-knobs: Add object[] support for select

matthewliuswims:
Issue: #810 (was closed long ago). I may be misinterpreting the post, but when OP said he desired his options to look like below

Screen Shot 2019-09-02 at 12 16 58 AM

I thought that was for (or included) the `select` knob. However, I was surprised to see that the `select` knob still could NOT take options as an array of objects (only an array of strings). Thus....

## What I did

1. I made it so that the `select` knob can take in an array OF objects.
2. modified README to document this change

## How to test

in `official-storybook/stories/addon-knobs/with-knobs.stories.js` I added an example for testing purposes and documentation.


matthewliuswims:
@shilman I was wondering why there is a `do not merge` label. I'm happy to make changes to fix it.

shilman:
@matthewliuswims Sorry I forgot to explain. We're in the final stages of 5.2 release, so we're not merging any features until that's out (Sep 16 target). Making this explicit with the label since we're a distributed team and it's hard to coordinate.

matthewliuswims:
@shilman understood. Thank you :)

matthewliuswims:
@shilman congrats on the 5.2 release! now that it's out, would it be possible to get this PR reviewe d+ remove do not merge tag?

ndelangen (approved):

addons/knobs/README.md
```diff
@@ -307,8 +307,37 @@ const groupId = 'GROUP-ID1';
 const value = select(label, options, defaultValue, groupId);
 ```
 
-> You can also provide options as an array like this: `['red', 'blue', 'yellow']`.
+Options can also be an array:
 
+```js
+import { select } from '@storybook/addon-knobs';
+const label = 'Cats';
+const options = ['linus', 'eleanor', 'lover']
```
carlosaherrera: Shouldn't `options` be an array of Objects that correspond with `arrayOfObjects`? `defaultValue` is `arrayOfObjects[0]`, but that object doesn't match anything within `options`. This appears *broken*, but perhaps I am not understanding something. 

Now, looking at https://github.com/storybookjs/storybook/pull/7957/files#diff-1b10673d7ef02f5f40b685bb83416afeR92, this appears to be configured correctly, where the `arrayOfObjects[0]` can be matched against something within `arrayOfObjects`.

matthewliuswims: yup you got it!

addons/knobs/src/components/types/Select.tsx
```diff
@@ -30,15 +30,23 @@ const SelectType: FunctionComponent<SelectTypeProps> & {
...
-    : (options as Record<PropertyKey, SelectTypeKnobValue>);
 
-  const selectedKey = Object.keys(entries).find(k => {
-    if (Array.isArray(knob.value)) {
-      return JSON.stringify(entries[k]) === JSON.stringify(knob.value);
+  const callbackReduceArrayOptions = (acc: any, option: any, i: number) => {
+    if (typeof option !== 'object') return { ...acc, [option]: option };
+    const label = option.label || option.key || i;
+    return { ...acc, [label]: option };
+  };
```
duhseekoh: Does this mean that my objects need to have a `label` or `key` attribute to have a display value? That's a little funky. The data passed to the react component shouldn't need to include storybook specific labels.

Something like this would keep the storybook integration separate from the data being passed to the component:
```
const arrayOfObjects = [
  {
    label: 'Sparky', // storybook specific label for the option
    data: { // passed to the react component
        dogName: 'Sparky',
        dogParent: 'Matthew',
        location: 'Austin',
    }
  },
  {
    label: 'Juniper',
    data: {
       dogName: 'Juniper',
       dogParent: 'Joshua',
       location: 'Austin',
    }
  },
];
```

matthewliuswims: > Does this mean that my objects need to have a label or key attribute to have a display value?

1. there's a fallback `const label = option.label || option.key || i;` to the index (i.e. `i`). So it'll always work (label might just not look very nice) if a user chooses to not pass a label or key.

2. Also, the `label` or `key` would just need to be passed in on the storybook level (react logic does NOT need to have a label). In fact: you show a great example of how to do this in your code snippet!
```
// using your `arrayOfObjects` example
const dogInfo = select(<label>, arrayOfObjects).data // notice we use.data`
// pass dogInfo to react object.
```

duhseekoh:
Ah, right good point, the knob functions return values can be reached into
/ manipulated before using that data as a prop.
👍

PR #7965, opened 2019-09-02
Addon-docs: Error handling for invalid Story id

atanasster:
Issue: #7964

## What I did
added a check for no children to Story.tsx
added an empty story example

## How to test
```
<Preview>
  <Story id="empty-story" />
</Preview>
```

### DIFF
--- a/examples/official-storybook/stories/addon-docs/addon-docs.stories.mdx
+++ b/examples/official-storybook/stories/addon-docs/addon-docs.stories.mdx
@@ -28,6 +28,10 @@ import DocgenButton from '../../components/DocgenButton';
 
 <Props of="." />
 
+<Preview>
+  <Story id="empty-story" />
+</Preview>
+
 ## A random color ColorPalette
 
 <ColorPalette>
--- a/examples/official-storybook/stories/addon-docs/addon-docs.stories.mdx
+++ b/examples/official-storybook/stories/addon-docs/addon-docs.stories.mdx
@@ -29,7 +29,7 @@ import DocgenButton from '../../components/DocgenButton';
 <Props of="." />
 
 <Preview>
-  <Story id="empty-story" />
+  <Story id="nonexistent-story" />
 </Preview>
 
 ## A random color ColorPalette
--- a/lib/components/src/blocks/Story.tsx
+++ b/lib/components/src/blocks/Story.tsx
@@ -13,7 +13,7 @@ export enum StoryError {
  * if the story id exists, it must be pointing to a non-existing story
  *  if there is assigned story id, the story must be empty
  */
-const MISSING_STORY = (id: string) => (id ? `Story "${id}" doesn't exist.` : StoryError.NO_STORY);
+const MISSING_STORY = (id?: string) => (id ? `Story "${id}" doesn't exist.` : StoryError.NO_STORY);
 
 interface CommonProps {
   title: string;
@@ -22,7 +22,7 @@ interface CommonProps {
 }
 
 type InlineStoryProps = {
-  storyFn: () => React.ElementType;
+  storyFn: React.ElementType;
 } & CommonProps;
 
 type IFrameStoryProps = CommonProps;
@@ -53,27 +53,13 @@ const InlineZoomWrapper: React.FC<{ scale: number }> = ({ scale, children }) => 
   );
 };
 
-/** render of the storyFn function.
- * catch for null function
- * catch for functional component stories using react hooks
- */
-
-const renderStoryFn = (storyFn: () => React.ElementType, id: string): React.ReactNode => {
-  // invalid story function, bail out
-  if (!storyFn) {
-    return <EmptyBlock>{MISSING_STORY(id)}</EmptyBlock>;
-  }
-  // let react do its magic for calling the render
-  // allows using react hooks in story functions
-  const ReactComponent = (storyFn as unknown) as React.ElementType;
-  return <ReactComponent />;
-};
-
 const InlineStory: React.FunctionComponent<InlineStoryProps> = ({ storyFn, height, id }) => (
   <div style={{ height }}>
     <ZoomContext.Consumer>
       {({ scale }) => (
-        <InlineZoomWrapper scale={scale}>{renderStoryFn(storyFn, id)}</InlineZoomWrapper>
+        <InlineZoomWrapper scale={scale}>
+          {storyFn ? React.createElement(storyFn) : <EmptyBlock>{MISSING_STORY(id)}</EmptyBlock>}
+        </InlineZoomWrapper>
       )}
     </ZoomContext.Consumer>
   </div>
--- a/lib/components/src/blocks/Story.tsx
+++ b/lib/components/src/blocks/Story.tsx
@@ -13,8 +13,7 @@ export enum StoryError {
  * if the story id exists, it must be pointing to a non-existing story
  *  if there is assigned story id, the story must be empty
  */
-const WRONG_OR_EMPTY_STORY = (id: string) =>
-  id ? `Story "${id}" links to a nen-existing story` : StoryError.NO_STORY;
+const MISSING_STORY = (id: string) => (id ? `Story "${id}" doesn't exist.` : StoryError.NO_STORY);
 
 interface CommonProps {
   title: string;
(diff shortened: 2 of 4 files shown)


examples/official-storybook/stories/addon-docs/addon-docs.stories.mdx
```diff
@@ -28,6 +28,10 @@ import DocgenButton from '../../components/DocgenButton';
 
 <Props of="." />
 
+<Preview>
+  <Story id="empty-story" />
```
shilman: Should this be `id="nonexistent-story"`? The `id` attribute is to refer to an existing story that's been defined somewhere else in storybook.

lib/components/src/blocks/Story.tsx
```diff
@@ -98,11 +98,14 @@ const IFrameStory: React.FunctionComponent<IFrameStoryProps> = ({
 const Story: React.FunctionComponent<StoryProps> = props => {
   const { error } = props as ErrorProps;
   const { storyFn } = props as InlineStoryProps;
-  const { id, inline, title, height } = props;
+  const { id, inline, title, height, children } = props;
 
   if (error) {
     return <EmptyBlock>{error}</EmptyBlock>;
   }
+  if (!children) {
```
shilman: I think this `Story` never gets any children passed. If you have access to the Chromatic build (linked in the Github status checks) you can see how it breaks some of the examples.

`addon-docs/blocks/Story` parses out the `storyFn` and that gets passed in. So I think the check should be done inside `InlineStory` & `IFrameStory` respectively.

shilman (changes requested):
@atanasster Thanks for this PR! It's the right idea, but I think the implementation needs work. Comments below.

examples/official-storybook/stories/addon-docs/addon-docs.stories.mdx
atanasster: thanks, changed it. I guess I have been looking at empty stories too much :)

lib/components/src/blocks/Story.tsx
atanasster: I might be missing something - I don't see this (but I dont really know the chromatic build). 

I added the fix there (show a "No component or story to display'" message) because I saw also the other error check is being in there. 
```
if (error) {
    return <EmptyBlock>{error}</EmptyBlock>;
  }
```

Also, on my end, it breaks as soon as I add the sample Story from the issue, doesn't to break for you too even outside the chromatic build. And child are passed as null in that case. 
```
<Preview>
  <Story id="nonexistent-story" />
</Preview>
```

shilman: I think something is wrong with your setup. Running this on my local machine matches with Chromatic is showing. Try running:

```
yarn bootstrap --reset --core
```

This will clear out your `node_modules` and reinstall from scratch. It will also delete ALL files in the monorepo that are not in git, so **BE CAREFUL** with this command.

atanasster: you are right that my config did not match yours, here is an update it might help
1. tried to rebuild gave me an error building node_modules/weak
`error /Users/atanasster/storybook/node_modules/weak: Command failed.` from:
`../src/weakref.cc:147:25: error: too few arguments to function call, expected 4, have 3` etc.
2. turns out I had `node --version: v12.1.0`
3. switched down to `node --version v10.16.3`
4. `yarn bootstrap --reset --core` - the build worked
 
No idea how/if the original build worked since I've had node 12 for several months now, maybe building storybook was so new I didn't really pay attention if some parts of the build failed

Now I see the same situation as you do, looks like the issue was introduced with the fix for react functional component stories with hooks, now invalid stories have the issue rendering i will update this PR as soon as I figure out what's happening, just wanted to give a heads up that it seems the build has issues under node 12.xx.

shilman: Hmm maybe somehow `yarn dev` succeeded where `yarn bootstrap` failed? `bootstrap` does more work than `dev`.

atanasster: aaaah, I think you are right - I probably never did a full build

atanasster:
@shilman added a `renderStoryFn` to take care both of invalid storyFn and also catch exception from using react hooks. 
Tried any other recommended method (https://overreacted.io/how-does-react-tell-a-class-from-a-function/) to determine if a function is actually a react component but they didn't seem to work with the empty/invalid story. 
If you have other ideas to try let me know, just this was the best I could come up at this moment.

shilman:
Thanks @atanasster! So would you consider this a bugfix for the hooks support PR? If so, I'll try to get it merged tomorrow my time.

atanasster:
@shilman yes it can be considered a fix for the hooks support and I think it should be merged.
From what I saw, there is a default storyFn that is not a react function/component that was used with the invalid story.
So this is both a safer way to enable react hooks (now using react magic only as a fallback), but also allow for storyFn functions that are not really react, so it should probably be better for non-react frameworks used to write inline stories.

shilman:
@atanasster I tried the latest version on this PR with a clean bootstrap and am seeing the following issues on the Docs tab in `official-storybook`:

1. **Addons|Docs/mdx**: The nonexistent story does not cause the page to crash. However no error is shown either.
2. **Docs|Story**: The docs page crashes. Since this component is special, I'm not sure if this matters.
3. **Others|Hooks**: The docs page crashes. I think this should work.

The error in all crashes is: `Invalid hook call. Hooks can only be called inside of the body of a function component.`

atanasster:
@shilman ouch, very sorry about that. It seems the exception handler might be too specific for your config and you might have a different exception name ` if (e.name === 'Invariant Violations') ` in `Story.tsx` ? Can you try to just comment out the check for the e.name?

atanasster:
Sorry - a misspelling crept in - 'Invariant Violation' was spelled 'Invariant Violations'. Maybe I pressed ctrl+s to save the file or something, because it was working for me before. 
Pushed the fixed spelling now, sorry again.

shilman:
@atanasster no worries. fixing the typos fixes the hooks issues. however, i'm still seeing the following on `Addons|Docs/mdx`:

Storybook

Does it work on your machine?

atanasster:
@shilman glad to hear it fixes the issue. 

As for the empty message, no it doesn't show here, I didn't re-add it (I just added a fix for the exception if storyFn was null) - since the empty messages were in the other file (Story.tsx I think?)instead of the InlineStory?

atanasster:
@shilman I pushed back the empty error message, the UI is a bit bigger, since its enclosed in a `<Preview>`

screenshot819

shilman:
@atanasster Mind including the `storyId` in the error message?

shilman:
Also, unused var in DeepScan 😉

atanasster:
Pushed again, can you check the wording of the message

Removed the unused children /weird thing to say in anything but react programming :)/

Shouldn't the linter catch those btw?

shilman (approved):
@atanasster This is working well for me now, thanks for your patience!

I worry a little bit about the hardcoded hooks error--I wonder why can't we just call `<StoryFn />` (equivalent to `React.createElement`?) every time.

shilman:
Also, why do we need the try/catch at all? Isn't the non-existent story handled by the check at the beginning of the function?

atanasster:
@shilman 
The try catch is currently for react hooks, to fallback on react magic
I also don't like the hardcoded error handling, but the situation made me worried about always using react rendering function and potentially breaking stories that used to work under non-react environments. Tried to find any other way to determine if its actually a react component to be rendered or not, but only the try/catch seemed to always work.
Your call on this one?

shilman:
What happens if we just always run `<StoryFn />`? Does it break other frameworks?

atanasster:
- For just `<StoryFn />` - I can't yet run and test the other frameworks, so  I just chickened out :) Also, it was only needed for React hooks, `storyFn()` works fine for class components and it works fine with storybooks hooks (the hooks example in the official storybook example, where I added a react hooks example that will have problems).

lib/components/src/blocks/Story.tsx
```diff
@@ -10,14 +9,20 @@ export enum StoryError {
...
+
 interface CommonProps {
   title: string;
   height?: string;
   id: string;
 }
 
 type InlineStoryProps = {
-  storyFn: React.ElementType;
+  storyFn: () => React.ElementType;
```
shilman: @atanasster Why not just make this `React.ElementType`? It will clean up the typing below too.

shilman: And you can do everything inline too

atanasster: thanks for the feedback - simplified the code now

shilman:
Awesome!

PR #7991, opened 2019-09-04
Addon-docs: Doc blocks fixes for dark theme

atanasster:
Issue: #7982

## What I did
set color in DocPage to `theme.color.defaultText,` to accommodate for dark theme

## How to test
set `options.theme: themes.dark,` (wait if local storage still not refreshed)

### DIFF
--- a/lib/components/src/typography/DocumentFormatting.tsx
+++ b/lib/components/src/typography/DocumentFormatting.tsx
@@ -1,11 +1,12 @@
 import { styled, CSSObject, Theme } from '@storybook/theming';
 import { withReset } from './withReset';
 
-const headerCommon: CSSObject = {
+const headerCommon = ({ theme }: { theme: Theme }): CSSObject => ({
   margin: '20px 0 8px',
   padding: 0,
   cursor: 'text',
   position: 'relative',
+  color: theme.color.defaultText,
   '&:first-of-type': {
     marginTop: 0,
     paddingTop: 0,
@@ -16,7 +17,7 @@ const headerCommon: CSSObject = {
   '& tt, & code': {
     fontSize: 'inherit',
   },
-};
+});
 
 const withMargin: CSSObject = {
   margin: '16px 0',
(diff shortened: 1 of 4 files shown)


atanasster:
screenshot after change:

screenshot820

shilman (changes requested):
I don't think this is the right solution unfortunately.

We have a complicated situation where we don't want the global styles to interfere with the user's story styles.

As far as I understand, this means that every doc block should be independently themable. So AFAIK the solution is to fix them theming for `Description` and `PropsTable` separately, rather than fix it at the top level as you've don here (although that's simpler).

@domyen can you chime in here?

atanasster:
@shilman Initially I did change the PropsTable and Markdown elements, but then noticed that `DocsWrapper` already has css styling (thats how the background is set to `theme.background.content`), so the color prop does not change the behavior:
```
export const DocsWrapper = styled.div<{}>(({ theme }) => ({
  background: theme.background.content,
  color: theme.color.defaultText,
...
```

ndelangen:
This color property will cascade into user's components, where they might expect it to be the browser default.

I think there should be a place where we set the font-family. Color belongs there too probably.

atanasster:
@ndelangen - `DocsWrapper` was already setting the background color and it was propagating to the user components and other elements of the docs page, causing actually the initial issue with text not visible on dark theme. This PR is only adding the text color prop to match the background.

ISTM that either background should be removed from here or color is rightly placed also here.

shilman:
@atanasster Does the background also propagate like that? Can you double check? I don't think it does.

domyen:
As Michael said we have a complex situation (due to inline rendering + style bleeding). That means we'll need to theme doc blocks independently. 

For DocsWrapper the background will not cascade into child elements but `color` will. I'd suggest adding `color` to each doc block you want to theme.

atanasster:
@domyen - thanks for the input, switched to setting colors for the individual elements (hope I didn't miss too many :)).

It seems that there is an issue with user text in stories once we are not setting it anymore, ie:
```
storiesOf('Basics|Button', module).add('all buttons', () => (
  <div>
    <p>Button that is used for forms</p>
    <FormButton>Form.Button</FormButton>
    <br />
```
will be like that: 
screenshot821

the same on the preview window:
screenshot822

domyen:
Thanks Atanas, unfortunately the first screenshot is an unavoidable situation because we shouldn't actually style any of the users code ourselves. When using the dark theme users need to write their stories to accommodate dark backgrounds.

In the case of SB design system, we wrote our stories for the light theme and supply a background when possible, that's why they look that way. 

It's not perfect as you pointed out, but that's probably something we need to fix in the design system code.

atanasster:
So what about the background from `DocWrapper` ?

The background color is set in multiple places (but not the text color), here is the DOM:
screenshot823

IframeWrapper - outside the iFrame:
```
const IframeWrapper = styled.div(({ theme }) => ({
  position: 'absolute',
  top: 0,
  left: 0,
  bottom: 0,
  right: 0,
  width: '100%',
  height: '100%',
  background: theme.background.content,
}));
```
DocsWrapper - inside the iFrame, also changes the color (but if its only removed then the color stays because of higher-up in hierarchy):
```
export const DocsWrapper = styled.div<{}>(({ theme }) => ({
  background: theme.background.content,
  display: 'flex',
  justifyContent: 'center',
  minHeight: '100vh',
  padding: '4rem 20px',

  [`@media (min-width: ${breakpoint * 1}px)`]: {},
}));
```

lib/components/src/typography/DocumentFormatting.tsx
```diff
@@ -25,24 +25,29 @@ const withMargin: CSSObject = {
 export const H1 = styled.h1<{}>(withReset, headerCommon, ({ theme }) => ({
   fontSize: `${theme.typography.size.l1}px`,
   fontWeight: theme.typography.weight.black,
+  color: theme.color.defaultText,
```
shilman: This can be put in headerCommon

atanasster: thanks, great feedback. Updated `headerCommon` and pushed

shilman (approved):
LGTM!

PR #7993, opened 2019-09-04
Refactor: Use Array.includes instead of indexOf

jamesgeorge007:
Use `includes()` to check for existence.


jamesgeorge007:
not stale

shilman (approved):
LGTM!

shilman (changes requested):
@jamesgeorge007 Do we have the right polyfill for IE?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility

jamesgeorge007:
@shilman I guess we could make use of this polyfill:-
```js
if (!String.prototype.includes) {
  String.prototype.includes = function(search, start) {
    'use strict';
    if (typeof start !== 'number') {
      start = 0;
    }

    if (start + search.length > this.length) {
      return false;
    } else {
      return this.indexOf(search, start) !== -1;
    }
  };
}
```

Is it fine including the respective code-block just above where we use `includes()` :thinking:

ndelangen (approved):

ndelangen:
> @jamesgeorge007 Do we have the right polyfill for IE?
> 
> [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility)

Should be auto polyfilled by core-js

jamesgeorge007:
@ndelangen hope we're all set to land this one :thinking:

ndelangen (approved):

PR #8006, opened 2019-09-05
Addon-knobs: Handle undefined array value

### LINKED ISSUE #8005: liElement Array property breaks storybook knobs when when value is undefined
**Describe the bug**
A story testing a lit Element like this: 
```js
import { LitElement, html } from 'lit-element';

class Test  extends LitElement {

  static get properties() {
    return {
       anArray : {
         type: Array
       }
    };
  }
  
  render() {
    return html`
      <!-- template content -->
      <p>A paragraph</p>
    `;
  }
}

// Register the new element with the browser.
customElements.define('my-test', Test **);**
```

breaks storybook knobs when Array has not been initialized with a value. 

Root cause is knob component Array render function calling `join` on undefined `knob.value`: 
```js
 render() {
    const { knob } = this.props;
    const value = knob.value.join(knob.separator);

    return (
      <Form.Textarea
        id={knob.name}
        name={knob.name}
        value={value}
        onChange={this.handleChange}
        size="flex"
      />
    );
  }
```
Will submit a PR shortly

christophe-g:
Fix #8005

### DIFF
--- a/addons/knobs/src/components/types/Array.tsx
+++ b/addons/knobs/src/components/types/Array.tsx
@@ -62,7 +62,7 @@ export default class ArrayType extends Component<ArrayTypeProps> {
 
   render() {
     const { knob } = this.props;
-    const value = knob.value.join(knob.separator);
+    const value = knob.value && knob.value.join(knob.separator);
 
     return (
       <Form.Textarea


shilman (approved):
LGTM!

PR #8027, opened 2019-09-07
Addon-knobs: allow array values in select and options knobs

emilio-martinez:
Issue: #7600 

## What I did

Allow array values in `select` and `options` knob types. To be clear, changes are only to types, no runtime/functional changes.

## How to test

Added types test cases to validate passing types for both `select` and `options` knob types.


shilman (commented):
LGTM! thanks for the fix!

tinymachine:
This is great, but is there any reason arrays can't be used with the radio and other types of knobs besides select and options?

PR #8036, opened 2019-09-09
Angular: remove webpack-env from tsconfig types

kroeder:
Issue: https://github.com/storybookjs/storybook/issues/7855

## What I did
Removed `webpack-env` from tsconfig.json in `app/angular`. I'm not sure why it was added but it seems to work locally plus it removes the tripple-slash reference path from `dist` which causes the related build issue for all @storybook/angular consumers

## How to test
- Green CI
- Running app/angular locally
- Check if the tripple-slash reference is now removed from `dist/`

### DIFF
--- a/app/angular/tsconfig.json
+++ b/app/angular/tsconfig.json
@@ -3,7 +3,7 @@
   "compileOnSave": false,
   "compilerOptions": {
     "outDir": "dist",
-    "types": ["webpack-env"],
+    "types": [],
     "rootDir": "./src",
     "resolveJsonModule": true
   }


shilman (approved):
LGTM. Do you know why this was needed? @kroeder @ndelangen

app/angular/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compileOnSave": false,
   "compilerOptions": {
     "outDir": "dist",
-    "types": ["webpack-env"],
+    "types": [],
```
ndelangen: If it's working, that's great, happy to remove this.

ndelangen (approved):

sami616:
im seeing the webpack-env build error in my react setup.

PR #8056, opened 2019-09-12
UI: Fix document title not updating

lonyele:
Issue: #866 

## What I did
I extracted storyInfo and passed it to `<Helmet />` so that it can refresh document title.

## How to test

- Is this testable with Jest or Chromatic screenshots? Should later add tests to `lib/ui` package


ndelangen (approved):

lonyele:
@ndelangen  I should have explained more context for better code review experience, I can't really spend much time on opensource right now...(recruiting process...) 
  Just more info, It would have been better if I could find why `selected` is `undefined` and fix this part. Instead, I used `storiesHash` for finding the story info we are looking for. The reason for this implementation is that as I know of `storiesHash` is the core data structure and `selected` is some derived state from clicking or routing. I simply thought using this core data is more reliable in the long run.

ndelangen:
@lonyele thanks for the clarification.

I tested the solution and it worked. I'll revisit the code and the reason why we had to do this at a later stage. Thanks for the time you've invested already. You're welcome back anytime! 💯 👏

PR #8097, opened 2019-09-16
Core: Fix error handling

Hypnosphi:
Issue: #8087 

## What I did
– Added a try/catch
– Added an error boundary for React


app/react/src/client/preview/render.tsx
```diff
@@ -16,37 +14,49 @@ const render = (node: React.ReactElement, el: Element) =>
...
   storyFn: StoryFn,
-  selectedKind,
-  selectedStory,
   showMain,
   showError,
   forceRender,
 }: RenderMainArgs) {
-  const element = <StoryFn />;
-
-  if (!element) {
```
Hypnosphi: This condition is never true

app/react/src/client/preview/render.tsx
```diff
@@ -16,37 +14,49 @@ const render = (node: React.ReactElement, el: Element) =>
...
-      title: `Expecting a React element from the story: "${selectedStory}" of "${selectedKind}".`,
-      description: stripIndents`
-        Did you forget to return the React element from the story?
-        Use "() => (<MyComp/>)" or "() => { return <MyComp/>; }" when defining the story.
-      `,
-    });
-    return;
-  }
-
-  if (!isReactRenderable(element)) {
```
Hypnosphi: Same here

app/react/src/client/preview/render.tsx
```diff
@@ -1,9 +1,7 @@
 import { document } from 'global';
-import React from 'react';
+import React, { ErrorInfo } from 'react';
```
tmeasday: Is this used (`ErrorInfo`)? What version of `react` does it mean we require?

Hypnosphi: not really, forgot to remove (and didn't get caught by eslint because of https://github.com/storybookjs/storybook/pull/6621/files#diff-e4403a877d80de653400d88d85e4801aR118, cc @ndelangen)

shilman (approved):
Great work @Hypnosphi! 🙇 🙇 🙇

KaySackey:
@dependabot ignore this major version

PR #8102, opened 2019-09-17
Typescript: Migrate `@storybook/polymer`

kroeder:
## What I did

Issue https://github.com/storybookjs/storybook/issues/5030


app/polymer/src/server/framework-preset-polymer.ts
```diff
@@ -10,7 +10,7 @@ export function webpack(config) {
         {
           test: /\.html$/,
           use: [
-            ...config.module.rules[0].use,
+            ...(config.module.rules[0].use as RuleSetUseItem[]),
```
kroeder: tslint complained about the typing of `use`. I had to cast it but it might be a bug

`rules[0].use` can be ```    type RuleSetUse =
        | RuleSetUseItem
        | RuleSetUseItem[]
        | ((data: any) => RuleSetUseItem | RuleSetUseItem[]);```

or are we 100% certain it is always `RuleSetUseItem[]`? Do we need to add a check here?

kroeder:
I could use some help migrating the client `storyFn` parts of polymer to typescript 😬 
Can you help or ping someone who can help with that? @ndelangen @shilman

ndelangen:
@daKmoR would you possibly be able to help with this one?

daKmoR:
with `storyFn` you mean what the user has in his `foo.stories.js` files right?

then it would be
```ts
import { TemplateResult } from 'lit-html';

export type StoryFnPolymerReturnType = string | Node | TemplateResult;
```

it's actually also "wrong" in `app/web-components` 🙈 
https://github.com/storybookjs/storybook/blob/next/app/web-components/src/client/preview/types.ts#L3

kroeder:
@daKmoR @ndelangen can you review this, please?

app/polymer/src/server/framework-preset-polymer.ts
ndelangen: We can't be not sure, we should probably check

ndelangen (approved):

shilman (approved):
LGTM. FYI, I'm ok with merging & releasing in 5.3 even tho we're in beta. Are you guys @kroeder @ndelangen @daKmoR ?

ndelangen:
Let's not take the chance @shilman.

Let's make this a 6.0.0 feature

Issue #8101, opened 2019-09-17
Changes in CSF name inference aren't reflected in codemod


Hypnosphi:
Run `sb migrate stories-to-csf` on following file:

```
storiesOf('Story', module)
  .add('camelCase', () => 'foo');
```

Expected:

```
export default {
  title: 'Story',
};

export const camelCase = () => 'foo';

camelCase.story = {
  name: 'camelCase',
};
```

Actual:

```
export default {
  title: 'Story',
};

export const startCase = () => 'foo';

startCase.story = {
  // Redundant, as default algorithm infers that anyway
  name: 'Start Case',
};

// Default algorithm infers "Camel Case" as story name
export const camelCase = () => 'foo';
```

PR #8106, opened 2019-09-17
CLI: Fix variable collisions in storiesof-to-csf

Hypnosphi:
## What I did
– Apply `storyNameFromKey` before deciding whether to add an explicit name #8101
– Prepend `_` if story name starts with a digit
– Check all the variable names (not only exports) when detecting collision. Preserve original name in case of collision

### DIFF
--- a/lib/cli/story.js
+++ b/lib/cli/story.js
@@ -1,0 +1,10 @@
+export default {
+  title: 'Story',
+};
+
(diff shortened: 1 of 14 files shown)


shilman (approved):
This is great @Hypnosphi. Thanks for the fix!!!

PR #8119, opened 2019-09-18
fix: add @types/webpack-env to apps that depend on it, fixes #8110

### LINKED ISSUE #8110: Typescript build fail when removing @types/storybook__react
My understanding is that the `@types` package is no longer needed for `@storybook/react`. However when i uninstall the module i get the below error when trying to build my project.

```
ERROR in [at-loader] ./node_modules/@storybook/react/dist/client/preview/index.d.ts:2:23 
    TS2688: Cannot find type definition file for 'webpack-env'.
```

Have i misunderstood ?

hipstersmoothie:
Issue: fixes  #8110

## What I did

add webpack-env types to react, they are needed for the build

## How to test

### DIFF
--- a/app/preact/package.json
+++ b/app/preact/package.json
@@ -36,6 +36,7 @@
     "@storybook/addons": "5.3.0-alpha.2",
     "@storybook/core": "5.3.0-alpha.2",
     "common-tags": "^1.8.0",
+    "@types/webpack-env": "^1.13.9",
     "core-js": "^3.0.1",
     "global": "^4.3.2",
     "regenerator-runtime": "^0.12.1"
--- a/app/html/package.json
+++ b/app/html/package.json
@@ -34,6 +34,7 @@
   "dependencies": {
     "@storybook/addons": "5.3.0-alpha.2",
     "@storybook/core": "5.3.0-alpha.2",
+    "@types/webpack-env": "^1.13.9",
     "common-tags": "^1.8.0",
     "core-js": "^3.0.1",
     "global": "^4.3.2",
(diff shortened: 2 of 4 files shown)


shilman:
@hipstersmoothie @kroeder Do we need to add these to all the other frameworks as well?

ndelangen:
Yeah I think if we need to add this to 1 package, we need to add them to all that have typescript in them.

hipstersmoothie:
Will do tomorrow!

ndelangen:
Thank you! Really really appreciated!

hipstersmoothie:
@ndelangen I added the types dep to all `app`s that had it in their tsconfig

shilman:
@hipstersmoothie mind fixing the linting error?

> /tmp/storybook/app/preact/src/client/preview/render.tsx
  3:1  error  'common-tags' should be listed in the project's dependencies. Run 'npm i -S common-tags' to add it  import/no-extraneous-dependencies

app/preact/package.json
```diff
@@ -35,7 +35,7 @@
     "@babel/plugin-transform-react-jsx": "^7.3.0",
     "@storybook/addons": "5.3.0-alpha.2",
     "@storybook/core": "5.3.0-alpha.2",
-    "common-tags": "^1.8.0",
```
ndelangen: Did you intent to remove the dependency on `common-tags`

ndelangen (changes requested):

app/preact/package.json
hipstersmoothie: whoops! all fixed now

ndelangen (approved):

PR #8133, opened 2019-09-19
CSF: Allow duplicate kinds

hipstersmoothie:
Issue: 

## What I did

Duplicates story titles used to work and actually work quite easily in the new code.

For example a we have a few components that have have multiple storyfiles.

*Example storybook sidebar:*

```txt
Choice
-- Checkbox
-- Radio
```

In the project these two subcomponent are in separate folders and their stories are in those folders.

In the old format you could:

*Checkbox.stories.js*

```js
storiesOf('Choice', module).add('Checkbox', () => <div />)
```

*Radio.stories.js*

```js
storiesOf('Choice', module).add('Radio', () => <div />)
```

This would produce the structure i mention (`Example storybook sidebar`). 

If I try to switch to the new CSF i get the error `Duplicate title`. It suggests combining the files, but in most cases this make the stories harder to read and less focused, it also wouldn't be able to match the structure we use.

Since it used to be support and works quite well, we shouldn't stop this behavior in CSF.

## How to test

- Is this testable with Jest or Chromatic screenshots? yeah?
- Does this need an update to the documentation? no (well maybe)

### DIFF
--- a/lib/core/src/client/preview/start.js
+++ b/lib/core/src/client/preview/start.js
@@ -336,9 +336,6 @@ export default function start(render, { decorateStory } = {}) {
 
   let previousExports = new Set();
   const loadStories = (loadable, framework) => () => {
-    // Make sure we don't try to define a kind more than once within the same load
-    const loadedKinds = new Set();
-
     let reqs = null;
     if (Array.isArray(loadable)) {
       reqs = loadable;
@@ -392,14 +389,6 @@ export default function start(render, { decorateStory } = {}) {
 
       const { default: meta, ...exports } = fileExports;
       const { title: kindName, parameters: params, decorators: decos, component } = meta;
-
-      if (loadedKinds.has(kindName)) {
-        throw new Error(
-          `Duplicate title '${kindName}' used in multiple files; use unique titles or combine '${kindName}' stories into a single file.`
-        );
-      }
-      loadedKinds.add(kindName);
-
       // We pass true here to avoid the warning about HMR. It's cool clientApi, we got this
       const kind = clientApi.storiesOf(kindName, true);
 


shilman:
This was an intentional constraint that we added, tho I can't entirely remember the rationale. cc @tmeasday

hipstersmoothie:
If it doesn’t break anything I think this is totally a valid pattern. From
the point of view of the user this also makes csf a breaking change. Code
that didn’t throw errror now throws when converted to the new format


A few common uses cases for repeated titles:

- your story file has gotten large and you want to split it up so it’s
easier to consume ( with the new error you have to restructure you
storybook when you do this)

- you have stories about the same thing that require drastically different
decorators and parameters. You could set this on each story but it adds a
lot oof boilerplate. If you allow multiple titles you can group stories
into files by the decorators/parameters they share

tmeasday:
I mean it sounds reasonable on the surface.

@shilman TBH I don't know if I remember discussing multiple files per component, if I recall we felt stronger about disallowing the multiple components in one file use case.

hipstersmoothie:
>  but I could also see people getting pretty confused if the settings they set in one file don't apply globally to the component.

To me it would be kind of odd if those settings became global. But i can see both points of view. If that's how more people see it though i could try to make it work that way. And the add a `isolated` prop to the meta. But I think that's messier that what is here now.

hipstersmoothie:
@ndelangen @shilman @tmeasday Any further though on this? I am currently using patch package to solve this for me but I think it makes sense as the default behavior.

shilman:
@hipstersmoothie have you tested this thoroughly with your patched package? Any problems with HMR?

hipstersmoothie:
We've been using "multiple stories register under the same title" far before CSF and it's always worked. Since upgrading to 5.2 and using my patch (this pr) we have encountered no problems and it still works like it used to

shilman:
@hipstersmoothie the HMR for CSF works fairly differently from storiesOf, so it's possible that "multiple stories under single title" works fine in storiesOf but has problems in CSF. But if you've tested it out in CSF and it's working reliably, then I'm ok with this change.

shilman (approved):
Tested it out and it's looking good!

Hypnosphi:
@hipstersmoothie does this problem sound familiar to you?

– You change the code of `Checkbox.stories.js`
– After HMR, "Choice -> Radio" story disappears from sidebar

shilman:
@Hypnosphi do you think we should revert this change?

Hypnosphi:
@shilman I think, ideally, we should adjust how HMR works in this case. To do that, we need some way to remember, which stories correspond to which kind _declaration_

hipstersmoothie:
@Hypnosphi I do not seem to be experiencing that

initial: 

Screen Shot 2019-10-14 at 11 04 53 AM

change code + save: 

Screen Shot 2019-10-14 at 11 05 02 AM

go back to browser without refreshing:

Screen Shot 2019-10-14 at 11 05 10 AM

shilman:
@hipstersmoothie I want to revert this change in 6.0 and recommend a single story file per kind, and re-exporting stories from other files if you want to break things up. Reverting this is a breaking change at this point, but we're in 6.0 and this has intermittently caused HMR issues for us ever since it was released. WDYT?

cc @ndelangen @tmeasday

hipstersmoothie:
That seems fine to me. That could probably get us around this issue. Does ‘components/bar/examples’ and ‘components/bar/features’ count as different kinds?

hipstersmoothie:
We've also started adding overview files like this

```mdx
import LinkTo from '@storybook/addon-links/react';
import { Meta, Description, Title } from '@storybook/addon-docs/blocks';
import { RelatedComponents } from 'storybook-doc-blocks';

import AppDocs from './App.mdx';
import notes from '../../README.md';

<Meta
  title="Components/App/Overview"
  parameters={{
    notes,
  }}
/>

<Title>@cgds/app</Title>

<RelatedComponents
  components={[
    'Components/Left Navigation',
    'Components/Product Header',
    'Components/Skip To Content',
  ]}
/>

<Description />

<br />
<AppDocs />
```

hopefully this doesn't break

shilman:
Yes, ‘components/bar/examples’ and ‘components/bar/features’ are different kinds. And your 'Components/App/Overview' example should continue to work provided that you don't define 'Components/App/Overview' in another CSF/MDX file.

I'll make the change and it would be great if you could kick the tires when it's read. Issue https://github.com/storybookjs/storybook/issues/11108

Issue #8185, opened 2019-09-24
Suggested TS config does not load files using CSF


good-idea:
**Describe the bug**
When using the suggested configuration in the "[Typescript Config](https://storybook.js.org/docs/configurations/typescript-config/)" section in the docs, it does not discover files that use CSF.

**To Reproduce**
Use the recommended config in `.storybook/config.ts`:

```ts
import { configure } from '@storybook/react';
// automatically import all files ending in *.stories.tsx
const req = require.context('../stories', true, /\.stories\.tsx$/);

function loadStories() {
  req.keys().forEach(req);
}

configure(loadStories, module);
```

Create two story files:

`stories/button-useStory.stories.ts`

```ts
import * as React from 'react'
import { storiesOf } from '@storybook/react'

storiesOf('Button', module)
  .add('With Text', () => <button>Hello Button</button>)
  .add('With Emoji', () => (
    <button>
      <span role="img" aria-label="so cool">
        😀 😎 👍 💯
      </span>
    </button>
  ))
```

`stories/button-csf.stories.ts`

```ts
import * as React from 'react'
import { storiesOf } from '@storybook/react'

storiesOf('Button', module)
  .add('With Text', () => <button>Hello Button</button>)
  .add('With Emoji', () => (
    <button>
      <span role="img" aria-label="so cool">
        😀 😎 👍 💯
      </span>
    </button>
  ))
```

With the default config, you will only see the story created with `storiesOf`.

**Expected behavior**
See both stories appear

**Fix**
Changing to the same config that is recommended in other sections will find both stories:

```ts
import { configure } from '@storybook/react'

configure(require.context('../src', true, /\.stories\.tsx?$/), module)
```

PR #8186, opened 2019-09-24
fix recommended typescript config in docs

good-idea:
Issue: https://github.com/storybookjs/storybook/issues/8185

## How to test

n/a, docs change only

Note: I don't know why the TS config was recommended in a different format in the first place. Someone who knows a little bit more about how this all works should take a look. I can provide a sample repo if needed


shilman (approved):
Thanks @good-idea. This is old documentation that didn't get updated when we changed the story format in 5.2. Thanks for fixing it!

PR #8206, opened 2019-09-26
UI: Move addon dependencies to devDependencies

ndelangen:
Issue: https://github.com/storybookjs/storybook/pull/7413#pullrequestreview-293585776


stof:
should this be done in master rather than next, to have it in a 5.2 release too ?

ndelangen:
That's what the patch label does, this will be picked unto master and released as a 5.2 patch

shilman (approved):
LGTM

PR #8218, opened 2019-09-27
UI: Improve code indentation

ndelangen:
Issue: indentation in code examples is sometimes removed

## What I did
CHANGE common-tags stripIndents to dedent, which is a lot smarter about removing indentation from code samples.

Removes the need for formatter in syntaxhighlighter


shilman (approved):
AWESOME!

PR #8225, opened 2019-09-30
Update index.md in /writing-stories

rwieruch:
Small error in documentation. `fname` wouldn't be defined otherwise.


shilman (approved):
😅

PR #8228, opened 2019-09-30
UPGRADE telejson to new version with typings included and bugs fixed

ndelangen:
## What I did
I upgraded telejson
It now includes types & fixed a few important bugs


shilman (approved):
Awesome!

PR #8234, opened 2019-09-30
FIX a critical api missing

ndelangen:
Issue: going fullscreen is broken, so is hiding the navigator

## What I did
I found there was an api missing on the channel, It was simple to add, I'm not sure why it broke/was removed or exactly what changed. I'll investigate.

Seems to be broken on master
EDIT: I tested master locally, it's OK.

### DIFF
--- a/lib/channels/src/index.ts
+++ b/lib/channels/src/index.ts
@@ -126,6 +126,10 @@ export class Channel {
     this.addListener(eventName, listener);
   }
 
+  off(eventName: string, listener: Listener) {
+    this.removeListener(eventName, listener);
+  }
+
   private handleEvent(event: ChannelEvent, isPeer = false) {
     const listeners = this.listeners(event.type);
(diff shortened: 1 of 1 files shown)


ndelangen:
I found the problem:
https://github.com/atanasster/storybook/commit/076f5dc52c603a769b5b30166d78de34e4a2b688#diff-4dd89de8d806fcb17634d4712c47c416R317-R318

channel.off never existed, it should have been api.off.

Or better yet, useChannel should probably be used, or useEffect

@atanasster @Hypnosphi

atanasster:
Ouch, very sorry about that - i assumed channel.on -> channel.off. Unfortunately i am not near a computer for another week or two to fix it.

ndelangen:
This PR already fixes it, and I made sure it's not on master. So crisis averted.

atanasster:
Great, relieved - sorry again will triple check next time.

PR #8236, opened 2019-10-01
Fix require cycles in addons hooks

benoitdion:
Found a require cycle in addons hooks that was also causing a warning in react-native storybook


shilman (approved):
LGTM

PR #8269, opened 2019-10-02
Angular: Fix angular2-template-loader / raw-loader version conflicts

kroeder:
Issue: https://github.com/storybookjs/storybook/issues/7877

## What I did

As suggested in https://github.com/storybookjs/storybook/issues/7877#issuecomment-536556755 I fixed this issue by modifying the code of `angular2-template-loader`

Though, instead of using `patch-package` I just copied the loader function from the `angular2-template-loader` repository. This package hasn't received any updates in the past 3 years and I contacted the owner to talk about the future of this package.

Also, copying the loader also means we don't have to install another dependency that also needs to be executed as a postinstall script.

## How to test

- Checkout this branch
- Execute the `prepare` script in `app/angular`
- Start the angular example in `examples/angular-cli`
- Go to the `AppComponent` story 
  - If it works you see the initial angular welcome page
  - It does not work if you see an error like this
```
@storybook/angular: Expected 'styles' to be an array of strings
```

## What else

We should make sure this works in Angular 7 and 8 before merging it


kroeder:
Working example: https://monorepo-git-fix-issue7877-webpack-styles-loader.storybook.now.sh/angular-cli/?path=/story/app-component--component-with-separate-template

shilman (approved):
LGTM! 😭

kroeder:
Added `DO NOT MERGE`. I want to improve this weird require string / `length` check

kroeder:
I found this test file in the angular internals https://github.com/angular/angular-cli/blob/0aae1476218c95a38527877e15ae55ba68913400/packages/ngtools/webpack/src/transformers/replace_resources_spec.ts and thought we can maybe just use what angular uses

This worked locally and I even could get rid of two webpack plugins but this is all still just WIP. I need to figure out a couple of things

[ ] Is this backwards compatible, if not, can I make it backwards compatible?
[ ] Can I really remove the `ForkTsChecker` Plugin that decreased the angular build time a lot?
[ ] Can I really remove? 
```
      new ContextReplacementPlugin(
        /@angular(\\|\/)core(\\|\/)(fesm5|bundles)/,
        path.resolve(__dirname, '..')
      ),
```

*Benefits of this approach*
- Getting rid of an unmaintained package (angular2-template-loader) by writing a small loader that uses angular internals 
- Maybe getting rid of additional webpack plugins
- Maybe being more stable that way, depending on if this is backwards compatible

kroeder:
I reverted my new approach. It's unsafe and I couldn't make it compatible for all my test projects. This implementation is a safe fix and solves peoples problems

PR #8246, opened 2019-10-01
Build: Upgrade chromatic

ndelangen:
upgrade chromatic

PR #8280, opened 2019-10-03
Tech/upgrades

ndelangen:
Issue:
I'm trying to rule out that some dependency update caused a visual diff here https://github.com/storybookjs/storybook/pull/8246

## What I did
Tons and tons of dependency upgrades


examples/angular-cli/src/stories/addon-jest.stories.ts
```diff
@@ -2,7 +2,7 @@ import { storiesOf } from '@storybook/angular';
 import { withTests } from '@storybook/addon-jest';
 
 import { AppComponent } from '../app/app.component';
-// eslint-disable-next-line import/no-unresolved
+// eslint-disable-next-line
```
Hypnosphi: Why?

BTW precommit hooks fails here when I try to merge next into my branch

ndelangen:
@Hypnosphi When you merge this in, the merge-commit will be using the new eslint config, but likely you still have an older version installed. I think that mismatch causes the problem.

examples/angular-cli/src/stories/addon-jest.stories.ts
ndelangen: Ow I remember:

Usually this file does NOT exist.. It's a generated file not checked into source-control. We generate it just before we start|build the storybook.

This rule ignores the fact that it's not there.

BUT when it IS there, there's a rule warning us that there's an eslint rule being ignored that isn't being reported. (because now the file is present).

I don't know a good fix for this. Sometimes the file is there, sometimes it isn't. Either cases, we don't want eslint to complain about it.

Would you know a good way of doing that?

Hypnosphi: Fixed with https://github.com/storybookjs/storybook/pull/8289/files#diff-46e956f0103273a39dd3c85b4093053a

Hypnosphi: But I meant, why the wildcard disable instead of particular rules

ndelangen: In the hopes that eslint would then not complain in either case, But I guess it did.

PR #8281, opened 2019-10-03
Don't use `global` package in templates

Hypnosphi:
Issue: We import `global` package in templates for html app, but don't add it as a dependency.

I removed those imports, because our own dev setup preferences shouldn't affect users


shilman (approved):
LGTM

ndelangen (approved):

Issue #8165, opened 2019-09-23
Knobs do not update in Storybook 5.2.0 release candidate, or 5.2.1


rey-wright:
@shilman 

**Describe the bug**
Knobs don't continue to update the component.

**To Reproduce**
https://cbinsights.github.io/form-design-system/fds-components/?path=/story/interactive-button--knobs

Go to the button and toggle the "disabled" knob on and off.

Our design system is thankfully open source, so you can try this out locally too:

https://github.com/cbinsights/form-design-system

Clone the above repo, and run `yarn && yarn build:full && yarn storybook` to install all dependencies and run storybook, and try to disable that button, same as on that URL above.

Now I did a bisect on this, and was able to nail down the commit that introduces this bug: `3fa319cb7bef80a82be1a1a462f7962b0dbc8379` which is the storybook upgrade to the release candidate 5.2.0 (note: updating to 5.2.1 does not fix this either)

So if you go to the commit prior: `be0ead9857a40e2c720f0b8d162dc40adfa77743` and `yarn install --force`, you'll notice that everything fixes itself. I believe this is pretty much a smoking gun (to be fair, maybe I'm doing something that isn't supported in newer versions of Storybook too).

**Expected behavior**
It's expected for the knobs to continue to update... they do not.

(final note: none of the errors in the console.log are the direct result of this problem. I've fixed a lot of those locally and the problem still occurs).

**Screenshots**
N/A

**Code snippets**
N/A

**System:**
```
Environment Info:

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
  Binaries:
    Node: 10.16.0 - /usr/local/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 77.0.3865.90
    Firefox: 67.0.1
    Safari: 12.1.1
  npmPackages:
    @storybook/addon-actions: ^5.2.0-rc.11 => 5.2.0-rc.11 
    @storybook/addon-centered: ^5.2.0-rc.11 => 5.2.0-rc.11 
    @storybook/addon-knobs: ^5.2.0-rc.11 => 5.2.0-rc.11 
    @storybook/addon-links: ^5.2.0-rc.11 => 5.2.0-rc.11 
    @storybook/addon-notes: ^5.2.0-rc.11 => 5.2.0-rc.11 
[...]

PR #8287, opened 2019-10-03
Addon-knobs: Fix issues caused by rerenders

Hypnosphi:
Issue: https://github.com/storybookjs/storybook/issues/8165

## What I did
– Made knob options update after each call, so that e.g. Button knob always uses the latest callback reference with latest closure
– Made sure that effects aren't duplicated because of React rerenders or a decorator calling `storyFn` more than once
– Made sure that there is at most one render subscription in `hooks.ts` at a time

## How to test

Open [Button With React Use State](https://monorepo-git-fix-button-knob-use-state.storybook.now.sh/official-storybook/?path=/story/addons-knobs-withknobs--button-with-react-use-state) story, click increment and decrement buttons in Knobs panel
Open [With Decorator Calling Story Function More Than Once](https://monorepo-git-fix-button-knob-use-state.storybook.now.sh/official-storybook/?path=/story/addons-knobs-with-decorators--with-decorator-calling-story-function-more-than-once) story, change text via knob a couple of times

### DIFF
--- a/addons/knobs/src/KnobManager.test.js
+++ b/addons/knobs/src/KnobManager.test.js
@@ -26,6 +26,7 @@ describe('KnobManager', () => {
       beforeEach(() => {
         testManager.knobStore = {
           set: jest.fn(),
+          update: jest.fn(),
           get: () => ({
             defaultValue: 'default value',
             name: 'foo',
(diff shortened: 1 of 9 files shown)


shilman (approved):
Phew! Looking good! 😅

PR #8289, opened 2019-10-03
Fix local `yarn lint`

Hypnosphi:
## What I did
– Tweaked `react/sort-comp` config to match the current repo state as much as possible
– Added a couple of `.eslintignore` files for stuff that's only present locally


ndelangen (approved):
Awesome! We could move as much of this config as possible into `@storybook/eslint-config-storybook`

PR #8300, opened 2019-10-05
Remove `weak` dependency for Node 12 compat

gaetanmaisse:
Issue: https://github.com/storybookjs/storybook/issues/8002

## Context

To give some context about weak dependency, it's used only for jest related stuff and only affects some developers. The good news is Jest has been working on that and the next release of Jest will certainly allow us to remove weak from our deps.
See facebook/jest#8686

## What I did

I removed `weak` dependency from SB project.

--

----

_OLD DESCRIPTION_

This is sadly not a fix for https://github.com/storybookjs/storybook/issues/8002 but a workaround to force dev to use Node <12 as long as the issue with `node-weak`  is not fixed.

What I did

I updated `engine` property of root `package.json` to force node to be between version 8 and 11 as `node-weak` dependency is not working on Node 12: https://github.com/TooTallNate/node-weak/issues/99
This node max version can be removed as soon as we found a fix/workaround to `node-weak` and Node 12 incompatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,8 +216,7 @@
     "trash": "^6.0.0",
     "ts-dedent": "^1.1.0",
     "ts-jest": "^24.0.2",
-    "typescript": "^3.4.0",
-    "weak": "^1.0.1"
+    "typescript": "^3.4.0"
   },
   "engines": {
     "node": ">=8.6.0",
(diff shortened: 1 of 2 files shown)


ndelangen:
I think we could just lazy install the `weak` dependency

gaetanmaisse:
To give some context about `weak` dependency, it's used only for jest related stuff and only affect some developers. The good news is Jest has been working on that and the next release of Jest will certainly allow us to remove `weak` from our deps. 
See https://github.com/facebook/jest/pull/8686

ndelangen:
"some developers" === @ndelangen 

Just remove the dependency, I'll add it when I need it.

shilman (approved):
LGTM!

PR #8310, opened 2019-10-07
Remove redundant dependency on hoist-non-react-statics (#6349)

mbpreble:
Issue: #6349 

## What I did
Removed explicit dependency of `addons/a11y` upon `hoist-non-react-statics`.
This was originally added to fix a build issue. The dependency `@types/react-redux` which necessitated the installation of `hoist-non-react-statics` has since added it as a direct dependency: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33690.

This is now being resolved through the `@types/react-redux` dependency and does not need to be included explicitly.

## How to test

Testable via successful CI build.

### DIFF
--- a/addons/a11y/package.json
+++ b/addons/a11y/package.json
@@ -41,7 +41,6 @@
     "axe-core": "^3.3.2",
     "core-js": "^3.0.1",
     "global": "^4.3.2",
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM!

PR #8346, opened 2019-10-08
Fix React typings

MichaelDeBoey:
Just like @gaetanmaisse suggested (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38961#issuecomment-539646061), I updated all React imports from `import React from 'react'` to `import * as  React from 'react'`, but I also tried to import as much as possible so the `* as React` part isn't necessary.

I left the stories and tests just like they were.

### DIFF
--- a/addons/a11y/src/components/ColorBlindness.tsx
+++ b/addons/a11y/src/components/ColorBlindness.tsx
@@ -1,5 +1,5 @@
 import { document } from 'global';
-import React, { ReactNode, useState } from 'react';
+import React, { FunctionComponent, ReactNode, useState } from 'react';
 import memoize from 'memoizerific';
 import { styled } from '@storybook/theming';
 
@@ -79,7 +79,7 @@ const getColorList = (active: string | null, set: (i: string | null) => void): L
   })),
 ];
 
-export const ColorBlindness: React.FC = () => {
+export const ColorBlindness: FunctionComponent = () => {
   const [active, setActiveState] = useState(null);
 
   const setActive = (activeState: string | null): void => {
--- a/addons/a11y/src/components/Report/HighlightToggle.tsx
+++ b/addons/a11y/src/components/Report/HighlightToggle.tsx
@@ -1,5 +1,5 @@
 import { document } from 'global';
-import React, { Component } from 'react';
+import React, { Component, createRef } from 'react';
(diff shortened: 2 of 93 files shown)


MichaelDeBoey:
@gaetanmaisse Can you point me in the right direction for fixing the errors please?
It seems like they're not related to this PR.

ndelangen (approved):

ndelangen:
I'll take a look at why this fails

MichaelDeBoey:
Thanks @ndelangen!

I see errors about missing `@emotion/native` types, but these were already missing before this PR

ndelangen:
The problem does not exist on the next branch.

It is something that has changed in this PR that is causing this.

```
declare module '@emotion/native';
```

Is still in `typings.d.ts` so I'm confused why it's not picking this up.

MichaelDeBoey:
Hmmm very strange indeed then.

The only thing I've done in this PR changing all `React.*` calls and just import the things directly.

`React.createRef` -> `createRef`

addons/a11y/src/components/ColorBlindness.tsx
```diff
@@ -1,5 +1,5 @@
 import { document } from 'global';
-import React, { ReactNode, useState } from 'react';
+import React, { FunctionComponent, ReactNode, useState } from 'react';
```
gaetanmaisse: With `import React, { FunctionComponent, ReactNode, useState } from 'react';` we are still importing React using default syntax I think you should use `import * as React from 'react';` and `React.FunctionComponent` etc everytime `import React from 'react';` is still needed.

gaetanmaisse: hmmm, in fact, is importing React still needed?

MichaelDeBoey: It's indeed needed for the TSX files.

ndelangen:
I think I fixed it

MichaelDeBoey:
Hi @ndelangen!

Thanks for fixing types!

Looking at your fixes, these weren't PR related were they? 🤔

ndelangen:
Well something caused typescript to stop loading in the `typings.d.ts`.

I found a way to include it back, and found some improvements to make typing stricter in the RN app.

PR #8348, opened 2019-10-08
Remove spurious flag to build in Ember

backspace:
There’s no need to specify a port to only build, and
doing so produces this error:

```
error: unknown option `-p'
```

Thanks for all the work on this!


shilman (approved):
Good catch! 🚀

PR #8363, opened 2019-10-09
Stop inserting HTML tags from error messages

hasparus:
Issue: https://github.com/storybookjs/storybook/issues/8362

## What I did

I have added `escapeXML: true` option to AnsiToHtml constructor call.

### Before
storybook-html-1

### After
storybook-html-2

## How to test

It's testable with Jest and Chromatic. I didn't add any test yet, because tests on next fail on my machine (Windows 😢) and to run them at all I had to modify https://github.com/storybookjs/storybook/blob/746e98f412ba59e6136a6de808c74a0598c17c27/scripts/test.js#L170 to `const jest = path.join(__dirname, '..', 'node_modules', 'jest', 'bin', 'jest.js');`
~I'm gonna clone the repo on other OS tomorrow and add one.~
Edit: ☝ this is harder than I expected. I'm _trying_ to add a test, I'm gonna commit once I get it done I suppose


lib/core/src/client/preview/start.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import deprecate from 'util-deprecate';
 import AnsiToHtml from 'ansi-to-html';
+import escape from 'lodash/escape';
```
hasparus: Oops. This is left from my previous attempt at tackling the problem. Gonna remove it in a sec.

ndelangen (approved):

PR #8404, opened 2019-10-13
Docs: Removed condescending language

### LINKED ISSUE #8394: More inclusive language in docs
:wave: Sorry I know I'm ignoring all of your well-thought out templates, but I wasn't sure where this would land! Here me out...

I noticed that there are many instances of the word "simply" in your documentation. Unfortunately, words like this can make people feel frustrated and isolated while reading (myself included, hence the issue) - especially if they are facing issues.

If it's alright with your team, I'd like to take a shot at updating the content to make it more inclusive by removing instances of "simply", "easy", "just", etc. I used Storybook _all_ the time at my last job, so I'm happy to help 😊  

Also if you're interested or want more information on this topic, I'd highly recommend watching Jim Fisher's [Don't Say Simply](https://www.youtube.com/watch?v=gsT2BBWBVmM) talk from Write the Docs Prague 2018 ✍️ 🤓 

Thank you for the work you do and please let me know if you need more details!

carolstran:
Issue: #8394 

## What I did

As discussed in the issue, this PR attempts to remove as many instances of condescending or isolating language (ie. simply, easy, just, of course, etc) as possible in the documentation. I updated all versions that were flagged in my search. While most of the changes are removing the word, some involved altering the text. 

Feel free to revert or ask me to fix changes as you see fit. If even a fraction of these changes can be merged then it will be a huge step towards making the documentation more inclusive!

## How to test

Only text changes, so as long as the documentation still builds and renders - I'd say we're good 👐


CONTRIBUTING.md
```diff
@@ -229,7 +229,7 @@ We use the following label scheme to categorize issues:
 
 All issues should have a `type` label. `bug`/`feature`/`question`/`discussion` are self-explanatory. `dependencies` is for keeping package dependencies up to date. `maintenance` is a catch-all for any kind of cleanup or refactoring.
 
-They should also have one or more `area`/`status` labels. We use these labels to filter issues down so we can easily see all of the issues for a particular area, and keep the total number of open issues under control.
+They should also have one or more `area`/`status` labels. We use these labels to filter issues down so we can see all of the issues for a particular area, and keep the total number of open issues under control.
```
tmeasday: Do you think this instance is unhelpful @carolstran? Although I am onboard with this PR generally (and personally am constantly fighting my inclination to use this unhelpful kind of language), I think in this case the "easily" is actually expressing something with some meaning.

Not to nitpick, it's not a big deal if we drop it, I'm just interested.

tmeasday (approved):
Generally these changes tighten the text up in any case so I'm all for it. @shilman I think you should do a pass over it.

shilman (approved):
LGTM. Thanks for tightening things up @carolstran!

CONTRIBUTING.md
carolstran: My working theory is to always assume that things aren't easy - so even if you're using labels for this purpose, someone who is brand new to GitHub, open source, etc. may not be able to do the same. However, reading this instance back, this statement is talking about (I'd assume) the maintainers who a task like this would be easy... so I think you're right that this works here.

The PR is already merged, but if you'd like me to add it back - I could do that! Thanks for flagging because context and who you're addressing in each statement is important in these cases.

tmeasday: Really, as I said I don't think in this particular instance it really matters, I was just more interested in the thought process. I guess it is true that there's no need to make a new maintainer feel like they are not one of "us" (i.e. the "we" in the sentence) because they don't feel it is easy to do what "we" do.

Probably "more easily" or "so we have a way to ..." would make more sense in this one case.

PR #8409, opened 2019-10-13
Update iframe.js to fix disappearing component caused by zoom 

CodeByAlex:
Issue: #7167

## What I did
Fix the issue where zoom functionality mixed with addon centered causes the component to disappear. This fix works for all browsers except Firefox.

@shilman, this was the solution we discussed a while back. If we find a solution that doesn't become impacted by multiple transforms and works for all browsers, this can be replaced at that time.

## How to test

- Is this testable with Jest or Chromatic screenshots? maybe


CodeByAlex:
looks like my branch name is causing issues with circleci - if necessary, I can rename and put out a new PR

PR #8416, opened 2019-10-14
React: Move Create React App support to external preset

mrmckeb:
Issue: #7221

## What I did

As we prepare for a final release of [`@storybook/preset-create-react-app`](https://github.com/storybookjs/presets/tree/master/packages/preset-create-react-app), Storybook will now install the preset and use that instead of the in-built preset.

The plan is to disable the in-built preset with Storybook 6.0.

## How to test

This should pass existing CRA tests.


shilman (approved):
Great job @mrmckeb -- this looks spectacular 👏 👏 👏

PR #8442, opened 2019-10-16
Addon-centered: Fix zoom issues for non-Firefox browsers

CodeByAlex:
Issue: #7167 

## What I did
Fix the issue where zoom functionality mixed with addon centered causes the component to disappear. This fix works for all browsers except Firefox.

@shilman , This was the solution we discussed a while back. If we find a solution that doesn't become impacted by multiple transforms and works for all browsers, this can be replaced at that time.

## How to test

### DIFF
--- a/lib/ui/src/components/preview/iframe.js
+++ b/lib/ui/src/components/preview/iframe.js
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types';
 
 import { styled } from '@storybook/theming';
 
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM! Will do a sanity check before I merge

CodeByAlex:
@shilman after reading through your comment again, I may have merged before you got a chance to do your sanity check. Please let me know if I should revert and I will do so.

PR #8445, opened 2019-10-16
Addon-docs: Fix React.forwardedRef/memo props

darmawanalbert:
Issue: #7933 

## What I did
Fix Props block rendering issue when using `React.forwardRef` by checking whether the target should be calling `render()` first to get the list of `propTypes` or not.

## How to test

### DIFF
--- a/addons/docs/src/blocks/Props.tsx
+++ b/addons/docs/src/blocks/Props.tsx
@@ -27,11 +27,13 @@ export const getPropsTableProps = (
     const params = parameters || {};
     const { component, framework = null } = params;
 
-    const target = of === CURRENT_SELECTION ? component : of;
+    let target = of === CURRENT_SELECTION ? component : of;
     if (!target) {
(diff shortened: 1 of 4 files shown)


shilman (approved):
Thank you @darmawanalbert -- this is really awesome!! 

This code makes sense, but since it's React-specific, would it be possible to put it inside the `getPropDefs` function returned from `inferPropDefs`?

Also, if it's a similar fix, would you mind also adding `React.memo` support and test case while you're at it?

shilman (changes requested):
Meant to request changes in my prev review :)

darmawanalbert:
@shilman sure, let me try it on `React.memo()`

darmawanalbert:
@shilman I have added another logic to fix #6298 and move it to `getPropDefs` function 🙇

shilman (approved):
This is freakin' awesome @darmawanalbert! Thanks so much for the fix!!!

PR #8454, opened 2019-10-17
UI: Remove emotion from NoDocs.js

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/8284

## What I did
I REMOVED the usage of emotion in the places where it's not allowed


shilman (approved):
Thanks for the fix @ndelangen! Any idea how to prevent this from happening again?

brendonco:
> Thanks for the fix @ndelangen! Any idea how to prevent this from happening again?

Should we add a test while building to check whether error below exist?

```
TypeError: Object(...) is not a function
```

PR #8491, opened 2019-10-20
React-native-server: Convert port to number in CLI options

96f3e4c:
Issue: #8487 
The issue is caused by comparison operation between string and number ie: port specified via CLI and available port

## What I did
Converting port type to number in CLI options parser

## How to test
This will fix the above-mentioned issue


shilman (approved):
LGTM! Thanks for the fix

Issue #639, opened 2016-12-05
Get features from CRA v0.8.0


arunoda:
CRA v0.8.0 just got released. So, we need to get some features and it ships.

* [x] [Update CSS Loader](https://github.com/storybooks/react-storybook/pull/648): https://github.com/facebookincubator/create-react-app/pull/1057
* [x] [Treat unknown file extensions](https://github.com/storybooks/react-storybook/pull/653) as input for url loader: https://github.com/facebookincubator/create-react-app/pull/1059
* [x] [Support yarn](https://github.com/storybooks/getstorybook/pull/24) in `getstorybook` if it's available in the system. - https://github.com/facebookincubator/create-react-app/pull/898

PR #8498, opened 2019-10-21
Core: Allow custom postcss config

thollander:
- Search for the nearest postcss config
- Maintain previous defaults if no postcss config

Issue: #2455
Old PR: #4411

## What I did
- Add `find-up` to dependencies to search for the nearest postcss.config.js
- Use the nearest postcss config (`.postcssrc.json`, `postcss.config.js`...) to determine postcss-loader options if present, maintain previous defaults if not.

## How to test
Run the example `storybook/examples/postcss-kitchen-sink`. 
The `h1` should have the `color: #639;` instead of `color: rebeccapurple;`.

No
I created one, not sure if it was mandatory.
Yes, I added a documentation in `/custom-postcss-config`


shilman (commented):
Thanks for putting this together @thollander !

A few thoughts:
- Would it be possible to add the example to an existing `examples/` project rather than creating a new one? I don't want to increase our build times if possible.
- Right now `postcss` support is implemented in core, but I'm wondering if we can move it into a preset? Any chance you might be able to take this on? I'd like to move as much as possible out of core to make the default storybook experience lean and mean. This would be a breaking change, so the first step would be to move this into a preset and always include that preset. In a subsequent release (6.0?) we'll make it easy for users to opt-in to those presets

shilman (approved):
Ace 🎯

thollander:
> Would it be possible to add the example to an existing examples/ project rather than creating a new one? I don't want to increase our build times if possible.

Yes for sure, I did it. You’re totally right it is not necessary to increase the build time just for this. 

> Right now postcss support is implemented in core, but I'm wondering if we can move it into a preset? Any chance you might be able to take this on? I'd like to move as much as possible out of core to make the default storybook experience lean and mean. This would be a breaking change, so the first step would be to move this into a preset and always include that preset. In a subsequent release (6.0?) we'll make it easy for users to opt-in to those presets

I get your point but not sure how to implement it yet. I’ll give it a try. 
If you’re ok with theses changes you can still merge this PR. 
If I succeed to manage as a preset I would open another PR.

PR #8506, opened 2019-10-21
Automate codemod tests creation and update

Hypnosphi:
Issue: There's no easy way to update `.output.js` files

## What I did
Used `jest-specific-snapshot` to turn those into Jest snapshots


shilman (approved):
This is awesome. I'll remember to use this next time!

PR #8545, opened 2019-10-23
Increase TooltipLinkList max-height to accommodate more links

KuSh:
Issue: #8509 

## What I did
Increase the max-height value to accommodate up to 20 items

## How to test

```shell
yarn storybook
```

## Note
The same hard limit also exists in docs/src/new-components/basics/tooltip/TooltipLinkList.js but it doesn't seem to be used (at least with a fast search in source code) so I didn't touch it


shilman (approved):
Nice one! Thanks for picking this up!!!

PR #8549, opened 2019-10-24
Remove outdated and unused `jsx-to-string` library

DiscoStarslayer:
Issue: The info storybook addon has a dependency for `jsx-to-string`. Looking at this module upstream, it seems like it hasn't been updated since May 2018, and has it's own dependencies for an outdated react release, 0.14.0.

Because of this older react dependency, when installing the `@storybook/addon-info` module with yarn you'll see the following warning:
```
warning @storybook/addon-info > jsx-to-string > react > fbjs > core-js@1.2.7: core-js@<2.6.8 is no longer maintained. Please, upgrade to core-js@3 or at least to actual version of core-js@2.
```

It seems like this dependency is not only outdated, but also unused. [A full code search for this module being imported reveals nothing.](https://github.com/storybookjs/storybook/search?q=jsx-to-string&unscoped_q=jsx-to-string) Also looking at the current code, it seems like `react-element-to-jsx-string` was added to replace `jsx-to-string` which helps to confirm this.

## What I did
Removed the unused dependency, updated the `yarn.lock` to comply with this change, and ran the full test suite as a smoke test.

### DIFF
--- a/addons/info/package.json
+++ b/addons/info/package.json
@@ -35,7 +35,6 @@
     "@storybook/theming": "5.3.0-alpha.26",
     "core-js": "^3.0.1",
     "global": "^4.3.2",
-    "jsx-to-string": "^1.4.0",
     "marksy": "^7.0.0",
     "nested-object-assign": "^1.0.3",
     "prop-types": "^15.7.2",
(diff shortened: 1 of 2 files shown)


shilman (approved):
Good catch! LGTM!

PR #8562, opened 2019-10-24
Added example of a decorator in Storybook config

ajones55555:
There was previously text on this page that recommends adding decorators to the Storybook config, but doesn't show how to do that. I was confused about this at first, so when I figured it out I decided to add it to the docs.

Issue:

## What I did
Added an example to the docs

## How to test
N/A


shilman (approved):
LGTM

PR #8584, opened 2019-10-27
React-native-server: Changed default port to number in CLI options

96f3e4c:
Issue: related to #8487
While fixing the above issue in #8491 I accidentally introduces same bug for default port. This PR is a fix for that

## What I did
Changed default port from string to number


shilman (approved):
🙈

PR #8588, opened 2019-10-27
Typescript: Use native package types now available

ndelangen:
minor cleanup, these packages have types now, and so putting them in the typings.d.ts removes types


shilman (approved):
LGTM! 🚀

PR #8598, opened 2019-10-28
Addon-docs/web-components: Add attributes to props table

steel:
Issue:

The new web-components framework does not render attributes in the doc block.

## What I did

I ran web-component-analyzer to get the most up to date custom-element.json file and added support for `attributes` when extracting props.

## How to test

Here is what web components kitchen sink app now renders:

Addons | Docs - Simple ⋅ Storybook 2019-10-28 15-16-58

### DIFF
--- a/examples/web-components-kitchen-sink/custom-elements.json
+++ b/examples/web-components-kitchen-sink/custom-elements.json
@@ -4,28 +4,39 @@
     {
       "name": "demo-wc-card",
       "description": "This is a container looking like a card with a back and front side you can switch",
-      "properties": [
+      "jsDoc": "/**\n * This is a container looking like a card with a back and front side you can switch\n *\n * @slot - This is an unnamed slot (the default slot)\n * @fires side-changed - Fires whenever it switches between front/back\n * @cssprop --demo-wc-card-header-font-size - Header font size\n * @cssprop --demo-wc-card-front-color - Font color for front\n * @cssprop --demo-wc-card-back-color - Font color for back\n */",
+      "attributes": [
+        {
+          "name": "back-side",
+          "type": "boolean"
+        },
         {
           "name": "header",
-          "type": "String",
-          "attribute": "header",
-          "description": "Shown at the top of the card",
-          "default": "Your Message"
+          "type": "string"
         },
         {
           "name": "rows",
-          "type": "Array",
-          "attribute": "rows",
-          "description": "Tabular data shown on the back of the card",
-          "default": []
+          "type": "never[]"
+        }
+      ],
+      "properties": [
+        {
+          "name": "side",
+          "description": "A card setter can have side A or B",
+          "jsDoc": "/**\n   * A card setter can have side A or B\n   *\n   * @param {(\"A\"|\"B\")} value\n   */",
+          "type": "\"A\" | \"B\""
         },
         {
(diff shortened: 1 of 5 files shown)


daKmoR:
`customElements.json` was handwritten before 🙈  - for testing purposes 😅  
awesome that auto-generation works that well 👍 

let's make it fully nice then 💪 
- remove `set side` and `get side` it's outdated code
- add a description to properties `backSide`, `header`, `rows` via the constructor like

```js
/**
 * Indicates that the back of the card is shown
 */
this.backSide = false;
```

`default` is yet not supported by the analyzer but it's ok 😅 

@runem be prepared for all the storybook users 💪  and yeah `default` and attribute on the properties would be nice 😬

shilman (approved):
This is pretty awesome guys 😻  - thanks @steel!!

examples/web-components-kitchen-sink/custom-elements.json
```diff
@@ -4,28 +4,51 @@
...
-          "description": "Tabular data shown on the back of the card",
-          "default": []
+          "description": "Data rows",
+          "jsDoc": "/**\n     * Data rows\n     */",
+          "type": "never[]"
+        }
+      ],
+      "properties": [
+        {
+          "name": "side",
```
daKmoR: we should get rid of this by removing the setter/getter for `side` 🤗

steel: I left this in there because it triggers the `side-changed` event which is documented by the `@fires side-changed`. This gives an example of all the things the `custom-elements.json` file can provide. 

I can still remove it though if you want.

daKmoR: ah right... for the event let's better do this 🤗 
```js
  set backSide(value) {
    this.__backSide = value;
    this.dispatchEvent(new Event('side-changed'));
    this.requestUpdate();
  }

  get backSide() {
    return this.__backSide;
  }
```

sorry for the many change request 🙈

daKmoR: dam this "disables" the attribute sync...

daKmoR: that works
```js
  _requestUpdate(name, oldValue) {
    // @ts-ignore
    super._requestUpdate(name, oldValue);

    if (name === 'backSide') {
      this.dispatchEvent(new Event('side-changed'));
    }
  }
```

but requires the private api `_requestUpdate`... 

so I'm not sure what is best in that case... feel free to choose 👍

steel: I think I have a solution, also fixed some css issues.

steel:
Addons | Docs - Wrapper ⋅ Storybook 2019-10-29 11-33-15

The buttons weren't clickable and the backside wasn't rendering properly for me in Safari. This fixes that.

examples/web-components-kitchen-sink/custom-elements.json
daKmoR: awesome 👍

daKmoR (approved):

ndelangen (approved):

daKmoR:
thaankkk you 🤗

PR #8625, opened 2019-10-29
UI: Fix unmount components on Canvas/Docs tab switch

patricklafrance:
Issue: https://github.com/storybookjs/storybook/issues/8622

React component are not unmount when switching from "canvas" to "docs"

## What I did

Here's a few observations before I updated any code:

- When switching between 2 stories in canvas mode, the previous story components will unmount once and the new story components will mount once
- When switching between 2 stories in docs mode, the previous story components will unmount (and then mount again?) and the new story components will mount
   - This is expected that the previous story component mount again since docs aggregates all the stories
- When switching from canvas to docs for the same story, the components will only mount once, unmount never happened.
- When switching from docs to canvas for the same story, the docs components will unmount twice and will then mount once.

Here's the same observations after I updated the code:

- When switching between 2 stories in canvas mode, the previous story components will unmount once and the new story components will mount once
- When switching between 2 stories in docs mode, the previous story components will unmount (and then mount again?) and the new story components will mount
- **When switching from canvas to docs for the same story, the components will unmount once and then mount once**
- **When switching from docs to canvas for the same story, the docs components will unmount once and will then mount once.**

Basically,  all I did is call `unmountComponentAtNode` on the `root` element when there is a switch between story and docs.

A few questions:

- The `docs-root` element is always empty, what's is purpose? Initially I though the fix would be to unmount the components in this element but the docs is always rendered in the `root` element, even with `--docs`.

- Should we instead unmount the components when the hooks are cleaned? (THIS IS WHAT I END UP DOING)


shilman:
@patricklafrance I think perhaps you're looking at the contents of `docs-root` **after** this code has already been run:

https://github.com/storybookjs/storybook/blob/next/lib/core/src/client/preview/start.js#L227

It definitely has content. I'm wondering if, instead of what's in this PR, you just need to add the following line to the `docs` case of the switch statement I just linked to:

```js
  ReactDOM.unmountComponentAtNode(document.getElementById('docs-root'));
```

What do you think? Can you try it out & test it?

shilman (changes requested):
@patricklafrance Thanks for putting together this fix!!!

However, I think perhaps you're looking at the contents of `docs-root` **after** this code has already been run:

https://github.com/storybookjs/storybook/blob/next/lib/core/src/client/preview/start.js#L227

It definitely has content. I'm wondering if, instead of what's in this PR, you just need to add the following line to the `docs` case of the switch statement I just linked to:

```js
  ReactDOM.unmountComponentAtNode(document.getElementById('docs-root'));
```

What do you think? Can you try it out & test it?

patricklafrance:
@shilman 

It really seems empty.

image

If you mean this `docs` case, it was already there and it's not working:

image

All I get is a "Mount".

I tryed the following and it's still not working:

image

patricklafrance:
The following is working thought:

image

I changed:

```
ReactDOM.unmountComponentAtNode(document.getElementById('docs-root'));
```

For:

```
ReactDOM.unmountComponentAtNode(document.getElementById('root'));
```

There seems to be something wrong with the `docs-root` element.

Let me know what you think.

shilman:
@patricklafrance That's strange. Let's pair on this when you're up. It's behaving as I'd expect on my machine--in your branch--so I think there must be some miscommunication OR something very different going on in your browser (Windoze?).

patricklafrance:
Changes are pushed @shilman

PR #8627, opened 2019-10-29
Clean up Channel test

adamyonk:
Issue: Missing `Channel.off` test. 

## What I did

Fixed the test for `Channel.on` and added one for `Channel.off`.

### DIFF
--- a/lib/channels/src/index.test.ts
+++ b/lib/channels/src/index.test.ts
@@ -51,8 +51,20 @@ describe('Channel', () => {
     it('should do the same as addListener', () => {
       const eventName = 'event1';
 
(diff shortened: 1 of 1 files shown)


ndelangen (approved):

ndelangen:
AWESOME, thank you @adamyonk

adamyonk:
@ndelangen Glad to help! I ended up in this test because it looks like the `off` method is missing in the type def for this class in the NPM package. Do you know why that would be? How are those generated?

adamyonk:
Actually, it looks like it's not just the type, the actual `off` function isn't there in the build (5.2.5): 
image
but it is in the source:
image
🤔

PR #8678, opened 2019-11-02
React-native: Make menu bar have its own position instead of absolute

Gongreg:
Issue: N/A.

Currently navigation bar is positioned absolutely and it is displayed on top of components. I saw some issue in github where someone asked to make it static, but I can't find it.


shilman (approved):
LGTM

PR #8679, opened 2019-11-02
React-Native: Add safe area wrapper for iphone x and later

Gongreg:
Issue: N/A.

Currently the top of addons panel is cut off when running on iphone x.

## What I did

Added safe area view wrapper.


shilman (approved):
LGTM

PR #8681, opened 2019-11-02
React-native: Fix layout in RN61 so addons no longer initially displayed

Gongreg:
Issue: #8613, #8274

## What I did

Fixed the layout


shilman (approved):
LGTM

Issue #7606, opened 2019-07-29
[React-Native][5.1.X] addon-ondevice-notes crash on initial story selection


kenny-inthirath-pear:
**Describe the bug**
Using the Notes add-on for the INITIAL Story that is loaded crashes the app and throws a JS error

Any Story selected after the initial one WILL NOT crash

**To Reproduce**
1. Go to 'ADDONS' on the INITIAL Story loaded
2. Click on 'NOTES'
3. See error `TypeError: Cannot read property of 'selection' of null`

**Expected behavior**
Should see the text passed in (or auto-generated notes)

**Screenshots**
(with bug and workaround)
![StorybookNotes](https://user-images.githubusercontent.com/53273174/62090131-9630f800-b220-11e9-8f3a-bc6fea34d53e.gif)

**Code snippets**
```js
// rn-addons.js
import '@storybook/addon-ondevice-notes/register';

...

// someStory.js
simpleComponentsStory.add('someName', SomeComponent, { notes: 'some notes'})
```

**System:**
```bash
Nexus 5X (Android 9.+) device emulator
Dev/debugging mode
"react-native": "0.59.8"
"@storybook/addon-backgrounds": "^5.2.0-beta.17"
"@storybook/addon-knobs": "^5.2.0-beta.17"
"@storybook/addon-ondevice-backgrounds": "^5.2.0-beta.17"
"@storybook/addon-ondevice-knobs": "^5.2.0-beta.17"
"@storybook/addon-ondevice-notes": "^5.2.0-beta.17"
"@storybook/react-native": "^5.2.0-beta.17"

System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
Binaries:
    Node: 8.11.2 - ~/.nvm/versions/node/v8.11.2/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.11.2/bin/npm
Browsers:
    Chrome: 75.0.3770.100
    Safari: 12.1.2
```

**Additional context**
No issues pre-migration on `4.1.6`
Reproducible on `5.1.9` and `5.20-beta-17`

ChuanxuMa:
meets the same error

ChuanxuMa:
It's because your default component on initial has no any notes. To avoid it, suggest to add note on every component.

perpetuatingcuriosity:
[...]

PR #8692, opened 2019-11-03
React Native: Fix on-device-notes

Gongreg:
Issue: #7606

Notes addon is not using state anymore, since creating event listener happens only after initial story is rendered. By using api we avoid that.


shilman (approved):
Nice!!!

PR #8694, opened 2019-11-03
React Native: On Device Knobs Groups Fix

Gongreg:
Currently if all knobs have a group, we will display no knobs text instead of grouped knobs. This fixes it.

Issue: N/A


shilman (approved):
LGTM

PR #8698, opened 2019-11-04
CHANGE the compilation target for RN - it was compiled for old browsers

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/8371

RN app is compiled for old browsers

## What I did
I added an override for `app/react-native` in the `.babelrc.js`

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -73,6 +73,11 @@ module.exports = {
       },
     },
     {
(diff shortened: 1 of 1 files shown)


dannyhw:
Sorry to comment on older PR's but this might be useful to have in a 5.3.x patch so that we can release the fix in a current version of react native storybook.

Issue #6807, opened 2019-05-17
Error adding custom theme,  exports is not defined.  @storybook/html


yoojene:
**Describe the bug**

Similar to https://github.com/storybooks/storybook/issues/6749, I'm trying to add a custom theme to change the logo and title, following the instructions https://storybook.js.org/docs/configurations/theming/

```js
// myTheme.js

import { create } from '@storybook/theming';

export default create({
  base: 'light',

  brandTitle: 'My Component Lib',
  brandUrl: 'www.mylib.com',
  brandImage: 'https://placehold.it/350x150'
});
```

```js
// config.js

import { configure, addParameters } from '@storybook/html';
import myTheme from './myTheme';
// automatically import all files ending in *.stories.js
const req = require.context('../src', true, /\.stories\.js$/);
function loadStories() {
  req.keys().forEach(filename => req(filename));
}

addParameters({
  options: { theme },
  html: {
    preventForcedRender: false, // or true
  },
});

configure(loadStories, module);

```

Error:
```
Uncaught ReferenceError: exports is not defined
    at Module../.storybook/myTheme.js
    at __webpack_require__ (bootstrap:781)
    at fn (bootstrap:149)
    at Module.<anonymous> (config.js:3)
    at Module../.storybook/config.js (config.js:18)
    at __webpack_require__ (bootstrap:781)
    at fn (bootstrap:149)
    at Object.0 (readme.md:1)
    at __webpack_require__ (bootstrap:781)
    at checkDeferredModules (bootstrap:45)
```

**System:**
 - OS: MacOS 10.14.5
 - Framework: HTML (Stencil JS)
```json
{
"dependencies": {
    "@stencil/core": "^0.16.1",
    "@stencil/router": "0.3.1",
    "cryptiles": "4.1.2",
    "date-fns": "1.29.0",
    "hammerjs": "2.0.8",
    "moment": "2.20.1"
  },

"devDependencies": {
    "@babel/core": "^7.4.3",
    "@stencil/dev-server": "latest",
    "@stencil/sass": "0.2.1",
    "@stencil/utils": "latest",
    "@storybook/addon-actions": "5.0.10",
    "@storybook/addon-knobs": "5.0.11",
    "@storybook/addon-notes": "5.0.10",
    "@storybook/addon-viewport": "5.0.10",
    "@storybook/html": "5.1.0-alpha.37",
    "@types/jest": "23.3.10",
    "@types/node": "10.12.12",
[...]

Issue #6749, opened 2019-05-09
Updating theme and brand - throwing Object(...) not a function


brendonco:
how do you change the storybook logo?

I'm getting below errors after applying the theme:

import { create } from '@storybook/theming';

yourTheme.js
```
export default create({
  base: 'light',

  brandTitle: 'My custom storybook',
  brandUrl: 'https://example.com',
  brandImage: 'https://placehold.it/350x150',
});
```

config.js
```
import yourTheme from './yourTheme';

const req = require.context('../stories', true, /\.stories\.js$/);
function loadStories() {
  req.keys().forEach(filename => req(filename));
}

addDecorator(story => (
  <ThemeProvider theme={theme}>
    {story()}
  </ThemeProvider>
));

addParameters({
  options: {
    theme: yourTheme,
  },
});

configure(loadStories, module);

```

```
Uncaught TypeError: Object(...) is not a function
    at Module../node_modules/@emotion/core/dist/core.browser.esm.js (core.browser.esm.js:15)
    at __webpack_require__ (bootstrap:781)
    at fn (bootstrap:149)
    at Module../node_modules/@emotion/styled-base/dist/styled-base.browser.esm.js (styled-base.browser.esm.js:1)
    at __webpack_require__ (bootstrap:781)
    at fn (bootstrap:149)
    at Module../node_modules/@emotion/styled/dist/styled.browser.esm.js (styled.browser.esm.js:1)
    at __webpack_require__ (bootstrap:781)
    at fn (bootstrap:149)
    at Object../node_modules/@storybook/theming/dist/index.js (index.js:9)
```

package.json

```
  "@storybook/addon-actions": "^5.0.11",
    "@storybook/addon-info": "^5.0.11",
    "@storybook/addon-links": "^5.0.11",
    "@storybook/addon-storysource": "^5.0.11",
    "@storybook/addons": "^5.0.11",
    "@storybook/react": "^5.0.11",
    "babel-core": "^6.26.3",
    "babel-eslint": "^10.0.1",
    "babel-jest": "^23.0.1",
    "babel-loader": "^7.1.4",
    "babel-plugin-styled-components": "^1.10.0",
    "babel-plugin-transform-object-assign": "^6.22.0",
    "babel-plugin-transform-object-rest-spread": "^6.26.0",
    "babel-polyfill": "^6.26.0",
    "babel-preset-env": "^1.7.0",
    "babel-preset-es2015": "^6.24.1",
[...]

PR #8699, opened 2019-11-04
FIX incorrect documentation about theming

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/6749 https://github.com/storybookjs/storybook/issues/6807

## What I did
CHANGE the references from '@storybook/theming/create' to '@storybook/theming/create', so it's react15 compatible


shilman (approved):
👏 👏 👏

PR #8702, opened 2019-11-04
Addon-docs: Render params description in props table, support @ignore

patricklafrance:
Issue: https://github.com/storybookjs/storybook/issues/8532 https://github.com/storybookjs/storybook/issues/8538

## What I did

- When the @param and @returns tag provide a description, it will be displayed under the function type in the PropTables. It works for JS, TS, Flow and "unknown".
- Also added support for @ignore to prevent a prop from being displayed in the PropsTable when it's marked with @ignore. It also works for JS, TS, Flow and "unknown.

You should be aware that the design is not final.

I coded a temporary UI until domyen design is available to prevent this branch from getting too far away from `next`. Since this is under development and JSDoc support is still an undocumented feature I assumed it's fine. 

## How to test

- Jest tests are avaiable: `yarn jest --testPathPattern=type-system-handlers --watch`
- For @ignore, stories are available in cra-ts-kitchen-sink under the story "docgen/Props types/jsdoc"
- For the params & returns description modifications to the PropsTable, a new story called "func" has been added to the "official-storybook" under the PropsTable component.


shilman (approved):
LGTM - only skimmed

PR #8705, opened 2019-11-05
Addon-docs: Inline rendering for web-components as default

daKmoR:
Issue:

## What I did

- Allow rendering web-components inline in docs mode.
- Enable it by default

## How to test

Open docs mode page on kitchen sink for web components.

I am not really a react pro so I am not sure if that is the best way to do it... but it works really nice and seems quite simple. 🤗 

feedback welcome 🙏


shilman (approved):
Bravo 🔥

PR #8707, opened 2019-11-05
Addon-docs: Render array of shape properly in props table

patricklafrance:
Issue: https://github.com/storybookjs/storybook/issues/8483

## What I did

The update of react-docgen to `4.0.0-beta.1` seem to have resolve most the of the issue.

Still, it could be improve a little bit IMO by rendering the name of the shape instead of "custom". This is closer to what TS renders for a similar case.

With a named shape:

image

With an inline shape:

image

Before this PR changes, always outputting "custom":

image

This is how it's done in TS:

image

**NOTE:**

I created my branch from `8532-render-params-and-returns-tags-optional-description-under-func-type` that's why there is a lot of changes.

The important code for this feature is this part in the PropRow component:

```js
    case PropType.ARRAYOF: {
      let shape = type.value.name;

      if (shape === 'custom') {
        if (type.value.raw) {
          shape = type.value.raw.replace(/PropTypes./g, '').replace(/.isRequired/g, '');
        }
      }

      return `[ ${shape} ]`;
    }
```

## How to test

- New example in the `cra-ts-kitchen-sink` in story "/types".


shilman (approved):
LGTM - only skimmed

PR #8720, opened 2019-11-05
Storybook ingestion feature (load external storybooks in storybooks)

ndelangen:
**Intro**
The scenario for this feature:
Team A building a UI is using storybook, they also use components from the Open Source space (team C), and possible components from another team within the company (Team B).
Team A, wants a complete overview of all components that they have at their disposal. So in fact they want to see the other team’s storybook; preferably within their own storybook.

**The problem**
This is currently not easy to do. Team B has a couple of options:
- Run both/all storybooks
  *This is not easy to do, when the number of teams increases (think OSS components)*
- Configure storybook to also build the external stories from source
  *This is slow, and it’s not common for stories to be published to NPM*
- Use deployed version of storybook when referencing stories from other teams
  *Note this is what most people do now, but it takes a lot of context switching, decreasing productivity.*

**The solution**
Team A will configure storybook to ingest the storybooks of Team B & Team C.

Team A can ingest many more storybooks within theirs. They do not need to scavenge online for docs, everything is in their Storybook.

A hard requirement is that both Team B & Team C deploy their storybook somewhere publicly.

**Technical**

Using `main.js`:

```js
module.exports = {
  refs: {
    identifier: url
  }
}
```

The storybooks will get merged at runtime.

The previews never get merged. Instead the manager will deal with multiple iframes, and need to understand which stories are in which iframe.
When switching stories we may need to switch which iframe to display.


examples/dev-kits/main.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  addons: [
+    '@storybook/addon-roundtrip/register',
+    '@storybook/addon-parameter/register',
+    '@storybook/addon-preview-wrapper/register',
+  ],
+  stories: ['./stories/**/*.js'],
+  refs: {
```
ndelangen: This is how to configure refs

examples/dev-kits/manager.js
```diff
@@ -0,0 +1,10 @@
+import { addons } from '@storybook/addons';
+import { themes } from '@storybook/theming';
+
+addons.setConfig({
+  theme: themes.dark,
+  mapper: ({ id, url }, { kind, ...rest }) => ({
```
ndelangen: You can use this function to map incoming stories from external urls to your own storybook

lib/api/src/index.tsx
```diff
@@ -173,7 +173,38 @@ class ManagerProvider extends Component<Props, State> {
...
     });
 
-    api.on(SET_STORIES, (data: { stories: StoriesRaw }) => {
+    api.on(SET_STORIES, function handleSetStories(data: { stories: StoriesRaw }) {
+      const { source } = this;
+      const { refs, mapper } = provider.getConfig();
+
+      const { origin, pathname } = location;
+      
+      if (refs) {
```
ndelangen: Here's where refs are merged into the storiesHash

PR #8752, opened 2019-11-07
Fix import statement in mdx example

kroeder:
`addon-` was missing in `import { Meta, Story, Props } from '@storybook/addon-docs/blocks';`


shilman (approved):
LGTM! 🙈

PR #8773, opened 2019-11-09
Source-loader: Warn if applied to non-stories file

donaldpipowitch:
Issue: https://github.com/storybookjs/storybook/issues/8294

## What I did

I added a warning for people like me who misconfigured the `@storybook/source-loader`. If I understand correctly, it should only be used for files containing stories. In my webpack config it was applied to all source code files.

## How to test

Checkout my PR, `cd lib/source-loader && yarn prepare && yarn link`. After that `$ git clone git@github.com:donaldpipowitch/storybook-bug-demo-source-loader-context-null.git && cd storybook-bug-demo-source-loader-context-null && yarn && yarn link "@storybook/source-loader" && yarn storybook`. You should see the warning.

image

### DIFF
--- a/lib/source-loader/package.json
+++ b/lib/source-loader/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@storybook/addons": "5.3.0-alpha.42",
-    "@storybook/node-logger": "5.3.0-alpha.42",
+    "@storybook/client-logger": "5.3.0-alpha.42",
     "@storybook/router": "5.3.0-alpha.42",
     "core-js": "^3.0.1",
     "estraverse": "^4.2.0",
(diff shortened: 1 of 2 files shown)


shilman (changes requested):
Thanks @donaldpipowitch, this is great! Mind updating this to use `@storybook/node-logger` instead?

donaldpipowitch:
@shilman sure :) Thank you for the fast review.

donaldpipowitch:
> Mind updating this to use `@storybook/node-logger` instead?

@shilman did you meant `@storybook/client-logger` by any chance? Because I changed `lib/source-loader/src/client/preview.js` (inside `client`).

shilman (approved):

shilman (changes requested):
@donaldpipowitch You're right, it's client-logger

donaldpipowitch:
@shilman Fixed 👍

PR #8784, opened 2019-11-11
Move 'others' to the top of the tree

ghengeveld:
## What I did

Most storybooks don't use roots, but when you embed/import a 3rd party storybook, that one gets a root. The result is that your own stories are no longer at the top, but buried under the imported stories and slotted under an "others" category. Besides the fact this label might not always make sense (and it's not translatable either), these stories are often the most important ones and therefore should be at the top of the tree. This has the added benefit of being able to drop the "others" label because it's not unusual to have ungrouped items at the top of such a list.

## How to test

See the [Sidebar stories](https://www.chromaticqa.com/build?appId=5a375b97f4b14f0020b0cda3&number=12917).


shilman (approved):
I love this. ❤️ 

I'd like to hear feedback from other community members. And I'd like to save it for 6.0 since it's a very visible breaking change.

ndelangen (approved):

tmeasday:
This is directly relevant to something else I am looking at so can we hold off merging until we have a resolution there.

tmeasday:
Also, we need to check things like keyboard shortcuts still make sense (does "next story" work properly at the boundary). Perhaps it is already broken though.

ghengeveld:
> This is directly relevant to something else I am looking at so can we hold off merging until we have a resolution there.
Is that the storySort?

tmeasday:
Nope, although it is related to that too ;) -- https://github.com/storybookjs/storybook/issues/8793

shilman:
@tmeasday I've proposed we don't merge this until 6.0

tmeasday:
As discussed -- this probably doesn't break keyboard shortcuts anymore than they were already broken. There is an orthogonal issue (that could probably be fixed in this PR or earlier) that could be fixed by moving the "others" stories to the start (or currently the end) of the stories hash around these line of code: https://github.com/storybookjs/storybook/blob/c65a9db74f730acc7692ee25129227a49027c0eb/lib/api/src/modules/stories.ts#L306:L323

ndelangen:
@tmeasday is still relevant since https://github.com/storybookjs/storybook/issues/8793

tmeasday:
@ndelangen yes! The scenario where you end up with "others" even if you are using `showRoots:true` (which may become the default in 6.0) is if you have stories that have *no* path separators. 

This is actually probably going to be pretty common, especially if you combine the following two things:

 1. Users that never use separators in their own stories (I suspect maybe there are a lot of them)
 2. Storybook ingestion, as @ghengeveld indicated at the start of this ticket.

So if you take the vanilla starting storybook ("Button" and "Welcome", no separators) and ingest anther storybook (which will be rooted, I think based on our latest discussions), you'll end up in exactly this scenario.

So, TLDR, yes, it will happen, and thanks to ingestion as @ghengeveld says, more important than ever.

ndelangen:
Thanks for the excellent explanation

jamesgeorge007 (approved):

PR #8786, opened 2019-11-11
UI: Store layout state in sessionStorage

ghengeveld:
## What I did

This adds `{ persistence: 'session' }` to all `store.setState` calls in `layout.ts`. This will make sure panel visibility and other layout settings are retained between page reloads.

## How to test

Toggle one of the layout settings (fullscreen, nav, addons, toolbar) and reload the page. Notice your settings will be retained.

### DIFF
--- a/lib/api/src/modules/layout.ts
+++ b/lib/api/src/modules/layout.ts
@@ -157,86 +157,107 @@ let hasSetOptions = false;
 export default function({ store, provider }: { store: Store; provider: Provider }) {
   const api = {
     toggleFullscreen(toggled?: boolean) {
-      return store.setState((state: State) => {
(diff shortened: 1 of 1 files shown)


lib/api/src/modules/layout.ts
```diff
@@ -157,86 +157,107 @@ let hasSetOptions = false;
...
+          { persistence: 'session' }
+        );
+      }
+
+      return store.setState(
+        (state: State) => ({
           layout: {
             ...state.layout,
-            panelPosition: position,
+            panelPosition: state.layout.panelPosition === 'right' ? 'bottom' : 'right',
```
ghengeveld: This is just GitHub rendering a weird diff.

ndelangen (approved):

shilman (approved):
NICE!!!

PR #8791, opened 2019-11-11
Framework-specific presets are deprecated

Peeja:
The Readme told me to use `@storybook/addon-docs/react/preset`, but Storybook told me to use `@storybook/addon-docs/preset`. I think this correctly brings the docs up to date.


shilman (approved):
Thanks @Peeja!

PR #8815, opened 2019-11-13
Addon-docs: Add default prop value when it is null in docgen

gmukul01:
Issue: #8191 

## What I did
- Add component default prop value to prop table when it is coming null from docgen
## How to test

- Is this testable with Jest or Chromatic screenshots? **yes**
- Does this need a new example in the kitchen sink apps? **no**
- Does this need an update to the documentation? **no**


gmukul01:
@shilman Can you please help me understand the failure in the PR.

shilman:
@gmukul01 Thanks for this!! Will try to get it reviewed & merged ASAP. Please ignore the failure ... 😬

patricklafrance:
@shilman let's merge the refactor first.

patricklafrance:
I was wondering how to fix this. This is an amazing idea @gmukul01.

There was major changes to this code in https://github.com/storybookjs/storybook/pull/8826 that is waiting to be merged.

Would you mind adapting your code once it's merged?

gmukul01:
@patricklafrance Yeah sure. :)

shilman:
@gmukul01 The massive PR has been merged into `next` 😬😬😬  https://github.com/storybookjs/storybook/pull/8826

shilman:
@gmukul01 after the merge, the "Files changed" tab of this PR shows only changes unrelated to this PR. Do you know what's going on?

ndelangen:
There are now 0 changes? 💭

gmukul01:
@ndelangen, @shilman  Yeah actually the files which i have changed were deleted after the merge. So in my latest commit you can find out the required changes in the new files.

ndelangen:
Awesome! hopefully @shilman will be able to review soon.

addons/docs/src/lib/docgen/createPropDef.ts
```diff
@@ -4,31 +4,45 @@ import { TypeSystem, DocgenInfo, DocgenType } from './types';
 import { JsDocParsingResult } from '../jsdocParser';
 import { createDefaultValue } from './createDefaultValue';
 
-export type PropDefFactory = (
-  propName: string,
-  docgenInfo: DocgenInfo,
-  jsDocParsingResult?: JsDocParsingResult
-) => PropDef;
+export interface PropDefFactoryProps {
+  name: string;
```
patricklafrance: I do prefer `propName`, `name` doesn't give enough context in this case.

addons/docs/src/lib/docgen/createPropDef.ts
```diff
@@ -4,31 +4,45 @@ import { TypeSystem, DocgenInfo, DocgenType } from './types';
...
+  type: DocgenType;
+}
 
-function createBasicPropDef(name: string, type: DocgenType, docgenInfo: DocgenInfo): PropDef {
+function createBasicPropDef({
+  name,
+  defaultPropValue,
+  type,
+  docgenInfo,
+}: Omit<CreatePropDefProp, 'jsDocParsingResult'>): PropDef {
```
patricklafrance: Not sure about that coupling between the `createPropDef` and this function.

I understand it is convenient to re-use the interface but I think those are 2 separate things that should evoluate separately. The `Omit` you added prove it.

addons/docs/src/lib/docgen/createPropDef.ts
```diff
@@ -4,31 +4,45 @@ import { TypeSystem, DocgenInfo, DocgenType } from './types';
...
   const { description, required, defaultValue } = docgenInfo;
+  const defaults = defaultValue || { value: defaultPropValue };
```
patricklafrance: Could you please inline that code in `createDefaultValue`. `defaults` make me thing there is multiple default values

addons/docs/src/lib/docgen/createPropDef.ts
patricklafrance: I am not sure we need this interface.

I feel like it's only adding complexity without much gain. If you want to prevent redefining the argument for every factories maybe you could use rest + spread operators?

addons/docs/src/lib/docgen/extractDocgenProps.test.ts
```diff
@@ -71,6 +77,31 @@ TypeSystems.forEach(x => {
...
+      {
+        ...createStringType(x),
+        description: 'Hey! Hey!',
+      },
+      { [PROP_NAME]: 'Default' }
+    );
+
+    const { propDef } = extractPropsFromDocgen(component, DOCGEN_SECTION)[0];
+
+    expect(propDef.defaultValue.summary).toBe('Default');
```
patricklafrance: Awesome!

addons/docs/src/lib/docgen/extractDocgenProps.test.ts
```diff
@@ -71,6 +77,31 @@ TypeSystems.forEach(x => {
     expect(propDef.defaultValue.summary).toBe('Default');
   });
 
+  it('should map defaultProp value properly when it is missing in docgen', () => {
```
patricklafrance: typo: properly -> property.

Same typo in the other test

addons/docs/src/lib/docgen/extractDocgenProps.test.ts
```diff
@@ -71,6 +77,31 @@ TypeSystems.forEach(x => {
...
+
+  it('should map defaultProp value properly when it is missing in both docgen and default prop object', () => {
+    const component = createComponent({
+      ...createStringType(x),
+      description: 'Hey! Hey!',
+    });
+
+    const { propDef } = extractPropsFromDocgen(component, DOCGEN_SECTION)[0];
+
+    expect(propDef.defaultValue).toBe(null);
```
patricklafrance: ❤️

patricklafrance (changes requested):

lib/components/src/blocks/PropsTable/PropValue.tsx
```diff
@@ -25,9 +25,13 @@ const PropSummary: FC<PropSummaryProps> = ({ value }) => {
     style.borderBottom = '1px solid blue';
   }
 
-  return (
+  return typeof summary === 'object' ? (
+    <pre title={title} style={{ ...style, whiteSpace: 'pre-wrap' }}>
+      {JSON.stringify(summary, null, 1)}
+    </pre>
+  ) : (
```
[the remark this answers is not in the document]
gmukul01: I have removed specific rendering but i hope below scenario is covered. 

Scenario: When default prop is object it is coming like this.

image

shilman (approved):
Looks great! 💯💯💯

patricklafrance:
After testing the PR, I do think this is way more complicated than this and I am not sure this is a good idea anymore.

The thing is that the default value has to be stringify to be renderer and if we want to gracefully display the value as we do for the values that comes from react docgen we need to implement a smart way to stringify the value.

I am not sure at this point I am ready to deal with that since that would require a whole new code path.

Here's an example of the problem:

image

shilman:
Let's put this on the back burner and pick it up again after the new prop tables have stabilized a little bit. Thanks for your patience @gmukul01 @patricklafrance

gmukul01:
@shilman @patricklafrance Yes you are correct, while dealing with default prop we have handle each type of prop like number|string|boolean|object|symbol|undefined.

patricklafrance:
Closing this PR since https://github.com/storybookjs/storybook/pull/8950 has been merged

PR #8839, opened 2019-11-15
8793 do not use old separators in docs

tmeasday:
Issue: We were still using `|` as a separator in a couple of doc files


shilman (approved):
LGTM!! Thanks @tmeasday

PR #8852, opened 2019-11-16
Changes links to absolute URLs

schalkventer:
Relative links break when README file is viewed at https://www.npmjs.com/package/@storybook/addon-docs.

Issue: Images break when viewing at https://www.npmjs.com/package/@storybook/addon-docs

## What I did

Changed image `src` to absolute URL and not relative URL (which breaks on other domains).

## How to test


shilman (approved):
LGTM thanks!

PR #8813, opened 2019-11-12
DocsPage: Add subcomponents parameter

enagy27:
Issue: #7811 

## What I did
Added `subcomponents` parameter to `DocsPage` to allow named props tables

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, TODO
- Does this need a new example in the kitchen sink apps? Yes, TODO
- Does this need an update to the documentation? Yes, TODO

PR #8855, opened 2019-11-17
Addon-docs: Customizable DocsPage doc blocks

atanasster:
Issue: #8723 #7811

Now users can customize the DocsPage more easily, or can entirely re-create the DocsPage in MDX stories:

Syntax: 
```
      page: () => (
        <>
          <Title />
          <Subtitle />
          <Description />
          <Primary />
          <Props />
          <Stories />
        </>
      ),
```
## What I did
Refactor DocsPage blocks - created new blocks:
* `<Title />` - will display the current documentation title. Accepts optional  slot function or a children (text) parameter
* `<Subtitle>` - will display the `parameters.componentSubtitle` if set. Accepts optional  slot function or a children (text) parameter
* `<Description />` - reused the previous docs/blocks/Description component, except set the current component (`'.'`) as default. Accepts now optional children (text) parameter
* `<Primary />` - displays the primary story (the first one)
* `<Props />` - displays the primary story (the first one) props
* ` <Stories />` - displays all the current doc stories, except the first one. Accepts optional  title (text) parameter, by default 'Stories'.
* `<Heading />` - a section heading for docs page - it simply uses a H2 element but could be later customized.
* `<Subheading />` - a sub-section heading for docs page - it simply uses a H3 element but could be later customized.
* `<DocsStory />` - displays a `<Story />` within a `<Preview />` block with options: **expanded,** and **withToolbar**. If expanded, it will also display the story name and description above the story block.

## How to test

added a new story to official storybook `addon-docs-blocks.stories.js` which uses the new doc blocks to configure the DocsPage:

```
export default {
  title: 'Addons/Docs/stories docs bocks',
  component: DocgenButton,
  parameters: {
    docs: {
      page: () => (
        <>
          <Title>Custom Title</Title>
          <Title />
          <Subtitle>Custom sub-title</Subtitle>
          <Description />
          <Description>Custom markdown</Description>
          <Primary />
          <Props />
          <Stories title="Stories title" />
        </>
      ),
    },
  },
};
```
^*update**: changed `<PrimaryProps />` to `<Props />` if this is used for documentation


atanasster:
The ‘<DocsPage />’ from ‘@storybook/components/blocks’ is not used anymore from docs addon and just kept to avoid breaking changes. 

I think we can remove it before 5.3 final?

shilman:
@atanasster TS error:

```
src/blocks/DocsPage.tsx(21,18): error TS2322: Type 'StringSlot' is not assignable to type 'DescriptionSlot'.
  Types of parameters 'context' and 'description' are incompatible.
    Type 'string' has no properties in common with type 'SlotContext'.
```

ndelangen:
Looking good so far!

shilman:
@atanasster Before we merge this, could you possibly work this in?

https://github.com/storybookjs/storybook/pull/8813#issuecomment-554762254

sb-sub-components

I'm thinking the UI would not contain the toggle in @patricklafrance 's example, but would contain something like the tabbed UI.

I think we should do this first to avoid a breaking change in the beta. What do you think?

ndelangen:
+1 for tabbed interface!

atanasster:
Sounds good, i will add the tabbed interface and a a few examples. It shouldnt be a breaking change btw.

atanasster:
Added support for multiple components props and updated the `Addons/Docs/stories docs bocks` examples.

Two ways to display table with multiple props:

1. Using the story parameters:
```
  parameters: {
    component: ButtonGroup,
    components: {
      'Button Group': ButtonGroup,
      'Docgen Button': DocgenButton,
      'Base Button': BaseButton,
    },
```

2. Using the `<Props />` docs block:

```
<Props
  components={{
    'Button Group': ButtonGroup,
    'Docgen Button': DocgenButton,
    'Base Button': BaseButton,
  }}
/>
```

atanasster:
@shilman added example like @patricklafrance checkbox for display/hide props (can be used for any other docs block, its just regular JSX docs). It seems cleaner to not have the hide/show checkbox in the blocks themselves so everyone can implement it the way they prefer.

atanasster:
a few changes that I would like to make but that are slightly not backwards compatible, pls give me your feedback:

1. remove components/DocsPage
2. Make Props and DocsStory default to first story - this will remove the need for `<Primary />` and `<PrimaryProps />` - instead DocsPage will be simply:
```
page: () => (
        <>
          <Title />
          <Subtitle />
          <Description />
          <DocsStory />
          <Props />
          <Stories />
        </>
      ),
```
3. Change all slot functions to be first parameter the data (list of stories, or description the title) and the second parameter to be the context (DocsContext). Currently the params are a bit randomizes to preserve backwards compatibility

shilman:
@atanasster About to add some changes to components/Props

Regarding your changes:

> 1. remove components/DocsPage

Go for it

> 2. Make Props and DocsStory default to first story - this will remove the need for `<Primary />` and `<PrimaryProps />` - instead DocsPage will be simply:

Let's keep `Primary`. It also uses `withToolbar` and I think it's conceptually different from the rest of the stories. Let's get rid of `PrimaryProps`

> 3. Change all slot functions to be first parameter the data (list of stories, or description the title) and the second parameter to be the context (DocsContext). Currently the params are a bit randomizes to preserve backwards compatibility

I suspect we'll deprecate the slot functions soon after this change and some other changes I have in mind. Users will hate us if we change the signatures THEN deprecate them, so let's leave these alone for now.

atanasster:
@shilman thanks for the feedback. Latest check in removes components/DocsPage and makes Prop default to '.'  and removes PrimaryProps

shilman (approved):
THIS IS SO 🔥🔥🔥

Great work @atanasster 💯

NordlingDev:
Great feature! However, it doesn't seem like I'm able to configure a custom DocsPage inside `.storybook/config.tsx`?

shilman:
@NordlingDev Looks correct to me. Do you have a reproduction I could look at?

valentinbourqui:
There is a possibility to add directly these blocks inside a mdx file and if we define the same title inside Meta for js and mdx files all information about the component is displayed ?

   <>
          <Title />
          <Subtitle />
          <Description />
          <DocsStory />
          <Props />
          <Stories />
        </>

shilman:
@valentinbourqui should work

PR #8857, opened 2019-11-18
Addon-docs: Fix props table props sorting for PropTypes

patricklafrance:
Issue: https://github.com/storybookjs/storybook/issues/8539

## What I did

The sorting problems comes from react-docgen that mess up the original sorting order of the props by returning first the props that have a default value.

My fix consist of using the original definition of the PropTypes object to re-order the props based on their original definition order.

This only fix the problem for PropTypes. If other frameworks or type systems have the same problem, they will also need to be fixed.

## How to test

- Does this need a new example in the kitchen sink apps? Yes

Jest: `yarn jest --testPathPattern=handleProp`
cra-ts-kitchen-sink: ?path=/docs/propssort--page

image


shilman (approved):
LGTM

PR #8871, opened 2019-11-18
Triconfig - configure ui options overhaul

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/8860

## What I did
- ADD handling for extra options set in `setConfig`
- CHANGE readme of addon-options, so we can remove it in 6.0.0
- ADD a deprecation warning when user import the options addon, so we can remove it in 6.0.0


addons/options/README.md
```diff
@@ -1,136 +1,28 @@
...
-### Set options globally
-
-Import and use the `addParameters` + `options`-key in your `config.js` file.
+Global options can be specified with:
 
 ```js
-import { addParameters, configure } from '@storybook/react';
-
-// Option defaults:
-addParameters({
```
tmeasday: We don't want to keep these JSdocs in this file (for the types of options?)

addons/options/README.md
```diff
@@ -1,136 +1,28 @@
...
+import { themes } from '@storybook/theming';
+
+addons.setConfig({
+  theme: themes.dark,
+  showNav: true,
+  showPanel: true,
+  panelPosition: 'bottom',
+  enableShortcuts: true,
+  isToolshown: true,
+  selectedPanel: 'storybook/roundtrip',
```
tmeasday: What happened to the other options (e.g. `url`, `showAddonPanel` etc)?

tmeasday (approved):
LGTM, some documentation qns

addons/options/README.md
ndelangen: They are to be documented on the docs site instead.

I want to completely remove this addon.

tmeasday: Ok, but are these options documented somewhere else?

ndelangen: https://github.com/storybookjs/storybook/blob/651e058cee6cecbb8280da4025c8e56345c883a3/docs/src/pages/configurations/options-parameter/index.md

ndelangen: I updated to complete remove this content and just refer to the docs.

This prevents these docs from going out of sync.

addons/options/README.md
ndelangen: No, this was duplicated, I just refer to the docs site now.

PR #8883, opened 2019-11-19
Addon-a11y: Support manual run

donaldpipowitch:
Issue: Follow up to https://github.com/storybookjs/storybook/issues/8126#issuecomment-541376857

## What I did

- refactored `A11YPanel` a lot:
  - properly type icons (no `any`)
  - properly type `state` with multiple possible `status`
  - we now have an initial stat, so we don't show an empty result which isn't related to any story on startup
  - I removed the `ActionBar`  for the `running` state (personal opinion that this was confusing UX, because it was still clickable to trigger a `run` even though it was already `running` - I can revert that change, if you want to keep it)
- a11y checks can now run manually 🥳

It would be nice, if the ESLint rules `default-case,consistent-return,no-case-declarations` could be removed completely for `.ts/.tsx` files. They should not be relevant anymore as TypeScript should check, if everything is declared correctly/every switch branch is covered/return value is correct. For now I ignored these checks in this file 😔

## How to test

This is testable with Jest, however I probably need some guidance here. I had some troubles executing the tests locally. 😕 If I find time I try it again later this day or in the next days. But any help how to fix something properly would be useful.

I probably need to add `api.emit(EVENTS.MANUAL, true);` _and_ `api.emit(EVENTS.MANUAL, false);` at the right places.


addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -20,60 +20,70 @@ export enum RuleType {
...
+interface RanState {
+  status: 'ran';
+  passes: Result[];
+  violations: Result[];
+  incomplete: Result[];
+}
+
+interface ReadyState {
+  status: 'ready';
   passes: Result[];
```
CodeByAlex: Couldn’t these 4 be made into a single interface with options for status? In other words, extending the old interface.

donaldpipowitch: It could be merged, but I made this change on purpose. I wanted to use a state machine like approach (without introducing a dedicated lib like `xstate`) to make the code more easy to reason about.

The old interface looked simple, because it was just _one_ interface, but it was (imho) harder to read and understand, because it was not very explicit. There was no differentiation between the _initial state_ and the `ready` state for example, so there was the need to mock a proper result. But inside the component you never knew, if the result was mocked or not. (Also the user briefly saw "invalid" results, because of this.)

I _could_ have added mocked results to the `initial`, `manual` and `running` states. They wouldn't "hurt", but they also wouldn't have any purpose. But it get's harder if you introduce new states which need _other_ data (like `error`).

**tl;dr**: Yes, the interface became more verbose. But it also became more correct and explicit and I hope this will make it easier to understand and extend the logic of this component.

donaldpipowitch:
I updated the tests. I hope they are easier to read and extend now. 👋

donaldpipowitch:
@shilman could you have a look? I just tried to resolve the latest conflicts.

addons/a11y/src/components/A11YPanel.tsx
ndelangen: Long term, this addon should start using the `useAddonState` tbh.

ndelangen:
@donaldpipowitch I've added the 5.4 milestone to this, We're currently in a feature freeze so we can release a stable 5.3.

donaldpipowitch:
Alright. Thank you for the update :)

addons/a11y/src/components/A11YPanel.tsx
donaldpipowitch: @ndelangen Sorry, just saw your comment now. Did I understand you correctly that the approach is fine for now and `useAddonState` should be used in a future PR? Or should I add `useAddonState` to this PR?

ndelangen:
@donaldpipowitch If you think this is good to merge into the 6.0.0 branch, you can press the merge button 😄

donaldpipowitch:
Thank you @ndelangen for the update 😊 Sadly I can't press on merge as there is a an approval needed by someone with write access. Have nice holidays everyone. ❤️

addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -142,9 +149,18 @@ export class A11YPanel extends Component<A11YPanelProps, A11YPanelState> {
     });
   };
 
+  onManual = (manual: boolean) => {
+    if (manual) {
+      this.setState({
+        status: 'manual',
+      });
+    } else {
+      this.request();
```
gaetanmaisse: Are we sure that the story has been rendered before calling `this.request()`? 🤷‍♂ I'm just asking because `STORY_RENDERED` event was used before

donaldpipowitch: It seemed to work for me without problems, but I don't know the internals. Maybe @ndelangen could confirm if events emitted inside `makeDecorator -> wrapper` (see https://github.com/storybookjs/storybook/pull/8883/files/d73733d99e2250d840f35e16044e8de69bdc9b6c#diff-135e5fd1db00e7a987038977b6cb9148R70) happen after rendering? If not, do you know an example I could add to the tests to see when it would fail, so I can fix it?

ndelangen: STORY_RENDERED is called after the awaited render function resolved/returned:
https://github.com/storybooks/storybook/blob/38a3dc6ad1d19a98b1be2a7e33b907bbd3252879/lib/core/src/client/preview/start.js#L240-L242

ndelangen: So I do believe it's called later. Whether or not this impacts the a11y execution, I don't know.

donaldpipowitch: There actually seems to be a problem - not sure it's because of this or something else. If _another_ panel is _initially_ active and you switch to the a11y panel you only see the initial state.

E.g. if you visit https://monorepo-51csjv7sk.now.sh/cra-kitchen-sink/?path=/story/app--full-app for example (taken from a different pull request) and switch to the a11y you see results. While you only see the initial state (at least on _first try_ - if you reload it works) if you run my pull request locally.

I probably need to fix that 😭

gaetanmaisse: Good catch @donaldpipowitch, yes we need to fix this, I will try to take a look tonight if I have the time ;)

ndelangen:
@gaetanmaisse could you resolve the merge conflicts here, and merge if all seems well?

gaetanmaisse:
@ndelangen 👌I self-assigned me this issue.
After resolving conflicts there will still be an issue to fix:  https://github.com/storybookjs/storybook/pull/8883#discussion_r364712456 I will work on it too 😉 

I will do it as soon as I can.

donaldpipowitch:
Thank you for also working on that bug and getting everything to the finish line 👏

addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -20,26 +20,29 @@ export enum RuleType {
   INCOMPLETION,
 }
 
-const RotatingIcons = styled(Icons)(({ theme }) => ({
+const Icon = styled(Icons)({
```
gaetanmaisse: @donaldpipowitch  I needed to keep an `Icon` component with `height` and `width` otherwise the check icon from https://github.com/storybookjs/storybook/pull/8883/commits/e368e570989a85b7f8cb7565ecced624134ee4d2#diff-3250ce08c136d40dd47659131333a6a7R213 is not displayed

addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -144,9 +152,18 @@ export class A11YPanel extends Component<A11YPanelProps, A11YPanelState> {
...
+    if (manual) {
+      this.setState({
+        status: 'manual',
+      });
+    } else {
+      this.request();
+    }
+  };
+
   request = () => {
```
gaetanmaisse: @donaldpipowitch I succeeded to fix locally the issue we talk about a few days ago (https://github.com/storybookjs/storybook/pull/8883/files?file-filters%5B%5D=.snap&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#r364712456) by removing the if-check in the request function. Can you try on your side? and add a commit in your branch if it works?

```ts
request = () => {
  const { api } = this.props;
  this.setState(
    {
      status: 'running',
    },
    () => {
      api.emit(EVENTS.REQUEST);
      // removes all elements from the redux map in store from the previous panel
      store.dispatch(clearElements());
    }
  );
};
```

I don't well understand why this `active` check is done, I played with SB examples locally and didn't find any new issue 🤷‍♂

donaldpipowitch: @gaetanmaisse thank you for having a look. This seems to fix the issue - it now works for me locally.

I'm unsure about `active` as well. I mostly left this untouched. But there is an old comment which says that it might can be removed completely?
https://github.com/storybookjs/storybook/blob/2da3a4e1dd5a5f489624ebb0949e0b941bfd31e9/addons/a11y/src/components/A11YPanel.tsx#L100

For now I just added your fix and kept other usages of `active`.

gaetanmaisse: I tried to simply remove all `active` usages and it breaks everything so I think we need to keep it for now.

gaetanmaisse (approved):

gaetanmaisse:
As this PR is about adding manual run mode I will merge it 🎉 Thanks @donaldpipowitch 👏 

And some technical improvements can be done in another PR, for instance:
 - use `useAddonState`
 - check if `active` flag can be removed as indicated in a comment

donaldpipowitch:
Thank you so much 😊

PR #8890, opened 2019-11-19
Addon-docs: Improve basic support for Flow props

patricklafrance:
Issue:

## What I did

Improved the basic support for Flow.

Before the improvement for PropTypes the support for Flow was better than it is now. This PR fix that.

This is also a first step to generalize some concepts of PropTypes to other type systems.

## How to test

To test with Jest run: `yarn jest --testPathPattern=createPropDef`


shilman (approved):
GREAT! 😻

PR #8896, opened 2019-11-20
Addon-docs: Docgen lib maintenance

patricklafrance:
Issue:

## What I did

- Flow not filter out "null" and "undefined" default values
- Reworked the docgen lib utils to prepare for re-use
- Added more test cases for prop types from an external file

## How to test


shilman (approved):
LGTM!

PR #8897, opened 2019-11-20
FIX #8504 - HTML elements get their classes dropped in MDX

### LINKED ISSUE #8504: Using raw HTML in MDX files strips classes
**Describe the bug**
When using mdx as a source of doc, not wrapping the html in a story component strips all classes.

**To Reproduce**
1. Create a mdx file 
2. Paste code from code snippet 
3. No classes are applied to the DOM elements

**Expected behavior**
Using raw HTML in mdx would behave as straight HTML

**Screenshots**

image

**Code snippets**
```
import { Meta } from "@storybook/addon-docs/blocks";

<Meta title="Materials|Colours" />

<div>
    <div class="pa4 bg-desktop-500">
        <div className="flex justify-between">
            <span>500</span>
            <span>#F47321</span>
        </div>
    </div>
    <div className="pa4 bg-sunray-500">
        <div className="flex justify-between">
            <span>500</span>
            <span>#FCB736</span>
        </div>
    </div>
</div>
 ```

Note that nor class or ClassName are working

**System:**
  System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  Binaries:
    Node: 12.7.0 - ~/.nvm/versions/node/v12.7.0/bin/node
    Yarn: 1.19.1 - ~/Repos/sg-orbit/node_modules/.bin/yarn
    npm: 6.10.0 - ~/.nvm/versions/node/v12.7.0/bin/npm
  Browsers:
    Chrome: 77.0.3865.120
    Firefox: 69.0.1
    Safari: 13.0.2
  npmPackages:
    @storybook/addon-actions: 5.2.4 => 5.2.4 
    @storybook/addon-console: 1.2.1 => 1.2.1 
    @storybook/addon-docs: 5.2.4 => 5.2.4 
    @storybook/addon-knobs: 5.2.4 => 5.2.4

fraincs:
Issue: #8504

## What I did

Added an example story where classes are stripped by Storybook - in cra-ts-kitchen-sink > classes stripped

## How to test


fraincs:
See issue https://github.com/storybookjs/storybook/issues/8504

ndelangen:
So this is a reproduction that needs a fix?

fraincs:
> So this is a reproduction that needs a fix?

This is a reproduction of the issue https://github.com/storybookjs/storybook/issues/8504 where straight HTML classes are stripped.

ndelangen:
@shilman do you think we should address this before 5.3 release?

ndelangen:
I'm adding this to my todo list

ndelangen:
I think this is fixed now, and TS-typings for the html elements should be much improved as well!

patricklafrance:
@ndelangen pretty sure I fixed this when doing the MDX deep linking feature.

ndelangen:
Hey @patricklafrance I think the components are going to be much more strict after this PR. 

https://github.com/storybookjs/vs-code-plugin/issues/15

This just makes sure the className prop is accepted, but my change should make it so The `<A>` component actually only accepts props valid to a `a` tag PLUS the added `class` prop, that gets mapped to `className` (that's actually the fix for the issue).

ndelangen (approved):

PR #8900, opened 2019-11-20
Escape hatches

fraincs:
Issue:
Stylizing the doc page could be problematic when there is parts of the DOM without classes(escape hatches). This is true for the Preview component as well as the Props Table header and body.

## What I did

Added a class ``` sbdocs-preview ``` to the doc preview component

Added a class of  ``` docblock-propstable-head ``` and ``` docblock-propstable-body ``` to the props table doc component

## How to test


lib/components/src/blocks/Preview.tsx
```diff
@@ -154,7 +156,11 @@ const Preview: FunctionComponent<PreviewProps> = ({
   }
   const showToolbar = withToolbar && !Array.isArray(children);
   return (
-    <PreviewContainer {...{ withSource, withToolbar: showToolbar }} {...props}>
+    <PreviewContainer
+      {...{ withSource, withToolbar: showToolbar }}
+      {...props}
+      className="sbdocs-preview"
```
ndelangen: I think this should accept a className from props and merge with `"sbdocs-preview"` if present?

ndelangen (changes requested):

fraincs:
@ndelangen done

ndelangen (approved):

lib/components/src/blocks/PropsTable/PropJsDoc.tsx
```diff
@@ -7,24 +7,107 @@ interface PropJsDocProps {
   tags: JsDocTags;
 }
 
-const TBody = styled.tbody({ boxShadow: 'none !important' });
+export const Table = styled.table<{}>(({ theme }) => ({
```
domyen: I think this is now outdated – can you merge `next` back into the branch to get the latest.

domyen (changes requested):

domyen:
The classname stuff looks good to me. But the other changes have since been updated on `next` so will hold off commenting on that till you merge `next` back into this branch. Let me know and I'll review again.

lib/components/src/blocks/PropsTable/PropJsDoc.tsx
fraincs: Perfect

domyen (approved):

PR #8925, opened 2019-11-22
Revert "feat: use `puppeteer-core` instead of `puppeteer`"

gaetanmaisse:
This reverts commit aad29ed5

Issue:

See https://github.com/storybookjs/storybook/commit/aad29ed53d63daf145ed226dc2848913c6c3b9e5#commitcomment-36082241

> It looks like after using `puppeteer-core` instead of `puppeteer`, `imageSnapshot` doesn't work unless you specify `chromeExecutablePath`

## What I did

Revert https://github.com/storybookjs/storybook/commit/aad29ed53d63daf145ed226dc2848913c6c3b9e5

### DIFF
--- a/addons/storyshots/storyshots-puppeteer/package.json
+++ b/addons/storyshots/storyshots-puppeteer/package.json
@@ -32,10 +32,8 @@
     "@storybook/node-logger": "5.3.0-beta.3",
     "@storybook/router": "5.3.0-beta.3",
     "@types/jest-image-snapshot": "^2.8.0",
-    "@types/puppeteer-core": "^2.0.0",
     "core-js": "^3.0.1",
     "jest-image-snapshot": "^2.8.2",
-    "puppeteer-core": "^2.0.0",
     "regenerator-runtime": "^0.13.3"
   },
   "devDependencies": {
--- a/addons/storyshots/storyshots-puppeteer/package.json
+++ b/addons/storyshots/storyshots-puppeteer/package.json
@@ -32,20 +32,25 @@
     "@storybook/node-logger": "5.3.0-beta.3",
     "@storybook/router": "5.3.0-beta.3",
     "@types/jest-image-snapshot": "^2.8.0",
+    "@types/puppeteer-core": "^2.0.0",
     "core-js": "^3.0.1",
     "jest-image-snapshot": "^2.8.2",
+    "puppeteer-core": "^2.0.0",
     "regenerator-runtime": "^0.13.3"
+  },
+  "devDependencies": {
+    "@types/puppeteer": "^2.0.0"
   },
   "peerDependencies": {
     "@storybook/addon-storyshots": "5.3.0-beta.3",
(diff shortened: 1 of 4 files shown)


gaetanmaisse:
@ndelangen I'm unsure why `puppeteer` is not installed during `yarn install` because it is listed as `optionalDependencies` and is in `yarn.lock` 🤷‍♂  any idea?

@shilman as soon as it is ✅ and merged, this fix will have to be cherry-picked into 5.3 branch

Hypnosphi:
It still can make sense to use `puppeteer-core` in some cases. I'm thinking about this:

1. Remove `puppeteer` from `optionalDependencies`
2. Make `puppeteer` an optional peer dependency using `peerDependenciesMeta` feature supported both by npm and yarn:
    ```js
    {
      "peerDependencies": {
        "puppeteer": "^1.12.2 || ^2.0.0"
      },
      "peerDependenciesMeta": {
        "puppeteer": {"optional": true}
      },
    }
    ```
3. Require `puppeteer` dynamically, and only if `getCustomBrowser` is not passed:
    ```js
    testFn.beforeAll = async () => {
      if (getCustomBrowser) {
        browser = await getCustomBrowser();
      } else {
        const puppeteer = require('puppeteer')
        // add some options "no-sandbox" to make it work properly on some Linux systems as proposed here: https://github.com/Googlechrome/puppeteer/issues/290#issuecomment-322851507
        browser = await puppeteer.launch({
          args: ['--no-sandbox ', '--disable-setuid-sandbox', '--disable-dev-shm-usage'],
          executablePath: chromeExecutablePath,
        });
      }
    ```

That way, users that pass `getCustomBrowser` option anyway could safely remove the unused dependency on  `puppeteer`

As for `@types/puppeteer`, I think it can be just moved to `devDependencies`

Hypnosphi:
I'm not too good at TS, is there any way to use types from `@types/puppeteer` without requiring `puppeteer`? If there is, we can also remove `puppeteer-core` and `@types/puppeteer-core` dependencies

gaetanmaisse:
@Hypnosphi I was just trying what you said in your last comment. In fact, with the right `peerDependenciesMeta` and `devDependencies` we can get ride of `puppeteer-core`

Hypnosphi (approved):

Hypnosphi:
>  I was just trying what you said in your last comment

@gaetanmaisse sorry, what exactly didn't work with requiring `puppeteer` dynamically?

gaetanmaisse:
@Hypnosphi Both static `import` and dynamic `require` were working so I decided to keep `import` to be as close as possible of original code in this "revert" PR. I will take a look at https://github.com/storybookjs/storybook/pull/8932 as soon as I have time ;)

Hypnosphi:
Static `import` would not allow users to remove `puppeteer` depandency at all, while dynamic `require` would if they provide `getCustomBrowser `

PR #8929, opened 2019-11-22
FIX now deploy

ndelangen:
Issue:
Now deploy seems to use some cache, which has an older version of some typings, this conflicts and makes the deploy fail.

## What I did
FIX now deploy which seems to use some back cache or something, as locally it seems to work just fine


ndelangen:
There's something wrong still on the now side of things, somehow the install in now is different from the local or even github action and circleCI.

PR #8932, opened 2019-11-23
Fix image snapshots setup in official-storybook

Hypnosphi:
Issue: `yarn image-snapshots` in `examples/official-storybook` doesn't work

## What I did
1. Reused `toRequireContext` in storyshots to match the behaviour of `@storybook/core`
2. Replaced `glob-regex` with `micromatch.makeRe` as the former only supports very basic patterns
3. Added a customisable timeout for `storyshot-puppeteer` setup and tests


addons/storyshots/storyshots-core/src/frameworks/configure.ts
```diff
@@ -64,9 +64,16 @@ function getConfigPathParts(input: string): Output {
     if (main) {
       const { stories = [] } = require.requireActual(main);
 
-      const result = stories.reduce((acc: string[], i: string) => [...acc, ...glob.sync(i)], []);
-
-      output.stories = result;
+      output.stories = stories.map(
+        (pattern: string | { path: string; recursive: boolean; match: string }) => {
+          const { path: basePath, recursive, match } = toRequireContext(pattern);
+          return require.context(
```
tmeasday: 🤔 how does this work in Jest? I didn't think `require.context()` existed?

Hypnosphi: https://github.com/storybookjs/storybook/tree/next/addons/storyshots/storyshots-core#configure-jest-to-work-with-webpacks-requirecontext

In our own setup, it's Option 1: Plugin (`babel-plugin-require-context-hook`)

tmeasday: Oh, right but this isn't universal. For instance:

1. Users may be using the babel macro
2. This code was added to actually avoid needing to use babel at all if they are using the `main.js` entrypoint.

What was the issue with what was there before w/ more typical `require()`s?

tmeasday: Also it is unlikely that user's code is going to run their babel plugins over files in `node_modules`. Or is all this being compiled when we compile `addon/storyshots`?

Hypnosphi: > Also it is unlikely that user's code is going to run their babel plugins over files in

Oh I see. I think I'll require `babel-plugin-require-context-hook/register` and use `global.__requireContext` directly then

> What was the issue with what was there before w/ more typical require()s?

We used different libraries for turning globs into files which sometimes led to different result. For instance, "regex or" patterns (`.stories.(js|tsx|mdx)`) that we use in `main.js` aren't supported by `node-glob`

tmeasday: Ok, that makes sense.

gaetanmaisse (approved):

addons/storyshots/storyshots-puppeteer/src/imageSnapshot.ts
```diff
@@ -75,19 +79,22 @@ export const imageSnapshot = (customConfig: Partial<ImageSnapshotConfig> = {}) =
...
-    return browser.close();
   };
 
-  testFn.beforeAll = async () => {
+  const beforeAll = async () => {
     if (getCustomBrowser) {
       browser = await getCustomBrowser();
     } else {
+      // eslint-disable-next-line global-require
+      const puppeteer = require('puppeteer');
```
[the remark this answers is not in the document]
Hypnosphi: Does it work in Node 10?

PR #8942, opened 2019-11-25
Core/triconfig everywhere: migrate examples

ndelangen:
https://github.com/storybookjs/storybook/issues/8861

## What I did
I changed all example to use tri-config


tmeasday (approved):
I love it. What do you think @shilman?

Issue #8728, opened 2019-11-05
Addon-docs: Default value doesn't render in the props table for some TypeScript types


patricklafrance:
**Describe the bug**
The default value of a prop is not rendered for a few TypeScript types:

- func
- symbol
- array
- tuple
- alias

**To Reproduce**
View the "Docgen/types" story of cra-ts-kitchen-sink.

patricklafrance:
Matching issues in react-docgen-typescript: https://github.com/styleguidist/react-docgen-typescript/issues/199

patricklafrance:
Will close since the issue is with react-docgen-typescript.

PR #8950, opened 2019-11-25
Addon-docs: Increase width of props table type column

patricklafrance:
Issue: https://github.com/storybookjs/storybook/issues/8924 https://github.com/storybookjs/storybook/issues/8917 https://github.com/storybookjs/storybook/issues/8191
https://github.com/storybookjs/storybook/issues/8728 (only for arrays, tuples and symbols,partial fix for func)

## What I did

- Increased the type column to 70 characters (after a discussion with @domyen). To support this size, I had to refactor the code that determine if a value can be displayed as a summary or not. To discard unwanted complex objects / arrays and other types that are hard to read when compacted, I introduced the notion of depth. Everything with a depth greater than 1 is excluded from the summary.

- Function signatures with more than 70 characters are now displayed with a compact version instead of wrapping on multiple lines and being hard to read. (credits to @domyen for helping to choose the proper format)

- A prop row with JSDoc tags but without a description doesn't have a martin-top of 12px anymore.

- When a defaultValue is null but a defaultProp exists for the property, the defaultProp will be used as a fallback. This is useful for some use cases of PropTypes. For example, spreading an object into defaultProps is not supported by react-docgen. This is even more useful for TS since the defaultValue is omitted for a lot of types and I am not sure it will be supported anytime soon, view the [following PR](https://github.com/styleguidist/react-docgen-typescript/issues/199). Credits to @gmukul01 for the idea of using the defaultProp as fallback.

## How to test

Test with the `cra-ts-kitchen-sink` "docgen/types" story.
Test with Jest: `yarn jest --testPathPattern=docs/src`


shilman (approved):
LGTM

domyen:
Nice @patricklafrance 👏

PR #8953, opened 2019-11-26
Fix vertical alignment of props table expandable

patricklafrance:
## What I did

The props table expandable text and arrow were vertically centered for Windows users but not for Mac users. This PR fix that.

## How to test

Can be with the cra-ts-kitchen-sink


lib/components/src/typography/shared.tsx
```diff
@@ -20,8 +20,9 @@ export const headerCommon = ({ theme }: { theme: Theme }): CSSObject => ({
 });
 
 export const codeCommon = ({ theme }: { theme: Theme }): CSSObject => ({
+  lineHeight: 1,
```
shilman: This applies to all code blocks. Can you review the chromatic changes? I see 4 for proprow and then another one that looks like it might be a bug

patricklafrance:
@domyen can you review the changes on Chromatic please? I think this is fine.

Also, please consider that the way Chromatic rendered the expandable is not accurate. This is how it looks like on Mac:

https://media.discordapp.net/attachments/638061150755618827/648895006505172992/Docs___PropRow_-_Func__Storybook.png?width=1442&height=554

ndelangen:
This one seems like it's not quite an improvement:
https://www.chromaticqa.com/snapshot?appId=5a375b97f4b14f0020b0cda3&id=5ddcb0ff2394130020d68703

patricklafrance:
@ndelangen how come?

Look at the code blocks, way better:

image

domyen:
Reviewed code and Chromatic, I think this is an improvement on the way we do <code> inline.

domyen (approved):

ndelangen:
@patricklafrance It's because I'm no designer 🕶 

Good to merge! 👍👍👍

ndelangen (approved):

PR #8964, opened 2019-11-26
getSnapshotFileName is not a function

Tomekmularczyk:
I tried to run the example under `Stories2SnapsConverter` section but I was getting an error that `getSnapshotFileName` is not a function and I couldn't import it.

I played around and this seems to be working.

Issue:

## What I did
Updated README example

## How to test

- Does this need a new example in the kitchen sink apps? NO
- Does this need an update to the documentation? YES


shilman (approved):
LGTM! Thanks for fixing this

PR #8966, opened 2019-11-26
Core: Use logger in base-webpack.config.js

adamyonk:
## What I did

Just unifies the info output a bit.

image


shilman (approved):
LGTM! Thanks for fixing

PR #8971, opened 2019-11-26
Don't try to require .ts files from dist

Hypnosphi:
Issue: #8908

This reverts https://github.com/storybookjs/storybook/commit/83e9b5f1ab915d020f9c3fd8def07d6dc1548106

### DIFF
--- a/addons/storyshots/storyshots-core/src/frameworks/frameworkLoader.ts
+++ b/addons/storyshots/storyshots-core/src/frameworks/frameworkLoader.ts
@@ -4,7 +4,7 @@ import path from 'path';
 import { Loader } from './Loader';
 import { StoryshotsOptions } from '../api/StoryshotsOptions';
 
-const loaderScriptName = 'loader';
+const loaderScriptName = 'loader.js';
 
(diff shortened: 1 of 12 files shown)


addons/storyshots/storyshots-core/src/frameworks/frameworkLoader.ts
```diff
@@ -13,13 +13,7 @@ function getLoaders(): Loader[] {
     .readdirSync(__dirname)
```
Hypnosphi: When this code actually executes, `__dirname` is '<package-root>/dist/frameworks', so there are always compiled `__dirname/<framework>/loader.js` files available

gaetanmaisse (approved):

gaetanmaisse:
It looks like this addon's migration introduce more bugs than in many others 😞 Maybe because it is less used or tested in examples 🤷‍♂ Anyway, sorry to give you so much extra work @Hypnosphi

Hypnosphi:
I also updated the cleanup script, so that it also removes the `.ts` files copied by Babel from `dist`

PR #8989, opened 2019-11-27
UI: Escape hatch CSS on for "active" tablist buttons

fraincs:
Issue:
There is no way to theme a tablist button, nothing differentiate them between active and non active.

## What I did
I added a class of tabbutton-active when a tab is active.

## How to test


shilman (approved):
LGTM!

PR #9003, opened 2019-11-29
Ndelangen/remove css usage

ndelangen:
Issue: these styled.div`` string stay in the bundle as is, they cannot be minified, increasing the size of the bundle, especially in production mode.

## What I did
I migrated them to use object notation


ndelangen:
I did this be cause I was investigating the inclusion of all css of all languages of syntax-highlighter.

Though I wasn't able to reproduce #8972 I did see huge blobs of string in our bundle.

I figured it would save some bundle-size, because these blobs of string contained all sorts of comments that cannot be removed, but now they can.

Packtracker seems to agree this is an improvement, though smaller than I had hoped.

PR #9040, opened 2019-12-03
Update CSF docs to clarify how naming works

ghengeveld:
This also recommends starting a story identifier with a capital letter.


shilman (approved):
LGTM!

PR #9063, opened 2019-12-04
Addon-docs: MDX Octicon anchors should not be tabbable

patricklafrance:
anchor_tabbable

## What I did

- Added a tabindex of -1 to the Octicon anchors.

## How to test


shilman (approved):
LGTM

PR #8866, opened 2019-11-18
Add auto-syncing useAddonState for manager & preview

ndelangen:
@Joseph-Whiunui

PR #9079, opened 2019-12-05
Shared useaddonstate

atanasster:
Issue: #8866

## What I did
shared state between manager and preview

## How to test
in dev-kits a sample to exchange state

@ndelangen


atanasster:
@ndelangen - great changes to take care of state while changing stories and HMR on the preview side.

I checked in a few more changes after the tests I could think of:
* added now another state story so we can test both defaults set in the preview and defaults set in the manager.
* Start from a different story and update values for manager story
* Start from one of the state stories, navigate out and back
* From the state stories, invoke HMR
* Make changes to the state stories, move out and invoke HMR, then go back to the stories

ndelangen (approved):

atanasster:
To test out the shared state:

https://monorepo-ljhdfpukd.now.sh/dev-kits/?path=/story/addons-useaddonstate--preview-default

ndelangen:
```js
const useStoryState = (prefix = 'global') => {
  const { id } = useCurrentStory();
  const [state, setState] = useAddonState<number>(prefix + id);

  return [state, setState];
};
```

@PlayMa256 👆

PR #9086, opened 2019-12-06
Addon-docs: Fix prop table default value for web-components

steel:
Issue:

The work in progress specification for custom-elements.json uses the `defaultValue` key rather than `default`. Currently, the default value is not displayed in the props table because this is mapped incorrectly.

https://github.com/webcomponents/custom-elements-json/pull/8/files

## What I did

I updated the mapping.


shilman (changes requested):
Are there any peer dependencies that need to be updated? Or documentation?

Or something else? This change seems too small, unless it's just a bug.

steel:
Just a bug.

PR #9087, opened 2019-12-06
Core: Emit store render event synchronously if we can

ndelangen:
Previously we were always waiting a moment to emit the event as some frameworks (just React Native, we think) did not always have the channel setup in time when the stories are loaded.

Waiting to emit meant the first story can get rendered after the browser's load event, meaning you cannot necessarily rely on the load event to pick up all the resources used by the story.


lib/client-api/src/story_store.ts
```diff
@@ -145,9 +145,19 @@ export default class StoryStore extends EventEmitter {
       data === undefined ? this._selection : { storyId: data.storyId, viewMode: data.viewMode };
     this._error = error === undefined ? this._error : error;
 
+    let failed = false;
+    try {
+      if (this._channel) {
+        this._channel.emit(Events.STORY_RENDER);
+      }
+      failed = true;
```
tmeasday: Should this line be `failed = false`?

I don't really understand what "failed" means here. How do we know it's failed? Do we think there may not yet be a listener on the channel for the `emit` event?

If so I reckon there are better ways to deal with that. I don't see how this code would know that.

tmeasday (changes requested):

lib/client-api/src/story_store.ts
ndelangen: WDYT of this?

tmeasday: I'm not sure.

1. How do we expect it to fail? Will `this._channel` be undefined? Will `.emit()` throw somehow? Or will just no-one handle the event?

2. In which case, is a `try/catch` enough to detect it?

ndelangen: this._channel might be undefined

ndelangen: I think that's the case in RN

tmeasday: So maybe put the `isStarted=true` inside the if block then?

I want to do some more playing with the loading behaviour of SB before taking this further but thanks for taking a look.

ndelangen: updated the code, WDYT?

lib/client-api/src/story_store.ts
```diff
@@ -145,15 +145,20 @@ export default class StoryStore extends EventEmitter {
...
-      // should be deprecated in future.
-      this.emit(Events.STORY_RENDER);
-    }, 1);
+    } catch (e) {
+      setTimeout(() => {
+        if (this._channel) {
+          this._channel.emit(Events.STORY_RENDER);
+        } else {
+          // should be deprecated in future.
+          this.emit(Events.STORY_RENDER);
```
tmeasday: This was always getting called before but now is rarely if ever going to be called. I don't think that's right.

lib/client-api/src/story_store.ts
```diff
@@ -145,15 +145,20 @@ export default class StoryStore extends EventEmitter {
...
-      // preferred method to emit event.
+    try {
       if (this._channel) {
         this._channel.emit(Events.STORY_RENDER);
       }
-
-      // should be deprecated in future.
-      this.emit(Events.STORY_RENDER);
-    }, 1);
+    } catch (e) {
```
tmeasday: How would this catch happen if the `is (this._channel)` check is there?

tmeasday (changes requested):

tmeasday:
@ndelangen I'm still confused about this code! Am I reading it wrong?

If `this._channel` is not yet set, we'll still set `isStarted = true`, even though we never emitted the event!

tmeasday:
(The `this.emit()` part is good now though)

tmeasday:
@ndelangen I pushed a commit with what I think is the correct code and a test that exhibits the behaviour.

shilman (approved):
LGTM

PR #9110, opened 2019-12-10
Addon-docs: Typescript inherited prop types

atanasster:
Issue: #7943

grab24

## What I did
* added ability for Sections to manage expanded/collapsed state and hide/show "child" rows 
* extract parent property from react typescript docgen. If the parent prop is available, will collapse the property rows into propSections (sections of properties)
* added example for typescript in officialy-storybook
* added webpack-react-docgen-typescript to main.js in official-storybook
## How to test

replace `webpack-react-docgen-typescript` with `react-docgen-typescript-loader` in official-storybook/main.js

cc @shilman @domyen @patricklafrance


ndelangen:
@domyen this could use your design eye I think.

domyen:
Thanks @ndelangen! @atanasster we chatted about the design in the other issue.

Notes:
- Expandable prop "groups" should look similar to props except with an arrow beside them
- The nested props should be indented to illustrate the parent-child relationship
- When a prop group is collapsed show how many props are nested inside


image

atanasster:
Thanks @domyen - what about this:

prop-table-design

domyen:
I think we need to strip the letter-spacing and other styling from the expander items so that they look more like the default row items. 

Right now it seems like the expander item uses styling left over from the [section separator](https://user-images.githubusercontent.com/263385/70546532-764e0100-1b3d-11ea-8364-1afb15ef2780.png) when it should use the styling of the table row.

image

Check the Zeplin link to inspect the styles 👍

atanasster:
thanks @domyen , got the link and I think its getting close :)

collapsible-props

domyen:
This is a lot better! Still some UX things to resolve. 
- It's disorienting that the columns change in width when expanding and collapsing. What can we do to prevent this?
- For prop groups that have only 1 prop, I don't think they should be hidden in an expandable element. It adds an extra click for no reason. 


Is there a live preview of this component anywhere?

atanasster:
Thanks for the review @domyen 

collapse-props

The changes:
1. All sections with only 1 prop will be expanded by default
2. All sections that contain the name of the component will be expanded by default. In the example below the second page - component `Source` has `SourceCodeProps` and `SourceErrorProps`.
3. The table layout is changed to fixed
4. The table cells styling is changed to accommodate if an initially collapsed they have too much text:
```
      overflow: 'hidden',
      textOverflow: 'ellipsis',
```
5. Since the overflow hidden can hide some of the text, the Name column is now also adding the name to the title property, so a tooltip is shown with the full name:

grab27

grab26

atanasster:
@shilman - are the PR branches deployed so Dom can play around with the actual component?

atanasster:
@domyen - I found the deployment, it was just taking a bit longer:

The long property lists:
https://monorepo-lo2ei58cf.now.sh/official-storybook/?path=/docs/addons-docs-typescript--basic

The Source properties:
https://monorepo-lo2ei58cf.now.sh/official-storybook/?path=/docs/docs-source--jsx

patricklafrance:
Nice work!

I was wondering if the component own props shouldn't be in a collapsible row? After all, this is the props the user should care the most about.

Also, is there any way we could get the design to not use additional horizontal spacing to show the parent / child relationship? Not a big deal if we can't, but the lack of horizontal space is something I already struggle width to show the props.

atanasster:
Thanks @patricklafrance  - there isn’t anything to say what are the components own props vs other props - if oyu look at the code, there is a bit of a hack - expanding all prop types that contain the name of the component or expand the first props. on the other hand, if you look at the Source example, both SourceCodeProps and SourceErrorProps are its ‘own’ props in a way. 

For the design i do agree the horizontal space is at a premium - what is @domyen opinion?

domyen:
Thanks Atanas, will make style tweaks on this branch.

**re:Horizontal space**, the indentation is the neatest method of showing a nested relationship imo. I suspect @patricklafrance's PR to expand the default width of DocsPage will make the horizontal space constraint less of an issue or a non-issue (the 20px is a difference in 4 characters). I'll also make some style tweaks to address this.

atanasster:
I personally like both indentation and also coloring for the section as in my my first check- in, up to @domyen to decide what fits better for the storybook style.

domyen:
@atanasster Made some style refinements! Again, the PropsTable should be getting wider if @patricklafrance's PR to expand the DocsPage was merged. 



Can you create new stories for the PropTable component that document key states:
- Expandable: Items expanded  
- Collapsed: Items collapsed

ndelangen:
@domyen looking at https://www.chromaticqa.com/snapshot?appId=5a375b97f4b14f0020b0cda3&id=5e26d447ff540f0020a6ff47 it seems to be the before image is better.. 

Was it your intention to make the first column wider?

South-Paw:
@atanasster would it be possible for you to test a [styled-system](https://github.com/styled-system/styled-system) component (such as a custom [Box](https://github.com/South-Paw/spicy-ui/blob/master/src/Box/Box.tsx)) and see if it groups the props correctly? 👍

If I use the docs component currently on that Box component, there's a few hundred props in the table so this feature would sort that out nicely (if it works!) :)

Thanks!

atanasster:
@South-Paw my libs are having hundreds of props and thats what i tested with. When its merged i hope you can also give it a whirl before the official release

ndelangen:
We should probably add a story with a styled-components component so we can see what the effect is.

atanasster:
@ndelangen - i think it has to do with extending interfaces this way https://github.com/South-Paw/spicy-ui/blob/74b1c1b0e458adc0dc8e06393e4dc441a535473a/src/Box/Box.tsx#L25

Do you know if we have such example, or should we install spicy-ui in official-storybook?

What i tested is composing interfaces(similar but not exactly the same) - which also generates 100s of properties from the intrinisc element:
```
export const TypescriptButton: FC<ButtonProps & JSX.IntrinsicElements['button']> 
```

domyen:
@ndelangen good catch, that wasn't my intention. I was however trying to fix the width of columns so that expanding/collapsing columns didn't make the entire layout jump around. This was a regression caused by that effort. I will try a different method.

@atanasster Can you create new stories for the PropTable component that document key states:

- Expandable: Items expanded
- Collapsed: Items collapsed

These states would be super helpful for catching regressions as Norbert just did 😅 with me!

South-Paw:
@atanasster I'd be more than happy to give it a try then.

I did try to pull your fork down last night but I couldn't install due to node-gyp errors (which are probably to do with Node/Python/Windows rather than your fork!) I was going to pop this Box component into the storybook and see what this change resulted in.

atanasster:
Thanks @South-Paw - looking forward to your input. I am myself not very good at working from forks, so cant really help

@domyen - i will add those stories, but i might have git trouble if the fork is out of sync.

domyen:
@atanasster I don't think the PropsTables have been edited elsewhere but you might want to merge from `next` just in case. I also revised the PropsTable design to take your advice (from some time ago) and make the table rows (including the expandable rows) more compact across the board.

@ndelangen the column width was updated. It's not identical to the previous version because of the constraints of the expandable table row, but it's pretty close.

ndelangen:
Still a few stories left that show an big increase in the width of the first column, which makes it look weird to me.

domyen:
@ndelangen Updated and approved some of the stories. 

The [PropsTable:Empty](https://www.chromaticqa.com/snapshot?appId=5a375b97f4b14f0020b0cda3&id=5e2a12b52db0a30020aa221a) story is still broken, but that happened outside of my styling changes.

**Some extra detail (that folks might not care about)**
Note that the table layout now has fixed width columns so it doesn't quite behave like it used to where each column dynamically resized based on it's contents. That means on [some stories](https://www.chromaticqa.com/snapshot?appId=5a375b97f4b14f0020b0cda3&id=5e2a12b52db0a30020aa2212) where the name (first column) is shorter, the description will no longer expand to occupy the available space. 

I think this is an OK compromise in the context of the whole table layout. The previous table layout method took into account all rows in a column before determining the column width. That effectively normalized the widths to the widest row in each column. Now we're explicitly setting the width of the first column. In practice, this yields a pretty similar result as the previous method _except_ it allows us to dynamically add <tr>s to the DOM without the table layout reflowing in the browser causing the layout to "jump" around.

ndelangen:
@domyen sounds like a layout nightmare, sounds like we might want to use CSS grid instead maybe?

domyen:
Yea, that's probably a different PR 😅. Don't have enough experience with CSS grid's capabilities tbh.

patricklafrance:
@domyen does the fix width you added will play nice if someone increase the width of the docs container using css escape hatches?

domyen:
Yes! It’s a fluid width based on percentages so it will scale to the size
of its container.

ndelangen:
> Yea, that's probably a different PR 😅. Don't have enough experience with CSS grid's capabilities tbh.

Sure, I wasn't trying to increase the size of this PR.

I'll see if I can find someone that is interested in converting this to CSS-grid, so you can focus on other tasks 👍

johnnyBira:
@atanasster I came here from this gist and it's not clear to me how I would edit `.storybook/main.js` to work the way you describe.
https://gist.github.com/shilman/bc9cbedb2a7efb5ec6710337cbd20c0c

Could you please provide an example based on the config used in the gist?

```javascript
const path = require("path");

module.exports = {
  stories: ["../src/stories/**/*.stories.(ts|tsx|js|jsx)"],
  addons: [
    "@storybook/addon-actions",
    "@storybook/addon-links",
    {
      name: "@storybook/preset-create-react-app",
      options: {
        tsDocgenLoaderOptions: {
          tsconfigPath: path.resolve(__dirname, "../tsconfig.json")
        }
      }
    },
    {
      name: "@storybook/addon-docs",
      options: {
        configureJSX: true
      }
    }
  ]
};
```

Thank you.

ndelangen:
@shilman what's going to happen with this PR?

shilman:
@ndelangen I'll take care of it

ajkl2533:
Hi @shilman what's the state of this PR?

mlg87:
any updates on this? would be a fantastic feature

ndelangen:
@shilman please re-open if you think this is useful to keep.

PR #9119, opened 2019-12-10
Remove dead link

qoalu:
This removes a dead link from docs.
It's better to make a working link, but I am not aware of where this config is sitting now


qoalu:
Okay, I only tried to change one file
https://github.com/storybookjs/storybook/blob/master/docs/src/pages/addons/using-addons/index.md

at the end there is a dead link in 
> Here's an example of a production-ready config file from the cra-kitchen example.

I dont know what is happening here

PR #9122, opened 2019-12-10
fix documentation typo: add missing = in react config example

karlsander:
## What I did
I tried to copy & paste this example config, and noticed the missing equal sign

## How to test

not really testable


shilman (approved):
🙈

PR #9125, opened 2019-12-11
Addon-docs: Hide addons on  docs-only stories

atanasster:
Issue: #9009

## What I did

pass the docsOnly parameter to the Layout and Preview containers.
Layout simply does not display the addons panels if docsOnly
Preview - if viewMode is 'story' and docsOnly - simulates the viewMode is actually 'docs', since the match function is implemented in the adding themselves.

## How to test
The steps in the original issue


shilman (approved):
Excellent! 👏

PR #9141, opened 2019-12-12
UI: Add initialActive option parameter

alexandre-lelain:

## What I did

- Added `initialActive` option parameter to storybook-ui config.
- Active tabs in Mobile layout are now handled using constants instead of hardcoded numbers.
- Updated `Options Parameters` doc page.

Users can now set the default active tab on Mobile.

## How to test

To test this feature, simply set `initialActive` in the options of the storybook config, and check that the default active tab on Mobile matches your setup:

```javascript
// config.js
...

/**
 * 'sidebar' - Sidebar
 * 'canvas' - Canvas
 * 'addons' - Addons
 */
const initialActive = X // one of above values
addParameters({
  options: {
     initialActive,
  },
})
```


ndelangen (approved):

shilman (approved):
Can we make constants for this, something like this:

```js
export const ACTIVE = {
  SIDEBAR: 0,
  ...
}
```

shilman (dismissed):
Can we make constants for this, something like this:

```js
export const ACTIVE = {
  SIDEBAR: 0,
  ...
}
```

alexandre-lelain:
Thanks for the review 👍 

You are right, we should use constants. I pushed the update

alexandre-lelain:
Hello @shilman , do you have any updates on this ? 🙏

shilman:
@alexandre-lelain yeah we're basically in feature freeze while we ship 5.3, which is in RC now and scheduled to ship in the first two weeks of january

ndelangen (approved):

shilman (changes requested):
WDYT about making these strings? Strings are self-documenting. Consider:

```
initialActive: 'sidebar'
```

vs

```
initialActive: 0
```

alexandre-lelain:
Yup you got a point 👍

shilman (approved):
💥 thanks!!

kumarharsh:
+1 for this feature. I was looking for a way to set the default tab when storybook first loads... Right now, it always loads `Canvas` no matter the order of the tabs.

chavlji:
Setting `initialActive` isn't working for me. I've set it up in manager like this:

Screenshot 2020-11-11 at 14 30 54

PR #9143, opened 2019-12-12
feat(ui, preview): add loader to canvas's preview

alexandre-lelain:

## What I did

- Added a **Loader** element to the **Preview's canvas tab** during the loading time of the stories to give a visual feedback to the visitors on mobile.

## How to test

To test this feature, simply launch any storybook and check that a loader element is displayed on the canvas tab once the page has loaded and that the stories are being fetched, for both Desktop & Mobile modes.


alexandre-lelain:
The JS logic is implemented and working. Now I need to make that loader a little more sexy than a simple title with a `loading...`  text content ;)

ndelangen:
WOW this is such an crazy good first contribution @alexandre-lelain 👏👏👏

alexandre-lelain:
Thanks!

domyen:
Thanks @alexandre-lelain! Can you add a story for this state in the official-storybook so we can get some test coverage, that way it'll be easier to maintain in the future. 🙏

alexandre-lelain:
@domyen I added a loading state story in the `UI/Layout/App` section, which is included in the official-storybook. I hope this is what you were expecting :)

lib/components/src/Loader/Loader.tsx
```diff
@@ -0,0 +1,26 @@
...
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+`;
+
+export const Loader = styled.div`
```
ndelangen: Could you please change this to the object style of creating components.

We do not use the emotion babel transform, because we re-export emotion's exports, the babel-transform can't detect that.

Besides the fact this string will appear literally in the bundle, the object style has preference for a multitude of reason, one of which is it's how most of the storybook code is written.

```suggestion
export const Loader = styled.div(({ theme }) => ({
  animation: `${spin} 1s linear infinite`,
  borderRadius: '50%',
  height: 48,
  left: 'calc(50% - 24px)',
  position: 'absolute',
  top: 'calc(50% - 24px)',
  width: 48,
  zIndex: 4,
  border: `3px solid ${theme.color.secondary}`,
  borderTop: `3px solid transparent`,
});
```

ndelangen (changes requested):
Minor change-style change, then good to merge!

lib/components/src/Loader/Loader.tsx
ndelangen: I tried to make the code-suggestion to replace then multiple lines, but couldn't get that. it should obviously replace the whole thing, not just the first line 😄

alexandre-lelain: Done 👍

ndelangen:
@alexandre-lelain When I checkout this branch, and enable slow network throttle-ing, I'm not seeing the loading spinner over the preview area.

Did something break?

alexandre-lelain:
@ndelangen I have just checkout the branch with the latest changes and it seems to be working fine. If you build the project (I'm using `yarn dev` for that) and launch the official-storybook server you should be able to see it.

The loading spinner appears on the preview page when the stories are being fetched - it is even easier to see it on slow network connections since the loading time is longer. And it disappears when the story is displayed.

Here is what I have on my local machine, after I have built the project, and launched the official-storybook server:

loader

domyen:
Thanks for the screenshots! Can we use the spinner in Storybook's design system ([link](https://storybook-design-system.netlify.com/?path=/docs/design-system-spinner--default-story))? That way the design language stay similar between the app and the rest of the sites in the ecosystem.

I think emotion and styled-components (used by the DS) share the same template literal API so you can paste the code [here](https://github.com/storybookjs/design-system/blob/master/src/components/Spinner.js) in with some minor adjustments.

ndelangen (approved):

ndelangen:
Hey @alexandre-lelain if you could make the spinner look like this:
https://storybook-design-system.netlify.com/?path=/docs/design-system-spinner--default-story in a follow-up PR?

alexandre-lelain:
Thanks for merging my work @ndelangen :)

Here's your follow-up PR that styles the `Loader` according to the Spinner of `@storybook/design-system` : https://github.com/storybookjs/storybook/pull/9249

I wanted to import `@storybook/design-system` as a dependency and simply import the `Spinner` component, but it seems that the TS compiler didn't understand it - I think we should work on that in the future to prevent code duplication as early as we can

PR #9149, opened 2019-12-13
Addon-docs: Support for iframeWidth parameter on stories

alonsosfera:
Issue: https://github.com/storybookjs/storybook/issues/8816

## What I did
Enabled Story components to receive parameter `iframeWidth` to configure the width of the iFrame shown in the Docs page

## How to test
Add parameter iframeWidth to `/storybook/examples/angularcli/src/stories/app.component.stories.ts` like shown below:

image

Run Angular-cli example, go to `App Component -> Component with separate template -> Docs page` and watch width of the iFrame changed as shown below.

image

- Is this testable with Jest or Chromatic screenshots? ... no
- Does this need a new example in the kitchen sink apps? ... no


alonsosfera:
@tmeasday @ndelangen Any idea why that test fails? I did some research and danger-js seems to that have an ongoing issue related to that. (https://github.com/danger/danger-js/issues/918)

addons/docs/src/blocks/Anchor.tsx
```diff
@@ -7,5 +7,7 @@ export interface AnchorProps {
 }
 
 export const Anchor: FC<AnchorProps> = ({ storyId, children }) => (
-  <div id={anchorBlockIdFromId(storyId)}>{children}</div>
+  <div id={anchorBlockIdFromId(storyId)} style={{ textAlign: 'center' }}>
```
shilman: why text-align center?

addons/docs/src/blocks/Story.tsx
```diff
@@ -71,14 +73,15 @@ export const getStoryProps = (
       `Story '${storyName}' is set to render inline, but no 'prepareForInline' function is implemented in your docs configuration!`
     );
   }
-
-  return {
+  const newVar = {
```
shilman: what is `newVar`?

lib/components/src/blocks/Preview.tsx
```diff
@@ -54,6 +54,7 @@ const PreviewContainer = styled.div<PreviewProps>(
     position: 'relative',
     overflow: 'hidden',
     margin: '25px 0 40px',
+    display: 'inline-block',
```
shilman: is this needed?

lib/components/src/blocks/Story.tsx
```diff
@@ -65,8 +66,13 @@ const InlineStory: FunctionComponent<InlineStoryProps> = ({ storyFn, height, id
   </div>
 );
 
-const IFrameStory: FunctionComponent<IFrameStoryProps> = ({ id, title, height = '500px' }) => (
-  <div style={{ width: '100%', height }}>
+const IFrameStory: FunctionComponent<IFrameStoryProps> = ({
+  id,
+  title,
+  height = '500px',
+  width = '950px',
```
shilman: Can we make this completely optional?

shilman (changes requested):
Thanks for the contribution! Lots of small issues with the PR. Can you take a look?

addons/docs/src/blocks/Anchor.tsx
alonsosfera: I added this so no matter what size the iFrame has, the component will always be centered on the Docs Page...

addons/docs/src/blocks/Story.tsx
alonsosfera: After adding the `width` property to the StoryProps interface, TypeScript kept giving me the following error:
image

I kept getting the error whenever I would return the object like this:
image

After inserting the object into a constant and then returning it, the error cleared out.
image

lib/components/src/blocks/Preview.tsx
alonsosfera: It is necessary in order for the preview container to take adjust to the iFrame's width.

Without it, it would look like this:
image


Instead of this:
image

addons/docs/src/blocks/Anchor.tsx
alonsosfera: Without it:
image

With it:
image

lib/components/src/blocks/Story.tsx
alonsosfera: That's actually just a default height and width. In case the IFrameStory component does not receive those parameters, it would just set it to 950x500 px...

Could you explain a little bit more about how this could be optional?

alonsosfera:
@shilman Do you mind taking a look at it?

shilman:
@alonsosfera busy trying to get 5.3 out right now. will take a look as soon as i've worked through my backlog. thanks for your patience!

addons/docs/src/blocks/Anchor.tsx
ndelangen: that visual outcome looks ok to me (I'm no designer), but this shouldn't be done with inline styles IMHO, I'm wondering if centering should be done with flexbox rather then text-align.

@shilman @domyen does the centering of the iframe seem like the right thing to do to you?

lib/components/src/blocks/Preview.tsx
ndelangen: Using flexbox for this is a little more robust I think.

shilman:
Picking this up this week. Sorry for the delay and thanks for your patience @alonsosfera

ndelangen:
I can't work out the merge conflicts, what can we do about this PR @shilman ?

shilman:
I'll take this one @ndelangen

m1lk1way:
hi, guys, it would be nice to have this. Are you going to finally merge it?

shilman:
@m1lk1way yup, will merge in the next few days and release as 6.1-alpha

shilman:
Looks like the fork for this has been deleted and the PR is not in a mergeable state, so I'm going to close this for now.

PR #9157, opened 2019-12-15
React: Add support for CRA without overrides

mrmckeb:
## What I did

This update ensures that no additional config is inserted after the CRA preset is loaded. 

## How to test

Run this alongside the new CRA preset, and the config should be returned without additional loaders.

Related PR for the preset: https://github.com/storybookjs/presets/pull/77.

### DIFF
--- a/app/react/src/server/framework-preset-cra.ts
+++ b/app/react/src/server/framework-preset-cra.ts
@@ -7,14 +7,12 @@ type Preset = string | { name: string };
 
 // Disable the built-in preset if the new preset is detected.
 const checkForNewPreset = (presetsList: Preset[]) => {
-  try {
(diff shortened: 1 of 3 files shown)


app/react/src/server/framework-preset-cra.ts
```diff
@@ -7,21 +7,21 @@ type Preset = string | { name: string };
 
 // Disable the built-in preset if the new preset is detected.
 const checkForNewPreset = (presetsList: Preset[]) => {
-  try {
```
mrmckeb: This was left over from old code where we were using `require.resolve`.

shilman (approved):
LGTM!

Snaptags:
Am I supposed the get the "Storybook support for Create React App is now a separate preset." warning in a project *not* set up using CRA at all?

PR #9179, opened 2019-12-18
fix(a11y): fix highlight of selected blindness color filter

gaetanmaisse:
## What I did

Fix highlight of selected blindness color filter.

## How to test

 - Open a story
 - Select a blindness color filter 
 - Open the menu of blindness color filters
 - The previously selected filter should be highlighted

After the fix:
Dec-18-2019 08-11-53


gaetanmaisse:
@shilman I think this fix should to included to 5.3 RC ;)

gaetanmaisse:
It looks like `Puppeteer & A11y tests` are also 🔴 on next 🤔

shilman (approved):
LGTM!

PR #9180, opened 2019-12-18
Use TS strict mode in a11y addon sources

gaetanmaisse:
## What I did

Use TS strict mode in a11y addon sources. But why?
 
 - `strict` mode is included in the default `tsconfig.json` generated by tsc CLI
 - Stricter rules help to find bugs faster than ever, for instance I found and fix https://github.com/storybookjs/storybook/pull/9179 by using some of the rules I activated in `tsconfig.json`
 - Some of our users are using TS in strict mode and are expecting that the types we are exposing are working in strict mode (e.g. https://github.com/storybookjs/storybook/issues/8233)
 - `noUnusedLocals` put in light the fact that some variables were not used for at least 8 months but still there and still requiring maintenance.

## How to test

- All tests should pass
- As it is refactoring, everything should work like before


ndelangen (commented):
How does this relate to https://github.com/storybookjs/storybook/pull/8883?

gaetanmaisse:
@ndelangen I just take a look at #8883, there will be only minor conflicts because this PR focus on refactoring/cleaning some TS (strict mode) issues and not deeply modifying how the addon works. #8883 can be merged (but not immediately as we must wait in 6.0 I guess), then just tag me and I will fix the conflicts with this branch. 🎄🎅

ndelangen (approved):

PR #9182, opened 2019-12-18
Add react-scripts automatic setup to docs

mrmckeb:
Issue: #5183

## What I did
Users are short on time (as we all are) and often skip over the docs to the React section, without first trying automatic setup. 

As a result, `react-scripts` users install with `--type react`  - but they should pass `--type react_scripts`.

## How to test
N/A


shilman (approved):
Well said! 👏

PR #9187, opened 2019-12-18
Storyshots: update render w/ React.createElement

tmikeschu:
Issue: https://github.com/storybookjs/storybook/issues/8177

## What I did

Wrapped the `story.render()` calls with `React.createElement` for storyshots.

This allows you to use hooks in CSF without wrapping components in a thunk-like function.

## How to test

Jest

Nope

Nope

### DIFF
--- a/addons/storyshots/storyshots-core/src/frameworks/react/renderShallowTree.ts
+++ b/addons/storyshots/storyshots-core/src/frameworks/react/renderShallowTree.ts
@@ -1,8 +1,10 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
 import shallow from 'react-test-renderer/shallow';
 
 function getRenderedTree(story: any, context: any, { renderer, serializer }: any) {
-  const storyElement = story.render();
+  const storyElement = React.createElement(story.render);
   const shallowRenderer = renderer || shallow.createRenderer();
   const tree = shallowRenderer.render(storyElement);
   return serializer ? serializer(tree) : tree;
--- a/addons/storyshots/storyshots-core/src/frameworks/react/renderTree.ts
+++ b/addons/storyshots/storyshots-core/src/frameworks/react/renderTree.ts
@@ -1,8 +1,10 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
 import reactTestRenderer from 'react-test-renderer';
 
 function getRenderedTree(story: any, context: any, { renderer, ...rendererOptions }: any) {
-  const storyElement = story.render();
+  const storyElement = React.createElement(story.render);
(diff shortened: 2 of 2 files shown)


tmikeschu:
Hmmm all my tests pass locally with only these changes and only running the `yarn bootstrap` setup.

shilman:
@tmikeschu i think at this point you just need to update snapshots

```
yarn test --core --update
```

tmikeschu:
@shilman I had the same thought, but I don't get a diff/updates from that. I'm perplexed!

shilman:
Snapshots are updating in a kind of weird way. I think this qualifies as a breaking change, but if it seems correct we can work this into our next release, which will be 6.0.

GitHub_Desktop

@tmikeschu @Hypnosphi please lmk what you think!

tmikeschu:
Interesting. Is it possible the discrepancy is yarn version?

tmikeschu:
I'm not sure what the best solution is, but let me know how I can help!

shilman:
@tmikeschu that's a screenshot from github desktop. there were a bunch of diffs, i just chose one at random.

Hypnosphi:
I think we can't use shallow rendering with this approach, because now we snapshot the story return value, not the one of component

Hypnosphi:
> I just see a new <storyFn> wrapping node. <button> appears to be nested within <Button>

That's the case for default renderer and enzyme's mount

> replacing it.

And that's the case for shallow renderers: `react-test-renderer/shallow` which is default in `renderShallowTree`, and enzyme's shallow. In this case we lose the information about actual component output which is a large regression. Unfortunately, I don't have a clear solution in mind.

Hypnosphi:
> I don't get a diff/updates

@tmikeschu this can happen if your `dist` directories are outdated. You should either run `yarn bootstrap --core` once again, or have `yarn dev` running in background to transpile all your changes on the fly

shilman:
@Hypnosphi thanks for all the investigation on this. nothing's ever easy, is it? i'm not sure what a good solution is here. current solution with workaround is non-ideal, but it sure is simpler!

Hypnosphi:
Just to be clear. Current solution makes shallow storyshots test a comletely different thing, comparing to what they tested before. Imagin you have a component like that:

```js
export const MyComponent = ({someProp}) => (
  <SomeOtherComponent
    someImportantProp={someLogic(someProp)}
  />
)
```

and a story like that:

```js
export const MyStory = () => <MyComponent someProp={someValue}  />
```

Before the change, the shallow storyshot of the story would output something like that:

```js
<SomeOtherComponent
  someImportantProp={valueDependingOnComponentLogic}
/>
```

And after the change, it's

```
<MyComponent someProp={someValue}  />
```

So now, instead of testing your component, you only test your story. But it's the component what you actually use in your product, and therefore want to test

shilman:
@Hypnosphi thanks for clarifying. then maybe we should do this change only for deep rendering and not for shallow rendering?

Hypnosphi:
Maybe so. By my suggestion with two-pass rendering may work as well

PR #9194, opened 2019-12-19
Fix localeCompare() examples. Fixes #9193

### LINKED ISSUE #9193: storySort docs show incorrect usage of localeCompare
**Describe the bug**

The `localeCompare()` examples in the docs show a `{numeric: true}` option that is ignored because it is in the wrong parameter place. This is what the docs currently have:

```js
storySort: (a, b) =>
  a[1].kind === b[1].kind
    ? 0
    : a[1].id.localeCompare(b[1].id, { numeric: true }),
```

But the second parameter of `localeCompare` should be a locales string and it's the third parameter that is the options object. See MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

Following an example from the MDN site, this is how the `localeCompare()` method should be called:
```js
storySort: (a, b) =>
  a[1].kind === b[1].kind
    ? 0
    : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
```
 
**To Reproduce**
1. Check out the code in PR #9194 
2. Run the tests to see the correct usage of localeCompare().
3. Delete the second parameter (`undefined,`) from the localeCompare() in the new test (line 64 of `lib/client-api/src/story_store.test.ts`) 
4. Run the tests to see it fail to sort "b-b10--1" after "b-b9--1".

**Expected behavior**

The docs to have code snippets that work. :)

JohnAlbin:
Issue: #9193 

## What I did

- Fixes docs where localeCompare() is used incorrectly.
- Adds test for `storySort` option that uses a custom function where localeCompare() is used correctly.

## How to test

In this PR, you'll see a test that currently passes. If you want to see how the current docs don't work, you can delete the `undefined,` parameter from the `localeCompare()` in the new test and watch it fail to sort  "b-b10--1" after "b-b9--1".

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -256,7 +256,7 @@ For example, here's how to sort by story ID using `storySort`:
 addParameters({
   options: {
     storySort: (a, b) =>
-      a[1].kind === b[1].kind ? 0 : a[1].id.localeCompare(b[1].id, { numeric: true }),
+      a[1].kind === b[1].kind ? 0 : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
   },
 });
 ```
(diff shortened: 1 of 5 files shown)


shilman (approved):
Thank you!!! 🙏

PR #9201, opened 2019-12-19
ViewportAddon: Add EVENT_VIEWPORT_CHANGED event

hgenru:
Issue: https://github.com/storybookjs/storybook/issues/8230

## What I did
I provide a new event `EVENT_VIEWPORT_CHANGED` that you can use to react to viewport change.

## How to test
You can test this with example from documentation

### DIFF
--- a/addons/viewport/src/components/Tool.tsx
+++ b/addons/viewport/src/components/Tool.tsx
@@ -6,10 +6,11 @@ import { styled, Global, Theme, withTheme } from '@storybook/theming';
 
 import { Icons, IconButton, WithTooltip, TooltipLinkList } from '@storybook/components';
 
-import { useParameter, useAddonState } from '@storybook/api';
-import { PARAM_KEY, ADDON_ID } from './constants';
-import { MINIMAL_VIEWPORTS } from './defaults';
-import { ViewportAddonParameter, ViewportMap, ViewportStyles, Styles } from './models';
+import { useParameter, useAddonState, API } from '@storybook/api';
+import { PARAM_KEY, ADDON_ID, EVENT_VIEWPORT_CHANGED } from '../constants';
+import { MINIMAL_VIEWPORTS } from '../defaults';
+import { ViewportAddonParameter, ViewportMap, ViewportStyles, Styles } from '../models';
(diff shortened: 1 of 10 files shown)


addons/viewport/src/components/Tool.tsx
```diff
@@ -0,0 +1,258 @@
...
+          <ActiveViewportLabel title="Viewport height">
+            {styles.height.replace('px', '')}
+          </ActiveViewportLabel>
+        </ActiveViewportSize>
+      ) : null}
+    </Fragment>
+  );
+};
+
+export const ViewportTool = withTheme(memo(ViewportToolComponent));
```
hgenru: HOCs inlining cause problem with typings. But now git thinks that I deleted and created file. Huh, I separate move and edit to avoid this but it did not help j_j.

addons/viewport/src/models/ViewportAddonEvents.ts
```diff
@@ -0,0 +1,3 @@
+import { ViewportItem } from '../components/types';
+
+export type ViewportChangedEventProps = { item: ViewportItem; isRotated: boolean };
```
hgenru: Maybe expose this typing to index.ts?

addons/viewport/src/components/Tool.tsx
ndelangen: perhaps you could try using `useTheme` from `@storybook/theming`?

I'm not sure it's used anywhere yet, but it should be available, I think

ndelangen: This remove & add makes it super hard to review right

addons/viewport/src/components/Tool.tsx
```diff
@@ -0,0 +1,258 @@
...
+  if (styles === null) {
+    return null;
+  }
+  const result = typeof styles === 'function' ? styles(prevStyles) : styles;
+  return isRotated ? flip(result) : result;
+};
+
+type Props = {
+  theme: Theme;
+  api: API;
```
ndelangen: I don't think there should be a need for this, if you can do `useStorybookApi` ?

PR #9249, opened 2019-12-26
feat(ui, components): style the loader with the design-system

alexandre-lelain:

## What I did

- The `Spinner` style from [@storybook/design-system](https://github.com/storybookjs/design-system) was applied to the `Loader` component.

## How to test

You can either open the preview deploy and test that the Loader's style is following [this one](https://storybook-design-system.netlify.com/?path=/docs/design-system-spinner--default-story), or checkout the branch, build the project and run the official storybook.

## What it looks like

styled_loader


domyen:
Thanks @alexandre-lelain, I'll piggy back on this to add some theming support 🙏

domyen:
I found a regression where the loading spinner shows up in cases that it's not supposed to.  

This shouldn't appear in the App:default story. 
image

The `App:default`, `Preview:no tabs`, and `Preview:with tabs` should not render the Loader. [This build in Chromatic](https://www.chromaticqa.com/build?appId=5a375b97f4b14f0020b0cda3&number=13660) is where it was originally accepted (by mistake).

The only stories that should be affected by the Loader work are the ones @alexandre-lelain added `App:loading state` and `Loader:infinite state`.

Can you take a look at why Loader is being rendering in the other components and update the stories if needed? 🙏

domyen (dismissed):
See comment https://github.com/storybookjs/storybook/pull/9249#issuecomment-569337116

alexandre-lelain:
The fact the `Loader` is being displayed on those stories is "expected" - just like the placeholders for the stories' names on the left.

The Loader relies on the same core logic as the stories' names placeholders. It will be displayed automatically when the UI is in loading state (ie after page has loaded, when the stories' JS bundles are being fetched).

Maybe we can override this in the stories, but I'm not sure we can do this without adding some code in the components' sources - which, in my opinion, will add potential source of errors.

domyen:
Hey Alex, the stories are presentational and used for visual testing. Our
goal is to qa the expected experience from the eyes of our users.

In this case, we should mock the loader somehow so it doesn’t show up in
the other cases. For example, the app:default should not show the loader
and app:loading should. Or change how the logic works.

Ultimately, we can’t merge this or the other PR without updating the tests
in a project of this size.

alexandre-lelain:
Alright, I'll see how I can mock the loader in those cases 👍

alexandre-lelain:
Hey @domyen I pushed some changes to hide the Loader when needed.

However it seems I cannot update the chromatic snapshots from my machine (the `scripts/test.js` crashes) - I suspect it's because I'm on Windows (I have nodejs@10.16.3).

But you can see that the loader was hidden correctly in the [App:default](https://www.chromaticqa.com/component?appId=5a375b97f4b14f0020b0cda3&name=UI%2FLayout%2FApp&buildNumber=13866&specName=default) story.

domyen:
Thanks alexandre, on holiday right now but will review when I get back to a
computer 🙏 🙇

ndelangen:
LGTM
Screenshot 2020-01-08 at 15 14 29

ndelangen (approved):

PR #9095, opened 2019-12-07
configure tabs: title, visibility, order and disable

atanasster:
Issue: #8631 #8630 #7066

## What I did
added the following global configuration/parameters settings:

```
previewTabs: {
   addon_id: string: {
    title?: string; // new title for the addon tab
    index?: number; // override the place/order of the tabs (by default the change uses the order of the addon_id keys but index can be used to override in the story parameyters
    hidden?: boolean; // wether to hide the addon tab
    disabled?: boolean; // wether to disable the addon tab
  }
},
```

following formats for configuration: 
`manager.js`:
```
import { addons } from '@storybook/addons';
addons.setConfig({
  previewTabs: {
    // the order of the tabs is configured by the order here
    canvas: null, // canvas is here simply listed so its place as first tab
    'storybook/docs/panel': 'MDX', // the configuration is either an object or a title string
    'storybookjs/notes/panel': { title: 'Annotations', hidden: true },
    graphiql: {
      hidden: true,
    },
  },
});
```

and also story specific parameters:

```
export default {
  title: 'Addons/GraphQL',
  parameters: {
    previewTabs: {
      graphiql: { hidden: false, title: 'GraphiQL' },
    },
  },
};

```

## How to test

1. in official storybook - the manager.js hides by default the Notes and the GraphiQL tabs

grab16

2. by going to the Addons/Notes stories, the notes tabs is made visible and shows the globally set custom label:
```
export default {
  title: 'Addons/Notes',
  parameters: {
    previewTabs: {
      'storybookjs/notes/panel': { hidden: false },
    },
  },
};
```
grab15

3. by going to the Addons/GraphQL stories, the graphic tab is made visible and custom title is set locally (just for the stories here)
```
export default {
  title: 'Addons/GraphQL',
  parameters: {
    previewTabs: {
      graphiql: { hidden: false, title: 'GraphiQL' },
    },
  },
};
```
grab17

4. by going to the Addons/markdown docs page, the order of the tabs is custom set for the stories here (the Docs tab is first, the Canvas tab is second):

```
<Meta
  title="Addons/Docs/markdown-docs"
  parameters={{ previewTabs: { 'storybook/docs/panel': { index: -1 }}}}
/>
```
grab18

PR #9323, opened 2020-01-04
Official example: try moving options to `manager.js`

Hypnosphi:
I tried applying https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#using-managerjs to our own example storybook

I found two issues:
1. `showRoots` doesn't work
2. `manager.js` doesn't hot reload: e.g. if you change it to `theme: themes.dark`, it applies only after page reload


ndelangen:
For 6.0.0 the documentation is correct.

For 5.3 we need to reverse this documentation a bit, to tell users to keep setting the global parameter

ndelangen:
I can confirm that:
> showRoots doesn't work

I added a FIX in `api/modules/stories.ts`

ndelangen:
@tmeasday I did find what @Hypnosphi found too:

When a user would make migrate all options to `manager.js` certain things break.

- storySort (I added a fix for this, in this PR)
- showRoots (I added a fix for this, in this PR)
- theme *(the preview never gets the theme, meaning addon-docs isn't themed the same)* (I opened a new ticket for this: https://github.com/storybookjs/storybook/issues/9662)

@Hypnosphi WDYT?

after this PR, this becomes perfectly valid config:
```js
// manager.js
import { themes } from '@storybook/theming';

addons.setConfig({
  showRoots: true,
  theme: themes.light, // { base: 'dark', brandTitle: 'Storybook!' },
  storySort: (a, b) =>
    a[1].kind === b[1].kind ? 0 : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
});
```

ndelangen:
`manager.js` not hot-module-reloading is fine by me for now. I don not consider this an critical important feature. I'd be nice to add in the future.

lib/api/src/modules/stories.ts
```diff
@@ -229,7 +230,7 @@ const initStoriesApi = ({
         hierarchyRootSeparator: rootSeparator = undefined,
         hierarchySeparator: groupSeparator = undefined,
         showRoots = undefined,
-      } = (parameters && parameters.options) || {};
```
tmeasday: @ndelangen do you think we should come up with a principled way to pull configuration from both parameters and `.getConfig()`? It seems like this comes up a bit and maybe there should be a single API for it?

lib/api/src/modules/stories.ts
```diff
@@ -223,94 +223,99 @@ const initStoriesApi = ({
...
-
-        // 3. If the user passes showRoots, or doesn't match above, do a simpler splitting.
-      } else {
-        const parts: string[] = kind.split('/');
-        if (showRoots && parts.length > 1) {
-          [root, ...groups] = parts;
+    const { storySort = () => 0 } = provider.getConfig();
+
+    const storiesHashOutOfOrder = Object.entries(input)
+      .sort(storySort)
```
tmeasday: Am I right in saying that if you set the `storySort` options parameter it sorts the story on the preview side (and so they'll be sorted e.g. in `store.extract()`), but if you set the addon option it sorts them only in the UI here?

That seem... super confusing, and likely to cause problems (if for Chromatic + other integrators if nothing else).

I would vote you just **CANNOT** set `storySort` in `manager.js`. It is a property of the store, not a UI option. With that in mind it probably shouldn't live in `parameters.options`. 🤷‍♂

tmeasday (changes requested):

lib/api/src/modules/stories.ts
tmeasday: For instance @atanasster does something very similar in https://github.com/storybookjs/storybook/pull/9095

lib/api/src/modules/stories.ts
ndelangen: I'm considering putting it in `main.js` and making it a hard requirement it's a pure function.

That way I can inject the function into both environments...

tmeasday: I'm not necessarily against putting it in `main.js` but why do we need it in both environments?

ndelangen: I'm looking into this today.

lib/api/src/modules/stories.ts
```diff
@@ -222,94 +223,96 @@ const initStoriesApi = ({
...
+
+        // Kind splitting behaviour as per https://github.com/storybookjs/storybook/issues/8793
+        let root = '';
+        let groups: string[];
+        // 1. If the user has passed separators, use the old behaviour but warn them
+        if (typeof rootSeparator !== 'undefined' || typeof groupSeparator !== 'undefined') {
+          warnRemovingHierarchySeparators();
+          if (usingShowRoots) warnUsingHierarchySeparatorsAndShowRoots();
+          ({ root, groups } = parseKind(kind, {
+            rootSeparator: rootSeparator || '|',
```
shilman: Are we removing these in 6.0? Or are we planning to patch this back to 5.3? Or will we do that in a separate PR?

ndelangen: Yeah removing those deprecations is on my wishlist

shilman: I know a single key on your keyboard that could make your wish come true 😉

ndelangen: I'm well aware of existence of the `delete` key on my keyboard

ndelangen: Screenshot 2020-02-13 at 14 59 27

ndelangen (approved):

tmeasday:
👏

PR #9354, opened 2020-01-08
Core: Fix HMR for global decorators in main.js config

ndelangen:
Issue: #9192 

## What I did

Adding a _StorybookPreserveDecorators unto the module for flagging when to preserve global decorators

## How to test

- open the official storybook in dev mode
- change a story, it should HMR correctly without erros
  /?path=/story/basics-syntaxhighlighter--bash uses the global theme from a global decorator so it's a good reproduction
- change preview.js file, it should reload the iframe, not the manager


shilman (approved):
LGTM! 💯

PR #9359, opened 2020-01-08
CLI: Fix `sb init` to use spawn.sync if creating package.json

atanasster:
Issue: #9358

## What I did
Not familiar with it, but seems the cross-spawn has changed? 

I moved to use the sync method and also there were multiple unneeded parameters being returned

## How to test
like in the issue, `sb init --type html` or other storybook types

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -17,6 +17,22 @@
     "skipFiles": [
       "<node_internals>/**"
     ]
-  },
+  }, {
+    "type": "node",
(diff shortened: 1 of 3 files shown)


ndelangen (approved):

atanasster:
it turns out the newly created package.json was never re-read after a yarn/npm init spawn. 
Not sure how it used to work. I also now added a new VSCode debug profile so we can easier debug the cli.

PR #9360, opened 2020-01-08
cssresources-addon: Do not use SyntaxHighlighter with the very long code-content.

tomilaurell:
Issue:

## What I did

Adds support to use very long css content with **cssresources-addon**. The current implementation tries to render (forever) all the content with SyntaxHighlighter. This PR will restrict the use of SyntaxHighligher with content length up to 100k. In case the SyntaxHighligher is not used, the first 100k of the content will be rendered within plain pre-tag and warning of partly missing content is added to the end of it.

## How to test

This is example with less-files.

webpack.config.js
```
...
// do not include style-loader to inject css content manually
config.module.rules.push({
    test: /\.less?$/,
    loaders: [require.resolve('css-loader'), require.resolve('less-loader')], 
  })
...
```

config.js
```
...
import globalStyles from '../styles/index.less'

cssresources: [
    {
      id: `Global toggleable styles`,
      code: `<style>${globalStyles}</style>`,
      picked: true,
    },
  ],
...
```

## Screenshot

image

### DIFF
--- a/addons/cssresources/src/css-resource-panel.tsx
+++ b/addons/cssresources/src/css-resource-panel.tsx
@@ -1,7 +1,8 @@
 import React, { Component, Fragment } from 'react';
-import { SyntaxHighlighter } from '@storybook/components';
+import { SyntaxHighlighter, Placeholder, Spaced, Icons } from '@storybook/components';
 import { STORY_RENDERED } from '@storybook/core-events';
 import { API } from '@storybook/api';
+import { styled } from '@storybook/theming';
 
 import { EVENTS, PARAM_KEY } from './constants';
 import { CssResource } from './CssResource';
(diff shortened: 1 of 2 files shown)


addons/cssresources/src/css-resource-panel.tsx
```diff
@@ -96,7 +98,24 @@ export class CssResourcePanel extends Component<Props, State> {
...
                 <span>#{id}</span>
               </label>
-              {code ? <SyntaxHighlighter language="html">{code}</SyntaxHighlighter> : null}
+              {code && code.length < maxLimitToUseSyntaxHighlighter && (
+                <SyntaxHighlighter language="html">{code}</SyntaxHighlighter>
+              )}
+              {code && code.length >= maxLimitToUseSyntaxHighlighter && (
+                <div>
+                  <pre>{code.substring(0, maxLimitToUseSyntaxHighlighter)} ...</pre>
+                  <div
```
ndelangen: Thank you @tomilaurell 

This should probably use a Placeholder component:
https://storybookjs.now.sh/official-storybook/?path=/story/basics-placeholder--two-children

This way it will style along nicely with the rest of storybook, and be possible for users to theme it.

There are also Icons you can use:
https://storybookjs.now.sh/official-storybook/?path=/story/basics-icon--labels

And we have this component for spacing:
https://storybookjs.now.sh/official-storybook/?path=/story/basics-spaced--row

ndelangen (changes requested):

addons/cssresources/src/css-resource-panel.tsx
ndelangen: You can import all these components from `@storybook/components`

tomilaurell: I will make those changes tomorrow.

ndelangen: Awesome, thanks!

ndelangen:
We use yarn in our repo, so if you could also please remove the `package-lock.json`?

ndelangen (approved):

ndelangen:
beautiful

PR #9361, opened 2020-01-08
REGEN lockfile && FIX errors from upgrades

ndelangen:


addons/knobs/src/__types__/knob-test-cases.ts
```diff
@@ -15,7 +15,6 @@ import {
 } from '../index';
 
 // Note: this is a helper to batch test return types and avoid "declared but never read" errors
-// eslint-disable-next-line @typescript-eslint/no-empty-function
 function expectKnobOfType<T>(..._: T[]) {}
```
mrmckeb: This could also be resolved by returning undefined instead... just a thought.

app/react/package.json
```diff
@@ -49,15 +49,15 @@
...
+    "react-dev-utils": "^10.0.0",
     "regenerator-runtime": "^0.13.3",
     "semver": "^6.0.0",
     "ts-dedent": "^1.1.0",
     "webpack": "^4.33.0"
   },
   "devDependencies": {
     "@types/mini-css-extract-plugin": "^0.8.0",
-    "@types/node": "^12.12.11",
+    "@types/node": "^13.1.5",
```
mrmckeb: Should we lock these to the supported minimum, like 10.x? I didn't check the minimum.

examples/angular-cli/src/stories/moduleMetadata/dummy.service.ts
```diff
@@ -2,7 +2,7 @@ import { Injectable } from '@angular/core';
 
 @Injectable()
 export class DummyService {
-  // eslint-disable-next-line @typescript-eslint/no-empty-function, no-useless-constructor
+  // eslint-disable-next-line no-useless-constructor
   constructor() {}
```
mrmckeb: Could this just be omitted?

mrmckeb (approved):
Looks good, a few thoughts - but no changes needed.

PR #9367, opened 2020-01-09
Upgrade html-webpack-plugin

ndelangen:
Issue: #9026

## What I did
UPGRADE the dependency

@pllee


shilman (changes requested):
@ndelangen Image loading now appears to be broken in Polymer on Chromatic. I re-ran the build just in case it was a random network thing, but it looks like a problem with the upgrade (or with our polymer implementation?)

nickpresta:
FWIW, our project had the same issue and upgrading to `html-webpack-plugin@4.0.0-beta.11` fixed it. That doesn't mean this upgrade doesn't _also_ break other things in the Storybook ecosystem, but at least for a fairly normal React project, upgrading to `4.0.0-beta.11` (via yarn resolutions) fixed things.

PR #9368, opened 2020-01-09
Document showRoots for 5.3 by parameters

ndelangen:
Issue: https://github.com/storybookjs/storybook/pull/9323

## What I did
corrected the documentation so it's clear to users that for now they should configure showRoots in parameters


shilman (approved):
LGTM

PR #9408, opened 2020-01-13
Remove autoprefixer as peer dependency.

ruimarques:
Issue:

`autoprefixer` is already a dependency from `@storybook/core` which is a dependency from this  **@storybook/angular** package.

The versions mismatch, causing a:

`npm WARN @storybook/angular@5.3.1 requires a peer of autoprefixer@^8.1.0 but none is installed. You must install peer dependencies yourself.`

`$ npm ls autoprefixer
`
Screenshot 2020-01-13 at 13 20 24

## What I did

Removed `autoprefixer` as peer dependency.
Alternative would be to keep `autoprefixer` version matching with the one from `@storybook/core` ...


ruimarques:
Wrong base branch

PR #9416, opened 2020-01-13
Addon-knobs: Fix null knob values in select

ndelangen:
Issue: #9405

## What I did

ADD a story to test
FIX the bug by adding a `=== null` check


shilman (approved):
LGTM!

PR #9513, opened 2020-01-17
Addon-docs: Preserve Source indentation by default

atanasster:
Issue: #8078

## What I did

- Removed default decent (remove template strings spacing) 
- code styling prevent white space wrapping style

## How to test

I didn't get much stories to try with, so mostly tested indentation. 
Possibly find stories that are using template strings?

### DIFF
--- a/lib/components/src/blocks/Source.tsx
+++ b/lib/components/src/blocks/Source.tsx
@@ -69,4 +69,7 @@ const Source: FunctionComponent<SourceProps> = props => {
   return <ThemeProvider theme={convert(overrideTheme)}>{syntaxHighlighter}</ThemeProvider>;
 };
 
+Source.defaultProps = {
+  format: false,
+};
 export { Source };
--- a/lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
+++ b/lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
@@ -80,7 +80,6 @@ const Code = styled.code({
   flex: 1,
   paddingRight: 0,
   opacity: 1,
-  whiteSpace: 'pre',
 });
 
 export interface SyntaxHighlighterProps {
--- a/lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
+++ b/lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
@@ -105,7 +105,7 @@ export const SyntaxHighlighter: FunctionComponent<Props> = ({
   copyable = false,
   bordered = false,
   padded = false,
-  format = false,
+  format = true,
   className = null,
   ...rest
 }) => {
--- a/lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
+++ b/lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
@@ -80,6 +80,7 @@ const Code = styled.code({
   flex: 1,
   paddingRight: 0,
   opacity: 1,
+  whiteSpace: 'pre',
 });
 
 export interface SyntaxHighlighterProps {
@@ -104,7 +105,7 @@ export const SyntaxHighlighter: FunctionComponent<Props> = ({
   copyable = false,
   bordered = false,
   padded = false,
-  format = true,
+  format = false,
   className = null,
   ...rest
 }) => {


shilman:
This is "kind of" a breaking change. I think it's probably better than what it's replacing in most cases. Here's a case where it's not:

Addons___Actions_-_Multiple_actions_⋅_Storybook

Addons___Actions_-_Multiple_actions_as_object_⋅_Storybook_and_Storybook_Roadmap_Management_-_Google_Docs

shilman:
@atanasster Oh, see also the failing chromatic tests! 😱

atanasster:
> This is "kind of" a breaking change. I think it's probably better than what it's replacing in most cases. Here's a case where it's not:

Can you please elaborate, the new version is outputting it as it is in the code, while before it was mangling the code> no ?

Here is the code for the story:

```
export const MultipleActionsObjectConfig = () => (
  <Button
    {...actions({ onClick: 'clicked', onMouseOver: 'hovered' }, { clearOnStoryChange: false })}
  >
    Moving away from this story will persist the action logger
  </Button>
);
``

**Comment by atanasster** [comment]
chromatic tests - no idea what they say :) 
Like one issue seems to be a styled-components className? And some issues are expectedly the code output changes?

**Comment by atanasster** [comment]
> Can you please elaborate, the new version is outputting it as it is in the code, while before it was mangling the code> no ?

If the issue is the long line, I specifically removed the white-space wrapping style, so the code stays the same ad the story was written and the user has control over the code.  I can put it back in (to force-warp lines)

**Comment by shilman** [comment]
I took a look at all the different permutations. What I think:

1) Leave the `SourceHighlighter` alone
2) Modify `format=false` as the default on the `lib/components` Source component instead

That way we still get line wrapping (debatable, I know) but most of the annoying bugs that people are reporting go a way. WDYT?

**Comment by atanasster** [comment]
sounds good, checking in

**Review by shilman — APPROVED** [review]
LGTM 😅

**Comment by patricklafrance** [comment]
Not sure if these changes are a good idea.

Before 5.3.7 about 15% of my source were not formatted well. Not it's about 90%.

Most of them are like the following:

image

**Comment by shilman** [comment]
@atanasster WDYT

**Comment by atanasster** [comment]
@patricklafrance - can you paste your code that breaks here?

@shilman yes, lets keep for the 6.0 branch, its too much of a risk to push out. In general, i think we should not touch the source code on our end (including in source-loader) and use the syntax highlighter’s prettier configuration that would be also available to the end-user - ie he prefers 2 vs 4 spaces etc.

**Comment by patricklafrance** [comment]
@atanasster for this particular case:

```
<Preview>
    <Story name="size">
        <div className="flex flex-row items-end">
            <Button type="button" size="tiny">Tiny</Button>
            <Button type="button" size="small">Small</Button>
            <Button type="button" size="medium">Medium</Button>
            <Button type="button" size="large">Large</Button>
        </div>
    </Story>
</Preview>
```

I have a few other cases:

image

```
<Preview>
    <Story name="default">
        <RemoteSearchInput
            onFetchResults={useDefaultResultsFetcher("https://wft-geo-db.p.rapidapi.com/v1/geo/countries?limit=5", "namePrefix", { requestOptions: API_REQUEST_OPTIONS })}
            onResults={response => {
                return response.data.map(x => searchInputResult(x.code, x.name));
            }}
            onValueChange={logValueChanged}
            placeholder="Type a country name"
        />
    </Story>
</Preview>
```

image

```
<Preview>
    <Story name="disabled">
        <MultiSelect
            items={DEFAULT_ITEMS}
            defaultValues={[GROUP_RESTORED_VALUE, GROUP_NAME_CHANGED_VALUE]}
            disabled
            onValuesChange={logValuesChanged}
        />
    </Story>
</Preview>
```

image

```
<Preview>
    <Story name="custom-color">
        <Label
            naked
            style={{
                backgroundColor: "#FFF6E7",
                color: "var(--marine-500)",
                border: "1px solid #FEE9C3"
            }}
        >
            Notification Sent
        </Label>
    </Story>
</Preview>
```

shilman:
@atanasster This has already been released in 5.3.7. Should I revert the change there?

atanasster:
Ah sorry didnt realize. Based on what @patricklafrance is saying, i guess we should revert it if possible?

shilman:
👍 will revert it @patricklafrance @atanasster and let's figure out the right fix on the `next`/`6.0` branch

PR #9538, opened 2020-01-19
Node-logger: Move `@types/npmlog` to dependencies 

gaetanmaisse:
Issue: #7855

## What I did

As `npmlog` is exported [here](https://github.com/storybookjs/storybook/blob/28fed37f0d7d6a3c1276d0e54060205858f54c29/lib/node-logger/src/index.ts#L29) --`export { npmLog as instance };`, TS users of this lib need to have `npmlog` types too, so `@types/npmlog` have to be in `dependencies` and not in `devDependencies`.

### DIFF
--- a/lib/node-logger/package.json
+++ b/lib/node-logger/package.json
@@ -27,6 +27,7 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


shilman (approved):
LGTM

PR #9557, opened 2020-01-20
fix(webcomponents): make types play nicely with lit-element

urish:
## What I did

Before this change, returning html`<sometag>...</sometag>` from `storiesOf(...).add` would result in the following error: Type 'TemplateResult' is not assignable to type 'StoryFnHtmlReturnType'


ndelangen (approved):

ndelangen:
I think we need a eslint-ignore rule for:
```
/tmp/*********/app/web-components/src/client/preview/types.ts
  2:1  error  'lit-element' should be listed in the project's dependencies. Run 'npm i -S lit-element' to add it  import/no-extraneous-dependencies
```

urish:
@ndelangen do you want me to amend the pull request and add this?

ndelangen:
done

ndelangen (approved):

urish:
Even better :-)

p.s. you can see it in action here: https://elements.wokwi.com/
source code: https://github.com/wokwi/wokwi-elements

PR #9560, opened 2020-01-20
Addon-docs:  Fix TS false default value in prop table

atanasster:
Issue:  #9394

## What I did
Conditional check for null and undefined default prop
Added example to cra-ts-kitchen-sink


shilman (approved):
LGTM

PR #9566, opened 2020-01-20
api: adding useSharedState, useStoryState

atanasster:
Issue:

## What I did
- isolated useSharedState in addition touseAddonState
- created new useStoryState that uses the story id as the key 
- added new useStoryState stories for examples/dev-kits

## How to test

examples/dev-kits

### DIFF
--- a/lib/api/src/index.tsx
+++ b/lib/api/src/index.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, Component, useContext, useEffect, useMemo } from 'react';
+import React, { ReactElement, Component, useContext, useEffect, useMemo, useRef } from 'react';
 import memoize from 'memoizerific';
 // @ts-ignore shallow-equal is not in DefinitelyTyped
 import shallowEqualObjects from 'shallow-equal/objects';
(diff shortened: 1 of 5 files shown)


lib/api/src/index.tsx
```diff
@@ -389,14 +389,25 @@ export function useAddonState<S>(addonId: string, defaultState?: S) {
...
 }
+
+export function useAddonState<S>(addonId: string, defaultState?: S) {
+  return useSharedState<S>(addonId, defaultState);
+}
+
+export function useStoryState<S>(defaultState?: S) {
+  const {
+    state: { storyId },
+  } = useContext(ManagerContext);
```
ndelangen: Can use `useStorybookState` ?

ndelangen:
WILD 👏

lib/api/src/index.tsx
atanasster: sure, just a sec

ndelangen (approved):

lib/api/src/index.tsx
atanasster: oops - spoken too early. There is already a useStorybookState :

https://github.com/storybookjs/storybook/blob/c35919c497ec03c8d20ec3c074a974ec04876142/lib/api/src/index.tsx#L300

ndelangen (approved):

PR #9574, opened 2020-01-21
CLI: Improve Rax template

SoloJiang:
Framework Support Table typo


shilman (changes requested):
Doesn't this change put `React` as a heading for the addons column?

SoloJiang:
> Doesn't this change put `React` as a heading for the addons column?

Sorry, I wil fix it soon.

shilman (approved):
LGTM!

PR #9579, opened 2020-01-21
Bumped react-dev-utils dependency to v10.

robcaldecottvelo:
## What I did

Bumped the `react-dev-utils` dependency in `storybook/core` and `storybook/react` to v10 (which is the version now used by `create-react-app`.

See issue #9551


shilman (approved):
LGTM!

PR #9590, opened 2020-01-21
Fix/use story state

ndelangen:


atanasster (commented):
I would suggest we leave out the story hooks/ store state for now, since they didn't work for the 'knobs-2' and the whole implementation is a bit over the top. The story_store already had built-in mechanism for HMR and sharing story states between preview and manager and seems the right place to store story state.
We will still have the addon shared state and its a great addition - it will simplify communication in addons btw manager and preview and make those things way simpler.

PR #9603, opened 2020-01-22
FIX some dependencies & ts problems on next

ndelangen:


examples/cra-react15/.storybook/main.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
   stories: ['../src/stories/welcome.stories', '../src/stories/**/button.stories.js'],
-  addons: ['@storybook/addon-actions', '@storybook/addon-links'],
+  addons: [
+    '@storybook/preset-create-react-app',
```
ndelangen: This starts using the new cra preset, we'll be removing the included one in the app/react in 6.0.0

The embedded preset has a problem with the css-mini-extract plugin

examples/cra-ts-kitchen-sink/src/stories/Button.tsx
```diff
@@ -14,7 +14,7 @@ interface ButtonProps {
   /**
    * default is false
    */
-  secondary: boolean;
+  secondary?: boolean;
```
ndelangen: In the story this prop isn't provided, so apparently it's optional

lib/components/src/tabs/tabs.tsx
```diff
@@ -114,7 +114,7 @@ export const panelProps = {
 
 const childrenToList = (children: any, selected: string) =>
   Children.toArray(children).map(
-    ({ props: { title, id, color, children: childrenOfChild } }, index) => {
+    ({ props: { title, id, color, children: childrenOfChild } }: React.ReactElement, index) => {
```
ndelangen: Seems like before in the react typings this was only React.ReactElement, but in the new typings it can also be things like portals, and strings.

We'll likely have to deal with that at some point, but for now make the typings work.

shilman (approved):
LGTM. Are these causing problems?

ndelangen:
Yes, when removing polymer, the lockfile got changed, and I was hitting these issues.

PR #9613, opened 2020-01-23
Update Custom Body Documentation

smithambera:
I'm using Storybook to document and build a design system. The design system uses relative sizing, but it is based on a `15px` base font-size rather than the `16px` font-size in the storybook preview. 

## What I did

I referred to the Storybook docs to try and solve my problem, but there was no explicit documentation for how to change the base font-size in the preview. It took me a bit to figure out, but ended up being an easy fix. I updated the Storybook documentation to explain how to update the base font-size in the preview window because I imagine a lot, lot, lot of people are using Storybook just how I am - to build a relatively-sized design system. 😄 

## How to test

Not much to test here, just an addition to the docs 🎉

### DIFF
--- a/docs/src/pages/configurations/add-custom-body/index.md
+++ b/docs/src/pages/configurations/add-custom-body/index.md
@@ -11,7 +11,7 @@ You can accomplish this by creating a file called `preview-body.html` inside the
 <div id="custom-root"></div>
 ```
 
-If using relative sizing in your project (like `rems` or `ems`), you may update the base `font-size` by adding a `style` tag to `preview-body.html`:
+If using relative sizing in your project (like `rem` or `em`), you may update the base `font-size` by adding a `style` tag to `preview-body.html`:
 
 ```html
 <style>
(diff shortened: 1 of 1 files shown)


ndelangen:
> Also of Note: I didn't pull down Storybook to suggest this update; I did all of this through GitHub because I'm only adding to the docs. All of that is to say ... I'm not sure how all those commits that aren't from me are a part of this PR. 🤷‍♀ Let me know if I need to do something differently 👍 happy to resubmit if need be!

No worries we can fix it.

I think you based off the 'master' branch (not the default branch), then made a PR against the 'next' branch (the default branch).

The preferred method is to base off of 'next' PR against 'next' and then we cherry-pick PRs to 'master'.

ndelangen (approved):

smithambera:
Thanks @ndelangen - I think that's exactly what happened 🤦‍♂My bad!

shilman (approved):
💯

ndelangen:
@smithambera don't worry about it, it's all good. Thank you so much for the contribution!

PR #9616, opened 2020-01-23
fix(addon-storyshots): Remove excess slashes from jest transform warning

callumacrae:
It's printing all of them, so I copied all of them, and that didn't work!

before:

image

## How to test

Testable by running storyshots tests with this config:

```
import initStoryshots, { multiSnapshotWithOptions } from '@storybook/addon-storyshots';

initStoryshots({
  test: multiSnapshotWithOptions()
});
```

Then if you're using CSFs, the warning will be thrown.

The example in the docs is already correct.


shilman (approved):
💯

PR #9617, opened 2020-01-23
Improve Typescript Configuration documentation page

South-Paw:
Issue: n/a

## What I did

When I used the Typescript Configuration page recently, I thought that it didn't even mention the preset but as it turned out it was just mentioned in a single line near the top which I missed.

Some others that I know and have pointed to this page have also missed this line and link - which then leads you into all the details further down the page and makes Typescript + Storybook seem like a configuration nightmare (which its really not 😄)

This PR is an attempt to make the preset options more predominate on the Typescript Configuration page, so that people are less likely to skip over them when trying to get Storybook set up.

## How to test

n/a docs only

### DIFF
--- a/docs/src/pages/configurations/typescript-config/index.md
+++ b/docs/src/pages/configurations/typescript-config/index.md
@@ -7,13 +7,15 @@ This is a central reference for using Storybook with TypeScript.
 
 ## Typescript configuration presets
 
-The easiest way to write and configure your stories in TypeScript is by using [Storybook presets](../../presets/introduction).
+The fatest and easiest way to write and configure your stories in TypeScript is by using a Storybook preset.
 
-If you're using Create React App (CRA) and have configured it to work with TS, you should use the [CRA preset](https://github.com/storybookjs/presets/tree/master/packages/preset-create-react-app), which configures Storybook to reuse CRA's TS handling.
+If you're using Create React App (CRA) and have configured it to work with TS, you should use [`@storybook/preset-create-react-app`](https://github.com/storybookjs/presets/tree/master/packages/preset-create-react-app), which configures Storybook to reuse CRA's TS handling.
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/configurations/typescript-config/index.md
```diff
@@ -7,13 +7,15 @@ This is a central reference for using Storybook with TypeScript.
 
 ## Typescript configuration presets
 
-The easiest way to write and configure your stories in TypeScript is by using [Storybook presets](../../presets/introduction).
+The fatest and easiest way to write and configure your stories in TypeScript is by using a Storybook preset.
```
ndelangen: "fastest"

Why remove the link?

ndelangen (changes requested):
I found a little typo

docs/src/pages/configurations/typescript-config/index.md
South-Paw: Oh dammit! 🤣

I just moved the link down rather than removing it completely - you'll find it on L18.

As for the reason for moving it - I think that pointing people to this general page about presets is confusing when their purpose for being on this page is to get Typescript working (no matter 'how its done'). This is at least what I found myself falling into every time I visited this page.

So instead, point people to the 2 options for getting Typescript working and then optionally let them follow through on the 'What is a preset' link after that.

There's a tiny discussion (Sat, 18th Jan) between @shilman and myself in `#typescript` on Discord that may have some context on where this change came from.

South-Paw:
@ndelangen typo fixed 🙂

ndelangen (approved):

docs/src/pages/configurations/typescript-config/index.md
ndelangen: sounds good!

PR #9647, opened 2020-01-27
Add Paddings addon to addon gallery

rbardini:
## What I did

Add [Paddings](https://github.com/rbardini/storybook-addon-paddings) addon to the _Community Addons_ section in addon gallery.

### DIFF
--- a/docs/src/pages/addons/addon-gallery/index.md
+++ b/docs/src/pages/addons/addon-gallery/index.md
@@ -90,6 +90,10 @@ You can even run these tests inside a CI box.
 With this addon, you can showcase multiple components (or varying component states) within 1 story.
 Break your stories down into smaller categories (chapters) and subcategories (sections) for more organizational goodness.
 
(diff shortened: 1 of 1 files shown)


ndelangen:
hey @rbardini this looks cool, and very much related to a change I did for 6.0.0 a few days ago here:
https://github.com/storybookjs/storybook/pull/9229

This has yet to be documented. Perhaps you'd like help us out with that?

The addon looks polished, definitely worth an add to the list!

ndelangen (approved):

rbardini:
Thanks @ndelangen!

I wasn't aware something related was being discussed already. Good to know there may be demand for an addon like this. I'll need to do some changes to make it work well with the `layout` parameter though, as the addon pads the iframe element instead of the preview document.

Meanwhile, I can give documentation a try. Any suggestions where this should be placed? I'm thinking of a _Layout parameter_ section under _Configurations_.

ndelangen:
@rbardini we can always move the page, your suggestion sounds good to me!

PR #9687, opened 2020-01-30
check if docsOnly is set to hide the addon panels

atanasster:
Issue: #9684

## What I did
added additional check for docsOnly for hiding addon panels
the use case is what user goes to a 'story' path out of a mdx page

## How to test
the flow of this code is really complicated, I have the use cases in official-storybook but comes also that depending on the setup it can get even more complicated. 
At some point, you might decide to not set 'story' by default if user navigates to the main age - and show addons/toolbars only after the first story was loaded, so we know if its a documentation only story. There will be a slight delay to displaying the addon panel if its actually a canvas story, but long term will avoid UI complications

### DIFF
--- a/lib/ui/src/components/layout/desktop.js
+++ b/lib/ui/src/components/layout/desktop.js
@@ -43,7 +43,7 @@ const Desktop = React.memo(
                   <Preview id="main" debug={previewProps} />
                 </S.Preview>
 
-                <S.Panel {...panelProps} hidden={viewMode !== 'story'}>
+                <S.Panel {...panelProps} hidden={viewMode !== 'story' || docsOnly}>
                   <Panel debug={panelProps} />
                 </S.Panel>
 


tmeasday (approved):
Seems right to me. I'm a little confused by your comment though -- is there an example in the official storybook that demonstrates this behaviour?

atanasster:
@tmeasday - the example is provided by the referenced issue. 

for example the first page of a storybook is an mdx documentation page and they don't want to show the canvas tab at all. I think its a common practice with a mdx welcome page. 

However storybook sets 'story' as the default viewMode and displays addon panels, toolbars even before an actual story is loaded to know f this is a "docsOnly" the story. So currently there is a flicker showing the addon panels and them when the story actually loads, it hides the addon panels.

ndelangen (approved):

PR #9688, opened 2020-01-30
Upgrades

ndelangen:


ndelangen:
@kroeder @gaetanmaisse would you have a suggestion how to fix:
```
Summary of all failing tests
 FAIL  examples/angular-cli/angularshots.test.js
  ● Test suite failed to run

    Unable to find tsconfig file given "src/tsconfig.spec.json". If you gave a relative path, it'll be relative to the resolved "rootDir".
    To avoid issues, use <rootDir> followed by a relative path to it in "tsConfigFile" config key.

      at findTSConfigPath (node_modules/jest-preset-angular/node_modules/ts-jest/src/utils/get-ts-config.ts:118:13)
      at Object.memoized [as default] (node_modules/lodash/lodash.js:10552:27)
      at Object.module.exports.process (node_modules/jest-preset-angular/preprocessor.js:13:10)
      at ScriptTransformer.transformSource (node_modules/@jest/transform/build/ScriptTransformer.js:444:35)
      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:522:40)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:560:25)
```
?

kroeder:
@ndelangen seems like something has changed in either `jest-preset-angular` or `ts-jest` 

Setting an absolute path fixed the issue

package.json
```diff
@@ -197,7 +197,7 @@
     "jest-vue-preprocessor": "^1.5.0",
     "jest-watch-typeahead": "^0.4.0",
     "lerna": "^3.19.0",
-    "lint-staged": "^9.4.1",
+    "lint-staged": "^10.0.6",
```
gaetanmaisse: As of `lint-stage` 10 there is no need to add `git add` in tasks definition:
> Prior to version 10, tasks had to manually include git add as the final step. This behavior has been integrated into lint-staged itself in order to prevent race conditions with multiple tasks editing the same files. If lint-staged detects git add in task configurations, it will show a warning in the console. Please remove git add from your configuration after upgrading.

So you can safely removed all of them in: https://github.com/storybookjs/storybook/blob/402c3f7cb1cf090923256b0d8cf6eea93554e428/package.json#L83

ndelangen: great!

Issue #9294, opened 2020-01-02
React native fails with JSON.serialization issue


zhigang1992:
**Describe the bug**
Default configuration with sb stopped working after `beta.25`

**To Reproduce**
1. `npx react-native init testSB --template react-native-template-typescript`
2. `npx -p @storybook/cli sb init --type react_native`
3. Launch the app
4. Launch the storybook server

**Expected behavior**
Should not throw error

**Screenshots**
![Simulator Screen Shot - iPhone 11 - 2020-01-02 at 10 32 25](https://user-images.githubusercontent.com/1057756/71649706-da827000-2d4b-11ea-9e44-a33fed7b582c.png)

**System:**
```
Environment Info:

  System:
    OS: macOS 10.15.2
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
  Binaries:
    Node: 10.16.3 - /var/folders/xx/4pg__j5s58585b86mzz27z540000gn/T/fnm-shell-1707442/bin/node
    Yarn: 1.21.1 - ~/.yarn/bin/yarn
    npm: 6.9.0 - /var/folders/xx/4pg__j5s58585b86mzz27z540000gn/T/fnm-shell-1707442/bin/npm
  Browsers:
    Chrome: 79.0.3945.88
    Safari: 13.0.4
  npmPackages:
    @storybook/addon-actions: 5.3.0-beta.25 => 5.3.0-beta.25
    @storybook/addon-links: 5.3.0-beta.25 => 5.3.0-beta.25
    @storybook/addons: 5.3.0-beta.25 => 5.3.0-beta.25
    @storybook/react-native: 5.3.0-beta.25 => 5.3.0-beta.25
    @storybook/react-native-server: 5.3.0-beta.25 => 5.3.0-beta.25
```

**Additional context**
It still works after downgrade to `beta.25` and it stops working after `beta.26` and up.

doctyper:
I just ran into this issue as well.

The error comes from the use of [json-fn](https://github.com/storybookjs/storybook/blob/next/lib/channel-websocket/src/index.ts#L2), which does not guard against circular references in JSON objects.

There is a circular reference in [`hooks.ts`](https://github.com/storybookjs/storybook/blob/next/lib/addons/src/hooks.ts#L179). `context` includes a reference to `hooks`, which has a parameter `currentContext`, which itself includes a reference to `hooks`, hence the circular ref.

I'm using the following patch to workaround the issue:

[...]

PR #9698, opened 2020-01-31
Patch for Issue 9294

brotsky:
Patch was suggested here: https://github.com/storybookjs/storybook/issues/9294#issuecomment-580582681
Credit: @arlyon

Issue:

## How to test


lib/addons/src/hooks.ts
```diff
@@ -176,7 +176,7 @@ export const applyHooks = (
     const { hooks } = context;
     hooks.prevMountedDecorators = hooks.mountedDecorators;
     hooks.mountedDecorators = new Set([getStory, ...decorators]);
-    hooks.currentContext = context;
+    hooks.currentContext = Object.assign({}, context, { hooks: null });
```
ndelangen: This would ensure the ref to hooks.currentContext is always new/different.

What about:
```suggestion
hooks.currentContext = context || { hooks: null };
```

?

PR #9706, opened 2020-02-02
Cra-ts-kitchen-sink: Fix stories glob pattern

swlee93:
Issue:
When we started the storybook of cra-ts-kitchen-sink , we had a situation loading only the mdx format files, unlike our intention to verify the story of the components written with type scripts.

## What I did
The matching pattern of the stories defined in main.js has been changed.


shilman (approved):
LGTM, thanks for fixing this!! There's something up with the lockfile which is breaking the build. I'll follow up and get it sorted out.

PR #9759, opened 2020-02-05
Examples: Fix build on master

ndelangen:
Issue: https://app.netlify.com/sites/storybookjs/deploys/5e3a8b3054f52300084af24c

The examples aren't building on master, because of some version mismatch with MiniCssExtractPlugin & cra.

## What I did

- I REMOVED the MiniCssExtractPlugin for CRA. It's only giving grief :(


ndelangen:
@shilman the changes in the lockfile shouldn't affect users.

The snapshots & typings issues are just caused by a minor version bumps of some deps.

The removal of the MiniCSSExtract plugin I did, because I'm tired of that thing breaking.

ndelangen:
This will make it so the examples on netlify start working

shilman (approved):
N/A

shilman (changes requested):
Is removing the CSS minification actually necessary to get this working? If not, can you put it in separate PR? That's a user-facing change on `master`

ndelangen:
@shilman I was not able to get it working without disabling the CSSMinificationPlugin :(

shilman (approved):
👍

mtliendo:
Since upgrading to 5.3.13, my scss files are no longer loading and thus my build is breaking. I believe it has something to do with this being removed.

image

Downgrading to 5.3.12 resolves my issue instantly. Additionally, upgrading to 6-alpha breaks the build with the same error.

Issue #9765, opened 2020-02-05
Add support for `.storybook/main.ts` file


sentilesdal:
**Is your feature request related to a problem? Please describe.**
The current documentation shows using a `main.js` file for configuring storybook.  However, using a `main.ts` file is not currently supported which makes importing from other typescript files impossible.  Since the new `main.js` file includes an option to customize the webpack configuration, I'd like to be able to import rules from the project's `webpack.config.ts`.

**Describe the solution you'd like**
Support for a `main.ts` file out of the box.

**Describe alternatives you've considered**
Currently the other configuration files (`addons.ts`, `config.ts` etc) support using typescript).  This works for now, but it is hard to find documentation for this old pattern.

**Are you able to assist bring the feature to reality?**
Possibly, haven't looked into what's necessary to get this to work.

shilman:
@sentilesdal `main.js` loading is an iteration on top of `preset.js` loading, which was originally written by @igor-dv. maybe Igor can help guide you on what might be required here?

sentilesdal:
@shilman, @igor-dv yeah if you could point me in the right direction i could find some time this weekend.

igor-dv:
hey, it's actually possible. check this out  #9775

shilman:
Yay!! I just released https://github.com/storybookjs/storybook/releases/tag/v6.0.0-alpha.9 containing PR #9775 that references this issue. Upgrade today to try it out!

You can find this prerelease on the `@next` NPM tag.

Closing this issue. Please re-open if you think there's still more to do.

sentilesdal:
Hey thanks!  This gave me enough clues to get it to work for us.

sentilesdal:
@shilman did you mean to reopen this issue?

shilman:
Yeah I think @igor-dv also wanted to make some documentation updates

vasilii-kovalev:
Is there a type for the `main.ts` config?

shilman:
Not yet, coming in 6.1 #11843

vasilii-kovalev:
Ok, thanks!

I used `require` to get the type, but `import` solved the problem and made it accessible:
`import { StorybookConfig } from '@storybook/core/types';`

mlshv:
Hi everyone! I've just updated Storybook to 6.1.1 and `main.ts` doesn't seem to work for me

I have the following error:
```sh
yarn run v1.22.4
$ start-storybook -p 6006 -c storybook
info @storybook/react v6.1.1
[...]

PR #9775, opened 2020-02-06
Examples: Change main.js to main.ts to show it's possible

igor-dv:
Issue: #9765
Hi all =)

The issue claims it's not possible to use `main.ts` inside .storybook dir while it's actually possible since it's required with `serverRequire`. 

## What I did
renamed `main.js` to `main.ts` in `cra-ts-kitchen-sink` and changed a bit tsconfig. 
`tsignore` there is because of the `"isolatedModules": true,` attribute that prevents using 
 commonjs. 

TBD: gonna add this to docs soon.


jiraiya103:
How to export ?
image
in main.ts
image

PR #9797, opened 2020-02-09
Addon-viewport: Fix Galaxy S9's viewport size

behrangsa:
Galaxy S9's viewport size is 740x360. See https://yesviz.com/devices/s9/.

Issue:

## What I did

Fixed Galaxy S9 viewport size in the viewport addon's defaults.

## How to test

### DIFF
--- a/addons/viewport/src/defaults.ts
+++ b/addons/viewport/src/defaults.ts
@@ -92,8 +92,8 @@ export const INITIAL_VIEWPORTS: ViewportMap = {
   galaxys9: {
     name: 'Galaxy S9',
     styles: {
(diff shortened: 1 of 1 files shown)


behrangsa:
I think Pixel and Pixel XL sizes are also incorrect. See https://material.io/resources/devices/.

We should also add Galaxy Note 9 or 10 to the list too.

ndelangen (approved):

ndelangen:
Is it useful to have so many devices in the default list?

behrangsa:
Not if it remains flat. If we group by manufacturer, then we can expand the defaults.

PR #9807, opened 2020-02-10
CLI: transpile `@storybook/cli` to CJS

gaetanmaisse:
## Description 
🏗  Preparatory work for Yarn 2 compatibility https://github.com/storybookjs/storybook/issues/9527.

📓  Until this PR, `@storybook/cli` ES6 modules were not transpiled with Babel but "read" with `esm` NPM package at runtime. I was facing a lot of issues when trying to use CLI + `esm` + Yarn v2 all together. These are kinds of magical errors related to Node resolve process 💥 (which is customized both by Yarn 2 and esm).

🔍  After hours of investigation, I'm decided to work around the problem by using Babel to export CJS modules that will be readable natively by Node. As far as I understand the only files that must not be transpilated are the ones inside `generators/**/template*/` folders because they are copied to SB user when he/she is running `sb init`, everything else can be transpiled and used by Node directly.

## What I did

 - move all source code in `src` as it's, by convention in SB monorepo, the source folder for Babel
 - update `.babelrc.json` to ignore all templates files
 - add `prepare` NPM script as in all other libs
 - update `cleanup` stage of `prepare` script to avoid removing anything from `@storybook/cli/dist/generators/**/template*` folder because there are the template files that will be copied to initialize SB on users' projects
 - remove `esm` as source code is now transpiled to CJS
 - replace some default exports with named ones 

## How to test

All CLI tests must be 🟢


shilman (commented):
Good lord. Thanks for all your hard work on this @gaetanmaisse 🙏 

This is closer to @ndelangen 's wheelhouse so I'm gonna let him make the call. Please feel free to loop me in if you need to discuss.

ndelangen (approved):

PR #9817, opened 2020-02-11
Core: Add story state to store

tmeasday:
For #9811. Based on https://github.com/storybookjs/storybook/pull/9810 (need to change base when this is merged).

This is just the basics in the story store (doesn't do anything on the manager side, and no hooks exist yet). Opening PR now in case there is any discussion. Some notes, questions:

 - Previously the story store has not listened to events itself (even though it has access to the channel), and all communication w/ the manager has been mediated by `preview/start.js`. I kind of feel like that complicates things unnecessarily in this case (when the manager is sending a message to change the state) and probably other existing cases too, TBH.

 - I don't fully understand why the story store both is an event emitter and also emits events on the channel. Is this a historical thing due to React Native (where the channel can not exist)? Do we still need it? It seems confusing and it might be the time to get rid of it.

 - Should I use the force re-render mechanism to re-render the story when the state changes, or should there be explicit state checks in the `renderMain` function? It kind of feels like forceRender is a reasonable fit for what we are doing here.

 - Should we merge this PR (which AFAICT doesn't break anything) and iterate or would people prefer me to keep going on this branch?


lib/client-api/src/story_store.test.ts
```diff
@@ -258,29 +319,4 @@ describe('preview.story_store', () => {
       expect(store.fromId(toId('kind-1', 'story-1.2'))).toBeTruthy();
     });
   });
-
-  describe('story sorting', () => {
```
tmeasday: There is a whole other, larger sorting test.

lib/client-api/src/story_store.ts
```diff
@@ -14,24 +14,13 @@ import {
...
-const toKey = (input: string) =>
-  input.replace(/[^a-z0-9]+([a-z0-9])/gi, (...params) => params[1].toUpperCase());
-
-let count = 0;
-
-const getId = (): number => {
-  count += 1;
-  return count;
-};
-
```
tmeasday: I forgot to delete this stuff in the other PR

atanasster:
Definitely not refactor of useStoryState - and good riddance :)
Some of things you have right, some you will find out once you travel that road. most everything in my original code was done for a reason - like the channel listeners attached to the store, story re-render. I am also testing to unhook the channel listeners on channel.set 
Haven't looked in depth but some notes
- The code on the manager will be mostly the same and i would suggest to create a shared package to manage the state (and other shared functionality)
- There are some subtle differences between manager and preview - any ‘functions’ will reside on the preview - like onClick and will not be actually serialized to the manager,

- The notation ({ state : { name } }) is really ugly in my opinion, esp. if you want to reuse it for controls
Compare to ({ name }) - a beauty :)

The best approach in my opinion is to merge the changes in small batches - and i guess *you* can do that

atanasster:
Further, in support of my comment above that `({ state : { name } })` is not the best design choice.
 
Here is a small video with the final result of my 'smart-controls' implementation - please note that just by adding props as a parameter, the controls are created and used automatically (counting just a minimal few keystrokes). This is pure perfection :)

smart-controls

atanasster:
This is the entirety of the story code with my architecture:
```
export const controls = props => <Avatar {...props} />;
```
and state would go into the second story parameter, with the story context:
```
export const controls = (props, { state: { name }}) => <Avatar {...props} />;
```

tmeasday:
>  I am also testing to unhook the channel listeners on channel.set

Yeah I am pretty sure we never change the channel on the story store so I am inclined not to add code that is never used. This is something else I want to confirm with @ndelangen

atanasster:
> Yeah, state is not going to include functions.
So is state going to replace my controls implementation or are the two going to co-exist
Tom can you please explain in details what you intend to accomplish as far as architecture goes. 

>I know, but as discussed in the original design document we will iterate our way there, probably changing to that as the default in 7.0 if it works in the wild. In the meantime we'll add an option that makes the state the first argument to the story function.

Hmm - my code is already handling both legacy and "new" way to pass the properties/controls. It works great.

In the meantime, I am finishing a version of component controls that works with storybook 5.3 - its a botched down version (lets call it the useStoryState version :)) - and that makes me even more enthusiastic with what I accomplished with my full implementation. Lets try to get to that level with 6.0

lib/core/src/client/preview/start.js
```diff
@@ -350,6 +350,7 @@ export default function start(render, { decorateStory } = {}) {
   });
 
   storyStore.on(Events.STORY_RENDER, renderUI);
+  storyStore.on(Events.STORY_STATE_CHANGED, () => renderUI(true));
```
shilman:
```suggestion
  storyStore.on(Events.STORY_STATE_CHANGED, forceReRender);
```

lib/client-api/src/story_store.ts
```diff
@@ -308,6 +303,17 @@ export default class StoryStore extends EventEmitter {
     this.pushToManager();
   }
 
+  setStoryState(id: string, newState: StoryState) {
+    if (!this._data[id]) throw new Error(`No story for id ${id}`);
+    const { state } = this._data[id];
+    this._data[id].state = { ...state, ...newState };
```
shilman: Is there any way to delete something from state? Do we care?

shilman (approved):
Looks good to me. We'd discussed some idiomatic ways to use this, and it would probably be useful to have those in a second PR before merging this.

lib/client-api/src/story_store.ts
tmeasday: Not currently. Let's wait until that use case comes up, maybe?

tmeasday:
> I am finishing a version of component controls that works with storybook 5.3

That's great, can't wait to see it!

I'll update this PR with some more details documenting explaining how state will work in 6.0, but the short story is that we want to do this thing incrementally so we can feel out the feature as we go without committing to a bunch of new APIs before the community have had a chance to use them.

atanasster:
> > I am finishing a version of component controls that works with storybook 5.3
> 
> That's great, can't wait to see it!

It's a bit of a disappointment really. I mean its better than knobs etc., but I honestly have seen perfection with my full architecture in place and can't stop thinking about it.  Did you actually run my branch? Everything was fully functioning with batteries included.
 
Anyway, regardless of anyone's reluctance I will find a way to make it work the right way.

> I'll update this PR with some more details documenting explaining how state will work in 6.0, but the short story is that we want to do this thing incrementally so we can feel out the feature as we go without committing to a bunch of new APIs before the community have had a chance to use them.

Yeah, no offense meant but thats a bit of a mistake - focus on implementation details and later fit in the end goal ( I will not mention useStoryState because that one even the implementation details were crooked)

As you remember, I wrote first the selection of end goals, and the best design was. Only then did I go into making it work (I did choose the hardest option to make, just because it was so much better architecture).
```
export const controls = props => <Avatar {...props} />;
```

This specific design was discussed with @shillman, including it was his suggestion was added the option to use legacy calling convention.

tmeasday:
> Yeah, no offense meant but thats a bit of a mistake - focus on implementation details and later fit in the end goal

🤷‍♂ we found this approached really well for SB Docs, TBH.

> ( I will not mention useStoryState because that one even the implementation details were crooked)

Can we stop doing that? Regardless of how close the current implementation of `useStoryState` is to being complete, I don't think it is in the spirit of this community to keep talking negatively about other's code (I realise you were involved in the code too, but still). 

> This specific design was discussed with @shillman, including it was his suggestion was added the option to use legacy calling convention.

Yep, I still plan to add this feature in this or a follow up PR.

atanasster:
>Can we stop doing that? Regardless of how close the current implementation of useStoryState is to being complete, I don't think it is in the spirit of this community to keep talking negatively about other's code (I realise you were involved in the code too, but still).

Actually I am all for it - I dont really care about code, or naming variables - just "what do you want to accomplish". Sooner or later all code becomes obsolete but ideas persist.

tmeasday:
> Actually I am all for it - I dont really care about code, or naming variables - just "what do you want to accomplish".

Well others do, and the code was written by real people in a community who in most cases are volunteering their time. It is important to set a tone that criticising others in this way isn't acceptable in the community.

tmeasday:
@atanasster I think you edited my comment by mistake, let me put what you said in quotes here:

> Not sure if we are disagreeing on this ? 
> Just in case - 100% agreed, and I have not criticized any code - useStoryState was just an idea and I did try my best to make it work but it simply wasn't the right idea or the right implementation. The current code is not related to useStoryState in any form or shape. 

Oh, maybe I misunderstood what you meant above when you said "Actually I am all for it", apologies.

I realise the history on `useStoryState`, I just don't want us to be publicly disparaging code or features here on GH or elsewhere because others reading it don't have all the context and can get the wrong impression about how the community operates. As prominent individuals we set the tone.

> I am also a member of the community, and I did spend a lot of time volunteering. I really hope my efforts are also appreciated - and I did spend an outsized amount of time on this feature. 

Of course they are!

> Anyway, I will leave you at this - if you need my input please let me know.

👍 thanks.

PR #9826, opened 2020-02-11
Downgrade TypeScript to ~3.6.0

### LINKED ISSUE #9463: TS1086: An accessor cannot be declared in an ambient context
**Describe the bug**
Upgraded to 5.3.3 just now and our Angular build is failing with the following:

```
[ng]
[ng] ERROR in ../../node_modules/@storybook/channels/dist/index.d.ts:25:9 - error TS1086: An accessor cannot be declared in an ambient context.
[ng]
[ng] 25     get hasTransport(): boolean;
[ng]            ~~~~~~~~~~~~
```

From a quick google, it appears other repos are facing the same issue due to the use of TypeScript 3.7.0 - but Angular is restricted to <3.5.0

**Similar errors**
https://github.com/nestjs/nest/issues/3513
https://github.com/googleapis/node-gtoken/issues/244

I believe the breaking change come from TypeScript directly https://github.com/microsoft/TypeScript/issues/33939 - but until a fix is released, the `index.d.ts` file will need regenerating with `npm i typescript@~3.6.0` (swap `^` for `~`)

**System:**
```
    OS: macOS Mojave 10.14.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  Binaries:
    Node: 12.14.1 - /usr/local/bin/node
    npm: 6.13.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.117
    Safari: 13.0.4
  npmPackages:
    @storybook/addon-actions: ^5.3.3 => 5.3.3
    @storybook/addon-knobs: ^5.3.3 => 5.3.3
    @storybook/addon-links: ^5.3.3 => 5.3.3
    @storybook/addon-notes: ^5.3.3 => 5.3.3
    @storybook/addon-storyshots: ^5.3.3 => 5.3.3
    @storybook/addons: ^5.3.3 => 5.3.3
    @storybook/angular: ^5.3.3 => 5.3.3
    @storybook/cli: ^5.3.3 => 5.3.3
```

lychyi:
**Possibly closing in favor of #9847**

---

Issue: Fixes #9463 

Storybook version 5.3.x should be on TypeScript 3.6 and not on 3.7.

The "minor" bump in TS versions caused a breaking change. A class of type declarations generated with TS3.7 would not be compatible with TS3.6 and below. This causes users of Storybook 5.3 to either upgrade TS to 3.7 or stay on an older version of Storybook.

For library maintainers, if they upgrade TS to 3.7, they would force their consumers to do so as well to avoid this issue. Given that the leap from TS 3.6 to 3.7 is technically a breaking change, it would make sense for the Storybook 5.3 version line to just hold at `~3.6.0` and wait to upgrade to 3.7 when Storybook 6 gets released (or held until future major versions). This PR does that.

## What I did
I specified the workspace `package.json` dependency to `"typescript": '~3.6.0'`. This generates declaration files using `tsc` using 3.6.5 and should play nice with Storybook consumers still using TypeScript 3.6 and below.

A few other changes were made to accommodate this and I've notated it as comments in the file diffs themselves.

### DIFF
--- a/addons/storyshots/storyshots-puppeteer/src/axeTest.ts
+++ b/addons/storyshots/storyshots-puppeteer/src/axeTest.ts
@@ -17,7 +17,8 @@ export const axeTest = (customConfig: Partial<CommonConfig> = {}) =>
     ...customConfig,
     async testBody(page, options) {
       const parameters = options.context.parameters.a11y;
-      const include = parameters?.element ?? '#root';
+      const element = (parameters === null || parameters === undefined) ? undefined : parameters.element;
+      const include = (element === null || element === undefined) ? '#root' : element;
       await expect(page).toPassAxeTests({ ...parameters, include });
     },
   });
--- a/app/angular/package.json
+++ b/app/angular/package.json
@@ -33,6 +33,8 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "dependencies": {
+    "@angular/compiler-cli": "^8.2.8",
+    "@angular-devkit/build-angular": "~0.803.6",
     "@storybook/addons": "6.0.0-alpha.10",
     "@storybook/core": "6.0.0-alpha.10",
     "@storybook/node-logger": "6.0.0-alpha.10",
--- a/app/angular/src/server/angular-cli_config.ts
+++ b/app/angular/src/server/angular-cli_config.ts
@@ -161,7 +161,7 @@ export function applyAngularCliWebpackConfig(baseConfig: any, cliWebpackConfigOp
   // todo add type for acc
   const entry = [
     ...baseConfig.entry,
-    ...Object.values(cliStyleConfig.entry).reduce((acc: any, item) => acc.concat(item), []),
+    ...Object.values<any>(cliStyleConfig.entry).reduce((acc: any, item) => acc.concat(item), []),
   ];
 
   const module = {
--- a/lib/ui/src/components/sidebar/treeview/treeview.tsx
+++ b/lib/ui/src/components/sidebar/treeview/treeview.tsx
@@ -49,6 +49,7 @@ const linked = (
   }: { onClick: Function; onKeyUp: Function; Link: ComponentType; prefix: string }
 ) => {
   const Linked = React.memo(p => (
+    // @ts-ignore
     <L
       prefix={prefix}
       {...p}
--- a/package.json
+++ b/package.json
@@ -215,7 +215,7 @@
     "trash": "^6.1.1",
     "ts-dedent": "^1.1.1",
     "ts-jest": "^24.0.2",
-    "typescript": "^3.4.0",
+    "typescript": "~3.6.0",
     "wait-on": "^4.0.0"
   },
   "optionalDependencies": {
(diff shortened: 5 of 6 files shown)


addons/storyshots/storyshots-puppeteer/src/axeTest.ts
```diff
@@ -17,7 +17,8 @@ export const axeTest = (customConfig: Partial<CommonConfig> = {}) =>
     ...customConfig,
     async testBody(page, options) {
       const parameters = options.context.parameters.a11y;
-      const include = parameters?.element ?? '#root';
+      const element = (parameters === null || parameters === undefined) ? undefined : parameters.element;
```
lychyi: This was the only place in the codebase that I could find that took advantage of the optional chaining and nullish coalescing operator introduced in 3.7.

I replaced it with the equivalent code here.

app/angular/package.json
```diff
@@ -33,6 +33,8 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "dependencies": {
+    "@angular/compiler-cli": "^8.2.8",
```
lychyi: These two dependencies are required for this package but hoisted from a different location (`examples/angular-cli` [here](https://github.com/storybookjs/storybook/blob/next/examples/angular-cli/package.json#L23) and [here](https://github.com/storybookjs/storybook/blob/next/examples/angular-cli/package.json#L36)).

These two additions to `dependencies` guarantees that they're here in this package. If these are missing, the build and tests will fail since `yarn install` nests them in `example/angular-cli` after changing TypeScript to `~3.6.0` for some reason.

🤷‍♂

app/angular/src/server/angular-cli_config.ts
```diff
@@ -161,7 +161,7 @@ export function applyAngularCliWebpackConfig(baseConfig: any, cliWebpackConfigOp
   // todo add type for acc
   const entry = [
     ...baseConfig.entry,
-    ...Object.values(cliStyleConfig.entry).reduce((acc: any, item) => acc.concat(item), []),
+    ...Object.values<any>(cliStyleConfig.entry).reduce((acc: any, item) => acc.concat(item), []),
```
lychyi: After installing TypeScript ~3.6.0, this line errors out: `TS2461: Type 'unknown' is not an array type.`

This is because `unknown` is the type for `cliStyleConfig.entry`. This is defined as `unknown` if you don't specify so I just went ahead and updated the generic to match

lib/ui/src/components/sidebar/treeview/treeview.tsx
```diff
@@ -49,6 +49,7 @@ const linked = (
   }: { onClick: Function; onKeyUp: Function; Link: ComponentType; prefix: string }
 ) => {
   const Linked = React.memo(p => (
+    // @ts-ignore
```
lychyi: For some reason, compiling TS doesn't quite ignore the `onKeyUp`, it references line 52 instead so I had to add a `// @ts-ignore` here. I'm not sure if this is different in TS3.7 vs TS3.6.

shilman (commented):
Thanks @lychyi, really great work!!! Per discussion in the issue, I'm not sure whether this should target `next` or `master`. Hope one of the typescript guys can comment cc @kroeder @gaetanmaisse

gaetanmaisse:
Superseeded by https://github.com/storybookjs/storybook/pull/9847

PR #9833, opened 2020-02-12
Fix module.export syntax error in 'Storybook for HTML' README

th-ad:
Issue:

## What I did
There is a small syntax error in the `Storybook for HTML` README. 

The example js should be `module.exports = { ...` rather than `module.exports { ...`


shilman (approved):
LGTM thanks!

PR #9842, opened 2020-02-13
FIX deepscan issues

ndelangen:
FIX some deepscan detected issues, which seem like minor issues easily resolved.


ndelangen:
Self merge, all seems good

PR #9843, opened 2020-02-13
FIX more deepscan issues

ndelangen:
more deepscan issues resolved


ndelangen:
Self merge, all seems good

PR #9844, opened 2020-02-13
Theming: Update link to addon-docs theming

tskarhed:
## What I did
Updated the link about addons-docs theming to the actual information about theming addon-docs.


shilman (approved):
🙈 thanks!!!

PR #9860, opened 2020-02-14
Form/textarea maxHeight : 400

Raphe-dev:
Issue: [9849](https://github.com/storybookjs/storybook/issues/9849)

## What I did
Add a maxHeight of 20 lines (400px) to the textarea form component to avoid infinite textarea with big data.


ndelangen:
I think it's a good default, should probably be configurable though via a prop.

Do you think you could add that? 👍

Raphe-dev:
Yeah I believe it should be configurable too ( eg: on a 4k screen 20 lines can be quite small ), I should have some time later this week to look into that.

ndelangen (approved):

PR #9861, opened 2020-02-14
Ts/better api consumer

ndelangen:
Issue: Using Consumer component with a state filter isn't as optimized as it could be

## What I did
I improved the typings of lib/api's Consumer component.

It now automatically does typing for the children function
It also automatically only re-renders the component when some has changed in the return value of the state filter function.


gaetanmaisse (approved):
LGTM TSly speaking, I'm not good enough at React to give my opinion on that 😉

PR #9866, opened 2020-02-15
CLI: Add Yarn 2 compatibility

gaetanmaisse:
Issue:
Part of #9527 

## What I did
 - Make SB CLI compatible with Yarn 2 by handling command used to fetch package versions on NPM registry. => Will need a refactoring, much easier to do if I migrate the package to TS first.
 - Add tests to check CLI and Yarn 2 compatibility: These tests mimic the ones done to test SB CLI with "classic" yarn. However, some magic setup and `package.json` customization are needed in order to make Yarn 2 works inside a Yarn v1 workspace.

⚠️ CLI is working with Yarn 2 but it does not mean that running `yarn sb init` will not fails because all frameworks or the SB integration are currently not compatible with Yarn 2. Especially projects using:
 - Angular as Angular is not working with Yarn 2
 - Marko => ` Cause: Error: Unable to load tag ([.yarn/berry/cache/marko-npm-4.18.46-d693c4c719-2.zip/node_modules/marko/src/core-tags/cache/marko.json → <cached-fragment>])`
 - React Static => `Error: Couldn't find preset "./babel-preset.js" relative to directory ".yarn/berry/cache/react-static-npm-5.9.12-e804e3b66e-2.zip/node_modules/react-static"`
 - Vue => https://github.com/nickmessing/babel-plugin-jsx-event-modifiers is not compatible with Yarn 2, it is deprecated too, so we will need to do something about it

## How to test

I add a new GitHub Actions workflow running tests related to SB CLI with Yarn 2. 
It should be 🟢.

### DIFF
--- a/lib/cli/src/generators/METEOR/index.js
+++ b/lib/cli/src/generators/METEOR/index.js
@@ -14,6 +14,7 @@ export default async (npmOptions, { storyFormat = 'csf' }) => {
     storybookVersion,
     actionsVersion,
     linksVersion,
+    knobsVersion,
     addonsVersion,
     reactVersion,
     reactDomVersion,
@@ -24,6 +25,7 @@ export default async (npmOptions, { storyFormat = 'csf' }) => {
     '@storybook/react',
     '@storybook/addon-actions',
     '@storybook/addon-links',
+    '@storybook/addon-knobs',
     '@storybook/addons',
     'react',
     'react-dom',
@@ -43,6 +45,7 @@ export default async (npmOptions, { storyFormat = 'csf' }) => {
     `@storybook/react@${storybookVersion}`,
     `@storybook/addon-actions@${actionsVersion}`,
     `@storybook/addon-links@${linksVersion}`,
+    `@storybook/addon-knobs@${knobsVersion}`,
     `@storybook/addons@${addonsVersion}`,
   ];
 
--- a/lib/cli/src/generators/MITHRIL/index.js
+++ b/lib/cli/src/generators/MITHRIL/index.js
@@ -8,11 +8,18 @@ import {
 } from '../../helpers';
 
 export default async (npmOptions, { storyFormat = 'csf' }) => {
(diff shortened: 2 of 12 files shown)


lib/cli/src/generators/MITHRIL/index.js
```diff
@@ -35,6 +42,7 @@ export default async (npmOptions, { storyFormat = 'csf' }) => {
     `@storybook/mithril@${storybookVersion}`,
     `@storybook/addon-actions@${actionsVersion}`,
     `@storybook/addon-links@${linksVersion}`,
+    `@storybook/addon-knobs@${knobsVersion}`,
```
gaetanmaisse: I added `knobs` because it is listed in `addons` of the configuration: https://github.com/storybookjs/storybook/blob/3dd714d3532687d958c41e093187b84ffca603ee/lib/cli/src/generators/MITHRIL/template-csf/.storybook/main.js but it is not used in template stories so maybe it should be deleted from configuration instead? 🤷‍♂ 

At the opposite `@storybook/addon-links` is added and used in template stories but is not declared in `addons` config, so we can maybe remove it?

lib/cli/src/generators/METEOR/index.js
```diff
@@ -43,6 +45,7 @@ export default async (npmOptions, { storyFormat = 'csf' }) => {
     `@storybook/react@${storybookVersion}`,
     `@storybook/addon-actions@${actionsVersion}`,
     `@storybook/addon-links@${linksVersion}`,
+    `@storybook/addon-knobs@${knobsVersion}`,
```
gaetanmaisse: Same as https://github.com/storybookjs/storybook/pull/9866/files#r379884879 for Mithril

gaetanmaisse:
@Hypnosphi there are 2 broken TeamCity checks, should I mind or rely on all other checks only?

ndelangen:
@gaetanmaisse you can ignore the TeamCity checks for now.

PR #9867, opened 2020-02-15
Core: Use telejson for websockets channel

Gongreg:
Issue: https://github.com/storybookjs/react-native/issues/13

## What I did

As @Hypnosphi recommended, using telejson for serialization.

## How to test

No tests for react-native-server available yet. Tested it manually.

@ndelangen @shilman, I think react-native is the only user of this package. Maybe it should be moved to rn monorepo?


ndelangen:
Moving it sounds good to me!

ndelangen (approved):

shilman (approved):
LGTM - agree moving it would be 💯

PR #9881, opened 2020-02-17
chore(deps): migrate to leven

### LINKED ISSUE #9878: Migrate to leven which is more actively maintained
As of now [didyoumean.js](https://github.com/dcporter/didyoumean.js/) matching engine is used to suggest matching commands if the user mistypes. But it is not longer actively maintained.

 Instead I would suggest switching over to [leven](https://github.com/sindresorhus/leven) which is the fastest JS implementation of the Levenshtein distance algorithm, `leven` returns the distance between 2 strings which can be made use of to find the exact match.

For sure :+1:

**Additional context**
N/A

jamesgeorge007:
closes #9878

### DIFF
--- a/lib/cli/package.json
+++ b/lib/cli/package.json
@@ -42,7 +42,6 @@
     "commander": "^4.0.1",
     "core-js": "^3.0.1",
     "cross-spawn": "^7.0.0",
-    "didyoumean": "^1.2.1",
     "envinfo": "^7.5.0",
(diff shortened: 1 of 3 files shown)


ndelangen (approved):

ndelangen:
@jamesgeorge007 do you think you could have a look at https://github.com/storybookjs/storybook/issues/9887?

PR #9913, opened 2020-02-20
Update example main.js for Ember

bzf:
Adds a missing `=` in the example that made it fail to run when copy/pasting it into a project.


shilman (approved):
Thanks!

PR #9916, opened 2020-02-20
Add missing dependencies or peerDependencies

gaetanmaisse:
## What I did

As Yarn 2 is must stricter than Yarn 1 or NPM about missing `dependencies` or `peerDependencies` I used `npx @yarnpkg/doctor` (details [here](https://yarnpkg.com/advanced/migration#run-the-doctor)) to fix some of them. 

There are still a lot more, I will fix them step by step. 

--

Specific case of `regenerator-runtime`: as discussed with @ndelangen `regenerator-runtime` is used as polyfill to backport async/await and generator to old browser (like core-js). Any package that uses async/await or generator will require `regenerator-runtime` at runtime without any import in the source (it's added by babel during transpilation). To avoid any issue in the future `regenerator-runtime` is added to almost all packages, as it has been done for `core-js`

## How to test

 - All check should be green

### DIFF
--- a/addons/a11y/package.json
+++ b/addons/a11y/package.json
@@ -54,7 +54,8 @@
   },
   "devDependencies": {
     "@types/react-redux": "^7.0.6",
-    "@types/webpack-env": "^1.15.1"
+    "@types/webpack-env": "^1.15.1",
+    "enzyme": "^3.11.0"
   },
   "peerDependencies": {
     "react-dom": "*"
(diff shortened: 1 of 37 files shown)


ndelangen (approved):

JalilArfaoui:
Then, why `regenerator-runtime hasn’t been set as a regular `dependency` instead of a `peerDependency` ?

ndelangen:
@JalilArfaoui Where do you see that dependency still specified as a peerDependency?

JalilArfaoui:
I’m sorry @ndelangen ! I had an out-of-date @storybook/api in my modules, and I felt upon this PR and thought it was still like this !

You can close this, sorry for the inconvenience

PR #9945, opened 2020-02-24
Improve/addons hooks

ndelangen:
Migrate some addons to hooks


addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -90,187 +89,161 @@ type A11YPanelState =
...
-            });
-          }
-        }, 900);
+    setTimeout(() => {
+      const { status } = state;
+      if (status === 'ran') {
+        setState({
+          ...(state as RanState),
+          status: 'ready',
+        });
```
jonspalmer: why does this change from being called as the 2nd parameter of setState?

jonspalmer:
I like it. I've overlapped with you here #9951 on the removing of some of the deprecated cruft.

ndelangen:
@jonspalmer I'll close this PR in favour of yours, and you cam perhaps cherry-pick some of my changes?

jonspalmer:
I didn't do any of the hooks upgrade. Maybe keep it open for that?

PR #9950, opened 2020-02-25
Cleanup docs wrt registering addons

jonspalmer:
## What I did
Cleaned up the docs for addons wrt no longer needing `/register` in main.js

### DIFF
--- a/addons/a11y/README.md
+++ b/addons/a11y/README.md
@@ -18,20 +18,15 @@ Add this line to your `main.js` file (create this file inside your storybook con
 
 ```js
 module.exports = {
-  addons: ['@storybook/addon-a11y/register'],
+  addons: ['@storybook/addon-a11y'],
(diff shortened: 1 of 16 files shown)


addons/a11y/README.md
```diff
@@ -18,20 +18,15 @@ Add this line to your `main.js` file (create this file inside your storybook con
 
 ```js
 module.exports = {
-  addons: ['@storybook/addon-a11y/register'],
+  addons: ['@storybook/addon-a11y'],
```
jonspalmer: Docs should lead with the preset IMO. 

If you want the old behavior of 'off by default'  instead of using register why not set disable: true in main.js and then enable it selectively?

shilman (approved):
Looks great @jonspalmer !! Thanks so much for cleaning this up 🙏

PR #9965, opened 2020-02-26
Misc: Add missing dependencies

gaetanmaisse:
Follows https://github.com/storybookjs/storybook/pull/9916

## What I did

As Yarn 2 is much stricter than Yarn 1 or NPM about missing `dependencies` or `peerDependencies` I used `npx @yarnpkg/doctor` (details [here](https://yarnpkg.com/advanced/migration#run-the-doctor)) to fix some of them. Added them with version range already used in the workspace to avoid downloading new deps (that's why yarn.lock is not modified)

### DIFF
--- a/addons/docs/package.json
+++ b/addons/docs/package.json
@@ -40,6 +40,7 @@
   "dependencies": {
     "@babel/generator": "^7.8.4",
     "@babel/parser": "^7.8.4",
+    "@babel/preset-env": "^7.8.4",
     "@babel/plugin-transform-react-jsx": "^7.3.0",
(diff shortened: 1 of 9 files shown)


gaetanmaisse:
As `@reach/router` is compatible with React 15+ only, SB should maybe inherit from this constraint? And so 
we should update the version range in `lib/router`:  https://github.com/storybookjs/storybook/blob/dac6752d479e3be806cd57dcbbc40a442f759b35/lib/router/package.json#L42

ndelangen (approved):

PR #9969, opened 2020-02-27
Experiment to make CircleCI tests a faster and more stable

jonspalmer:
## What I did

Experiment to see if we can make CircleCI a little faster by not schlepping a Chromium binary around all the build steps 

Puppeteer comes with an installed copy of chromium which we then tar and pass around in our node_module caches in GitHub actions and CircleCI workspaces. Removing the hard dependency saves about 90MB of tared node modules we don't have to cache.

Only a single set of tests (which are failing) need Puppeteer and just install in when we need it. It's also likely that we can simply use Chrome installed on the Docker images we using.

 [ ] Add documentation on running puppeteer tests locally (needs to install Chomre/Chromium)
 [ ] Explore using Chrome on CI vendor's docker images


shilman (approved):
❤️❤️❤️

ndelangen (approved):

jamesgeorge007 (approved):

PR #9971, opened 2020-02-27
CI: Fix GitHub unit test workflow

jonspalmer:
## What I did

Fixed the GitHub unit test workflow and added tests for Node 12

### DIFF
--- a/.github/workflows/tests-cli.yml
+++ b/.github/workflows/tests-cli.yml
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: '10.x'
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: Cache node modules
       uses: actions/cache@v1
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests-unit.yml
```diff
@@ -5,22 +5,26 @@ on: [push]
 jobs:
   build:
 
-    name: Test on node ${{ matrix.node_version }} and ${{ matrix.os }}
```
jonspalmer: The previous version failed because we weren't using the matrix but trying to reference it in the name.

shilman (approved):
💯💯💯

.github/workflows/tests-cli.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: '10.x'
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
jonspalmer: v2 is faster. Looks like its 2s vs 25s to checkout the repo. Small things add up :)

ndelangen (approved):

PR #10003, opened 2020-03-01
CLI: Support js / jsx / ts / tsx stories in React CSF template

thakursachin467:
Updated Regex to accommodate both Js and JSX or ts and TSX.

Issue:

What I did
Update default template generation in storybook cli to add jsx/ts/tsx file.

Issue:
When Using cli tool to initialise storybook inside a project it only take .js files and not jsx/ts/tsx.


shilman (approved):
LGTM! 💯 thank you!

PR #10013, opened 2020-03-02
Core: Fix StoryInput parameters typings

wKich:
Issue:

Story parameters properties doesn't match with declared types for `StoryInput`. In type declaration used parameter name `filename` but in implementation [here](https://github.com/storybookjs/storybook/blob/next/lib/core/src/client/preview/makeConfigure.ts#L105) and [here](https://github.com/storybookjs/storybook/blob/next/lib/client-api/src/client_api.ts#L183) parameter named as `fileName`.

## What I did

Update types to be more consistent.


tmeasday (approved):
LGTM

PR #10016, opened 2020-03-02
Upgrade/reach router 9705

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/9705

## What I did
- UPGRADE reach-router
- FIX issues concerning dependencies version
- FIX resulting ts errors

### DIFF
--- a/addons/docs/src/frameworks/react/lib/inspection/acornParser.ts
+++ b/addons/docs/src/frameworks/react/lib/inspection/acornParser.ts
@@ -24,6 +24,7 @@ interface ParsingResult<T> {
 }
 
 const ACORN_WALK_VISITORS = {
+  // @ts-ignore
   ...acornWalk.base,
   JSXElement: () => {},
 };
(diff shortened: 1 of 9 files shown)


app/react/package.json
```diff
@@ -57,7 +57,7 @@
     "webpack": "^4.33.0"
   },
   "devDependencies": {
-    "@storybook/client-api": "6.0.0-alpha.15",
```
ndelangen: This seems like a bad issue currently on next

addons/docs/src/frameworks/react/lib/inspection/acornParser.ts
```diff
@@ -24,6 +24,7 @@ interface ParsingResult<T> {
 }
 
 const ACORN_WALK_VISITORS = {
```
ndelangen: @shilman I assume the typings improved upstream from any to something specific, which might still be incomplete, OR we're doing something wrong perhaps?

I added ignores, because i assume it works as expected (test will tell?)

PR #10020, opened 2020-03-02
Add build-storybook script during setup as well

ghengeveld:
This updates the manual getting started guides for each framework to also include the `"build-storybook": "build-storybook"` script in the snippet to copy-paste into `package.json`.


ghengeveld:
Maybe it should be the same script as suggested in the [Exporting Storybook](https://storybook.js.org/docs/basics/exporting-storybook/) guide?

```
"build-storybook": "build-storybook -c .storybook -o .out"
```

ndelangen (approved):

PR #10040, opened 2020-03-03
Tech/improvements extracted from composition

ndelangen:
Issue: https://github.com/storybookjs/storybook/pull/9210

## What I did
Extracted some improvements I made into it's own PR, to make review easier:

- ADD white background for dev-kits example
- REMOVE unnecessary propTypes on TS files
- CHANGE svg for dummy logo
- REFACTOR sidebar menu construction & memoization
- CHANGE getDescription to properly typed & have a fallback for if there's no kind
- CHANGE exactMatch to case insensitive && ADD safety to parameters being undefined check
- CHANGE FUZZY_SEARCH_THRESHOLD to 0.35
- CHANGE to use useMemo over memoizerific
- REMOVE log statements that serve no value to the end-user
- IMPROVE typings of SVG icon in lib/components
- ADD ability for listeners to get the original message event via `this`
- CHANGE postmessage to use native Message type over a custom one
- CHANGE addon module to use fullAPI.getData
- FIX viewMode in docsMode in lib/api state
- FIX mutation of storieshash
- ADD fullAPI to module in lib/api
- IMPROVE typings of addons.setConfig
- CHANGE api.getdata of manual destructure of state

@tmeasday


lib/api/src/index.tsx
```diff
@@ -124,7 +125,7 @@ type StatePartial = Partial<State>;
 export type ManagerProviderProps = Children & RouterData & ProviderData & DocsModeData;
 
 class ManagerProvider extends Component<ManagerProviderProps, State> {
-  api: API;
+  api: API = {} as API;
```
ndelangen: @tmeasday This is the change in API you needed before to gain access to fullAPI in the modules subAPI.

lib/api/src/index.tsx
```diff
@@ -178,15 +179,17 @@ class ManagerProvider extends Component<ManagerProviderProps, State> {
       initStories,
       initURL,
       initVersions,
-    ].map(initModule => initModule({ ...routeData, ...apiData, state: this.state }));
+    ].map(initModule =>
+      initModule({ ...routeData, ...apiData, state: this.state, fullAPI: this.api })
+    );
```
ndelangen: @tmeasday This is the change in API you needed before to gain access to fullAPI in the modules subAPI.

lib/api/src/modules/addons.ts
```diff
@@ -85,20 +85,20 @@ export function ensurePanel(panels: Panels, selectedPanel?: string, currentPanel
...
+export default ({ provider, store, fullAPI }: Module) => {
   const api: SubAPI = {
     getElements: type => provider.getElements(type),
     getPanels: () => api.getElements(types.PANEL),
     getStoryPanels: () => {
       const allPanels = api.getPanels();
-      const { storyId, storiesHash } = store.getState();
-      const storyInput = storyId && (storiesHash[storyId] as StoryInput);
+      const { storyId } = store.getState();
+      const story = fullAPI.getData(storyId);
```
ndelangen: @tmeasday here's an example usage

tmeasday:
@ndelangen can you explain a little at the top of this PR what changes are included here so I know what I am looking at?

lib/addons/package.json
```diff
@@ -33,6 +33,7 @@
     "@storybook/client-logger": "6.0.0-alpha.20",
     "@storybook/core-events": "6.0.0-alpha.20",
     "@storybook/router": "6.0.0-alpha.20",
+    "@storybook/theming": "6.0.0-alpha.20",
```
tmeasday: Ohh, I don't think we can have this dependency can we? `@storybook/theming` depends on `emotion`, no? 

Or is it OK because we don't require the files that do that in addons?

lib/api/src/modules/addons.ts
tmeasday: Thanks @ndelangen, makes sense to me.

lib/ui/src/components/sidebar/treeview/utils.ts
```diff
@@ -192,12 +191,20 @@ const fuse = memoize(5)(
     })
 );
 
-const exactMatch = memoize(1)(filter => (i: Item) =>
-  (isStory(i) && i.kind.includes(filter)) ||
-  (i.name && i.name.includes(filter)) ||
-  (i.parameters && i.parameters.fileName && i.parameters.fileName.includes(filter)) ||
-  (i.parameters && typeof i.parameters.notes === 'string' && i.parameters.notes.includes(filter))
-);
+const exactMatch = memoize(1)((filter: string) => (i: Item) => {
```
tmeasday: Why do we only `memoize(1)` in this function I wonder?

tmeasday (approved):
LGTM couple of small comments

lib/addons/package.json
ndelangen: It's only used for types, it should not appear in the dist files.

PR #10045, opened 2020-03-03
docs: update addons info

yannbf:
Issue: #9973 

## What I did
Main intention of this PR is to fix outdated documentation related to addons. What it does:

- Add query params and docs addons to the addons table
- Create an extra section for deprecated addons and move notes/info there, with a link to the migration guide
- Fix broken links for notes and info (pointing to master, previously was broken pointing to next)
- Organize addons alphabetically
- Add query params addon to the support table *
 
*Would be great if a maintainer could check if this information is correct. I could not find information about it, and assumed that given it's just a replaceState on history, it should be supported on all web frameworks.

## How to test
Check how it looks like in my branch for [addons](https://github.com/yannbf/storybook/tree/docs/update-addons-info#addons) and [framework support](https://github.com/yannbf/storybook/blob/docs/update-addons-info/ADDONS_SUPPORT.md)

There is definitely more work that can be done regarding documentation and I am willing to go through with it (such as improving the documentation on note/info migration to docs), I just need more input from @shilman.


shilman (changes requested):
@yannbf thanks for these improvements! any chance you can create a separate table for deprecated addons? it might be more clear.

yannbf:
@shilman thanks for reviewing! I've added the separate table as requested.

shilman (approved):
LGTM! Great job @yannbf !!! 💯

PR #10052, opened 2020-03-04
Tech/improvements

ndelangen:
- ADD safety check for docs-mode without a story (can happen on initial render)
- FIX ember docs when description is missing for whatever reason

.

- CHANGE all css template tags to css objects
- ADD cors headers to netlify

.

- REMOVE debug props
- IMPROVE the navigation e2e cypress test

### DIFF
--- a/addons/docs/src/frameworks/ember/jsondoc.js
+++ b/addons/docs/src/frameworks/ember/jsondoc.js
@@ -26,5 +26,10 @@ export const extractProps = componentName => {
 export const extractComponentDescription = componentName => {
   const json = getJSONDoc();
   const componentDoc = json.included.find(doc => doc.attributes.name === componentName);
+
+  if (!componentDoc) {
+    return '';
+  }
+
   return componentDoc.attributes.description;
 };
--- a/cypress/integration/navigation.spec.ts
+++ b/cypress/integration/navigation.spec.ts
@@ -1,13 +1,15 @@
+/* eslint-disable jest/expect-expect */
 import { visitExample } from '../helper';
 
 describe('Navigation', () => {
-  beforeEach(() => {
+  before(() => {
     visitExample('official-storybook');
   });
 
   it('should search navigation item', () => {
     cy.get('#storybook-explorer-searchfield')
       .click()
+      .clear()
       .type('persisting the action logger');
 
     cy.get('.sidebar-container a')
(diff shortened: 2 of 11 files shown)


addons/docs/src/frameworks/ember/jsondoc.js
```diff
@@ -26,5 +26,10 @@ export const extractProps = componentName => {
 export const extractComponentDescription = componentName => {
   const json = getJSONDoc();
   const componentDoc = json.included.find(doc => doc.attributes.name === componentName);
+
+  if (!componentDoc) {
```
ndelangen: I've discovered that sometimes the parameter.component isn't there, and then this code throws because componentDoc is `undefined`

cypress/integration/navigation.spec.ts
```diff
@@ -1,13 +1,15 @@
+/* eslint-disable jest/expect-expect */
 import { visitExample } from '../helper';
 
 describe('Navigation', () => {
-  beforeEach(() => {
+  before(() => {
```
ndelangen: This should make the test run a little faster

lib/components/src/placeholder/placeholder.tsx
```diff
@@ -1,18 +1,18 @@
 import React, { Children, FunctionComponent } from 'react';
 import { styled } from '@storybook/theming';
 
-const Title = styled.div`
-  font-weight: ${props => props.theme.typography.weight.bold};
-`;
+const Title = styled.div(({ theme }) => ({
```
ndelangen: This makes the bundle smaller, and compilation faster

lib/core/src/client/preview/StoryRenderer.tsx
```diff
@@ -276,8 +276,10 @@ export class StoryRenderer {
   }
 
   renderDocs({ context, storyStore }: { context: RenderContext; storyStore: StoryStore }) {
-    const { kind, parameters } = context;
-
+    const { kind, parameters, id } = context;
```
ndelangen: When storybook initially renders, the url contains id="*".

This is a non-valid storyId, but it's because the manager doesn't know a valid ID yet.

This code makes it so docs-mode doesn't throw when that happens.

lib/ui/src/components/layout/desktop.tsx
```diff
@@ -54,15 +54,15 @@ const Desktop = React.memo<DesktopProps>(
           {({ navProps, mainProps, panelProps, previewProps }) => (
             <Fragment>
               <S.Sidebar {...navProps}>
-                <Sidebar debug={navProps} />
+                <Sidebar />
```
ndelangen: Useful for debugging, but bad for performance.

PR #10056, opened 2020-03-05
docs(README): add ts to module.exports

bovandersteene:
Issue:
In angular we work with typescript files, but this part was missing in the `main.js`file

## What I did
Fixing the README file

## How to test

No it is just documentation


shilman (approved):
LGTM, thanks! 💯

PR #10063, opened 2020-03-05
docs: Update preset-typescript documentation

albshin:
Issue:
With the changes in https://github.com/storybookjs/presets/pull/68, `tsDocgenLoaderOptions` no longer exists on `@storybook/preset-typescript`. 

## What I did
`tsDocgenLoaderOptions` has been replaced with `tsLoaderOptions` instead as shown [here](https://github.com/storybookjs/presets/tree/master/packages/preset-typescript#advanced-usage). I have updated the docs to reflect these changes.

PR #10065, opened 2020-03-05
docs: change "tsConfigPath" to "configFile"

albshin:
## What I did
While updating the docs in #10063, I forgot to change "tsConfigPath" to "configFile" as well [as shown here](https://github.com/storybookjs/presets/tree/master/packages/preset-typescript#advanced-usage). The docs should now be correct.


shilman (approved):
thanks! 💯

PR #10071, opened 2020-03-06
Fix: Remove usage of a vulnerable version of serialize-javascript

mattmoreira:
Issue: 

## What I did
I fixed the issue #9173, by just forcing the resolution of the dependency `serialize-javascript` to version 2, in which they fixed the XSS vulnerability.

According to their [release notes](https://github.com/yahoo/serialize-javascript/releases/tag/v2.0.0), their major change is basically allowing `undefined` values to be serialized, which shouldn't affect Storybook.

## How to test

As Storybook transitively depends on this dependency through `terser-webpack-plugin`, I'd recommend trying to run storybook and build the project. If everything goes well, then there shouldn't be any problem.

- `@storybook/react > @storybook/core > corejs-upgrade-webpack-plugin > webpack > terser-webpack-plugin > serialize-javascript`
- `@storybook/react > @storybook/core > webpack > terser-webpack-plugin > serialize-javascript`
- `@storybook/react > webpack > terser-webpack-plugin > serialize-javascript`
- `@storybook/react > @storybook/core > terser-webpack-plugin > serialize-javascript`

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,9 @@
     ]
   },
   "browserslist": "defaults",
+  "resolutions": {
+    "serialize-javascript": "2.1.2"
+  },
   "devDependencies": {
     "@angular/common": "^8.2.8",
     "@angular/compiler": "^8.2.8",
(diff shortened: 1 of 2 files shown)


shilman (approved):
LGTM, thanks! Will merge to `next` and release on 6.0-alpha, and then patch after a week or two if everything looks stable

PR #10079, opened 2020-03-09
docs(migration): fix syntax for main.js step

yannbf:
Issue: syntax is incorrect in migration step. Fix is basically adding the closing brackets/parenthesis to the configuration.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -130,11 +130,13 @@ module.exports = {
       rules: [
         ...config.module.rules,
         {
+          test: /\.tsx?$/,
           loader: require.resolve('react-docgen-typescript-loader'),
           options: {}, // your options here
         },
+      ]
     }
-  }
+  })
 }
 ```
 


MIGRATION.md
```diff
@@ -130,11 +130,13 @@ module.exports = {
       rules: [
         ...config.module.rules,
         {
+          test: /\.stories\.tsx?$/,
```
yannbf: If the test is not provided, the build process will take way longer than it's supposed to

MIGRATION.md
```diff
@@ -130,11 +130,13 @@ module.exports = {
       rules: [
         ...config.module.rules,
         {
+          test: /\.tsx?$/,
```
yannbf: If the test is not provided, the build process will take way longer than it's supposed to

shilman (approved):
Good catch, thanks! 🙏

PR #10083, opened 2020-03-09
Tech/improvements 2

ndelangen:
CHANGE so all interfaces & types are exported in lib/ui
REMOVE really weird href = object type
ADD ability for Tooltip onClick event access to clicked item + event


lib/components/src/tooltip/TooltipLinkList.tsx
```diff
@@ -15,31 +15,49 @@ const List = styled.div<{}>(
...
   links: Link[];
   LinkWrapper?: LinkWrapperType;
 }
 
+const Item: FunctionComponent<TooltipLinkListProps['links'][number]> = props => {
+  const { LinkWrapper, onClick: onClickFromProps, ...rest } = props;
+  const { id, title, href, active } = rest;
+  const onClick = useCallback(
+    (event: SyntheticEvent) => {
+      onClickFromProps(event, rest);
```
ndelangen: It now gets called with 2 arguments, the first is the event, as it was before, the second (new) argument is the object, with the id, href, etc on it.

PR #10085, opened 2020-03-09
MOVE event handlers into module init

ndelangen:


lib/api/src/index.tsx
```diff
@@ -195,31 +184,6 @@ class ManagerProvider extends Component<ManagerProviderProps, State> {
 
     initProviderApi({ provider, store, api });
 
-    api.on(STORY_CHANGED, (id: string) => {
```
ndelangen: MOVED into stories module init call, I think this is a lot cleaner

lib/api/src/tests/stories.test.js
```diff
@@ -531,49 +532,31 @@ describe('stories API', () => {
       expect(initialStoriesHash['a--1'].args).toEqual({ a: 'b' });
       expect(initialStoriesHash['b--1'].args).toEqual({ x: 'y' });
 
-      const api = new EventEmitter();
-      init({ api });
-
+      init();
```
ndelangen: init call has no arguments anymore, fullAPI is provided in module's construction.

lib/api/src/tests/stories.test.js
```diff
@@ -531,49 +532,31 @@ describe('stories API', () => {
...
-        isLeaf: true,
-        parent: 'b',
-        kind: 'b',
-        name: '1',
-        parameters: {},
-        path: 'b--1',
-        id: 'b--1',
-        args: { x: 'y' },
-      });
+      expect(changedStoriesHash['a--1'].args).toEqual({ foo: 'bar' });
```
ndelangen: REMOVED some data is the equals as the only thing we care about in THIS test is the args property

lib/api/src/index.tsx
```diff
@@ -243,7 +207,7 @@ class ManagerProvider extends Component<ManagerProviderProps, State> {
     // a chance to do things that call other modules' APIs.
     this.modules.forEach(({ init }) => {
       if (init) {
-        init({ api: this.api });
+        init();
```
ndelangen: Use the fullAPI instead

PR #10097, opened 2020-03-10
Core: Warn about deprecated config files

yannbf:
Issue: There's no issue specifically, but it's to avoid having more issues like #9854 

## What I did
This PR throws an error if the user does not have presets.js or main.js in their project. 
It could happen either if the file is really not there or if the file is being used in an unsupported extension. 
This could help users migrate by giving them a direction when upgrading storybook.

## How to test
Yes. I'm pointing to 6.0 docs in the log, so the docs from next need to be merged to master.


shilman (approved):
LGTM! Great fix!

shilman (changes requested):
What if the user is still using `config.js`?

yannbf:
> What if the user is still using `config.js`?

You are right. The intention would be to make people move away to the new configuration setup but it's better to warn with a deprecated message instead. Changes on the way!

yannbf:
Oops! sorry for inactivity, corona side effects 😓 Work will be done soon, just need to get some feedback from @shilman  👍

yannbf:
Hey @shilman changes are added! Please check if they all make sense to you.
ps: this PR will have a really small conflict with  #10288

lib/core/src/server/utils/interpret-files.js
```diff
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import { extensions } from 'interpret';
 
-const boost = new Set(['.js', '.jsx', '.ts', '.tsx']);
+export const boost = new Set(['.js', '.jsx', '.ts', '.tsx']);
```
yannbf, line 4: This is where the conflict with #10288 will happen

shilman (changes requested):
Can you update `exists` to handle the extensions code, i.e. `exists('main')`? will make the code a lot cleaner.

Also the other PR has been merged, so you can resolve the conflicts while you're at it! 😘

yannbf:
> Can you update `exists` to handle the extensions code, i.e. `exists('main')`? will make the code a lot cleaner.
Thanks for you review! Done!

> Also the other PR has been merged, so you can resolve the conflicts while you're at it! 😘
Also done!

shilman (approved):
Looking great!! 😍

ndelangen (approved):

PR #10102, opened 2020-03-10
Fix typo in addon-contexts docs

tinymachine:
Issue: I know addon-contexts will be deprecated in favor of addon-toolbars, but in the meantime others might benefit from a minor typo correction: the sample code for setting per-story context settings uses the key `context`, when it should be `contexts` (presumably to match the addon's `parameterName`?).

## What I did

Changed the key `context` to `contexts`.

## How to test

- Is this testable with Jest or Chromatic screenshots? **Don't know**
- Does this need a new example in the kitchen sink apps? **Don't think so**
- Does this need an update to the documentation? **Err, not beyond the change itself**

### DIFF
--- a/addons/contexts/README.md
+++ b/addons/contexts/README.md
@@ -84,7 +84,7 @@ Finally, you may want to modify the default setups at per story level. Here is h
 export const defaultView = () => <div />;
 defaultView.story = {
   parameters: {
-    context: [{}]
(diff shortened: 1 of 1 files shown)


shilman (approved):
Thanks @tinymachine !

tinymachine:
Happy to help, @shilman

PR #10112, opened 2020-03-11
Docs: add detail to addon-contexts per-story settings

tinymachine:
Issue:

The addon-contexts readme currently says, "...you may want to modify the default setups at per story level". But there's no example provided, and it appears that "modifying" is a stretch. You can create per-story contextual environments, and you can disable default setups, but there doesn't seem to be a way to actually modify a setup defined in context.js at the story level.

## What I did
Clarified that you can create per-story contexts or disable default (global) contexts defined in contexts.js, and provided sample code for the latter. I also linked to my explanation on Discord of how I went about "modifying" a default setup at the story level, which involved disabling the default setup and creating a new, nearly identical setup with a different default option. Not a lovely solution (and maybe only relevant until addon-toolbars is released), which is why I didn't include the code here in the readme itself.

## How to test

- Is this testable with Jest or Chromatic screenshots? **No.**
- Does this need a new example in the kitchen sink apps? **No.**
- Does this need an update to the documentation? **No -- it is an update to the documentation.**

### DIFF
--- a/addons/contexts/README.md
+++ b/addons/contexts/README.md
@@ -78,16 +78,35 @@ export default {
 };
 ```
 
-Finally, you may want to modify the default setups at per story level. Here is how you can do this:
+Finally, you may want to create new contextual environments or disable default setups at the story level. To create a new contextual environment at the story level:
+
+```js
+export const defaultView = () => <div />; // sample story in CSF format
+defaultView.story = {
+  parameters: {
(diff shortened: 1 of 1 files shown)


tinymachine:
(BTW feel free to delete the link to the Discord message, or to make any changes. Just wanted to submit this PR in case it could save other folks some time.)

shilman (approved):
Great! Thanks for the update @tinymachine 💯

PR #10152, opened 2020-03-18
Feature/async html

jzeltman:
Issue: Async HTML Stories

## What I did
Updated the HTML Render method to accept asynchronous stories. Discussed with Norbert and found that support had mostly been added already, but just needed to update the render method. I also updated the story snapshots with the new async story.

## How to test
Run tests, or view the async welcome story: http://localhost:9006/?path=/story/welcome--async-welcome

- Does this need a new example in the kitchen sink apps? Yes. It's been added: http://localhost:9006/?path=/story/welcome--async-welcome


jzeltman:
Closing for now. I branched from master, and I think that's causing the conflicts.

ndelangen:
Yeah branch off of next please :)

PR #10187, opened 2020-03-23
Core: Remove useStoryState

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10113


tmeasday:
Cool. What about the other changes to `useSharedState` that were added to support it? Do we still want to keep them?

ndelangen:
@tmeasday I think useSharedState is good as it is.

tmeasday (approved):

Issue #10153, opened 2020-03-18
(Documentation) glob pattern support for stories config option


matthias-ccri:
Hi folks! Thanks for the awesome product.

TLDR: This issue is a request to document the fact that storybook uses glob for the `stories` config.

It seems like the `stories` option in `.storybook/main.js` is being parsed by [glob](https://www.npmjs.com/package/glob), but I don't think this is documented. 

In the [React Guide](https://storybook.js.org/docs/guides/guide-react/#step-3-create-the-main-file), it hints at some pattern matching support but doesn't exactly say what's supported:

Under "Step 3: Create the main file", it uses a pattern: 
```js
module.exports = {
stories: ['../src/**/*.stories.[tj]s'],
};
```

P.S. My use case is that I wanted to match all of these file extensions: js, jsx, ts, tsx. I was able to do it with this pattern: `../src/**/*.stories.{js,ts,jsx,tsx}`

ndelangen:
It seems this page is the right place to place this info:
https://github.com/storybookjs/storybook/blob/next/docs/src/pages/configurations/overview/index.md

I think the page is missing on current docs and also on next docs
https://storybook.js.org/docs/basics/introduction/
https://deploy-preview-93--storybook-frontpage.netlify.com/docs/basics/introduction/

We should fix that. @matthias-ccri would you be able to assist?

matthias-ccri:
Sure thing — here's a PR: https://github.com/storybookjs/storybook/pull/10196

ndelangen:
Thank you!

Domiii:
This information from this issue is outdated, and at the same time, one of the top google results. If you (like me) are confused about Storybook's `glob` resolution algorithm:

1. The official documentation on this is now [here](https://storybook.js.org/docs/react/configure/overview#configure-story-loading)
2. As explained on that page, the glob syntax is that of [micromatch](https://github.com/micromatch/micromatch#extended-globbing) (and not that of [glob](https://www.npmjs.com/package/glob)).
3. Here is the storybook code that parses and resolves stories: https://github.com/storybookjs/storybook/blob/master/lib/core-common/src/utils/to-require-context.ts#L31

Hope this helps! (it took me a bit of digging to find all of that out)

PR #10196, opened 2020-03-23
Fix configurations/overview header & add glob info

matthias-ccri:
Issue: https://github.com/storybookjs/storybook/issues/10153

## What I did

This is a documentation change.
1) From https://github.com/storybookjs/storybook/issues/10153, I added wording about `glob`.
2) The configurations/overview page was not being rendered on the site. I removed a leading space above the file header. Hopefully that fixes the issue.

I went over the document and tried to improve the wording as much as I could. If you consider any of these changes to be regressions, then feel free to revert. 

I decided to move the bottom section (about the relationship of main.js to Presets) to the top, because there was starting to be two sections for the same thing --- explaining main.js properties. And I chose to put this at the top, because people probably want to see that information most of the time.

## How to test

Not sure --- how do you test these documentation changes?

### DIFF
--- a/docs/src/pages/configurations/overview/index.md
+++ b/docs/src/pages/configurations/overview/index.md
@@ -1,4 +1,3 @@
-
 ---
 id: 'overview'
 title: 'Configuration overview'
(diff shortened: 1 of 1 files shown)


shilman (approved):
Great catch and great changes @matthias-ccri 💯

Issue #7116, opened 2019-06-17
Options knob throws 'Object(...) is not a function'


robbertolierook:
**Describe the bug**
Options knob throws `Object(...) is not a function` error

**To Reproduce**
I used this knob directly on my component:
```
    defaultValue={options(
          "Default Value",
          { First: 1, Second: 2, Third: 3, },
          1,
          { display: 'check', }
        )}
```
The Knob did not render in the addon panel, and the story display threw this error: 

```
TypeError: Object(...) is not a function
    at webpackHotUpdate../src/stories/checkboxGroup.stories.js._index_stories__WEBPACK_IMPORTED_MODULE_2__.default.add.readme.content (checkboxGroup.stories.js:29)
    at client_api.js:114
    at wrapper (index.js:179)
    at make-decorator.js:17
    at make-decorator.js:29
    at client_api.js:112
    at client_api.js:114
    at jsxDecorator (index.js:98)
    at client_api.js:112
    at client_api.js:114
```

Since it also mentions the 'readme' addon, it might be a problem with the combination of both, which is weird, because all other knobs work perfectly.

**Expected behavior**
The knob to work and render checkboxes

**System:**
 - OS: [MacOS]
 - Device: [Macbook Pro 2018]
 - Browser: [chrome]
 - Framework: [react]
 - Addons: [knobs, storybook-readme]
 - Version: [5.0.5]

zackphilipps:
@robbertolierook how are you importing the options knob?

robbertolierook:
Like the rest of the knobs:

```
import {
  text, boolean, select, options,
} from '@storybook/addon-knobs';
```

zackphilipps:
@robbertolierook according to the docs, it's actually `optionsKnob`, so `options` is probably something else:

```js
import { optionsKnob as options } from '@storybook/addon-knobs';
```

Not intuitive, but 🤷‍♂

robbertolierook:
Oh golly, thanks for that! Now it works:)

PR #10218, opened 2020-03-26
Changes import of options/optionsKnob

Spielboerg:
Issue: https://github.com/storybookjs/storybook/issues/7116

## What I did
The import, which is different from all other imports, is easy to overlook. Therefore I have adapted the import to the other button types and added a note for an alternative import that corresponds to the naming of the other button types (without the suffix `Knob`).

### DIFF
--- a/addons/knobs/README.md
+++ b/addons/knobs/README.md
@@ -396,7 +396,7 @@ const value = radios(label, options, defaultValue, groupId);
 Configurable UI for selecting a value from a set of options. 
 
 ```js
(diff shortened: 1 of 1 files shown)


Spielboerg:
@shilman Why are the tests failing? I can't imagine that has anything to do with my changes, does it?

Spielboerg:
Not fixed yet

Spielboerg:
Comment for stale bot

shilman (approved):
LGTM

PR #10226, opened 2020-03-27
Feature add data-tags attribute to sidebar items

thehuijb:
Issue: [9209](https://github.com/storybookjs/storybook/issues/9209) Add an option to hide a story from the sidebar

## What I did
I made an adjustment to how stories are written out to the sidebar to include a `data-tags` attribute which defaults to `data-tags="none"` 

the tags can then be used to show or hide stories based on plain css selectors:
for instance one could add a rule to hide all stories with `data-tags="hidden"`:
```
[data-tags=hidden] {
  display: none;
}
```
or more complex to hide all stories that do not have a certain TAG so you could show just the stories that belong to that TAG:
```
[data-tags]:not([data-tags~=TAG] {
  display: none;
}
```
> NOTE: - these css options are not included in the PR one must add these depending on the solution they are looking for.

to add the tags to the stories you only need to add it to the default like so:
```
export default {
  title: 'Welcome',
  tags: 'hidden',
};
```
or to a specific story like this:
```
Button.story = {
  tags: 'hidden'
}
```

you can specify multiple tags using the pipe symbol `|` e.g. `tags: 'web|app'` to specify that the component or story belongs to the `web` and to the `app`
if you would add `tags: 'web|app'` to default and then add `tags: 'web'` to a story and `tags: 'app'` to another, one could show and hide different stories based on a selected TAG
if components are created under the same branch, the branch will receive the tags from all component and pipe `|` them together e.g. if one storyfile has 
```
export default {
  title: 'Components/Category/Headings',
  tags: 'web'
}
```
and another storyfile has
```
export default {
  title: 'Components/Category/Buttons',
  tags: 'app'
}
```
the resulting link in the sidebar to `Category` will have `data-tags="web|app"`

## How to test

1. start one of the example apps
2. add `tags: 'hidden'` to one of the stories files like described above
3. refresh page and inspect the stories in the sidebar with the devtools
4. see that the links in the sidebar contain `data-tags` attibutes

- Does this need a new example in the kitchen sink apps? probably NOT
- Does this need an update to the documentation? YES included

### DIFF
--- a/lib/addons/src/types.ts
+++ b/lib/addons/src/types.ts
@@ -18,6 +18,7 @@ export function isSupportedType(type: Types): boolean {
 
 export type StoryId = string;
 export type StoryKind = string;
+export type StoryTags = string;
 export type StoryName = string;
 export type ViewMode = 'story' | 'docs';
 
@@ -54,6 +55,8 @@ export interface StoryIdentifier {
 
 export type StoryContext = StoryIdentifier & {
   [key: string]: any;
+  componentTags?: StoryTags;
+  storyTags?: StoryTags;
   parameters: Parameters;
   args: Args;
   globalArgs: Args;
@@ -124,7 +127,9 @@ export interface StoryApi<StoryFnReturnType = unknown> {
   add: (
     storyName: StoryName,
     storyFn: StoryFn<StoryFnReturnType>,
-    parameters?: Parameters
+    parameters?: Parameters,
+    componentTags?: StoryTags,
+    storyTags?: StoryTags
   ) => StoryApi<StoryFnReturnType>;
   addDecorator: (decorator: DecoratorFunction<StoryFnReturnType>) => StoryApi<StoryFnReturnType>;
   addParameters: (parameters: Parameters) => StoryApi<StoryFnReturnType>;
--- a/lib/api/src/lib/stories.ts
+++ b/lib/api/src/lib/stories.ts
@@ -15,6 +15,7 @@ export interface Root {
   depth: 0;
   name: string;
   refId?: string;
+  componentTags?: string;
   children: StoryId[];
   isComponent: false;
   isRoot: true;
@@ -30,6 +31,7 @@ export interface Group {
   id: StoryId;
   depth: number;
   name: string;
+  componentTags?: string;
   children: StoryId[];
   refId?: string;
   parent?: StoryId;
@@ -51,6 +53,8 @@ export interface Story {
   name: string;
   kind: StoryKind;
   refId?: string;
+  componentTags?: string;
+  storyTags?: string;
   children?: StoryId[];
   isComponent: boolean;
   isRoot: false;
@@ -74,7 +78,9 @@ export interface StoryInput {
   id: StoryId;
   name: string;
   refId?: string;
-  kind: StoryKind;
+  kind: string;
+  componentTags?: string;
+  storyTags?: string;
   children: string[];
   parameters: {
     fileName: string;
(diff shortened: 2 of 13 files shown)


lib/client-api/src/decorators.ts
```diff
@@ -9,6 +9,8 @@ const defaultContext: StoryContext = {
   id: 'unspecified',
   name: 'unspecified',
   kind: 'unspecified',
+  componentTags: 'unspecified',
```
ndelangen: tagS suggest plural to me, shouldn't this be an array?

lib/api/src/lib/stories.ts
```diff
@@ -228,11 +235,24 @@ export const transformStoriesRawToStoriesHash = (
...
+        isComponent,
+        isLeaf,
         ...(child && { children: [child] }),
       });
+      acc[id].isRoot = group.isRoot;
+      if (group.componentTags) {
+        if (!acc[id].componentTags) {
+          acc[id].componentTags = group.componentTags;
+        } else if (!acc[id].componentTags.includes(group.componentTags)) {
+          acc[id].componentTags += `|${group.componentTags}`;
```
ndelangen: I think this concat should be done at render time.

ndelangen (changes requested):
Overall great change, I just wish the data-structure was an array, so it will be easier in the future to do more with this data.

lib/addons/src/types.ts
```diff
@@ -124,7 +127,9 @@ export interface StoryApi<StoryFnReturnType = unknown> {
   add: (
     storyName: StoryName,
     storyFn: StoryFn<StoryFnReturnType>,
-    parameters?: Parameters
+    parameters?: Parameters,
+    componentTags?: StoryTags,
```
ndelangen: What's the reason for having 2 distinct fields for this?

ndelangen: why not just `tags`?

lib/core/src/client/preview/loadCsf.test.ts
```diff
@@ -246,13 +254,19 @@ describe('core.preview.loadCsf', () => {
...
+      'X',
+      input.a.x,
+      {
+        x: 'y',
+        decorators: [decorator],
+        __id: 'a--x',
+        args: { b: 1 },
+        argTypes: { b: 'string' },
+      },
+      undefined,
```
ndelangen: This doesn't seem right to me, seems like tags should be part of the 3rd argument (the object)

lib/ui/src/components/sidebar/Tree/Tree.tsx
```diff
@@ -24,7 +27,7 @@ const branchOrLeaf = (
     depth,
   }: { root: string; dataset: Dataset; expanded: ExpandedSet; selected: SelectedSet; depth: number }
 ) => {
-  const node = dataset[root];
+  const node = dataset[root] as Story;
```
ndelangen: if no tags are provided, I'd say it should `undefined` ?

lib/ui/src/components/sidebar/Tree/Tree.tsx
```diff
@@ -106,7 +119,9 @@ const Tree: FunctionComponent<{
       return <List>{children.map(mapNode)}</List>;
     }
     case node.isLeaf: {
-      return <Leaf key={node.id} {...node} depth={depth} isSelected={selected[node.id]} />;
+      return (
+        <Leaf key={node.id} tags={tags} {...node} depth={depth} isSelected={selected[node.id]} />
```
ndelangen: There should be no need to pass in `tags` explicitly here, should be part of `node`

thehuijb:
> Overall great change, I just wish the data-structure was an array, so it will be easier in the future to do more with this data.

Thanks for reviewing this PR Norbert, I will respond to your comments as soon as possible, bit busy at the moment.

thehuijb:
Abandoning this PR

PR #10243, opened 2020-03-28
fix stories pattern in angular guide

wKich:
Issue: None

## What I did

Fix copy-pasted stories pattern from `react` guide.


shilman (approved):
Thanks!

PR #10245, opened 2020-03-28
components: format Source code with Prettier

hasparus:
Issue: Relevant to https://github.com/storybookjs/storybook/issues/8078 (doesn't close), https://github.com/storybookjs/storybook/issues/8078#issuecomment-605357603, https://github.com/storybookjs/storybook/issues/8078#issuecomment-605430645

## What I did

- [x] Augmented formatter with `prettier/standalone`.
  - Parsers are imported with `import()` function, so they should be loaded at runtime only when needed.
- [x] Allowed passing SourceProps through `withSource` prop of Preview to opt-out of formatting.
  - Example:
    ```jsx
    <Preview
        withSource={{
            language: 'jsx',
            format: false,
            state: SourceState.OPEN
        }}
    >
      <Story name="unformatted JSX in MDX">{/* ... */}</Story>
    </Preview>
    ```

### Notes

- The tests are in: lib/components/src/syntaxhighlighter/formatter.test.js
- It's used only in: lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
- Prettier usually infers parser from filename.
In the browser, we have to pass it explicite.
Supported parsers include: `angular`, `vue`,  `babel`, `mdx`, `css`, `html`.
    - https://prettier.io/docs/en/options.html#parser
    - https://prettier.io/docs/en/browser.html#es-modules
    - Parsers are published as separate entry points in prettier package.

    - I plan to add support for `babel`, `css` and `html` in this pull request.
I believe following up on this won't be hard.
- It's worth to note the significant bundle size of prettier/standalone (~**1mB unpacked**) and parsers (prettier-plugin-svelte is **3.4kB gzipped**, prettier-plugin-reasonml is **13kB gzipped**).
    - It might be beneficial to import `prettier/standalone` only when the format prop is set to `true`.
    - **Question:** Should storybook packages for particular framework (e.g. @storybook/vue) add parsers explicitly? Or should they be dynamically imported depending on the `language` prop passed to `Source`.

## How to test

  - I've tested my changes in `formatter.test.js` and added relevant stories to 
    - `syntaxhighlighter.stories.tsx`
    - `Source.stories.tsx`
   - `formatted-preview.stories.mdx`
     - These stories use the topmost API, so this might be a good place for Chromatic screenshots. I'm not familiar with Chromatic, so I'm not sure how to proceed further.
  - I'm not sure.
  - [x] A pararaph in [addons/docs/docs/mdx.md](https://github.com/storybookjs/storybook/blob/54757192c5bd8235c6343b4e0efff120fe61b5d4/addons/docs/docs/mdx.md) would be useful.


ndelangen:
Can we lazy-load the prettier dependency at runtime?

hasparus:
> Can we lazy-load the prettier dependency at runtime?

That makes sense.

hasparus:
Thanks 😅

hasparus:
Screenshot for comparison:
image

**Quick question: How should the user opt-out / opt-in from Prettier formatting?**

I assume we'd like to format with Prettier by default in the doc block for all supported languages.

Reasons:
1. It feels like a good default. Prettier sceptics are probably the minority.
2. MDX story source isn't preserved 1-1, what causes https://github.com/storybookjs/storybook/issues/8078.

However, a possibility to opt out may be valuable.
How does `<Preview format={false}>` or `<Story format={false}>` look like?

The part I'm interested in right now is the line with my TODO comment in addons/docs/src/blocks/Source.tsx.

```tsx
export const getSourceProps = (
  props: SourceProps,
  { id: currentId, storyStore }: DocsContextProps
): PureSourceProps => {
  // ...
  return source
    ? {
        code: source,
        language: props.language,
        dark: props.dark || false,
        format: true, // TODO: A way to opt-out
      }
    : { error: SourceError.SOURCE_UNAVAILABLE };
};
```

shilman:
@hasparus I think prettier formatting should def be the default. I can imagine cases where the user wants to format code manually and have NO formatting applied at all, so I think it couldn't hurt to support that as well.

hasparus:
image

DeepScan seems to have an old Prettier version.

caspardue:
I tried using `prettier/standalone` (with the babel parser). Please keep in mind that it's not compatible with IE 11/Edge out of the box :(

hasparus:
> I tried using prettier/standalone (with the babel parser). Please keep in mind that it's not compatible with IE 11/Edge out of the box :(

AFAIK Prettier uses `Symbol` and `String.prototype.startsWith`. 
Symbol seems to be already polyfilled in https://github.com/storybookjs/storybook/blob/ef14ecf3b5a61e0045ceaa4a310f7ec7c0a0e2fb/lib/core/src/server/common/polyfills.js#L3
We could add startsWith there, if it's needed.

I think they're both supported in Edge, though.

I'll get my Windows notebook from repair and try running the code in IE.
Not sure if it works even before my changes https://github.com/storybookjs/storybook/issues/4099.

caspardue:
It uses classes and arrow functions that need to be handled I guess. Storybook v5.3.18 works fine for me in IE11

shilman:
cc @ndelangen @tooppaaa any thoughts on IE11 compat in this one?

hasparus:
> It uses classes and arrow functions that need to be handled I guess.

Yeah, that's correct. Just skimmed through https://unpkg.com/prettier@2.0.5/standalone.js.
We'd need to transpile this for IE.

tooppaaa:
@hasparus We use DLL to transpile node_modules into es5. I'm unsure how this behave with splitChunks. I was going to checkout your PR and have a look :)

hasparus:
@tooppaaa My change to splitChunks config doesn't really affect anything.
`@storybook/components` are compiled to ES5 in TypeScript, before they reach webpack, so we can't use dynamic import for lazy loading. ATM all parsers just go into vendor bu

I suppose that the ideal scenario would be to compile to ES5 at the app level and leave ES6 imports in libraries to let the bundler do its magic.

JGJP:
Will this be merged sometime soon? really wanting this

lib/components/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "module": "ESNext",
```
ndelangen, line 4: This is broken, for some reason

lib/core/src/server/manager/manager-webpack.config.js
```diff
@@ -155,7 +155,23 @@ export default async ({
     },
     optimization: {
       splitChunks: {
-        chunks: 'all',
+        cacheGroups: {
+          prettier: {
+            test: /[\\/]node_modules[\\/]prettier[\\/]/,
+            chunks: 'async',
+            filename: '[name].bundle.js',
```
ndelangen, line 162: Is it possible to do this in the DLL as well? Or completely move this to the DLL?

lib/components/tsconfig.json
ndelangen: What was the reason this was added?

as-zlynn-philipps:
pour one out for this one :(

as-zlynn-philipps:
@hasparus @ndelangen Y'know, the two features:

1. `withFormat={false}`
2. Formatting with Prettier

Really should've been split into two PRs. Precisely because 2 is holding up 1 here, and 1 is useful on its own (I'd like to use it.) I'll attempt to make a PR for 1 based on the work here.

ndelangen:
@hasparus @caspardue @tooppaaa @JGJP @as-zlynn-philipps @shilman What can we do to move this forwards?

Do we start a new branch/PR based on next?

JGJP:
Can somebody confirm if this is still an issue or not?

ndelangen:
Yeah it is.

PR #10251, opened 2020-03-30
Adding children prop to button Button.tsx

alexreardon:


lib/components/src/Button/Button.tsx
```diff
@@ -13,6 +13,7 @@ export interface ButtonProps {
   small?: boolean;
   outline?: boolean;
   containsIcon?: boolean;
+  children?: React.ReactNode;
```
ndelangen: I think we shouldn't do this, in fact I recently learned about:

```js
import React, { ComponentProps } from 'react';

interface Props {
  isLink?: boolean;
  primary?: boolean;
  secondary?: boolean;
  tertiary?: boolean;
  gray?: boolean;
  inForm?: boolean;
  disabled?: boolean;
  small?: boolean;
  outline?: boolean;
  containsIcon?: boolean;
}

export type ButtonProps = ComponentProps<typeof ButtonWrapper>;
```

ndelangen: Want to use that instead of adding children prop 'manually'?

alexreardon: I think I have a better solution:

alexreardon: Actually, looking at the source I think you might be right

alexreardon: Right now the Button does not know about children, onClick and other standard button props

alexreardon: On my local I fixed it by adding:
```diff
export declare const Button: React.ForwardRefExoticComponent<ButtonProps & React.RefAttributes<any> 
+ & React.HTMLAttributes<HTMLButtonElement>
> & {
    defaultProps: {
        isLink: boolean;
    };
};
```

ndelangen: `React.HTMLAttributes<HTMLButtonElement>` is included in `styled.button`, the best solution is to pass those types up correctly IMHO.

I'll take any improvement PR though 👍

ndelangen (approved):

PR #10276, opened 2020-03-31
Misleading information on upgrade: react-docgen

Vanuan:
Unfortunately, RTDL supports imported typescript interfaces when generating property table.

## What I did

Changed a notice about a regression in 6.0

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -165,7 +165,8 @@ type NewType = SomeType & { foo: string };
 const MyComponent: FC<NewType> = ...
 ```
 
-This was also an issue in `RDTL` so it doesn't get worse with `react-docgen`. There's an open PR for this https://github.com/reactjs/react-docgen/pull/352 which you can upvote if it affects you.
+This isn't an issue with `RDTL` so unfortunately it gets worse with `react-docgen`.
+There's an open PR for this https://github.com/reactjs/react-docgen/pull/352 which you can upvote if it affects you.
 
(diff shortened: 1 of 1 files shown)


Vanuan:
Here's an example:

main.js:

    webpackFinal: async config => {
      config.module.rules.push({
        test: /\.(ts|tsx)$/,
        use: [
          {
            loader: require.resolve('ts-loader'),
            options: {
              configFile: path.resolve(__dirname, './tsconfig.json'),
              transpileOnly: true,
            },
          },
          // Automatics props
          {
            loader: require.resolve('react-docgen-typescript-loader'),
          },
        ],
      });
      config.resolve.extensions.push('.ts', '.tsx');
      return config;
    },


tsconfig:

```
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "baseUrl": "../src",
    "noEmit": false
  },
  "include": ["../src", "../declarations.d.ts"]
}

```

Button.tsx:

```
import { IButtonProps } from './types';

const Button = (props: IButtonProps) => <button>{props.children}</button>;

export { Button };
```

types.ts:
```
export interface IButtonProps {
  children: JSX.Element
}
```

Vanuan:
The example above works with RDTL but doesn't work with babel-plugin-react-docgen

shilman (approved):
Oops, didn't realize this was supported in RDTL. Thanks for the fix!

Issue #10282, opened 2020-04-01
Document addon-viewport disable paramter


buscape:
Had to dig into the [source](https://github.com/storybookjs/storybook/blob/next/addons/viewport/src/Tool.tsx#L130) to find the `disable` option.
It's missing in the Readme https://github.com/storybookjs/storybook/tree/next/addons/viewport#configuration

buscape:
Fixed by https://github.com/storybookjs/storybook/pull/10283

PR #10283, opened 2020-04-01
Document addon-viewport disable paramter #10282

buscape:
Issue: #10282

## What I did
Added it to the docs


shilman (approved):
LGTM, thanks!

PR #10288, opened 2020-04-02
ADD .cjs & .mjs to interpret-files

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10075#issuecomment-607436110

## What I did
ADD .cjs & .mjs to the list of extentions that storybook will try to load as config files

WDYT @andrewshawcare


andrewshawcare:
I'm good If it picks up the `main.mjs` extension -- thanks for this!

PR #10321, opened 2020-04-06
Explanation for template and addDecorators

seveves:
Issue:
For Angular devs it would be helpful to see how to use template and addDecorators and moduleMetadata with the MDX format because it is very common with the storiesOf-API.

## What I did
Added an example how to use template, moduleMetadata and addDecorators.

## How to test
Only documentation update.


seveves:
I'm not a native speaker. So feel free to correct my English where needed :)

shilman (approved):
Thanks @seveves !!! 🙏

PR #10343, opened 2020-04-07
Dev: Build script for package development

yannbf:
## What I did
This is a developer experience contribution for storybook contributors. It's an easy way to build packages within the monorepo. It presents all lerna packages that use the `@storybook/name` pattern and you can select multiple or single to build, also being able to toggle watch mode.

This is how it looks like:
cli

## How to test
In the order of the video above

* Present a prompt for options:
`yarn build`

* Build all @ storybook packages:
`yarn build --all`

* Build a specific @ storybook package:
`yarn build --@storybook/package-name`

* Build on watch mode (works for specific package or all):
add --watch to the command, like 
`yarn build --all --watch` or `yarn build --@storybook/package-name --watch`

**P.S.** if you select more than 5 packages to build on watch mode, it asks if you really want to do that, given that it's an expensive process that might eat up RAM really quick. It's also there for people who might mistakenly select all or more packages than they wanted.

Please hit me up with whatever requests and I'll be happy to do them to improve this contribution! 👍


shilman (changes requested):
This looks awesome!! I love it.

My one question is whether we can make the packages regular args instead of hyphened args? I.e. `yarn build @storybook/core @storybook/addon-docs --watch`

WDYT?

yannbf:
> This looks awesome!! I love it.

🤸‍♀ 

> My one question is whether we can make the packages regular args instead of hyphened args? I.e. `yarn build @storybook/core @storybook/addon-docs --watch`
> 
> WDYT?

Definitely! Just did it. Before I push, I was thinking of maybe making it shorter even, given that all packages are `@storybook/name`, just use `name` instead:
`yarn build core addon-docs --watch`

I didn't do it initially because I thought making it more explicit would be better, but it's all a matter of preference.

ndelangen:
WOW

ndelangen (approved):

ndelangen:
> just use name instead

would love that!

ndelangen (approved):

PR #10345, opened 2020-04-07
Marko: Always destroy old component when switching stories

DylanPiercey:
Issue:
In storybook@5 the Marko plugin was updated to prefer reusing component instances (passing in new input) if the last rendered component was the same. This change improved support for knobs and other plugins which rerender a story to preserve the Marko components state.

This issue is that this can be a bit confusing when switching between stories since you likely expect things to be in a fresh state when doing so.

## What I did

I updated the Marko plugin to keep track of the last rendered story and it now only does a rerender if the same component was previously rendered _and_ the story function has not changed.


shilman (commented):
LGTM! Thanks!

PR #10358, opened 2020-04-09
FIX .cjs files for config as main

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10075

## What I did

I improved the serverResolve & serverRequire


lib/core/src/server/utils/server-require.js
```diff
@@ -68,16 +69,15 @@ function getCandidate(paths) {
...
+  const candidatePath = serverResolve(filePath);
 
-  if (!existingCandidate) {
+  if (!candidatePath) {
     return null;
   }
 
-  const { path: candidatePath, ext: candidateExt } = existingCandidate;
-
+  console.log({ candidatePath });
```
yannbf, line 79: Are both console.logs from line 79 and 96 supposed to be there?

shilman (changes requested):
Nice catch @yannbf on the `console.log`

yannbf (approved):

PR #10387, opened 2020-04-11
Add build script steps to CONTRIBUTING

yannbf:
Related PR: #10343 

## What I did

Added steps to use build scripts CLI command on contributing docs. Making the devs lives easier! 😄


shilman (approved):
LGTM! But please remember to run prettier on the files before checking in. If you use VSCode this happens automatically so you don't need to remember.

PR #10216, opened 2020-03-26
Tech/remove deprecations for 6.0.0

ndelangen:
Issue:

## What I did
I removed all deprecation warnings

PR #10402, opened 2020-04-13
Fixed `webpackFinal` being called twice

Yama-Tomo:
Issue:

relate: https://github.com/storybookjs/storybook/pull/10216

Breaking the build because `webpackFinal` will be called twice if write  not idempotent custom webpack configuration.

## What I did

Fix createFinalDefaultConfig call to one time only.


shilman (approved):
Looks reasonable to me. @ndelangen @tmeasday can you take a look? Can we add a test for this?

ndelangen (approved):

PR #10404, opened 2020-04-13
FIX legacy redirect

ndelangen:


lib/core/src/client/preview/start.ts
```diff
@@ -65,15 +65,19 @@ export default function start(
 
   // Only try and do URL/event based stuff in a browser context (i.e. not in storyshots)
   if (isBrowser) {
-    // Initialize the story store with the selection in the URL
-    const { storyId, viewMode } = initializePath(storyStore);
+    const afterStoriesSet = () => {
```
ndelangen, line 68: @tmeasday The problem was that before `configure` is called the storystore is empty, and so the `initializePathe` would never return a story.

ndelangen: I've delayed that code to run after `setStories` event

tmeasday: Ok, I can see why we need to do this. Does this mean the `STORY_RENDER` event ends up running twice in this case?

ndelangen: Yea potentially, I think what SHOULD happen is that it only is called when we're sure the correct story rendered.

PR #10456, opened 2020-04-17
Addon-a11y: use channel to highlight elements in preview

tooppaaa:
Issue:

The a11y addon is using a getIframe method to access the preview which won't work with storybook composition.

## What I did
 
- Refactor A11yPanel & HighlightToggle to use a context to move the logic out of these components
- Create A11yContext in charge of handling results, what's highlighted, some addon behavior
- Unit test
- Use channel to highlight element in the preview
- use additional stylesheet to highlight instead of playing directly with element's outline style

## How to test

- Highlight elements in a11y addon


addons/a11y/src/a11yHighlight.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const sheet = document.createElement('style');
+  sheet.setAttribute('id', id);
+  sheet.innerHTML = infos.elements
+    .map((target) => `${target}{ outline: 1px dotted ${infos.color}!important; }`)
+    .join(' ');
+  document.head.appendChild(sheet);
+};
+
+channel.on(EVENTS.HIGHLIGHT, highlight);
```
ndelangen, line 32: Does this need to be reset when `story_change` hanppens?

addons/a11y/src/redux-config.tsx
```diff
@@ -1,44 +0,0 @@
-import { createStore } from 'redux';
```
ndelangen, line 1: Can we remove these dependencies now?

addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -49,228 +48,133 @@ const Centered = styled.span<{}>({
...
-          if (status === 'ran') {
-            this.setState({
-              status: 'ready',
-            });
-          }
-        }, 900);
+type Status = 'initial' | 'manual' | 'running' | 'error' | 'ran' | 'ready';
+
+export const A11YPanel: React.FC = () => {
+  const [status, setStatus] = React.useState<Status>('initial');
```
ndelangen, line 54: We can use `useAddonState` perhaps, to have it persist automatically by storing the state in the storybook global state?

addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -49,228 +48,133 @@ const Centered = styled.span<{}>({
...
+          <RotatingIcon inline icon="sync" /> Please wait while the accessibility scan is running
+          ...
+        </Centered>
+      )}
+      {(status === 'ready' || status === 'ran') && (
+        <>
+          <ScrollArea vertical horizontal>
+            <Tabs
+              key="tabs"
+              tabs={[
```
ndelangen, line 113: this creates a new list every render, we might want to extract into a `useMemo`?

addons/a11y/src/components/A11YPanel.tsx
```diff
@@ -49,228 +48,133 @@ const Centered = styled.span<{}>({
...
-            {error}
-          </Centered>
-        );
-    }
-  }
-}
+          </ScrollArea>
+          <ActionBar
+            key="actionbar"
+            actionItems={[
```
ndelangen, line 155: extract into `useMemo` as well?

addons/a11y/src/a11yHighlight.ts
```diff
@@ -0,0 +1,32 @@
...
+interface HighlightInfo {
+  /** html selector of the element */
+  elements: string[];
+  color: string;
+}
+
+const channel = addons.getChannel();
+
+const highlight = (infos: HighlightInfo) => {
+  const id = 'a11yHighlight';
```
ndelangen, line 18: extract into addon `constants.ts`

addons/a11y/src/a11yHighlight.ts
```diff
@@ -0,0 +1,32 @@
...
+  const sheetToBeRemoved = document.getElementById(id);
+  if (sheetToBeRemoved) {
+    sheetToBeRemoved.parentNode.removeChild(sheetToBeRemoved);
+  }
+
+  const sheet = document.createElement('style');
+  sheet.setAttribute('id', id);
+  sheet.innerHTML = infos.elements
+    .map((target) => `${target}{ outline: 1px dotted ${infos.color}!important; }`)
```
ndelangen, line 27: Ow nice, so the color in dependent on the priority?

Real meta, but how do we make sure those outlines are actually visible / accessible?

We should add a box-shadow perhaps of 1px background behind the dotted  outline?

addons/a11y/src/a11yHighlight.ts
tooppaaa: Good catch. Done by A11yContext & unit tested

addons/a11y/src/redux-config.tsx
tooppaaa: Yes

addons/a11y/src/a11yHighlight.ts
tooppaaa: That's actually a good question. I'm not a big fan of the outline, I can't see the difference between red / green. Maybe we can have this is the a11y parameters ? (different PR), not sure how the theme impacts this

addons/a11y/src/components/A11YPanel.tsx
tooppaaa: Yep, good idea

addons/a11y/src/a11yHighlight.ts
ndelangen: Assume no theming for now, just static colors.

tooppaaa: What do you think of this ?
image

addons/a11y/src/a11yRunner.ts
```diff
@@ -18,6 +18,7 @@ const getElement = () => {
 };
 
 const run = async (storyId: string) => {
+  console.log('Running', storyId);
```
shilman: nope 😉

tooppaaa: oh :)

addons/a11y/src/a11yHighlight.ts
ndelangen: Made a tweak to this, WDYT?

tooppaaa: Good for me ;)

addons/a11y/package.json
```diff
@@ -46,13 +46,10 @@
     "global": "^4.3.2",
     "lodash": "^4.17.15",
     "memoizerific": "^1.11.3",
-    "react-redux": "^7.0.2",
     "react-sizeme": "^2.5.2",
-    "redux": "^4.0.1",
```
ndelangen, line 51: reducing the weight of the monorepo, 1 dependency at a time 💪

addons/a11y/src/a11yHighlight.ts
```diff
@@ -0,0 +1,44 @@
...
+const channel = addons.getChannel();
+
+const higlighted = (color: string) => `
+  outline: 2px dashed ${color};
+  outline-offset: 2px;
+  box-shadow: 0 0 0 6px rgba(255,255,255,0.6);
+}
+`;
+
+const highlight = (infos: HighlightInfo) => {
```
ndelangen, line 24: Let's extract into a different module and re-sue it in a story so we can test this visually.

addons/a11y/src/a11yHighlight.ts
```diff
@@ -0,0 +1,44 @@
...
+  color: string;
+}
+
+const channel = addons.getChannel();
+
+const higlighted = (color: string) => `
+  outline: 2px dashed ${color};
+  outline-offset: 2px;
+  box-shadow: 0 0 0 6px rgba(255,255,255,0.6);
+}
```
ndelangen, line 21: Let's extract into a different module and re-sue it in a story so we can test this visually.

tooppaaa: Done !

ndelangen (approved):

CodeByAlex:
Great work on this! Excited to see it switched over to use some better practices

PR #10504, opened 2020-04-21
Composition: Update UI for refs

ndelangen:
Relates to #10340.

This PR adds collapsible sections, updates title capitalization and spacing, and adds five loaders.

Loaders are randomly indented by 0 or 1 levels, with the first always being level 0.


lib/api/src/modules/refs.ts
```diff
@@ -82,7 +85,7 @@ export const init: ModuleFn = ({ store, provider }) => {
     findRef: (source) => {
       const refs = api.getRefs();
 
-      return Object.values(refs).find(({ url }) => `${url}/iframe.html`.match(source));
+      return Object.values(refs).find(({ url }) => url.match(source));
```
ndelangen, line 88: This filename matching business is now handled by the channel

PR #10506, opened 2020-04-21
[FIX] switch zeit now to Vercel Now for prisma docs

Developerayo:
- Zeit just rebranded to Vercel and this PR changes the deployment guidelines to deployment
- Now still remains Now but ZEIT is no longer zeit but Vercel as of today


ndelangen (approved):

Developerayo:
🔥

Issue #10509, opened 2020-04-21
Composition: Remote stories are not loading due to refs being parsed as `"[object Object]"`


evanroane:
**Describe the bug**
I first experienced this behavior in my own project where I have the storybook version 6 alpha package installed, but I figured I probably missed something, so I sought to reproduce the issue in `examples/dev-kits` since it illustrates the behavior I’m interested in using.

When I run the `dev-kits` example with refs, the storybooks defined on the `refs` key in main.js do not load.
In the browser console and network tabs I can see repeated attempts to fetch `http://localhost:9011/undefined/metadata.json` and `http://localhost:9011/undefined/stories.json`
I tracked down the issue to https://github.com/storybookjs/storybook/blob/next/lib/api/src/modules/refs.ts#L158

Here, I was able to see that the value of `refs` here is `"[object Object]"`. When `refs` is looped over, `checkRef` is consequently getting called on each character of `"[object Object]"`. Interestingly, TypeScript did not catch that a character string is not the `SetRefData` type since it is being cast with `as` in that context.

**To Reproduce**
1. Clone the storybook repo
2. compile core `yarn bootstrap --core`
3. move into the dev-kits example `cd examples/dev-kits` and  and install deps `yarn install`
4. Run this storybook example `yarn storybook`
5. When you navigate to the running storybook, you will see the failed attempts to access `stories.json` and `metadata.json` in the JS and network tabs.

**Expected behavior**
I expect remote storybooks to load and be populated in the manager sidebar.

**Screenshots**
<img width="1266" alt="image" src="https://user-images.githubusercontent.com/8975323/79916594-a01e0b00-83ee-11ea-809c-88969646961b.png">

**Code snippets**
See links provided above

**System:**
```
Environment Info:

  System:
    OS: macOS 10.15.3
    CPU: (4) x64 Intel(R) Core(TM) i7-7660U CPU @ 2.50GHz
  Binaries:
    Node: 12.10.0 - ~/.nvm/versions/node/v12.10.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.11.3 - ~/.nvm/versions/node/v12.10.0/bin/npm
  Browsers:
    Chrome: 81.0.4044.113
    Firefox: 74.0
    Safari: 13.0.5
```

eroane-rms:
Pull request with fix: https://github.com/storybookjs/storybook/pull/10510

PR #10510, opened 2020-04-21
Composition: Fix remote stories by passing refs to virtual module safely

evanroane:
Issue: https://github.com/storybookjs/storybook/issues/10509

## What I did
In `manager-webpack.config.js` I changed ln. 58 to make sure that the single quote characters (`'`) are replaced in the template file and on ln. 59 I added a call to `JSON.stringify` so that the `refs` object is safely converted to a string before `String.prototype.replace` is called since calling it with an object returns the string `"[object Object]"`.

## How to test

To test follow the steps in https://github.com/storybookjs/storybook/issues/10509 on the `next` branch and with the code in this pull request. You will notice that the `undefined` value is no longer in the URLs of remote storybooks.

### DIFF
--- a/lib/core/src/server/manager/manager-webpack.config.js
+++ b/lib/core/src/server/manager/manager-webpack.config.js
@@ -55,8 +55,8 @@ export default ({
       refs
         ? new VirtualModulePlugin({
             [path.resolve(path.join(configDir, `generated-refs.js`))]: refsTemplate.replace(
-              '{{refs}}',
-              refs
+              /'{{refs}}'/g,
+              JSON.stringify(refs)
             ),
           })
         : null,
--- a/lib/core/src/server/manager/manager-webpack.config.js
+++ b/lib/core/src/server/manager/manager-webpack.config.js
@@ -55,7 +55,7 @@ export default ({
       refs
         ? new VirtualModulePlugin({
             [path.resolve(path.join(configDir, `generated-refs.js`))]: refsTemplate.replace(
-              /'{{refs}}'/g,
+              /'{{refs}}'/,
               JSON.stringify(refs)
             ),
           })


lib/core/src/server/manager/manager-webpack.config.js
```diff
@@ -55,7 +55,7 @@ export default ({
       refs
         ? new VirtualModulePlugin({
             [path.resolve(path.join(configDir, `generated-refs.js`))]: refsTemplate.replace(
-              `'{{refs}}'`,
+              /'{{refs}}'/g,
```
ndelangen, line 58: I'm unsure why this is necessary, `'{{refs}}'` only occurs once in the template?

ndelangen:
I think we both found the same bug, and fixed it, Your PR seems to also change the search-term to a regex, did you have a reason for that? I'd like to know.

evanroane:
I went for a regex here because I wanted to capture the single quote characters (`'`) that are in the template. I typically don't use a template literal to form a string that doesn't need multi-line support or expression interpolation. 

While the template literal seems to work just fine, I prefer to use the regex since it seems the more apt object for the job.

lib/core/src/server/manager/manager-webpack.config.js
evanroane: true, the `g` is not needed

evanroane:
Feel free to merge this if you think a regex object is a better option, otherwise I'm just glad the bug is fixed 👍

evanroane:
I had actually thought to set refs in the template file to something nullish but still valid JS
```
addons.setConfig({
  refs: undefined,
});
```

and then in the `replace` call in the manager webpack can use the string `'undefined'` a lot of the quotation wrangling disappears that way

ndelangen:
That feels a lot more risky to me tbh, the `'{{refs}}'` makes it pretty clear to me there's something getting replaced, and you can search the codebase for where that might be happening.

ndelangen:
Thank you for the effort, time and help regardless @evanroane 

It's appreciated!

PR #10512, opened 2020-04-22
Addon-Docs: Fix Ember extractArgTypes default value

jguze:
Issue: #10511 

## What I did
Fixed it so the default value is being exposed as the expected object type, in the summary field.
Also added an `@defaultValue` snippet in the jsdoc for the welcome-banner.js component, just to see the value appear in the props table.


shilman (approved):
Also slightly buggy, will fix in followup PR

PR #10525, opened 2020-04-22
Addon-Docs: Fix Ember extractArgTypes

jguze:
Issue: #10524

## What I did
Modified the return type of `extractArgTypes`. Seems somewhere in the @next package, it now expects a list of items rather than an object with a `rows` property.

This also fixes the broken docs examples in the Ember live example.

You can now see the props:
image


ddoria921:
@jguze I ran into the same issue. I applied the same fix you did locally and it worked. Although for me I kept getting an empty row at the bottom of the prop table. Did you run into this? 
Screen Shot 2020-07-11 at 1 06 30 AM

shilman (approved):
LGTM. there's actually a bug, but will update it in a follow-on PR!

PR #10534, opened 2020-04-23
Addon-a11y: Code cleanup

tooppaaa:
## What I did

Removed the last getIframe call from a11y (even if it didn't really accessed the iFrame..)
Reused logic from addon-backgrounds

## How to test

Switch color emulation


shilman (approved):
LGTM!

PR #10539, opened 2020-04-24
Fix: Search stories

Vlad160:
Issue: #10519 

Search throws errors on the current release in prod mode. The issue is that for dev mode webpack uses filenames as values of modules found using `require.context`. But in prod mode, it uses numbers to minimize the size.

Submitting these changes to `master` because this bug is not reproducible of `next`. On the `next` branch, there is a check that `parameters.fileName` is a string.
However, these changes should be cherry-picked to `next`

## What I did

Use filename for `require.key()` instead of `require.resolve()`

## How to test

1. Bootstrap
2. Build `examples/official-storybook`
3. `cd ../storybook-static`
3. `npx http-server .`
4. Search in Strorybook


ndelangen (approved):

Vlad160:
Thanks @ndelangen

PR #10561, opened 2020-04-27
CLI: Clarify error when `sb init` fails on angular

yannbf:
Issue: #10549 

## What I did
Changed the error message to make the fix step more clear to understand.


shilman (approved):
LGTM!

PR #10571, opened 2020-04-28
React: Remove deprecated CRA preset

mrmckeb:
Issue: N/A

## What I did

Removed the deprecated CRA preset, now replaced by a standalone addon/preset.

## How to test

The in-built CRA preset should no longer work, and users should see a warning when using using Storybook with CRA and no standalone preset.


shilman (approved):
Looks great! And apparently works!!

PR #10588, opened 2020-04-29
feat(components): add ids to panels tab and content

yannbf:
This PR adds identifiers to the tab panel elements, it makes it easier to select them when needed (e.g. e2e tests)
Add @storybook/csf in the dependency list as well to access `sanitize`

image

image

### DIFF
--- a/lib/components/src/tabs/tabs.tsx
+++ b/lib/components/src/tabs/tabs.tsx
@@ -8,6 +8,7 @@ import React, {
   ReactNode,
 } from 'react';
 import { styled } from '@storybook/theming';
+import { sanitize } from '@storybook/csf';
 
 import { Placeholder } from '../placeholder/placeholder';
 import { FlexBar } from '../bar/bar';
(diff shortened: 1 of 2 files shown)


lib/components/src/tabs/tabs.tsx
```diff
@@ -158,26 +159,30 @@ export const Tabs: FunctionComponent<TabsProps> = memo(
...
-                  actions.onSelect(id);
-                }}
-                role="tab"
-                className={`tabbutton ${active ? 'tabbutton-active' : ''}`}
-              >
-                {typeof title === 'function' ? title() : title}
-              </TabButton>
-            ))}
+            {list.map(({ title, id, active, color }) => {
+              const tabTitle = typeof title === 'function' ? title() : title;
```
ndelangen, line 163: if title is a function, might it be a FunctionComponent?

ndelangen (approved):

lib/components/src/tabs/tabs.tsx
yannbf: as discussed, I just moved what was already there into a constant, so I could reuse it

PR #10599, opened 2020-04-30
CHANGE isToolshown logic so it hides toolbar if there's only (hidden) canvas

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10592

## What I did

Fixed a bug where the toolbar might be shown with nothing in it.


hipstersmoothie:
Will this also make storybook navigate to the first non hidden tab?

shilman (commented):
Not obvious to me how this corresponds to the original issue. Can you provide a test, or a comment on why this logic makes sense?

lib/ui/src/components/preview/preview.tsx
```diff
@@ -135,11 +135,13 @@ const Preview: FunctionComponent<PreviewProps> = (props) => {
     baseUrl,
     withLoader = true,
   } = props;
-  const { isToolshown } = options;
   const { getElements } = api;
 
   const tabs = useTabs(previewId, baseUrl, withLoader, getElements, story);
 
+  const isToolshown =
+    !(viewMode === 'docs' && tabs.filter((t) => !t.hidden).length < 2) && options.isToolshown;
```
ndelangen, line 143: When the viewMode is docs, we only show tabs
When the number of tabs is 1, there's no point in showing them
When we don't show anything, the toolbar should be collapsed

AFAIK this was exactly the issue in #10592

This code makes it so when there's only 1 tab shown in docsmode the toolbar collapses

shilman: I didn't repro #10592 but @hipstersmoothie 's issue is that "nothing is rendered". AFAICT this change renders even less than what was there before, so I don't understand how this could solve that.

hipstersmoothie: I can set up an example repo if neeed

shilman: Best is repro in official-storybook

matheo:
@ndelangen I've confirmed that one of my issues (https://github.com/storybookjs/storybook/issues/10294)
was caused by moving the docs panel to the left with:
```
addons.setConfig({
  previewTabs: {
    'storybook/docs/panel': { index: -1 },
  },
});
```

lib/ui/src/components/preview/preview.tsx
matheo: @ndelangen @shilman I've ran this locally and the only issue I see is that on initial load,
the MDX `docs-only` page that I select on `SET_STORIES` shows the toolbar like:

image

and after navigate another story with Canvas/Docs tab, and going back, it's not shown:

image

the tabs/reordering seems to work fine with `beta.22` :D
now I'd like to know how to set the `selectedTab` in a similar way as `selectedPanel` hehe

ndelangen:
I'm just gonna close this if those doesn't fix any issues people are having.

matheo:
@ndelangen this fixes an empty toolbar for docsOnly pages:

image

ndelangen:
@matheo would you have some time to make any changes to this PR you deem necessary?

PR #10605, opened 2020-05-01
doc: update readme of addon viewport about configure section more accurately

Brew-Brew:
Issue:

Set **defaultViewport** to **iphonex** or **iphone6** works if developers set viewports to [INITIAL_VIEWPORTS](https://github.com/storybookjs/storybook/blob/master/addons/viewport/src/defaults.ts). If follow the readme, the default viewports doesn't work

(because default viewport works when match with key, i had trouble when set viewport with following this readme)

## What I did
So, the example should be change more accurately.

## How to test


shilman (approved):
thanks!

PR #10612, opened 2020-05-01
FIX background color of the panel

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10597

## What I did
Panel content should be content background color.


shilman (commented):
Any idea what change broke this?

ndelangen:
I must have removed some div-wrapper I though was unnecessary in the past @shilman

PR #10614, opened 2020-05-01
CHANGE the maxDepth default to 25

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/9534

## What I did
increase the default maxDepth


shilman (approved):
WFM

PR #10623, opened 2020-05-02
CLI: Add automatic detection for svelte

yannbf:
Add detection for SVELTE projects.


shilman (approved):
LGTM -- good catch! 💯

PR #10638, opened 2020-05-04
Future - Rename Storybook DOM root IDs

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10403

## What I did

I changed `#root` => `#storybook-root`
I changed `#docs-root` => `#storybook-docs`


ndelangen:
Do we want to put this in v6.0.0?

If we're not doing this now, it'll be 7.0.0 at the earliest.

ndelangen:
@tmeasday @shilman WDYT, should we push this forward for 6.0.0 or delay for 7.0.0?

ndelangen:
@shilman this is something we pushed to 7.0 back in the day.
I resolved conflicts and it's ready to be reviewed again now.
Are we going to commit to doing this? I'm having a doubts, based on what I am about to describe:

@tmeasday this is a problem for chromatic it seems, likely because it's got `#root` hard-coded into it.
@ghengeveld perhaps this is something you could discuss with the team? Would it be easy to support this in such a way it's  backwards compatible in chromatic?

tmeasday:
It would be easy enough for chromatic to search for either id

ghengeveld:
If we are going to do this, please include a config option to use the old root ID (e.g. `legacyRootElement: true`) so that people who rely on it (e.g. for portals) have an escape hatch when it blows up on them.

ghengeveld:
I'm seriously in doubt whether we should want to do this at all. Going to be a painful transition, most likely.

ndelangen:
Why would rendering portals be influenced by this change @ghengeveld ?
I'd rather not introduce feature flags for things we want to remove inside of a breaking change release, that would mean introducing something new with the intention of removing it a year later.

I've always considered the dom-nodes stories get mounted on as an internal API.

The reason we want to change it, is because it's TOO generic, and users could potentially end up with multiple #root divs; resulting in possible bugs for them.

Other than chromatic, why would anything specifically target #root selector?

ndelangen:
What pain do you think users or us storybook maintainers would endure if we pushed this into 7.0?

ndelangen:
So the possible foreseeable pain for users is re seeing to portals that target #root.
That's an anti-pattern anyway, because that can't work for docs.

So I think breaking that is the right thing to do.

Another point is that we don't document what div target storybook renders to at all. The fact we target #root is an implementation detail that's not supposed to be leveraged by users.

I'd expect users to render portals to either document.body OR to use the canvasElement provided in docs OR wrap the story in a decorator setting up a selector to render to.

I think there should be a task to make targeting the current story container real easy for users.

ndelangen:
So we need 2 things to get this merged/ready for release:

- chromatic needs to be updated, otherwise we lose a very significant part of our testing strategy. @skitterm is this something you could discuss with the apps team?
- we need a change to the migration document @jonniebigodes is this something you could put on your task list?

Additionally we should have a task to create a real api (on the storyContext?) to get the selector for the container of the current story. @shilman that is probably something the CSF team (possibly myself?) will have to do. Let's chat about that soon.

tmeasday:
@ndelangen - the `StoryContext` already has `CanvasElement` which I think is what you want here: https://github.com/ComponentDriven/csf/blob/4566f4da20ea41a599548128cf993d1bae7d1d46/src/story.ts#L106

Also @jmhobbs or @codykaup would be the ones to talk about changing Chromatic's capture to handle a different `#root` element/

ndelangen:
@tmeasday @shilman @yannbf shall I merge this?

ndelangen:
@tmeasday could I get a review on this?

ndelangen:
@yannbf we need a minor change in the test-runner so it checks both the IDs.

PR #10659, opened 2020-05-05
Fix/error handling

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10577 https://github.com/storybookjs/storybook/issues/10664

when there's an error in one of the stories initially, the preview & sidebar will go in a perpetual loading-state & error doesn't show up in the preview

## What I did

- made sure that if an error occurs storybook preview renders an error.
- Re-use the UI from refs to display the error in sidebar
- Add the error that occurred to setStories event


ndelangen:
This is what an error now looks like when loading stories fails:
Screenshot 2020-05-06 at 09 22 38
Screenshot 2020-05-06 at 09 23 43

This is what it looks like when a story-file is loaded successfully, but the story throws an error being rendered:
Screenshot 2020-05-06 at 09 21 39

lib/api/src/modules/stories.ts
```diff
@@ -382,6 +385,38 @@ export const init: ModuleFn = ({
       (storiesHash[id] as Story).args = args;
       store.setState({ storiesHash });
     });
+
+    fullAPI.on(CURRENT_STORY_WAS_SET, function handleSetStory({ storyId, viewMode }) {
```
tmeasday, line 389: We already have events to indicate that configuring is done:
  A) In 6.0, the `SET_STORIES` event, which is guaranteed to run once when the configure process has run. I'm ok with adding `error` to it though if that makes sense.

  B) In 5.x the `STORIES_CONFIGURED` event, I'm not sure what runs when there is an error however.

lib/client-api/src/config_api.ts
```diff
@@ -27,6 +27,7 @@ export default class ConfigApi {
       this._storyStore.finishConfiguring();
     } catch (err) {
       this._storyStore.setError(err);
+      this._storyStore.failConfiguring(err);
```
tmeasday, line 30: Why can't we call `finishConfiguring()`? The store knows there is an error itself.

lib/client-api/src/story_store.test.ts
```diff
@@ -793,16 +793,18 @@ describe('preview.story_store', () => {
     });
   });
 
-  describe('RENDER_CURRENT_STORY', () => {
+  describe('STORY_ERRORED', () => {
     it('is emitted when setError is called', () => {
       const onRenderCurrentStory = jest.fn();
-      channel.on(Events.RENDER_CURRENT_STORY, onRenderCurrentStory);
+      channel.on(Events.STORY_ERRORED, onRenderCurrentStory);
```
tmeasday, line 799: This is semantically wrong. No story errored, there was a problem configuring the store.

We use `RENDER_CURRENT_STORY` here to tell the `StoryRenderer` to figure out what to render to the screen.

lib/client-api/src/story_store.ts
```diff
@@ -152,10 +152,25 @@ export default class StoryStore {
     this._configuring = true;
   }
 
+  failConfiguring(error: Error) {
```
tmeasday, line 155: We can just do this in `finishConfiguring` if `this._error` is set.

lib/client-api/src/story_store.ts
```diff
@@ -481,7 +496,10 @@ export default class StoryStore {
 
   setError = (err: ErrorLike) => {
     this._error = err;
-    if (this._channel) this._channel.emit(Events.RENDER_CURRENT_STORY);
+
+    if (this._channel) {
+      this._channel.emit(Events.STORY_ERRORED, { message: err.message, stack: err.stack });
```
tmeasday, line 501: We shouldn't emit anything at this point I don't think.

tmeasday (changes requested):
I'm OK with the idea of including `errors` in `SetStoriesPayload` (if we need to know about it on the manager side).

But I don't like the changed event behaviour. Let's avoid incremental changes to events like this, that's how we ended up with the complex mess of incoherent events we had before. 

Let's be careful about what events are emitted when and document things. I tried to do that already although it could be more comphrehensive I think: https://github.com/storybookjs/storybook/blob/a9fc57fe490e5befcb4bdd4f77f6897d283d8263/lib/core/README.md#events-on-startup

lib/api/src/lib/stories.ts
```diff
@@ -107,6 +107,7 @@ export interface StoriesRaw {
 export interface SetStoriesPayload {
   v?: number;
   stories: StoriesRaw;
+  error?: Error;
```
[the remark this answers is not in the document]
ndelangen: ow true!

lib/client-api/src/story_store.ts
ndelangen: ok, great!

yannbf:
I don't have 100% context so can't help that much in code, but I ran the official storybook in a few scenarios:
1 - with a wrong glob in main.js:
```
stories: ['../../lib/ui/src/**/Wrong.stories.(js|tsx|mdx)']
```

2 - with a correct glob but everything commented out in the story:
```
stories: ['../../lib/ui/src/**/Sidebar.stories.(js|tsx|mdx)']
```

3 - with a correct glob, exporting a single story but not default exporting anything

Scenarios 1, 2 and 3 leave me in this screen:
image

4 - with a correct glob + the file only has a `throw new Error('error!!!');`
5 - with a correct glob + the file has a story that throws an error

Scenarios 4 and 5 work as you mentioned in the PR.

And a small observation, that you can ignore if you want, the component `/components/sidebar/RefBlocks -> ErrorBlock` has the following text:
> Ow now! something went wrong loading this storybook

Shouldn't it be 

> Oh no! Something went wrong loading this storybook.

lib/ui/src/components/sidebar/Sidebar.tsx
```diff
@@ -75,10 +77,22 @@ const useSearchResults = (refsList: [string, RefType][], filter: string) => {
   return { total: refsTotal || 0, list: refsLengths };
 };
 
-const useCombination = (stories: StoriesHash, refs: Refs) => {
+const useCombination = (
+  stories: StoriesHash,
+  ready: boolean,
+  error: Error | undefined,
```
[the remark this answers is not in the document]
ndelangen: This is not an interface?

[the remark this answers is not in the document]
yannbf: Indeed, but the optional notation can also be used in parameters. 
However, in this case it won't actually work unless `error` is the last parameter. Optionals have to be placed at the end of the parameters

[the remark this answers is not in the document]
yannbf: it was just a suggestion, feel free to ignore it!

ndelangen:
I'll investigate the scenarios:

- **all stories load fine, but they don't export/define any stories**
  should show a message in the sidebar
- **story-files are missing (glob matches nothing)**
  should show a warning in terminal?

lib/ui/src/components/preview/preview.tsx
```diff
@@ -57,7 +61,8 @@ const createCanvas = (id: string, baseUrl = 'iframe.html', withLoader = true): A
           ]);
 
           const isLoading = !!(
-            (storyId && !story) ||
+            (!story && !storiesFailed) ||
+            (!story && storiesConfigured === false) ||
             (story && story.refId && refs[refId] && !refs[refId].ready)
```
shilman, line 66: simpoler: `(refId && refs[refId] && ! refs[refId].ready)`

lib/client-api/src/story_store.ts
```diff
@@ -481,7 +484,14 @@ export default class StoryStore {
...
     this._error = err;
-    if (this._channel) this._channel.emit(Events.RENDER_CURRENT_STORY);
+
+    if (this._channel) {
+      // If the selection is set while configuration is in process, we are guaranteed
+      // we'll emit RENDER_CURRENT_STORY at the end of the process, so we shouldn't do it now.
+      if (!this._configuring) {
+        this._channel.emit(Events.RENDER_CURRENT_STORY);
+      }
+    }
```
tmeasday, line 494: I don't think it is allowed that you set error outside of configuration. In fact we should probably throw if you do that.

ndelangen: AFAIK the setError is also called when a story throws?

tmeasday: Who by?

ndelangen:
Here's what happens if a ref/frame results in a successful SET_STORIES, but the contents are empty:

Screenshot 2020-05-07 at 17 31 45

tmeasday:
Maybe we can make a better UI for the no stories case? WDYT?

ndelangen:
Screenshot 2020-05-08 at 20 35 56

new UI

tmeasday:
Looking good @ndelangen. I was actually thinking something similar in the preview pane. Wdyt?

tmeasday:
@domyen want to weigh in?

PR #10683, opened 2020-05-08
cli: add legacy flag that transpiles some node_modules

tooppaaa:
Issue: #10674 #10670 #10486 

## What I did

Add a legacy flag that take precedence over the dll and transpiles some node_modules detected to not be es5.

This is a suggestion, any idea is welcome !

## How to test

- Launch IE 11 with --compatibility=legacy flag

### DIFF
--- a/lib/core/src/server/manager/manager-webpack.config.js
+++ b/lib/core/src/server/manager/manager-webpack.config.js
@@ -10,6 +10,7 @@ import VirtualModulePlugin from 'webpack-virtual-modules';
 
 import themingPaths from '@storybook/theming/paths';
 import uiPaths from '@storybook/ui/paths';
+import { logger } from '@storybook/node-logger';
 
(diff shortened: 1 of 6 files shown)


lib/core/src/server/manager/manager-webpack.config.js
```diff
@@ -42,6 +45,17 @@ export default ({
     encoding: 'utf8',
   });
 
+  if (legacy) {
+    logger.info("=> Transpiling some node_modules, you're brave..");
+  }
+
+  // Don't know why webpack resolves esm modules...
+  const resolveExtension = legacy
```
tooppaaa, line 53: Any idea why I had to do this ? https://github.com/webpack/webpack/issues/5756

overlayscrollbar is the source of this hack..

tooppaaa:
Closed in favor of #10725

PR #10694, opened 2020-05-08
WIP react-refresh attempt

tooppaaa:
Issue:
#10602

## What I did

Try to setup react-refresh 

## How to test

- use react refresh story

Current state:

- [x] Story is loading
- [ ] Story refreshes with state maintained

image


tooppaaa:
image
I've got the feeling that it did "work" but because the entry is actually a require.context I do not think it can perform the fast refresh we expect

malixsys:
FYI, we use `customize-cra`, so I tried using info from this blog:
https://harrietryder.co.uk/blog/storybook-with-typscript-customize-cra
which should _just work™️_, but now SB can't load the stories anymore...

sibelius:
@pmmmwh any idea why this is not working well?

pmmmwh:
> @pmmmwh any idea why this is not working well?

Other than Webpack config issues - which I think is what the error is showing - I think it has something to do with Storybook's story format. Since FR depends on analysing exports, effectively each file will have a default export that is not react related, and it would propagate upwards which means to root and bail out?

PR #10703, opened 2020-05-09
cli: web-components stories homogenization

tooppaaa:
Issue:

We need similar stories after generation to run e2e against

Part of #10702


lib/cli/src/generators/WEB-COMPONENTS/template/stories/0-Welcome.stories.js
```diff
@@ -0,0 +1,49 @@
...
+    >
+      Writing Stories
+    </a>
+    section in our documentation.
+  </p>
+  <p class="note">
+    <b>NOTE:</b> <br />
+    Have a look at the <span class="inline-code">.storybook/webpack.config.js</span> to add webpack
+    loaders and plugins you are using in this project.
+  </p>
```
gaetanmaisse, line 44: I think we can remove that as I'm not sure there will be a `.storybook/webpack.config.js`

lib/cli/src/generators/WEB-COMPONENTS/index.js
```diff
@@ -15,17 +14,18 @@ export default async (npmOptions, { storyFormat }) => {
     '@storybook/web-components',
     '@storybook/addon-actions',
     '@storybook/addon-links',
+    '@storybook/addons',
```
gaetanmaisse, line 17: Is `@storybook/addons` really needed directly as it's already a dep of `@storybook/addon-actions` and `@storybook/addon-links`? Maybe @ndelangen have some info about that

gaetanmaisse (approved):

lib/cli/src/generators/WEB-COMPONENTS/template/stories/0-Welcome.stories.js
yannbf: You are right, the template does not contain such file:
image

However it's nice to tell people that if they add a `webpack.config.js` file it will be picked up, so maybe just rephrase that? Also, that sentence is in every framework's template, so if we were to make a change, it has to be synced with the others.

@shilman what's your take on this?

lib/cli/src/generators/WEB-COMPONENTS/template/stories/1-Button.stories.js
```diff
@@ -0,0 +1,42 @@
...
+`;
+
+export const Emoji = () => html`
+  <button @click=${action('clicked')}>
+    😀 😎 👍 💯
+  </button>
+`;
+
+Emoji.story = {
+  parameters: { notes: 'My notes on a button with emojis' },
```
yannbf, line 22: I think we should get rid of `notes` for every framework. It's deprecated or was even removed already.
@ndelangen what do you think?

gaetanmaisse: You're right @yannbf, `notes` has been moved to the repo `storybookjs/deprecated-addons` so no 6.0 version will be released.

lib/cli/src/generators/WEB-COMPONENTS/template/stories/0-Welcome.stories.js
shilman: AFAIK the current recommendation is to override `webpackFinal` in `main.js`, so we should probably update all the docs accordingly. `webpack.config.js` is still supported (for now, not sure how long we plan to keep this? But we should be  directing everybody towards the recommendation. @ndelangen please feel free to correct if I got any of this wrong.

lib/cli/src/generators/WEB-COMPONENTS/template/stories/1-Button.stories.js
shilman: Agree we should get rid of `notes` and pick a better example. Maybe `backgrounds`?

lib/cli/src/generators/WEB-COMPONENTS/template/stories/0-Welcome.stories.js
ndelangen: > the current recommendation is to override webpackFinal in main.js

indeed

lib/cli/src/generators/WEB-COMPONENTS/template/stories/1-Button.stories.js
ndelangen: I agree, we should leave it out for now.

lib/cli/src/generators/WEB-COMPONENTS/index.js
ndelangen: If it works without, let's remove it!

yannbf: @tooppaaa the other points have been resolved and will be done in a different PR, but can you check this one? then we're good to merge!

yannbf (changes requested):
@tooppaaa  the other points have been resolved and will be done in a different PR, but can you check the one about removing `@storybook/addons`? then we're good to merge!

gaetanmaisse:
Tested ✅ `@storybook/addons` is already a dependency of all `@storybook/addon-XXX` packages so there is no need to add it to the root project directly.

lib/cli/src/generators/WEB-COMPONENTS/index.js
gaetanmaisse: https://github.com/storybookjs/storybook/pull/10703#issuecomment-628227892

ndelangen (approved):

PR #10704, opened 2020-05-09
cli: svelte stories homogenization

tooppaaa:
Issue:

We need similar stories after generation to run e2e against

Part of #10702


gaetanmaisse (approved):
LGTM ✅ 

Some of the framework apps have `Welcome` and `Button` components, used in template stories, in `@storybook/FRAMEWORK/demo`, and some others define these components in CLI templates directly. I think it could be great to pick a way and be consistent across all apps. Personally I vote for components in `@storybook/FRAMEWORK/demo`. What's your opinion on that?

yannbf:
> LGTM ✅
> 
> Some of the framework apps have `Welcome` and `Button` components, used in template stories, in `@storybook/FRAMEWORK/demo`, and some others define these components in CLI templates directly. I think it could be great to pick a way and be consistent across all apps. Personally I vote for components in `@storybook/FRAMEWORK/demo`. What's your opinion on that?

I agree with that. We should be as consistent as possible across all frameworks. Shipping example components might be helpful for beginners but might also be considered noise which people would delete right away. I especially like that with the `Welcome` example shipped I can just change the text or use it as inspiration for a Welcome component, which might be nice for newcomers in my project to see as they open storybook.

But I guess I'd go for `@storybook/FRAMEWORK/demo` so we keep consistency and make maintenance simpler

@shilman @ndelangen which way do you think should be the way to go?

lib/cli/src/generators/SVELTE/template-csf/stories/welcome.svelte
```diff
@@ -0,0 +1,37 @@
...
+    border: 1px solid #eae9e9;
+    border-radius: 4px;
+    background-color: #f3f2f2;
+    color: #3a3a3a;
+  }
+</style>
+
+<main>
+  <h1>Welcome to storybook</h1>
+  <p>This is a UI component dev environment for your app.</p>
```
yannbf, line 26: Should we keep consistency on the content of the Welcome component? There is useful info in other components that I don't see here, such as telling users about hot module reloading, a link to storybook repo and explanation over the webpack config file.

yannbf:
Also about keeping things in sync, some templates generate stories under `/src/stories` and others under `/stories`

ndelangen (approved):

PR #10720, opened 2020-05-10
CLI: Remove CRA fixtures from Yarn 2 tests run

gaetanmaisse:
## What I did

When testing Yarn 2 compatibility of Storybook CLI, we are using Yarn 2 in fixture directories inside a Yarn 1 workspace. To access local versions of `@storybook` packages we use Yarn 2 `portal` [protocol](https://yarnpkg.com/features/protocols#table). 

However, it looks like CRA is not playing well with `portal` protocol and is throwing this kind of error: `Relative imports outside of src/ are not supported.`.

As new E2E tests will be soon available - see https://github.com/storybookjs/storybook/pull/10702, I just removed the CRA tests (that are always 🔴) instead of trying to hack them a bit more.

### DIFF
--- a/lib/cli/test/fixtures/react_scripts_ts/package.json
+++ b/lib/cli/test/fixtures/react_scripts_ts/package.json
@@ -24,13 +24,13 @@
     "extends": "react-app"
   },
   "dependencies": {
-    "@types/jest": "24.0.19",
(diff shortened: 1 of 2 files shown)


shilman (approved):
LGTM!!!

PR #10725, opened 2020-05-11
core: transpile minimum node_modules

tooppaaa:
Issue: Famous IE 11 #10674 #10670 #10486

## What I did

Transpile minimum node_modules.

This is a slighlty different approach than #10683. 

## How to test

- IE 11


lib/core/src/server/config/utils.js
```diff
@@ -25,7 +25,16 @@ const projectRoot = () => {
 
 export const includePaths = [projectRoot()];
 export const nodeModulesPaths = path.resolve('./node_modules');
-export const excludePaths = [/node_modules/];
+
+// Run `npx are-you-es5 check lib/core -r`
```
ndelangen, line 29: How long does it take to run this command, could we potentially run it before running storybook on the user's machine?

tooppaaa: It's pretty quick but it's only CLI and the output is as follow


$ npx are-you-es5 check lib/core -r
npx : 4 installé(s) en 2.09s
❌ @storybook/node-logger is not ES5
⚠️ @types/node-fetch was not checked because no entry script
 was found
❌ better-opn is not ES5
❌ boxen is not ES5
❌ chalk is not ES5
❌ commander is not ES5
❌ find-cache-dir is not ES5
❌ find-up is not ES5
❌ fs-extra is not ES5
❌ json5 is not ES5
❌ node-fetch is not ES5
❌ pkg-dir is not ES5
⚠️ react-dev-utils was not checked because no entry script w
as found
❌ resolve-from is not ES5
❌ semver is not ES5


Babel-loader exclude regex:
/[\\/]node_modules[\\/](?!(@storybook\/node-logger|better-opn|boxen|chalk|commander|find-cache-dir|find-up|fs-extra|json5|node-fetch|pkg-dir|resolve-from|semver)[\\/])/

ndelangen: if it's expensive, we should make it easier to update this code using a script?

lib/core/src/server/config/utils.js
```diff
@@ -25,7 +25,16 @@ const projectRoot = () => {
...
+// to get this regexp
+export const excludePaths = [
+  /[\\/]node_modules[\\/](?!(@storybook\/node-logger|are-you-es5|better-opn|boxen|chalk|commander|find-cache-dir|find-up|fs-extra|json5|node-fetch|pkg-dir|resolve-from|semver)[\\/])/,
```
shilman, line 32: why is `@storybook/node-logger` in this list?

tooppaaa: I unfortunately don't know. I was surprised as well but also didn't wanted to alter the output of the regexp

ndelangen: Because it only is run in node, and so ES6 is fine?

ndelangen: it should not be bundled into client-code ever

tooppaaa: being in this regexp doesn't mean you'll get bundled. It's only "If you're being bundled and going though babel, you'll get transpiled"

ndelangen: of course, the `are-you-es5` just detects the code being ES6 and adds it to the list.

tooppaaa:
Can't reproduce CLI error locally, can someone try on a mac ? :(

lib/core/src/server/preview/iframe-webpack.config.js
```diff
@@ -141,6 +142,26 @@ export default async ({
...
+                    { shippedProposals: true, useBuiltIns: 'usage', corejs: '3' },
+                  ],
+                ],
+                plugins,
+              },
+            },
+          ],
+          include: [/node_modules/],
+          exclude: excludePaths,
+        },
```
ndelangen, line 164: this is **always** added?

ndelangen (changes requested):

lib/core/src/server/manager/babel-loader-manager.js
```diff
@@ -16,18 +17,13 @@ export default () => ({
...
-          [require.resolve('@babel/plugin-proposal-class-properties'), { loose: true }],
-          require.resolve('@babel/plugin-proposal-export-default-from'),
-          require.resolve('@babel/plugin-syntax-dynamic-import'),
-          [
-            require.resolve('@babel/plugin-proposal-object-rest-spread'),
-            { loose: true, useBuiltIns: true },
-          ],
-          require.resolve('babel-plugin-macros'),
+          ...plugins,
+          // Shoud only be done on manager to to ember
```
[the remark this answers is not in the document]
tooppaaa: image
I don't know much about ember, just found that it is fine without that transform.
It can't be in the common plugins

lib/core/src/server/preview/iframe-webpack.config.js
tooppaaa: Yes, It look scary due to the include, but the exclude paths is excluding all but es6 modules.
This doesn't impact performances on startup

lib/core/src/server/common/babel.js
```diff
@@ -11,6 +11,26 @@ function createProdPresets() {
   ];
 }
 
+export const plugins = [
+  require.resolve('@babel/plugin-transform-shorthand-properties'),
+  require.resolve('@babel/plugin-transform-block-scoping'),
+
+  // [require.resolve('@babel/plugin-proposal-class-properties'), { loose: true }],
```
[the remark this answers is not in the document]
tooppaaa: my bad

lib/core/src/server/preview/iframe-webpack.config.js
ndelangen: I wish loader/rules could be given names or id's to tell them apart.

I think we need a comment here, telling what this loader is for.

tooppaaa: I can extract it to a es6-transform-loader

lib/core/src/server/config/utils.js
```diff
@@ -25,7 +25,16 @@ const projectRoot = () => {
...
+  // Never transpile again the following:
+  /[\\/]node_modules[\\/]webpack/,
+  /[\\/]node_modules[\\/]core-js/,
```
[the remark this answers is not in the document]
ndelangen: I'm guessing the include will not match anything in yarn2?

lib/core/src/server/manager/babel-loader-manager.js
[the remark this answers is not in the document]
ndelangen: ok

ndelangen (approved):

lib/core/src/server/config/utils.js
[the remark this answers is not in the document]
tooppaaa: As @gaetanmaisse once said: 

> it will not be worse than what we have today

PR #10753, opened 2020-05-12
Composition: Zero-config composition from dependencies

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10528

## What I did
Added the feature

### DIFF
--- a/lib/core/src/server/manager/manager-config.js
+++ b/lib/core/src/server/manager/manager-config.js
@@ -1,6 +1,32 @@
 import path from 'path';
+import fs from 'fs-extra';
+import findUp from 'find-up';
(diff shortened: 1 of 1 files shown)


tooppaaa (approved):

shilman (approved):
This looks great! And simple! I assume versioning and authentication will come separately?

tmeasday:
Can we add tests for this somehow @ndelangen ?

ndelangen:
auth & versioning is just part of the normal refs loading code in lib/ui @shilman 

@tmeasday I tested this locally of course. But I don't have any proper published packages with the storybook field in their `package.json`.

I can add some unit-tests possibly.

PR #10779, opened 2020-05-14
ADD a loglevel option for dev mode in CLI

ndelangen:
The loglevel cli options doesn't exists for dev-mode.

I think it should exist.

This doesn't add the feature, it just adds the cli-flag.

We're getting quite a few 'invalid cli option' errors, reported by users, because they try to use this cli-flag for both dev & build mode.

They expect this cli-flag to work for both.


shilman (commented):
Why would we add the flag without adding the feature?

ndelangen:
Because at least this means adding `--loglevel <something>` doesn't make storybook unable to run.

Currently if you pass in this flag, storybook will refuse to run at all.

ndelangen:
superseded by https://github.com/storybookjs/storybook/pull/10370 for 6.0.0

PR #10785, opened 2020-05-14
ADD checksums so building storybook is faster

ndelangen:
Issue:

`yarn bootstrap --build` is dumb it regenerates the dist, even if the src hasn't changed at all

## What I did
I added a checksum mechanism

The packages will now only be "prepared" if the src folder has any changes.


scripts/prepare.js
```diff
@@ -1,18 +1,51 @@
...
+
+  return fs.writeJSON(packagePath, { ...package, storybook: { checksum } }, { spaces: 2 });
+}
+
+async function compareChecksum(info) {
+  const { checksum: existing } = info.package.storybook || {};
+
+  const future = await createChecksum();
+
+  return existing === future ? existing : future;
```
yannbf, line 48: Will this equality check work with comparing objects? Won't you need to use something like `isDeepStrictEqual`?

scripts/prepare.js
```diff
@@ -28,48 +61,83 @@ const ignore = [
...
+    return Promise.resolve();
+  }
+  console.log(chalk.gray(`Skipped: ${chalk.bold(`${info.package.name}@${info.package.version}`)}`));
+};
 
-console.log(chalk.gray(`Built: ${chalk.bold(`${packageJson.name}@${packageJson.version}`)}`));
+run().catch((e) => {
+  console.error(e);
+  process.exitCode = 1;
+  // throw e;
```
yannbf, line 142: is this comment necessary?

yannbf (commented):
This is a game changer 🙌

scripts/prepare.js
ndelangen: it's not comparing objects it's comparing strings AFAIK.

yannbf: I assumed it was an object because of this line `const { checksum: existing } = info.package.storybook || {};`

ndelangen: info.package.storybook.checksum is a string, I call this string "existing"

ndelangen:
I changed the location of the checksums (no longer in package.json) thanks for the suggestion @tooppaaa 

I changed the reset script so it no longer removes the dist folder

ndelangen:
Thank you for the assist @KyleTryon !

It seem the bigger instance fixes it. I'm going to try a few optimizations

scripts/run-reset.ts
```diff
@@ -7,25 +7,25 @@ const logger = console;
...
         const [, uri] = i.match(/Would remove (.*)$/) || [];
 
         if (uri) {
           if (
+            uri.match(/dist/) ||
+            uri.match(/.vscode/) ||
+            uri.match(/.idea/) ||
+            uri.match(/reset.log/)
+          ) {
+            // ignore, don't remove these files
```
yannbf, line 27: I think you can extract all these uri match assertions to variables and as a result you both have more readable code plus you can re-structure these conditionals to be simpler. This empty if is a bit weird

scripts/utils/compile-babel.js
```diff
@@ -30,37 +29,42 @@ function getCommand(watch) {
...
+    return false;
   }
-}
+};
 
-function babelify(options = {}) {
-  const { watch = false, silent = true, errorCallback } = options;
+async function babelify(options = {}) {
+  const { watch = false, silent = !watch } = options;
+  const [src] = await Promise.all([exists('src')]);
```
yannbf, line 48: why not `const src = await exists('src');` ?

yannbf (changes requested):
Checked out and tried in my machine, works wonderfully!
However I'd HIGHLY recommend running this in a windows machine before merging. I can try that tomorrow and let you know!

My change request is simple, related to `build-package.js` please consider it and let me know if you need a hand 👍

lib/api/src/version.ts
```diff
@@ -1 +1,11 @@
-export const version = '6.0.0-beta.12';
+/* eslint-disable import/no-extraneous-dependencies */
+// @ts-ignore
+import preval from 'preval.macro';
+
+export const version = preval`
```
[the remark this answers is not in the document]
ndelangen: This is executed at build-time, the resulting bundle will only contain a string. And it's all automatic.

scripts/prepare.ts
```diff
@@ -0,0 +1,212 @@
+/* eslint-disable no-console */
+import path from 'path';
+import * as shell from 'shelljs';
+import chalk from 'chalk';
+import fse from 'fs-extra';
+import { PackageJson } from 'type-fest';
```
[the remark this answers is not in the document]
ndelangen: Never makes it into a bundle we ship?

scripts/prepare.ts
```diff
@@ -0,0 +1,212 @@
...
+  const files = info.filter(({ filePath, isDir }) => {
+    // Do not remove folder
+    // And do not clean anything for @storybook/cli/dist/generators/**/template* because these are the template files
+    // that will be copied to init SB on users' projects
+    if (isDir || /generators\/.+\/template.*/.test(filePath)) {
+      return false;
+    }
+
+    // Remove all copied TS files (but not the .d.ts)
+    if (/\.tsx?$/.test(filePath) && !/\.d\.ts$/.test(filePath)) {
```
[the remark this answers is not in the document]
ndelangen: it is?

scripts/build-package.js
```diff
@@ -126,44 +140,45 @@ function run() {
...
-            const command = `concurrently --kill-others "${watchTsc}" "${watchBabel}"`;
-            spawn(command);
-          };
-
-          runWatchMode();
-        } else {
-          spawn(`lerna run prepare --scope "${glob}"`);
+          let confirmation = true;
+
+          if (packageNames.length > 5) {
```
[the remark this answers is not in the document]
ndelangen: not a bad merge at all. I found creating the globs is more error prone then just listing all packages in separate flags.

lib/cli/package.json
```diff
@@ -52,8 +53,8 @@
     "fs-extra": "^9.0.0",
     "get-port": "^5.1.1",
     "inquirer": "^7.0.0",
+    "json5": "^2.1.3",
```
tooppaaa, line 56: Order was right before, final reference of this in Meteor generator. Worth removing it !

scripts/prepare.ts
```diff
@@ -0,0 +1,169 @@
...
+    package: await fse.readJSON(packagePath),
+    checksum: await fse.readJSON(checksumPath),
+  };
+}
+
+async function createChecksum({ modulePath }: Info) {
+  const { hash } = await hashElement(path.join(modulePath, 'src'), {
+    folders: { exclude: ['tests', '__mocks__', '__tests__'] },
+    files: {
+      include: ['*.js', '*.jsx', '*.ts', '*.jsx', '*.json'],
```
tooppaaa, line 48: **jsx is here twice. Missing tsx.**

Random question, is worth to create 2 hashes, one for src, one for ts, tsx which would only do tsc when this second hash changes.

It may add another level of complexity to this and maybe not needed !

tooppaaa (changes requested):
Only the hash include extensions to update.
Awesome otherwise !

Issue #10007, opened 2020-03-01
Custom addon not working on Windows.


jamime:
**Describe the bug**
I have a custom addon loaded from my `main.js` file. This works fine on Mac but fails on Windows with "Cannot use import statement outside a module". Looking at `@storybook\core\dist\server\presets.js`, it's probably `node` related rather than `storybook` but I wanted to check here first.

```sh
$ npm start

> carbon-react@13.6.0 start C:\Users\IEUser\carbon
> start-storybook -p 9001 -s .assets -c .storybook

info @storybook/react v5.3.4
info
info => Loading static files from: C:\Users\IEUser\carbon\.assets .
info => Loading presets
WARN   Failed to load preset: "C:\\Users\\IEUser\\carbon\\.storybook\\theme-selector\\register" on level 1
ERR! C:\Users\IEUser\carbon\.storybook\theme-selector\register.js:1
ERR! import React, { memo, useState } from 'react';
ERR! ^^^^^^
ERR!
ERR! SyntaxError: Cannot use import statement outside a module
ERR!     at wrapSafe (internal/modules/cjs/loader.js:1072:16)
ERR!     at Module._compile (internal/modules/cjs/loader.js:1122:27)
ERR!     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
ERR!     at Module.load (internal/modules/cjs/loader.js:1002:32)
ERR!     at Function.Module._load (internal/modules/cjs/loader.js:901:14)
ERR!     at Module.require (internal/modules/cjs/loader.js:1044:19)
ERR!     at require (internal/modules/cjs/helpers.js:77:18)
ERR!     at interopRequireDefault (C:\Users\IEUser\carbon\node_modules\@storybook\core\dist\server\presets.js:164:18)
ERR!     at loadPreset (C:\Users\IEUser\carbon\node_modules\@storybook\core\dist\server\presets.js:174:22)
ERR!     at C:\Users\IEUser\carbon\node_modules\@storybook\core\dist\server\presets.js:230:20
ERR!     at Array.reduce (<anonymous>)
ERR!     at loadPresets (C:\Users\IEUser\carbon\node_modules\@storybook\core\dist\server\presets.js:229:18)
ERR!     at loadPreset (C:\Users\IEUser\carbon\node_modules\@storybook\core\dist\server\presets.js:194:18)
ERR!     at C:\Users\IEUser\carbon\node_modules\@storybook\core\dist\server\presets.js:230:20
[...]

PR #10786, opened 2020-05-14
[fix] - Detect local addons for windows machines

RuneKR:
Issue: Local addons do not work on windows see [issue](https://github.com/storybookjs/storybook/issues/10007)

## What I did
I updated the local compare function to use the fs package to check if the file exists. Seemed to be the most simple solution. Wrong local paths to files that do not exists will now be interpreted as not being local files.

## How to test
- Use a local addon on a windows machine

## Note
Would like to get this change into a 5.x if possible. Can do another PR on that.

### DIFF
--- a/lib/core/src/server/presets.js
+++ b/lib/core/src/server/presets.js
@@ -1,6 +1,7 @@
 import dedent from 'ts-dedent';
 import { join } from 'path';
 import { logger } from '@storybook/node-logger';
+import fs from 'fs'
 import { resolveFile } from './utils/resolve-file';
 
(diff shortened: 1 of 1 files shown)


ndelangen (approved):

ndelangen:
That's actually a smart solution!

RuneKR:
Hi @ndelangen i realised i missed updating the test https://github.com/RuneKR/storybook/blob/next/lib/core/src/server/presets.test.js#L400

Do you want me to fix that and also do another PR for 5.x? I assume this PR got released in 6.x

PR #10819, opened 2020-05-18
FIX refs not authenticating

ndelangen:
## What I did

When refs are behind a authenticated page, the checkRef api should:

- add credentials
- pass content-type


lib/api/src/modules/refs.ts
```diff
@@ -128,10 +128,21 @@ export const init: ModuleFn = ({ store, provider, fullAPI }) => {
       };
 
       const [stories, metadata] = await Promise.all([
-        fetch(`${url}/stories.json`)
+        fetch(`${url}/stories.json`, {
+          headers: {
+            'Content-Type': 'application/json',
```
tmeasday, line 133: It's the `Accept` header you want: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

`Content-Type` would be if you were POST-ing data.

tmeasday (changes requested):

tmeasday (approved):

PR #10821, opened 2020-05-18
Vue: Add vue-cli example and remove babel-preset-vue

graup:
Issue: #10790 and #4475

## What I did
1. Added new example app, `vue-cli`. This is supposed to be a minimum working example using all the latest bells and whistles (typescript, class components, decorators). It is very similar to what someone gets when they run `vue create`.
2. I also FINALLY removed babel-preset-vue from storybook. As discussed in #4475 that is mostly obsolete and only needed for older apps. It is however needed for our vue-kitchen-sink example, so that gave me an immediate case to create a migration path, which is also included in this PR.
3. These two changes are in the same PR because 1 depends on 2.
4. Removed `localIdentName` option in vue-kitchen-sink webpack. It was throwing a warning and doesn't seem to be needed anyway.

## How to test

- example/vue-ci: `yarn storybook` and `yarn serve`
- example/vue-kitchen-sink: `yarn storybook` and `yarn dev`

## TODO
- Document migration path
- Change setup guide for vue (it's much easier now!) and remove references to babel-reset-vue, e.g. from [here](https://storybook.js.org/docs/presets/writing-presets/)

### DIFF
--- a/app/vue/package.json
+++ b/app/vue/package.json
@@ -46,7 +46,6 @@
   "devDependencies": {
     "@types/node": "^13.13.4",
     "@types/webpack": "^4.41.12",
-    "babel-preset-vue": "^2.0.2",
     "vue": "^2.6.8",
     "vue-loader": "^15.7.0",
     "vue-template-compiler": "^2.6.8"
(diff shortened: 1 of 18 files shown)


mrmckeb (approved):

shilman (approved):
Thanks @graup !! Merging 😍

PR #10825, opened 2020-05-18
Fix: upgrade and add some missing dependencies in core, router, api

gaetanmaisse:
## What I did

When building `@storybook/api` we are generating types that are based on `@types/reach__router` (see in dist/index.d.ts: `location: import("@reach/router").WindowLocation<{}>`).
And as in the monorepo we have the latest version of `@types/reach__router` we are using generic version of `WindowLocation` added by https://github.com/DefinitelyTyped/DefinitelyTyped/commit/410945551d356ddda19d268df0ad85ad6e8959f6

So we need to update the required version of `@types/reach__router` in the dependency otherwise users can try to run SB with an old version of `@types/reach__router` (like I did when upgrading to the latest beta 😁) and face the following error 💥 :
```
ERROR in /.../my-project/node_modules/@storybook/api/dist/index.d.ts(75,19):
TS2315: Type 'WindowLocation' is not generic.
```

I also updated the version in `@storybook/router` to be consistent across the different packages.

-- 

Also, added some missing dependencies in `@storybook/core`, making SB fails with Yarn 2. Related to https://github.com/storybookjs/storybook/pull/10725.

## How to test

- Everything should be 🟢 on CI, even Yarn 2 :wow:


yarn.lock
```diff
@@ -4794,14 +4794,7 @@
...
 
-"@types/puppeteer@*":
-  version "2.1.0"
-  resolved "https://registry.yarnpkg.com/@types/puppeteer/-/puppeteer-2.1.0.tgz#31367580654632f87f86df565f1bde0533577401"
-  integrity sha512-QIRQXl0VaSgnwOZ1LwxD321Tfb1jLOzCWuF2BrwjEkWq2IhxSicPOddUywLV7dRSO6mcU4sWKRdoGdci6gk0Aw==
-  dependencies:
-    "@types/node" "*"
-
-"@types/puppeteer@^2.0.0":
+"@types/puppeteer@*", "@types/puppeteer@^2.0.0":
```
gaetanmaisse, line 4797: Why not 🤷

ndelangen (approved):

PR #10838, opened 2020-05-19
UI: Don't emit storyChange event when viewmode not renderable by preview

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/10551

## What I did
Added a filter, so the event is only emitted if the viewMode matches `/(story|docs)/`.


shilman (approved):
WFM!

PR #10845, opened 2020-05-20
Fix/ref auth

ndelangen:
Issue: https://github.com/storybookjs/storybook/pull/10819

## What I did
The solution in the mentioned PR is incomplete:
This adds:

- startInjected is changed to a enum `type`
- refs start as `unknown` and then based on whether `stories.json` or `metadata.json` can be loaded change into a `lazy` or `auto-injected`
- `iframe.html` is attempted to be fetched with and without credentials. If it works without, we load `stories.json` without. if it only works with credentials, we fetch `stories.json` with it. If both fail, we show an error.
- improved the displayed error
- this adds an readable error when the ref returns a 404


tmeasday (changes requested):

lib/api/src/modules/refs.ts
```diff
@@ -127,32 +138,62 @@ export const init: ModuleFn = ({ store, provider, fullAPI }) => {
...
-          .catch(() => false)
-          .then(handler),
-        fetch(`${url}/metadata.json`, {
-          headers: {
-            Accept: 'application/json',
-          },
+      const loadedData: { error?: Error; stories?: StoriesRaw } = {};
+
+      const [included, omitted] = await allSettled([
+        fetch(`${url}/iframe.html`, {
```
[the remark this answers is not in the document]
ndelangen: fetch only throws if the browser doesn't allow the request.

If the server sends a 401, that's will still make the Promise resolve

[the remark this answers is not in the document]
ndelangen: What should we do? and could you explain the potential CSRF attack?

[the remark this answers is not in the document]
ndelangen: OK, got a proposal for you to review again..

lib/api/src/modules/refs.ts
```diff
@@ -113,46 +123,73 @@ export const init: ModuleFn = ({ store, provider, fullAPI }) => {
...
+          handle(
+            fetch(`${url}/metadata.json`, {
+              headers: {
+                Accept: 'application/json',
+              },
+              credentials,
+              cache: 'no-cache',
+            })
+          ),
+        ]);
```
tmeasday, line 183: If both `!included` and `!omitted` (i.e. we weren't able to get `stories.json` at all), do we bother with `metadata.json`?

lib/api/src/modules/refs.ts
```diff
@@ -113,46 +123,73 @@ export const init: ModuleFn = ({ store, provider, fullAPI }) => {
...
-          credentials: 'include',
-          cache: 'no-cache',
-        })
-          .catch(() => false)
-          .then(handler),
+          credentials: 'omit',
+        }),
+        fetch(`${url}/iframe.html`, {
+          credentials: 'omit',
+        }),
```
tmeasday, line 143: I'm not sure about this check. It won't work in the follow situations:

1. The server doesn't allow cross-origin fetch access to the `iframe.html` at all (this is the case for Chromatic, although in Chromatic's case `stories.json` is fetchable as `included`).

2. The server isn't setup for CORS at all (i.e. serves no `Access-Control-Allow-Origin` header). This is sort of the same as 1, functionally.

3. The server requires login to get `iframe.html`. We could probably assume such servers will serve `stories.json` however, maybe.

Given we aren't looking at the result of the `fetch`, perhaps we should use `mode: 'no-cors'`  for this request? It won't fix 3. above but maybe we are OK with that for the reason mentioned?

tmeasday (changes requested):
Getting closer!

lib/api/src/modules/refs.ts
ndelangen: good point!

lib/api/src/modules/refs.ts
ndelangen: > The server doesn't allow cross-origin fetch access to the iframe.html at all (this is the case for Chromatic, although in Chromatic's case stories.json is fetchable as included).

So it will work then?

> The server isn't setup for CORS at all (i.e. serves no Access-Control-Allow-Origin header). This is sort of the same as 1, functionally.

correct it won't work, that's the point.

> The server requires login to get iframe.html. We could probably assume such servers will serve stories.json however, maybe.

Yes, like chromatic, or storybook has no way to load it, so will show an error. It's the best we can do?

tmeasday: I'm not quite sure I am understanding, let me make a chart:

|Host type | Example | Resource public/private? | `included` | `omitted` | `iframe`|error?|
|-----|-----|-----|-----|-----|-----|-----|
|Authed, proper CORS | Chromatic | private, correct credentials | `true` | `false` | `false`|`false`|
|Authed, proper CORS | Chromatic | private, incorrect credentials | `false` | `false` | `false`|`true`|
|Authed, proper CORS | Chromatic | public | `true` | `true` | `false`|`false`|`false`|
|Simple CORS | Netlify | - | `false` | `true` | `true`|`false`|`false`|
|No CORS | S3 | - | `false` | `false` | `false`|`true`|

Doesn't this mean the S3 case is going to show an error in SB?

ndelangen: s3 no cors iframe will be true now with my latest changes

tmeasday (approved):

tmeasday:
OK, LGTM now. We better test this out properly though ;)

PR #10864, opened 2020-05-21
Addon-docs: Fix react forwardRefs with destructured props

tooppaaa:
Issue: #10655

## What I did
Add default props when rendering forward ref or component class 

## How to test

Check story `Addons|Docs/ForwardRef`


shilman (approved):
Thanks @tooppaaa ... you're awesome 💯

PR #10868, opened 2020-05-22
Core: Close server when e2e test failed

tooppaaa:
Issue:
When multiple configs are run on the same node, if the first one fail, the second will show "ADDR already in use"

## What I did
Close server on catch


shilman (approved):
Great fix!

PR #10881, opened 2020-05-22
Angular: Support `workspace.json` in nx workspace

isaacplmann:
Issue:

## What I did
Nx is a tool for managing a monorepo that can contain any kind of javascript framework.  Nx has its own CLI based on the angular CLI, but we also allow the root `angular.json` file to be named `workspace.json` for non-angular devs to be more comfortable.  However, the `@storybook/angular` package will only look for the `angular.json` file.  This PR checks for `angular.json` and falls back to `workspace.json`.  If neither file is present, the same error appears.

## How to test

```bash
npx create-nx-workspace storybook-with-workspace-json --preset=empty --cli=nx
cd storybook-with-workspace-json
yarn add @nrwl/angular @nrwl/storybook
npx nx g @nrwl/angular:ng-add
npx nx g @nrwl/angular:lib ui --no-interactive
npx nx g @schematics/angular:component --name=button --project=ui --no-interactive
npx nx g @nrwl/angular:storybook-configuration --name=ui --generateStories --no-interactive
npx nx run ui:storybook
```

You'll see an error that `angular.json` was not found.  After this PR, the error doesn't show up.

### DIFF
--- a/app/angular/src/server/angular-cli_config.ts
+++ b/app/angular/src/server/angular-cli_config.ts
@@ -47,14 +47,19 @@ function getTsConfigOptions(tsConfigPath: Path) {
 }
 
 export function getAngularCliConfig(dirToSearch: string) {
-  const fname = path.join(dirToSearch, 'angular.json');
+  const possibleConfigNames = ['angular.json', 'workspace.json'];
+  const possibleConfigPaths = possibleConfigNames
+    .map(name => path.join(dirToSearch, name))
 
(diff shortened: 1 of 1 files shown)


ZachJW34:
Thanks @isaacplmann this will be a huge help. Hope this get reviewed soon so I can start playing around with some Angular storybook components!

shilman (approved):
Thanks for adding this--looks great to me. Super excited about the NX integration and would love to chat about more ways to make it better!

PR #10889, opened 2020-05-22
Docs: fix example of story-level decorators with "storiesOf" API

okovpashko:
Issue:
There's an invalid example of using story-level decorators with "storiesOf" format in the docs.
We get an exception is we provide a single decorator instead of an array of decorators.

## What I did
Updated docs with the valid example

## How to test

It's just a documentation update so there's no need to test it


shilman (approved):
Yup. Definitely a typo! Thanks for fixing!

ndelangen:
Thanks for this

PR #10901, opened 2020-05-23
Core: colocate react stories

tooppaaa:
Issue: #10793

## What I did

Move components from `@storybook/<framework>/demo` to `cli/framework/<framework>/`
Copy them to user `/src/` on init


scripts/prepare.js
```diff
@@ -34,9 +34,16 @@ function cleanup() {
...
+      // And do not clean anything for:
+      // - @storybook/cli/dist/generators/**/template*
+      // - @storybook/cli/dist/framework/*
+      // because these are the template files
       // that will be copied to init SB on users' projects
-      if (fs.lstatSync(filePath).isDirectory() || /generators\/.+\/template.*/.test(filePath)) {
+      if (
+        fs.lstatSync(filePath).isDirectory() ||
+        /generators\/.+\/template.*/.test(filePath) ||
+        /dist\/framework\/.*/.test(filePath)
```
tooppaaa, line 45: absolutely not sure about this line

ndelangen: What is this line trying to do?

tooppaaa: Last build step is cleaning up folders.
Because cli/src/framework/* is meant to be copied and not compiled, it's in the original format in dist.
beacause it's .ts / .tsx, we need to tell `prepare` to not clean this folder.

The thing is prepare is valid for all projects so I can't only filter for `cli`

PR #10902, opened 2020-05-24
Fix screenshot path

renancouto:
This change fixes the screenshot path when the README.md is rendered not on Github for instance, on npm https://www.npmjs.com/package/@storybook/addon-viewport

Issue:

## What I did
Changed the image path to its `raw` version

## How to test
The screenshot should not be broken if viewed outside of Github


shilman (approved):
Thanks!

PR #10908, opened 2020-05-25
Core: Avoid re-render on HMR of other stories

tmeasday:
Issue: #10616 

## What I did

Refactored our story change detection to use reference equality check rather than a global revision

## How to test

Use the "core/rendering" story

1. Check it *does not* re-render if you change a different story
2. Check it does re-render if you change the current story.
3. Check it does re-render if you switch stories (and then back)

Note that (1) doesn't currently work due to https://github.com/storybookjs/storybook/issues/10893 (global args). Fixing that in a separate PR, but you can check if you comment out the global args event listener.

### DIFF
--- a/lib/client-api/src/client_api.ts
+++ b/lib/client-api/src/client_api.ts
@@ -145,7 +145,6 @@ export default class ClientApi {
         // and be handled by the HMR.allow in config_api, leading to a re-run of configuration.
         // So configuration is about to happen--we can skip the safety check.
         _storyStore.removeStoryKind(kind, { allowUnsafe: true });
-        _storyStore.incrementRevision();
       });
     }
 
@@ -184,7 +183,6 @@ export default class ClientApi {
           const { _storyStore } = this;
           // See note about allowUnsafe above
           _storyStore.remove(id, { allowUnsafe: true });
-          _storyStore.incrementRevision();
         });
       }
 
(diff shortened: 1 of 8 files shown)


lib/client-api/src/client_api.ts
```diff
@@ -184,7 +183,6 @@ export default class ClientApi {
           const { _storyStore } = this;
           // See note about allowUnsafe above
           _storyStore.remove(id, { allowUnsafe: true });
-          _storyStore.incrementRevision();
```
ndelangen, line 187: real nice!

lib/core/src/client/preview/StoryRenderer.tsx
```diff
@@ -155,14 +155,20 @@ export class StoryRenderer {
     const { forceRender, name } = context;
 
     const { previousMetadata, storyStore } = this;
+    console.log('renderStoryIfChanged', { previousMetadata, metadata, context });
```
ndelangen, line 158: remove the console.logs

ndelangen (approved):
Looks great! just remove some console.logs I think and then this looks great!

shilman (approved):
This is awesome @tmeasday 

Does chromatic look good to you? Any idea why the e2e tests are failing?

tmeasday:
@shilman Chromatic looks good (approved). No idea why e2e failed though, I kind of doubt it is relevant as I think I only changed behaviour on HMR and we were just talking today about how we don't test that in e2e tests ;0

ndelangen:
Work is being done on the e2e tests stability:
https://github.com/storybookjs/storybook/pull/10888

PR #10925, opened 2020-05-26
UPGRADE cypress && FIX lodash dependency

ndelangen:
## What I did
- UPGRADE cypress
- change dependency on 'lodash' to 'lodash/*'
- change the generators to not pass the event into the action


yarn.lock
```diff
@@ -11695,23 +11637,14 @@ cyclist@^1.0.1:
   integrity sha1-WW6WmP0MgOEgOMK4LW6xs1tiJNk=
 
 cypress@^4.5.0:
-  version "4.5.0"
-  resolved "https://registry.yarnpkg.com/cypress/-/cypress-4.5.0.tgz#01940d085f6429cec3c87d290daa47bb976a7c7b"
-  integrity sha512-2A4g5FW5d2fHzq8HKUGAMVTnW6P8nlWYQALiCoGN4bqBLvgwhYM/oG9oKc2CS6LnvgHFiKivKzpm9sfk3uU3zQ==
+  version "4.6.0"
```
tooppaaa, line 11640: This is causing e2e to fail

ndelangen: thanks, upgraded it in circle.yml too

gaetanmaisse:
Everything relevant to fix Yarn 2 compatibility is now in https://github.com/storybookjs/storybook/pull/11008. I did not pick "FIX full inclusion of lodash" as it was not directly related to the base issue

PR #10941, opened 2020-05-27
Core: Add missing babel plugin

tooppaaa:
Issue: #10939

## What I did
Add missing plugin with right configuration

## How to test

E2E should be green


shilman (approved):
LGTM

PR #10944, opened 2020-05-27
fix the focus not showing up on the button

mkamranhamid:
Issue: [#10862](https://github.com/storybookjs/storybook/issues/10862)

## What I did
update the global button CSS to have an outline on a button

## How to test

using e2e it will be testes correctly


lib/theming/src/global.ts
```diff
@@ -32,7 +32,7 @@ export const createReset = memoize(1)(
     },
 
     'button, input, textarea, select': {
-      outline: 'none',
+      // outline: 'none',
```
ndelangen, line 35: should just delete the line entirely

ndelangen (approved):

PR #10972, opened 2020-05-28
Tech/colocate more stories

tooppaaa:
Merging to #10913 to simplify review

This is only moving stories outside generators and using baseGenerator


gaetanmaisse (approved):
It LGTM, not sure why so many CI checks are 🔴  🤔

tooppaaa:
I'll have a look later :) Thanks for the review !

PR #10974, opened 2020-05-29
Core: Log swallowed errors when requiring stories

luke-j:
Issue: When stories are written in typescript and they contain type errors, requiring files can swallow the type errors emitted by packages like `fork-ts-checker-webpack-plugin`. Normally this wouldn't be an issue however, when running in certain test environments - like storyshots - this can cause the tests to not be found.

## What I did

This simply wraps the dynamic require in a try-catch block and logs the error. Re-throwing the error is not an option as that is also swallowed.

## How to test

Run storyshots (for example) with a story that has a type error in it in a package with webpack  type checking. The error should be emitted to the console.

- Is this testable with Jest or Chromatic screenshots? - No

### DIFF
--- a/lib/core/src/client/preview/loadCsf.ts
+++ b/lib/core/src/client/preview/loadCsf.ts
@@ -24,13 +24,17 @@ const loadStories = (
   if (reqs) {
     reqs.forEach((req) => {
       req.keys().forEach((filename: string) => {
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM! Thanks for the fix!!!

PR #10985, opened 2020-05-29
CLI: Fix `sb init` in Yarn workspace environment

### LINKED ISSUE #10915: automatic initialization fails in projects using yarn workspaces
**Describe the bug**
Attempting to run `npx -p @storybook/cli sb init` in projects using yarn workspaces fails with the following error:
```
 • Detecting project type. ✓
yarn add v1.22.4
error Running this command will add the dependency to the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
An error occurred while installing dependencies.
{
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 47115,
  stdout: null,
  stderr: null,
  error: null
}
```

This is because `yarn add` requires a `-W` flag as a safety measure when adding something to the root of a workspace.

**To Reproduce**
Steps to reproduce the behaviour:
1. Setup yarn workspaces in a project: https://classic.yarnpkg.com/en/docs/workspaces/
2. Try to run the storybook init command

**Expected behaviour**
Storybook should automatically add the -W flag so the init command continues working.

gaetanmaisse:
Issue: #10915 

## What I did

 - Add `--ignore-workspace-root-check` flag when CLI is running `yarn add` command.
 - Add E2E test for Yarn workspace. It's based on a simple React setup.
 - Remove `skip-install` option from `sb init` command used in E2E because it can hide some issues with the dependency installation process of @storybook/cli. For instance, missing Yarn flags needed to have the CLI working in Yarn workspaces.

--

Also, improve jobs assignment on CI nodes: 

With the current algorithm, some nodes have no jobs assigned to them due to rounding consideration. For instance, with 10 nodes and 13 jobs to assign:
 - 13 jobs / 10 nodes -> 1.3 jobs per node -> rounded to 2 -> only the 1st 6 nodes run 2 jobs, the 7th 1 job and the other 0 job.

Using a modulo ensures that each node will at least run 1 job and thus reduce the E2E run duration (in theory 🤞).

## How to test

- E2E tests should be 🟢 on CI, especially new E2E test run in `example-v2` job

### DIFF
--- a/scripts/run-e2e-config.ts
+++ b/scripts/run-e2e-config.ts
@@ -152,3 +152,13 @@ export const yarn2Cra: Parameters = {
     `yarn dlx create-react-app@{{version}} {{name}}-v{{version}}`,
   ].join(' && '),
 };
+
+export const reactInYarnWorkspace: Parameters = {
+  name: 'reactInYarnWorkspace',
+  version: 'latest',
+  generator: [
+    'cd {{name}}-v{{version}}',
+    'echo "{ \\"name\\": \\"workspace-root\\", \\"private\\": true, \\"workspaces\\": [] }" > package.json',
+    `yarn add react react-dom --silent -W`,
+  ].join(' && '),
+};
--- a/scripts/run-e2e.ts
+++ b/scripts/run-e2e.ts
@@ -85,7 +85,7 @@ const initStorybook = async ({ cwd, autoDetect = true, name }: Options) => {
   logger.info(`🎨 Initializing Storybook with @storybook/cli`);
   try {
     const type = autoDetect ? '' : `--type ${name}`;
-    await exec(`npx -p @storybook/cli sb init --skip-install --yes ${type}`, { cwd });
+    await exec(`npx -p @storybook/cli sb init --yes ${type}`, { cwd });
   } catch (e) {
     logger.error(`🚨 Storybook initialization failed`);
     throw e;
@@ -187,10 +187,10 @@ const runTests = async ({ name, version, ...rest }: Parameters) => {
     await generate({ ...options, cwd: siblingDir });
     logger.log();
 
-    await initStorybook(options);
+    await setResolutions(options);
     logger.log();
 
-    await setResolutions(options);
+    await initStorybook(options);
     logger.log();
 
     await addRequiredDeps(options);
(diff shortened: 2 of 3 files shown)


scripts/run-e2e-config.ts
```diff
@@ -152,3 +152,13 @@ export const yarn2Cra: Parameters = {
     `yarn dlx create-react-app@{{version}} {{name}}-v{{version}}`,
   ].join(' && '),
 };
+
+export const reactInYarnWorkspace: Parameters = {
+  name: 'reactInYarnWorkspace',
+  version: 'latest',
+  generator: [
+    'cd {{name}}-v{{version}}',
+    'echo "{ \\"name\\": \\"workspace-root\\", \\"private\\": true, \\"workspaces\\": [] }" > package.json',
```
gaetanmaisse, line 161: Init a Yarn workspace with the most simple way (i.e. without any tricks or new dependency)

scripts/run-e2e.ts
```diff
@@ -274,13 +274,18 @@ if (frameworkArgs.length > 0) {
...
-  const step = Math.ceil(narrowedConfigs.length / b);
-  const offset = step * a;
-
-  const list = narrowedConfigs.slice().splice(offset, step);
-
-  logger.info(`📑 Assigning jobs ${list.map((c) => c.name).join(', ')} to node ${a} (on ${b})`);
+  const nodeIndex = +process.env.CIRCLE_NODE_INDEX || 0;
+  const numberOfNodes = +process.env.CIRCLE_NODE_TOTAL || 1;
+
+  const list = narrowedConfigs.filter((_, index) => {
```
tooppaaa, line 280: I wanted to do that for a while now ^^
This logic is shared in other runners, chromatic and examples if I'm right

tooppaaa (approved):

scripts/run-e2e.ts
tooppaaa: https://github.com/storybookjs/storybook/pull/10989

shilman (approved):
LGTM!

PR #11035, opened 2020-06-03
Addon-controls: Add hideNoControlsWarning parameter

matheo:
Issue:

Right now the warning will be always visible even if we don't want to have interactive controls for a specific Story, and just have the expanded controls.

## What I did

I added a new parameter to be able to hide it in general or for a specific story with:
```
{ controls: { hideNoControlsWarning: true } }
```


shilman (changes requested):
Thanks @matheo, this is great! Can you add to the README too?

matheo:
done!

shilman (approved):
💯

PR #11048, opened 2020-06-04
Controls: Fix null entry in options array handling

matheo:
We have a component with `@Input size: null | 'large'` to generate a selector control for it.
In `beta.21` it throws an exception because `.toString()` is called for each item.

## What I did

Implemented a safer `String(item)` instead `item.toString()` for those cases.

image


shilman (approved):
Nice fix @matheo! In general please add an example to `official-storybook` somewhere to go along with the code 🙏

PR #11063, opened 2020-06-06
Core: Add babel plugin for typescript decorators

fernandopasik:
Issue: #11061 

## What I did
- Added the `@babel/plugin-proposal-decorators` to the babel config

## How to test

- I've tested it locally but I'd like to add a new example in the kitchen sink apps.

### DIFF
--- a/lib/core/package.json
+++ b/lib/core/package.json
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "@babel/plugin-proposal-class-properties": "^7.8.3",
+    "@babel/plugin-proposal-decorators": "^7.8.3",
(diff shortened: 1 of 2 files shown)


mrmckeb (approved):
Confirmed that this is the correct behaviour, reference:
https://babeljs.io/docs/en/babel-plugin-transform-typescript#typescript-compiler-options

mrmckeb:
@shilman This is good to merge IMO. I think once merged we could add a quick comment above the plugin referencing this doc:
https://babeljs.io/docs/en/babel-plugin-transform-typescript#typescript-compiler-options

mrmckeb:
I've added that comment, good to merge if you're OK with it @shilman!

PR #11072, opened 2020-06-07
feat(html): add JSX support for html stories

unrevised6419:
This is just an experiment feel free to close if this is not needed. I just created this PR if someone wants to learn something from this.

Today I studied how storybook internals work, and wanted to experiment with `JSX` in `HTML` stories. This is the final result.

Note:
Right now `import React from 'react'` is required in files where `JSX` is used. with [`@babel/plugin-transform-react-jsx@^7.9`][1] you can put it in automatic mode, and React import is not needed anymore.

[1]: https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#react-automatic-runtime

### DIFF
--- a/app/html/package.json
+++ b/app/html/package.json
@@ -34,12 +34,15 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "dependencies": {
+    "@babel/plugin-transform-react-jsx": "^7.3.0",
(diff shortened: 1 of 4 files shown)


iamandrewluca:
https://github.com/vadimdemedes/dom-chef

PR #11074, opened 2020-06-07
CLI: refactor to simplify works with multiple package managers

gaetanmaisse:
## What I did

I introduced a `JsPackageManager` class and specialized subclasses for NPM, Yarn 1, and Yarn 2. The goal of this is to use the Factory Pattern to encapsulate everything specific to a package manager in a single class.

The factory returns an instance of NPM, Yarn 1, or Yarn 2 proxy class according to the package manager used/preferred in the end project.

All CLI's code will be package manager agnostic and will just call functions of `JsPackageManager` interface.

The PR is quite big but reading commit by commit should be ok as it really follows how I worked on this.

There is still to do to centralize things but I prefer to open a working PR now and continue later than working on a branch for weeks and then have a 895612456756 conflicts to solve. 🙃 

## How to test

- All CI checks should be 🟢 especially CLI tests and examples-v2


gaetanmaisse:
⚠️ It will surely generate a lot of conflicts with https://github.com/storybookjs/storybook/pull/10913 but it worth it as with both these PRs we will greatly improve CLI codebase :slight_smile: :tada:

shilman (approved):
Nice one!! ❤️

lib/cli/src/js-package-manager/NPMProxy.ts
```diff
@@ -0,0 +1,47 @@
...
+      const parsedOutput = JSON.parse(commandResult);
+
+      if (parsedOutput.error) {
+        // FIXME: improve error handling
+        throw new Error(parsedOutput.error.summary);
+      } else {
+        return parsedOutput;
+      }
+    } catch (e) {
+      throw new Error(`Unable to find versions of ${packageName} using yarn`);
```
PiDelport, line 44: This refers to `yarn` instead of `npm`.

PR #11081, opened 2020-06-08
ADD design-system to official example

ndelangen:
Issue:  -we want to be able to view the storybook design system from within the official storybook example.-

## What I did

Adding this dependency should add a ref automatically


tmeasday (approved):
LGTM

PR #11118, opened 2020-06-10
Core: Fix default Babel config for CRA

mrmckeb:
Issue: #11045

## What I did
Our default Babel config is no longer returned for CRA, but user-provided config is still respected. We need a more scalable option for this in future.

## How to test

There should be no difference in behaviour for users, but the edge case reported in #11045 should be resolved - which was caused by duplicate presets/plugins.

### DIFF
--- a/lib/core/src/server/common/common-preset.js
+++ b/lib/core/src/server/common/common-preset.js
@@ -4,9 +4,14 @@ import babelConfig from './babel';
 
 export const babel = async (_, options) => {
   const { configDir, presets } = options;
+  const hasCRAPreset = options.presetsList.some(
+    (preset) => (preset.name || preset) === '@storybook/preset-create-react-app'
+  );
+
+  const babelDefault = hasCRAPreset ? { presets: [], plugins: [] } : babelConfig();
 
   return loadCustomBabelConfig(configDir, () =>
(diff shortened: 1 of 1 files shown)


shilman (commented):
Is there no way to do this in the CRA preset?

mrmckeb:
Sorry for the bad commit name. Now that I think about it, it made little sense - late night coding.

@shilman the alternative, from my discussion with @ndelangen, is to reimplement the logic from scratch in the preset - which means checking `main.js` for Babel config, checking `.babelrc.*` files, and then normalising paths and extended configs.

We're already doing the same thing in the base Webpack config for Storybook React. This is clearly not scalable, but will solve the immediate problem for v6 - and we can keep talking about how to solve it.

Long term, I think we need a type of addon that can tell the framework "I come with my own Webpack and Babel config".

ndelangen:
I would assume if the cra preset had this as an extra export:
```
export const babelDefault = () => ({ 
  presets: [],
  plugins: [],
});
```

It would reset the base babel config IF the cra preset is injected, starting it off from a clean slate.

mrmckeb:
Thanks @ndelangen, that's exactly what I was after.

PR #11129, opened 2020-06-11
simplify babel config and make polyfills lazyloaded

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/11126

## What I did
See what happens to storybook if we remove the global polyfills


ndelangen:
I've modified it so the polyfill CAN load, but they load async, and only when needed, thanks to @adamdoyle really

ndelangen:
I can likely turn this into a NON_BREAKING state 😮

ndelangen:
this makes the `configure` call async, unless an environment variable is set. (used by storyshots)

ndelangen:
@shilman I think this is mostly a backwards compatible change now. There are risks in the changes stated above.
And obviously this changes a load of babel config...

gaetanmaisse:
Will review it in the next few days but I will certainly have some question @ndelangen ;)

tmeasday:
> this makes the configure call async

This is a little scary for Chromatic which relies on stories being rendered within the page load even in certain cases. Do we need to worry @ndelangen?

ndelangen:
@tmeasday chromatic waits for the configure event, so it's already async

gaetanmaisse (commented):
It does look ok, however I'm really not a Babel expert and there are a lot of changes on that side

ndelangen:
understandable @gaetanmaisse, we have the e2e tests for a reason..

@shilman I'd love to know the perf impact of this.

shilman:
Looks like dev impact is negligible:

Storybook_Benchmark_›_Explorer__Install__Start__Build_

But production impact is good, about a 12% reduction in bundle size:

ndelangen:
I'm closing this because:
- `--modern` implements this already
- There are too many merge conflicts to salvage this PR

PR #11154, opened 2020-06-12
Addon-docs: Fix forwardRef & invalid hook call

tooppaaa:
Issue: #10895

## What I did
Make forward refs working

## How to test

Outside repo check component


shilman (approved):
LGTM! Thanks for your fix and patience @tooppaaa

PR #11156, opened 2020-06-13
Core: Extend router/utils test set

mbelsky:
## What I did

Add router/utils test set

### DIFF
--- a/lib/router/src/utils.ts
+++ b/lib/router/src/utils.ts
@@ -9,7 +9,7 @@ export interface StoryData {
 
 const splitPathRegex = /\/([^/]+)\/(?:(.*)_)?([^/]+)?/;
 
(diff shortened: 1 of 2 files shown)


lib/router/src/utils.ts
```diff
@@ -9,7 +9,7 @@ export interface StoryData {
 
 const splitPathRegex = /\/([^/]+)\/(?:(.*)_)?([^/]+)?/;
 
-export const parsePath: (path?: string) => StoryData = memoize(1000)(
+export const parsePath: (path: string | undefined) => StoryData = memoize(1000)(
```
mbelsky, line 12: I've made `path` not optional here, because  `memoize` throws `Memoizerific functions should always be called with the same number of arguments`

shilman (approved):
Looks great! Thanks @mbelsky!

PR #11159, opened 2020-06-13
UI: Remove unused modules

mbelsky:
## What I did

I've found some modules that look like unused code. So I've just removed them.


shilman (approved):
LGTM 🙈  thanks for tracking this down!

PR #11178, opened 2020-06-15
Core: Fix addon load order

ndelangen:
Issue: #9819

## What I did
Originally the addons property was split into register & presets, and then executed seperately

I edited the loadPresets code so it can deal with register type addons in order.


shilman (approved):
Awesome!!!! 💯

PR #11195, opened 2020-06-16
Addon-docs: Remove render for react components missing docgen

tooppaaa:
Issue: #11176

## What I did

Remove render completely -- it's unclear whether it solves any problems and it causes them in some cases


shilman (approved):
Thanks @tooppaaa

PR #11215, opened 2020-06-17
Fix/refs rename auth url & add tests for modules/refs

ndelangen:
@tmeasday


lib/api/src/modules/refs.ts
```diff
@@ -168,10 +170,11 @@ export const init: ModuleFn = ({ store, provider, fullAPI }) => {
...
         } as Error;
       } else if (omitted || included) {
-        const credentials = !omitted ? 'include' : 'omit';
+        const credentials = included ? 'include' : 'omit';
 
-        const [stories, metadata] = await Promise.all([
-          handle(omitted || included),
+        const [stories, storiesWithAuth, metadata] = await Promise.all([
+          handle(omitted),
+          handle(included),
```
tmeasday, line 177:
```suggestion
          included ? handle(included) : handle(omitted),
```

No reason to complicate things, they are the "same" data, only one of them will work.

lib/api/src/modules/refs.ts
```diff
@@ -180,10 +183,14 @@ export const init: ModuleFn = ({ store, provider, fullAPI }) => {
               credentials,
               cache: 'no-cache',
             })
-          ),
+          ).then((r: SetRefData) => (r?.error?.message === 'Failed to fetch' ? {} : r)),
```
tmeasday, line 186: I think we should only run this block if `included` is true.

tmeasday: Also -- do we have tests for this stuff?

tmeasday (changes requested):

lib/api/src/modules/refs.ts
ndelangen: Yea 1 of them works, this is how I know which one; and handle the data from the one that worked?

lib/api/src/modules/refs.ts
```diff
@@ -180,10 +183,14 @@ export const init: ModuleFn = ({ store, provider, fullAPI }) => {
...
         ]);
 
-        Object.assign(loadedData, { ...stories, ...metadata });
+        Object.assign(loadedData, { ...stories, ...storiesWithAuth, ...metadata });
+
+        if (storiesWithAuth && !storiesWithAuth.loginUrl) {
+          delete loadedData.loginUrl;
+        }
```
tmeasday, line 193: You won't need to do this is if you are careful not to include results that didn't work (i.e. non-credentialed requests to a server that requires them).

lib/api/src/modules/refs.ts
ndelangen: I think in principle metadata.json is optional right?
So if the request errors for whatever reason, I think it's ok to ignore that error, rather then showing it to the user?

lib/api/src/modules/refs.ts
ndelangen: Actually they both "work" 1 responds with `{ loginUrl: '' }` the other with `{ stories: {} }`.

ndelangen: I had `included ? handle(included) : handle(omitted),` before and it keps sending me to the login-state because the `handle(omitted),` succeeded with `{ loginUrl: '' }`

tmeasday: > Actually they both "work" 1 responds with { loginUrl: '' } the other with { stories: {} }.

If `included` doesn't error (with a CORS error) it should take priority over `omitted` no matter what. We should not even be looking at the result of `omitted` in such cases.

> I had included ? handle(included) : handle(omitted), before and it keps sending me to the login-state because the handle(omitted), succeeded with { loginUrl: '' }

I don't follow. If `included` worked then we would ignore `omitted` entirely.

lib/api/src/modules/refs.ts
tmeasday: It is fine to silently drop it if it errors, I agree. But given we pass credentials, and we already know going in if the credentials are going to work (based on `/stories.json`), then we can be a bit smarter no?

Or is it parallel? If so let's silently ignore errors.

tmeasday:
Thanks for adding the tests @ndelangen -- they look good, but I think we need some for the cases where the different requests return different things. I'm not sure what we are testing right now is realistic (all requests returning `stories: {}`).

lib/api/src/modules/refs.ts
ndelangen: Yes it's partially in parallel.

ndelangen:
I added a test for `metadata.json` returning something different and when `credentials:omit` returns a `loginUrl`. all seems to work well.

PR #11224, opened 2020-06-18
Composition QA

domyen:
Issue: 
This PR polishes Storybook Composition.

## What I did
I adjusted the Composition UI so that it matches the design. Also tweaked some Sidebar and Button styles that affected Composition.

## How to test

**Bugs that need fixing**
- Sidebar:Dark with refs does not get rendered and snapshotted by Chromatic
- Clicking on a new version collapses the ref when it should not
- "View Docs" link does not work


domyen:
@ndelangen found some issues that I could use a hand with:

**Bugs**

- `Sidebar:Dark With Refs` does not get rendered and snapshotted by Chromatic properly. Likely due to the theme config not being quite right for visual testing. I think you're addressing the theming in another PR so I think we're OK to remove this story? wdyt? 
image
- Clicking on a new version collapses the ref when it should not https://www.loom.com/share/5198402e9f7647cb8987cd25be8ff4d1
- "View Docs" link does not work for some reason
image

**Reorg "Refs" stories for visual testing**
Instead of having all the refs states printed in `Sidebar:With Refs`, can we decompose the refs into separate stories? That would make regression detection more granular/effective. 

Screenshot 2020-06-24 15 55 53

Strawman:
- Keep `Sidebar/Sidebar: With Refs`, but only show one representative Ref
- Add `Sidebar/Refs: ref name` for each ref

ndelangen:
Yes, I'll work on that @dom
I'll focus on getting https://github.com/storybookjs/storybook/pull/11295/ merged first

ndelangen:
I'm on it now, I'll do:
- [ ] Clicking on a new version collapses the ref when it should not
- [ ] "View Docs" link does not work
- [ ] Reorg "Refs" stories for visual testing

ndelangen (approved):

PR #11247, opened 2020-06-20
Fix invalid glob warning for absolute paths

vvanpo:
When passing absolute paths to the `stories` option (in `main.js`),
Storybook will build these stories just fine. However, it displays an
erroneous warning on startup asking you if your glob pattern is invalid.

## What I did

Check if the `stories` path is absolute before joining with the
configuration directory.

### DIFF
--- a/lib/core/src/server/preview/entries.js
+++ b/lib/core/src/server/preview/entries.js
@@ -58,7 +58,7 @@ export async function createPreviewEntry(options) {
     entries.push(path.resolve(path.join(configDir, `generated-stories-entry.js`)));
 
     const files = (
-      await Promise.all(stories.map((g) => glob(path.join(configDir, g))))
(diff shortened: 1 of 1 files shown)


lib/core/src/server/preview/entries.js
```diff
@@ -58,7 +58,7 @@ export async function createPreviewEntry(options) {
     entries.push(path.resolve(path.join(configDir, `generated-stories-entry.js`)));
 
     const files = (
-      await Promise.all(stories.map((g) => glob(path.join(configDir, g))))
+      await Promise.all(stories.map((g) => glob(path.isAbsolute(g) ? g : path.join(configDir, g))))
```
ndelangen, line 61: yeah that's a fine solution as well 👍

ndelangen (approved):

PR #11262, opened 2020-06-22
Client-logger: Replace magic numbers with constants

mbelsky:
## What I did
I've just found some magic numbers and replace them with constants defined above.

## How to test

- Is this testable with Jest or Chromatic screenshots? **yes**

Tests `lib/client-logger/src/index.test.ts` pass


shilman (approved):
❤️

PR #11265, opened 2020-06-23
Addon-backgrounds: Allow gradients in story preview

matus12:
Issue: #11231

## What I did
I have changed the `backgroundColor` property in backgrounds addon to the `background`, so we can also use gradients.


ndelangen (approved):

yannbf (approved):
Hey @matus12, thanks a lot for the contribution! Super nice :)
Indeed, with that change users can even add images as well!

PR #11282, opened 2020-06-23
CLI: use addon-essentials & args

tooppaaa:
Issue: #10795

## What I did

- Update stories to use args and follow `new` pattern with .bind({})
- Prepare addon-controls

## How to test

- E2E


tooppaaa:
@ndelangen I've not set back `() => action('onClick')()` to be closer from what's auto generated for some frameworks. Would that be an issue ?

PR #11295, opened 2020-06-24
ADD dual theme rendering

ndelangen:
Issue:

## What I did
I added a global decorator for the official storybook example that makes it so all stories get rendered twice.

Once in a dark theme context, once in a light.


tmeasday:
Would love to make this a built in feature of SB + Chromatic, but we'll get there. 

@ndelangen maybe we should use a globalArg to add a toolbar that turns this off and on in the UI?

tmeasday (approved):

domyen:
Thanks Norbert! 

Can stack the rendering instead of doing a side-by-side? Some of our components are responsive/fluid so they rely on the width of the viewport. 

I already found some theming bugs, I'll wait on the above to do them.

domyen:
Found some issues:

**For the "Controls" components**, when I try to do side-by-side or stacked I get the error below. Open local SB and go to this url [locally](http://localhost:9011/?path=/story/controls-array--basic) or in the [cloud](https://5a375b97f4b14f0020b0cda3-ckozzvtoca.chromatic.com/?path=/story/controls-array--basic).
image



**For the Heading component**, when I do side-by-side, the content doesn't seem to get updated with the theme vars.Open local SB and go to this url [locally](http://localhost:9011/?path=/story/ui-sidebar-heading--standard).
image

ndelangen:
Great catch @domyen I was able to find a fix!

PR #11303, opened 2020-06-24
CLI: Fix docs & essentials version on `sb@next init`

tooppaaa:
Issue: docs or essentials are installed using @latest tag instead of current version

## What I did
Quick fix to resolve right value

@gaetanmaisse I think we shoudn't check the version addons in peer dependency but base them on the framework version.

### DIFF
--- a/lib/cli/package.json
+++ b/lib/cli/package.json
@@ -65,6 +65,8 @@
   },
   "devDependencies": {
     "@storybook/addon-actions": "6.0.0-beta.35",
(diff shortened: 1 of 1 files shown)


tooppaaa:
https://github.com/storybookjs/storybook/blob/next/lib/cli/src/js-package-manager/JsPackageManager.ts#L140

shilman (approved):
AHA! 🤦

PR #11347, opened 2020-06-29
docs: remove deprecated addon

wormwlrm:
Issue: 

N/A

## What I did

I replaced deprecated addon links with [deprecated repository's](https://github.com/storybookjs/deprecated-addons) one.

- [Some addon hyperlinks](https://storybook.js.org/docs/basics/writing-stories/#parameters) are linked with `tree` branch, which already deprecated addons' were removed, so it shows 404 error. I think it can cause confusion.
- There is no meaningful feature addition on deprecated addons, so it should be replaced with repository which contains deprecated addons only, for their own roles.
- When `next` branch merged to `master`, then all links contain deprecated addons will be broken. 

## How to test

N/A


shilman (approved):
Thanks @wormwlrm !! 🙏

PR #11350, opened 2020-06-29
Addon-docs: Fix fixed-position inline stories

### LINKED ISSUE #8011: [addon-docs] Component with position:fixed leaks out from Preview
**Describe the bug**
When a component with css `position:fixed` is displayed on the Docs page, it's fixed to the viewport, leaks out of the preview box, and does not scroll with the page. I've noticed that adding `transform: scale(1)` to the parent of the `position:fixed` component (tried it by pressing the the zoom button) clips and bound the `position:fixed` element correctly.

**To Reproduce**
1. Create a div with `position: fixed`
2. Display it on the docs-page

**Expected behavior**
All element of the component that is rendered inside the Docs preview should be contained  inside the preview.

**Screenshots**
***The Bug***
image
***Expected (this is with `transform:scale(1)`)***
image

**System:**
```
Environment Info:

  System:
    OS: Linux 5.0 Ubuntu 19.04 (Disco Dingo)
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
  Binaries:
    Node: 12.4.0 - ~/.nvm/versions/node/v12.4.0/bin/node
    Yarn: 1.17.3 - /usr/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v12.4.0/bin/npm
  Browsers:
    Chrome: 76.0.3809.132
  npmPackages:
    @storybook/addon-a11y: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/addon-actions: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/addon-backgrounds: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/addon-console: ^1.2.1 => 1.2.1 
    @storybook/addon-docs: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/addon-knobs: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/addon-links: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/addon-viewport: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/addons: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/react: ^5.2.0-rc.6 => 5.2.0-rc.6 
    @storybook/theming: ^5.2.0-rc.6 => 5.2.0-rc.6 
```

**Additional context**
[Source Code](https://gitlab.com/decapos/termlink/blob/develop/silverhand/src/components/Layout/Layout.stories.tsx)
[Live Storybook](https://silverhand.decapos.io/?path=/docs/components-layout--basic)

frassinier:
Issue #8011 

## What I did

If you provide height to the story component it might want to deal with some fixed-positioned elements so we reset the position context.

## How to test

- Is this testable with Jest or Chromatic screenshots? yes, chromatic screenshots will be updated


lib/components/src/blocks/Story.tsx
```diff
@@ -34,12 +35,16 @@ type IFrameStoryProps = CommonProps;
 
 type StoryProps = InlineStoryProps | IFrameStoryProps;
 
-const InlineStory: FunctionComponent<InlineStoryProps> = ({ storyFn, height, id }) => (
+const InlineStory: FunctionComponent<InlineStoryProps> = ({ storyFn, height, id = uuidv4() }) => (
```
tooppaaa, line 38: Can id be undefined ? Seems weird ?

Appart from this, I don't think you need uuid for this. This adds another dependency for something _simple_. Interesting [reading](https://medium.com/javascript-in-plain-english/you-might-not-need-uuid-v4-for-generating-random-identifiers-89e8a28a7d77) about it
`Math.round(Math.random() * 1e5)` will be enough, unless someone plans to have that many story, we can take the risk of having the same id twice :)

lib/components/src/blocks/Story.tsx
```diff
@@ -34,12 +35,16 @@ type IFrameStoryProps = CommonProps;
...
   <Fragment>
-    {height ? <style>{`#story--${id} { min-height: ${height} }`}</style> : null}
-    <Fragment>
-      {storyFn ? createElement(storyFn) : <EmptyBlock>{MISSING_STORY(id)}</EmptyBlock>}
-    </Fragment>
+    {height ? (
+      <style>{`.sbdocs .story--${id} { min-height: ${height}; transform: translateZ(0); overflow: auto;`}</style>
```
tooppaaa, line 41: Why do you need to set transform ?

lib/components/src/blocks/Story.tsx
frassinier: Yes, it can be undefined, using InlineStory directly, that's why!
Really interesting reading and I'm not against removing this extra dependency (even if it's already used in other packages)

lib/components/src/blocks/Story.tsx
frassinier: I need to reset the wrapper context for dealing with `position: fixed` or `position: sticky` usage in the components.
Take a look at  #8011

lib/components/src/blocks/Story.tsx
tooppaaa: Thanks for the explanation ;) 
I'm sure we can apply the same logic for other use of uuid.

lib/components/src/blocks/Story.tsx
tooppaaa: Whoa that's impressive the impact of transform over position. Thanks again ;)

tooppaaa (approved):

lib/components/src/blocks/Story.tsx
```diff
@@ -34,12 +35,16 @@ type IFrameStoryProps = CommonProps;
...
+    ) : null}
+    <div className={`story--${id}`}>
+      <Fragment>
+        {storyFn ? createElement(storyFn) : <EmptyBlock>{MISSING_STORY(id)}</EmptyBlock>}
+      </Fragment>
+    </div>
```
ndelangen, line 47: There's no way to do this without the extra `<div>`?

frassinier: I don't see how actually but this Fragment is not necessary anymore if we keep this extra div 🤔

frassinier: @ndelangen Can it fit as it?

frassinier: @ndelangen any concerns?

ndelangen: WDYT @shilman, any concerns?

ndelangen:
I do fear the extra div is a change that will bite us later..
@shilman got an opinion on this? Do we have another plan to address this issue of height & width for previews in docs?

tooppaaa:
What about an absolute div **before** the story with the given constraints ?

I assume setting the parent of stories relative is no-impact or at least something we can explain.

tooppaaa:
@ndelangen @frassinier  any opinion on my potential solution ?

frassinier:
@tooppaaa I'm personally against using absolute position when it's not necessary but thank you for the proposal. I can rework it for resetting the z-index context for each story instead...

tooppaaa (commented):
Looks good ! ✔️

frassinier:
Wait Clément still in progress 🤞

frassinier:
@tooppaaa, @ndelangen, and @shilman 👋 I hope it will fit this time, I can't do it simpler!

frassinier:
Demo is here!
https://5a375b97f4b14f0020b0cda3-qwtnrhmjkl.chromatic.com/?path=/docs/addons-docs-mdx-id--fixed-layout-example

shilman (approved):
LGTM -- great work!!

PR #11354, opened 2020-06-29
CLI: Update mithril/rax/web-components templates, rework Yarn2 e2e

gaetanmaisse:
Related to https://github.com/storybookjs/storybook/issues/10797

## What I did

E2E:
 - introduce `--use-yarn-2` flag to run tests with Yarn 2
 - introduce `--use-local-sb-cli` flag to run E2E tests with local `@storybook/cli` package
 - use base CRA tests config for Yarn 2 instead of a custom one
 - remove riot@3 from e2e config list as example stories need to be updated and migrate to riot@4
 - remove HTML from e2e config list as example stories need to be updated (and I'm not sure about the code style we need to use: `document.createElement(...)` or plain text i.e. `"<div></div>"`)

CLI:
 - update Rax template with Button, Header and Page stories
 - update Mithril template with Button, Header and Page stories
 - update web component template with Button, Header and Page stories.

## How to test

- CI should be 🟢 and still contains a CRA test in `examples-v2-yarn-2` job


gaetanmaisse:
Templates for `mithril` and `web component` are outdated too, I will work on them

gaetanmaisse:
I updated mithril and web component stories. And everything is 🟢 🎉 

⚠️ For web component I kept the same code style as before i.e. using `html()` function from `lit-html`. However, I think we should instead create "real" web components using `lit-element` and then use these custom elements directly for the story template: `const Template = () => html'<custom-btn> </custom-btn>';`. As this is not trivial I would enjoy having your opinion before starting to work on that for nothing 😄 .

scripts/run-e2e.ts
```diff
@@ -92,7 +116,12 @@ const initStorybook = async ({ cwd, autoDetect = true, name }: Options) => {
   logger.info(`🎨 Initializing Storybook with @storybook/cli`);
   try {
     const type = autoDetect ? '' : `--type ${name}`;
-    await exec(`npx -p @storybook/cli sb init --yes ${type}`, { cwd });
+
+    const sbCLICommand = useLocalSbCli
```
yannbf, line 120: https://media3.giphy.com/media/l41YleLHixOeCWNe8/giphy.gif

scripts/run-e2e-config.ts
```diff
@@ -126,22 +130,22 @@ export const react_typescript: Parameters = {
...
+  generator: 'npx create-react-app@{{version}} {{name}}-{{version}}',
 };
 
 // TODO: there is a compatibility issue with riot@4
-export const riot: Parameters = {
-  name: 'riot',
-  version: '3',
-  generator: fromDeps('riot@3', 'riot-compiler@3', 'riot-tmpl@3'),
-};
+// export const riot: Parameters = {
```
yannbf, line 144: We really need to come back to these frameworks :s

yannbf (approved):
LGTM!

shilman (approved):
🙌

PR #11357, opened 2020-06-29
E2E: remove cli fixtures in favor of e2e

tooppaaa:


tooppaaa:
@ndelangen if this is all green, should we proceed like this and forgot failing tests from the other PR ?

ndelangen (approved):

scripts/run-e2e-config.ts
```diff
@@ -19,6 +19,14 @@ export const angular: Parameters = {
   additionalDeps: ['react', 'react-dom'],
 };
 
+export const angularv7: Parameters = {
+  ...baseAngular,
+  version: 'v7-lts',
+  additionalDeps: [...baseAngular.additionalDeps, 'core-js'],
```
gaetanmaisse, line 31: I'm wondering why `core-js` is needed, can you give me some hints?

gaetanmaisse (approved):

gaetanmaisse:
Just added 2 commits to fix Yarn 2 tests run, it was failing because we were trying to init project with a name invalid for NPM. Everything should be 🟢 .

Good job @tooppaaa 💪 👏

scripts/run-e2e-config.ts
tooppaaa: It's not "needed", but v7 (didn't checked above) installs core-js v2 which cause some issues with storybook

PR #11360, opened 2020-06-29
UI: Add version release notes

kylesuss:
# This depends on https://github.com/storybookjs/frontpage/pull/107

Issue: #11232 

## What I did

This PR sets up the release notes logic in the build process & also in the ui.

## How to test

- **Is this testable with Jest or Chromatic screenshots?**

Yeah. I added some stories for the release notes page states, though I left the iframe out as that will likely change quite often.

- **Does this need a new example in the kitchen sink apps?**

Not sure. I added a new CLI flag for disabling the auto-messaging around release notes. Does that warrant a new example?

- **Does this need an update to the documentation?**

There is a new CLI flag, so that should be documented. Does that happen automatically though? I couldn't quite tell. I added it to `lib/core/src/server/cli/dev.js` with a note, which seemed to be documentation, but not sure where that shows up.

### DIFF
--- a/lib/api/src/modules/release-notes.ts
+++ b/lib/api/src/modules/release-notes.ts
@@ -1,0 +1,60 @@
+import { RELEASE_NOTES_DATA } from 'global';
+import memoize from 'memoizerific';
+
+import { ModuleFn, State } from '../index';
+
+export interface ReleaseNotes {
+  success?: boolean;
+  currentVersion?: string;
+  showOnFirstLaunch?: boolean;
+}
+
+const getReleaseNotesData = memoize(1)(
+  (): ReleaseNotes => {
+    try {
+      return { ...(JSON.parse(RELEASE_NOTES_DATA) || {}) };
+    } catch (e) {
+      return {};
+    }
+  }
+);
+
+export interface SubAPI {
+  releaseNotesVersion: () => string;
+  showReleaseNotesOnLaunch: () => Promise<boolean>;
+}
+
+export const init: ModuleFn = ({ fullAPI, store }) => {
+  let initStoreState: Promise<State[]>;
+  const releaseNotesData = getReleaseNotesData();
+
+  const api: SubAPI = {
+    releaseNotesVersion: () => releaseNotesData.currentVersion,
+    showReleaseNotesOnLaunch: async () => {
+      // Make sure any consumers of this function's return value have waited for
+      // this module's state to have been setup. Otherwise, it may be called
+      // before the store state is set and therefore have inaccurate data.
+      await initStoreState;
+      const { showReleaseNotesOnLaunch } = store.getState();
+      return showReleaseNotesOnLaunch;
+    },
+  };
+
+  const initModule = () => {
+    const { releaseNotesViewed: persistedReleaseNotesViewed } = store.getState();
+    let releaseNotesViewed = persistedReleaseNotesViewed || [];
+    const didViewReleaseNotes = releaseNotesViewed.includes(releaseNotesData.currentVersion);
+    const showReleaseNotesOnLaunch = releaseNotesData.showOnFirstLaunch && !didViewReleaseNotes;
+
+    if (showReleaseNotesOnLaunch) {
+      releaseNotesViewed = [...releaseNotesViewed, releaseNotesData.currentVersion];
+    }
+
+    initStoreState = Promise.all([
+      store.setState({ showReleaseNotesOnLaunch }),
+      store.setState({ releaseNotesViewed }, { persistence: 'permanent' }),
+    ]);
+  };
+
+  return { init: initModule, api };
+};
--- a/lib/api/src/modules/release-notes.ts
+++ b/lib/api/src/modules/release-notes.ts
@@ -1,7 +1,7 @@
 import { RELEASE_NOTES_DATA } from 'global';
 import memoize from 'memoizerific';
 
-import { ModuleFn, State } from '../index';
+import { ModuleFn } from '../index';
 
 export interface ReleaseNotes {
   success?: boolean;
@@ -21,40 +21,38 @@ const getReleaseNotesData = memoize(1)(
 
 export interface SubAPI {
   releaseNotesVersion: () => string;
-  showReleaseNotesOnLaunch: () => Promise<boolean>;
+  setDidViewReleaseNotes: () => void;
+  showReleaseNotesOnLaunch: () => boolean;
 }
 
-export const init: ModuleFn = ({ fullAPI, store }) => {
-  let initStoreState: Promise<State[]>;
+export const init: ModuleFn = ({ store }) => {
   const releaseNotesData = getReleaseNotesData();
+  const getReleaseNotesViewed = () => {
+    const { releaseNotesViewed: persistedReleaseNotesViewed } = store.getState();
+    return persistedReleaseNotesViewed || [];
+  };
 
   const api: SubAPI = {
     releaseNotesVersion: () => releaseNotesData.currentVersion,
-    showReleaseNotesOnLaunch: async () => {
-      // Make sure any consumers of this function's return value have waited for
-      // this module's state to have been setup. Otherwise, it may be called
-      // before the store state is set and therefore have inaccurate data.
-      await initStoreState;
-      const { showReleaseNotesOnLaunch } = store.getState();
+    setDidViewReleaseNotes: () => {
+      const releaseNotesViewed = getReleaseNotesViewed();
+
+      if (!releaseNotesViewed.includes(releaseNotesData.currentVersion)) {
+        store.setState(
+          { releaseNotesViewed: [...releaseNotesViewed, releaseNotesData.currentVersion] },
+          { persistence: 'permanent' }
+        );
+      }
+    },
+    showReleaseNotesOnLaunch: () => {
+      const releaseNotesViewed = getReleaseNotesViewed();
+      const didViewReleaseNotes = releaseNotesViewed.includes(releaseNotesData.currentVersion);
+      const showReleaseNotesOnLaunch = releaseNotesData.showOnFirstLaunch && !didViewReleaseNotes;
       return showReleaseNotesOnLaunch;
     },
   };
 
-  const initModule = () => {
-    const { releaseNotesViewed: persistedReleaseNotesViewed } = store.getState();
-    let releaseNotesViewed = persistedReleaseNotesViewed || [];
-    const didViewReleaseNotes = releaseNotesViewed.includes(releaseNotesData.currentVersion);
-    const showReleaseNotesOnLaunch = releaseNotesData.showOnFirstLaunch && !didViewReleaseNotes;
-
-    if (showReleaseNotesOnLaunch) {
-      releaseNotesViewed = [...releaseNotesViewed, releaseNotesData.currentVersion];
-    }
-
-    initStoreState = Promise.all([
-      store.setState({ showReleaseNotesOnLaunch }),
-      store.setState({ releaseNotesViewed }, { persistence: 'permanent' }),
-    ]);
-  };
+  const initModule = () => {};
 
   return { init: initModule, api };
 };
--- a/lib/api/src/modules/settings.ts
+++ b/lib/api/src/modules/settings.ts
@@ -1,0 +1,49 @@
+import { ModuleFn } from '../index';
+
+export interface SubAPI {
+  changeSettingsTab: (tab: string) => void;
+  closeSettings: () => void;
+  isSettingsScreenActive: () => boolean;
+  navigateToSettingsPage: (path: string) => Promise<void>;
+}
+
+export const init: ModuleFn = ({ store, navigate, fullAPI }) => {
+  const isSettingsScreenActive = () => {
+    const { path } = fullAPI.getUrlState();
+    return !!(path || '').match(/^\/settings/);
+  };
+  const api: SubAPI = {
+    closeSettings: () => {
+      const {
+        settings: { lastTrackedStoryId },
+      } = store.getState();
+
+      if (lastTrackedStoryId) {
+        fullAPI.selectStory(lastTrackedStoryId);
+      } else {
+        fullAPI.selectFirstStory();
+      }
+    },
+    changeSettingsTab: (tab: string) => {
+      navigate(`/settings/${tab}`);
+    },
+    isSettingsScreenActive,
+    navigateToSettingsPage: async (path) => {
+      if (!isSettingsScreenActive()) {
+        const { settings, storyId } = store.getState();
+
+        await store.setState({
+          settings: { ...settings, lastTrackedStoryId: storyId },
+        });
+      }
+
+      navigate(path);
+    },
+  };
+
+  const initModule = async () => {
+    await store.setState({ settings: { lastTrackedStoryId: null } });
+  };
+
+  return { init: initModule, api };
+};
--- a/lib/api/src/modules/stories.ts
+++ b/lib/api/src/modules/stories.ts
@@ -226,17 +226,16 @@ export const init: ModuleFn = ({
     },
     selectFirstStory: () => {
       const { storiesHash } = store.getState();
-      const lookupList = Object.keys(storiesHash).filter(
+      const firstStory = Object.keys(storiesHash).find(
         (k) => !(storiesHash[k].children || Array.isArray(storiesHash[k]))
       );
-      const firstStory = lookupList[0];
 
       if (firstStory) {
         api.selectStory(firstStory);
         return;
       }
 
-      navigate('/story/*');
+      navigate('/');
     },
     selectStory: (kindOrId, story = undefined, options = {}) => {
       const { ref, viewMode: viewModeFromArgs } = options;
--- a/lib/api/src/modules/stories.ts
+++ b/lib/api/src/modules/stories.ts
@@ -46,6 +46,7 @@ export interface SubState {
 export interface SubAPI {
   storyId: typeof toId;
   resolveStory: (storyId: StoryId, refsId?: string) => Story | Group | Root;
+  selectFirstStory: () => void;
   selectStory: (
     kindOrId: string,
     story?: string,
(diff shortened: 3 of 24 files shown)


lib/api/src/modules/settings.ts
```diff
@@ -0,0 +1,41 @@
...
+  const api: SubAPI = {
+    closeSettings: () => {
+      const {
+        settings: { lastTrackedStoryId },
+      } = store.getState();
+
+      if (lastTrackedStoryId) {
+        fullAPI.selectStory(lastTrackedStoryId);
+      } else {
+        navigate('/story/*');
```
tmeasday, line 19: Does this have any effect? I didn't think that actually led to the first story being selected any more..

lib/api/src/modules/stories.ts
```diff
@@ -268,10 +268,17 @@ export const init: ModuleFn = ({
...
+    fullAPI.on(CURRENT_STORY_WAS_SET, async function handleCurrentStoryWasSet({
+      storyId,
+      viewMode,
+    }) {
       const { source }: { source: string } = this;
       const [sourceType] = getSourceType(source);
 
+      if ((await fullAPI.showReleaseNotesOnLaunch()) && !fullAPI.didViewReleaseNotes()) {
+        return;
+      }
```
tmeasday, line 280: Should this be more general? I can't imagine it is a good idea to close any settings page randomly when the SB loads.

lib/api/src/modules/url.ts
```diff
@@ -154,15 +137,35 @@ export const init: ModuleFn = ({ store, navigate, state, provider, fullAPI, ...r
...
     fullAPI.on(NAVIGATE_URL, (url: string, options: { [k: string]: any }) => {
       fullAPI.navigateUrl(url, options);
     });
   };
 
   return {
     api,
-    state: initialUrlSupport({ store, navigate, state, provider, fullAPI, ...rest }),
+    state: initialUrlSupport({ store, state, provider, fullAPI, ...rest }),
     init: initModule,
```
tmeasday, line 169: What's the idea behind moving this? I can't say I fully understand why this is where it is (or what's up with the `setTimeout(..., 1)`s, but can you explain why it's moving?

lib/api/src/modules/settings.ts
```diff
@@ -0,0 +1,41 @@
...
+      navigate(`/settings/${tab}`);
+    },
+    navigateToSettingsPage: async (path) => {
+      const { settings, storyId } = store.getState();
+
+      await store.setState({
+        settings: { ...settings, lastTrackedStoryId: storyId },
+      });
+
+      navigate(path);
```
tmeasday, line 32: I guess this means we decided we'd use URLs for settings after all?

lib/api/src/modules/settings.ts
tmeasday: I'm not sure what the best way to do it otherwise is 🤔

lib/api/src/modules/release-notes.ts
```diff
@@ -0,0 +1,94 @@
...
+        new Set([
+          ...releaseNotes.releaseNotesViewed,
+          ...(releaseNotesData.success ? [releaseNotesData.currentVersion] : []),
+        ])
+      );
+
+      await store.setState(
+        { releaseNotesViewed: uniqueVersionsViewed },
+        { persistence: 'permanent' }
+      );
```
tmeasday, line 67: I thought we were going to keep things simple and not track any persistent state in the browser. Did that change?

kylesuss: I wanted to, but when it came down to implementation, I wasn't sure how to do that & prevent the release notes from re-opening every time you re-loaded your Storybook (a browser refresh). It seemed like I needed to track that on the client to prevent them from getting "redirected" to the release notes every time they refreshed the page. Do you have any ideas on how to prevent doing this?

lib/api/src/modules/settings.ts
kylesuss: Yeah. I mean -- I'm quite neutral about it. I figured if the experience would work & be linkable as well, that it wouldn't hurt to keep the URLs around. Any thoughts on that?

lib/api/src/modules/url.ts
kylesuss: The whole `setTimeout` thing feels like a hack. TBH I am not sure why it is being done that way. I suspect it is some workaround for a timing issue related to postMessage events between the manager & preview. That said, I moved this stuff around because the `initialUrlSupport` function was triggering redirects _before_ all of the "modules" have been setup and therefore, I didn't have access to the state of the `release_notes` module to know if that required a redirect. By moving it, I could `await` the `release_notes` module & know whether or not we should redirect there.

lib/api/src/modules/stories.ts
kylesuss: Yeah, it would be nice to generalize. I think the hard part about that is that this callback handles all events for `CURRENT_STORY_WAS_SET` and therefore I can't just say "if on the settings page, don't redirect" because you might be intentionally navigating away from the settings page. Thoughts?

lib/api/src/modules/settings.ts
kylesuss: It has the effect of closing the settings page and showing this screen:

Screen Shot 2020-06-30 at 11 26 49 AM

I would have liked to navigate to the first story & spent some time trying to figure out how to do that but it seemed way more complicated than this. But, if anyone knows how to do that, I can change it up.

lib/api/src/modules/url.ts
kylesuss: That said, I am slightly confused as there are a bunch of redirects happening & given what I had to do here:

https://github.com/storybookjs/storybook/pull/11360/files#r447366304

There doesn't seem to be a dedicated place to isolate this logic.

lib/api/src/modules/settings.ts
tmeasday: We could add a `selectFirstStory` method to the stories module. 

How do you get into this state?

lib/api/src/modules/stories.ts
tmeasday: Should we make it the direct action that closes the settings, rather than the side effect? 

E.g close the settings when you click the close button or a story in the sidebar, instead of when the story change happens in the sidebar. 

I actually think this’ll happen pretty naturally if you block the routing here when settings is open. Here’s my thinking. There are two ways this event can fire:

1. Preview boots (at the very start OR on hot module reload)

2. In response to the manager selecting a new story

This code is here only for case 1, as:

If we break down what happens in 2:

A. Something triggers `SELECT_STORY`
B. That makes the manager change route (this would close settings)
C. Manager emits `SET_CURRENT_STORY` in response to route change
D. Preview changes story, emits `CURRENT_STORY_WAS_SET` which hits this code path
E. We try and route again to where we already are, which is a no-op

It’s a bit confusing for sure but hopefully that helps?

lib/api/src/modules/settings.ts
tmeasday: Sure, I’m not wedded to either approach.

lib/api/src/modules/release-notes.ts
tmeasday: Ahh, got it. Yeah that makes sense. Thanks for the clarification

lib/api/src/modules/settings.ts
kylesuss: You get into this state by triggering the release notes to open immediately when you start Storybook. When you do that, you don't have any story history & therefore there is no story to navigate "back" to.

kylesuss: `selectFirstStory` seems reasonable -- the thing I was struggling with was getting the state of the stories in order. There was a `storiesHash` but it wasn't sorted. I'm looking now at `jumpToStory` and think I can copy that logic to sort & find the first story.

lib/core/src/server/build-dev.js
```diff
@@ -124,6 +124,78 @@ const updateCheck = async (version) => {
...
+  return {
+    success: false,
+    currentVersion: getReleaseNotesVersion(version),
+    showOnFirstLaunch: false,
+  };
+};
+
+export const RELEASE_NOTES_CACHE_KEY = 'releaseNotesData';
+
+export const getReleaseNotesData = async (currentVersionToParse, fileSystemCache) => {
```
kylesuss, line 147: Of note here is that I decided to pass the cache in so it would be easier to mock when testing. I can never figure out how to mock a node module w/ jest & change the return value in each test.

lib/api/src/modules/settings.ts
kylesuss: Updated & got `selectFirstStory` to work 👌

lib/api/src/modules/url.ts
kylesuss: @tmeasday I moved these redirects back for now and commented them out. I was planning to take another look at how it affects things once you've had the chance to look into removing them.

lib/api/src/modules/stories.ts
kylesuss: Updated w/ a reference to the new settings module to determine if it is open. Seems to work well (though I have all of those `setTimeout` navigate calls commented out) -- I just need to circle back once those are removed.

lib/core/src/server/build-dev.js
lifeiscontent: @kylesuss have you tried https://jestjs.io/docs/en/manual-mocks ?

lib/api/src/modules/url.ts
```diff
@@ -79,19 +79,19 @@ const initialUrlSupport = ({
...
+  // } else if (selectedKind) {
+  //   // Create a "storyId" of the form `kind-sanitized--*`
+  //   const standInId = toId(selectedKind, 'star').replace(/star$/, '*');
+  //   setTimeout(() => navigate(`/${viewMode}/${standInId}`, { replace: true }), 1);
+  // } else if (!queryPath || queryPath === '/') {
+  //   setTimeout(() => navigate(`/${viewMode}/*`, { replace: true }), 1);
+  // } else if (Object.keys(query).length > 1) {
+  //   // remove other queries
+  //   setTimeout(() => navigate(`${queryPath}`, { replace: true }), 1);
+  // }
```
kylesuss, line 94: Some tests are failing because of this, which I expect.

lib/core/src/server/build-dev.js
kylesuss: I didn't. Mostly a lot of combinations of `jest.mock`, `jest.doMock`, `jest.mockReturnValue`, etc.

> If the module you are mocking is a Node module (e.g.: lodash), the mock should be placed in the __mocks__ directory adjacent to node_modules (unless you configured roots to point to a folder other than the project root) and will be automatically mocked. There's no need to explicitly call jest.mock('module_name').

Do we want to mock this module all the time? That seems a bit heavy handed.

tmeasday: @kylesuss something like

```js
import foo from 'bar';

jest.mock('bar');

it('...', () => {
  foo.mockReturnValue('...');
});
```

Is that what you mean?

lib/api/src/modules/release-notes.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const initModule = () => {
+    const { releaseNotesViewed: persistedReleaseNotesViewed } = store.getState();
+    let releaseNotesViewed = persistedReleaseNotesViewed || [];
+    const didViewReleaseNotes = releaseNotesViewed.includes(releaseNotesData.currentVersion);
+    const showReleaseNotesOnLaunch = releaseNotesData.showOnFirstLaunch && !didViewReleaseNotes;
+
+    if (showReleaseNotesOnLaunch) {
+      releaseNotesViewed = [...releaseNotesViewed, releaseNotesData.currentVersion];
+    }
```
tmeasday, line 51: Is there a reason this logic needs to be here and not inside `showReleaseNotesOnLaunch`?

lib/api/src/modules/release-notes.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    initStoreState = Promise.all([
+      store.setState({ showReleaseNotesOnLaunch }),
+      store.setState({ releaseNotesViewed }, { persistence: 'permanent' }),
```
tmeasday, line 55: Maybe we should just write this when the release notes page renders?

I think that would avoid the issue you are running into here with the `async`-ness

lib/api/src/modules/release-notes.ts
```diff
@@ -0,0 +1,60 @@
...
+  showReleaseNotesOnLaunch: () => Promise<boolean>;
+}
+
+export const init: ModuleFn = ({ fullAPI, store }) => {
+  let initStoreState: Promise<State[]>;
+  const releaseNotesData = getReleaseNotesData();
+
+  const api: SubAPI = {
+    releaseNotesVersion: () => releaseNotesData.currentVersion,
+    showReleaseNotesOnLaunch: async () => {
```
tmeasday, line 33: This is sort of an unusual pattern. I feel like it's pretty confusing to have a `async` function called from a module's `initModule` given that we don't actually `await` initModule...

lib/api/src/modules/stories.ts
```diff
@@ -197,6 +198,20 @@ export const init: ModuleFn = ({
         storiesFailed: error,
       });
     },
+    selectFirstStory: () => {
+      const { storiesHash } = store.getState();
+      const lookupList = Object.keys(storiesHash).filter(
+        (k) => !(storiesHash[k].children || Array.isArray(storiesHash[k]))
+      );
```
tmeasday, line 205: Can we just use a `find` here rather than filtering the whole list?

lib/api/src/modules/stories.ts
```diff
@@ -197,6 +198,20 @@ export const init: ModuleFn = ({
...
+      const firstStory = lookupList[0];
+
+      if (firstStory) {
+        api.selectStory(firstStory);
+        return;
+      }
+
+      navigate('/story/*');
```
tmeasday, line 213: This doesn't do anything, I don't think. If there's no stories at all I think should just set the URL to `/`.

tmeasday (changes requested):
Some comments about modules

lib/core/src/server/build-dev.js
kylesuss: Yeah I tried that.

```js
import Cache from 'file-system-cache';
import { getReleaseNotesData, RELEASE_NOTES_CACHE_KEY } from './build-dev';

jest.mock('file-system-cache');

it('handles errors gracefully', async () => {
  Cache.mockReturnValue(() => ({}));
  // The rest of the test
});
```

The `cache` in the module is `undefined`.

This works:

```js
jest.mock('file-system-cache', () => ({
  __esModule: true,
  default: () => ({}),
}));
```

But it means every test will have the same return value. I can't seem to override it with `mockReturnValue` inside the test.

Jest's docs [suggest](https://jestjs.io/docs/en/jest-object#jestdomockmodulename-factory-options) using `doMock` in these cases for ES6 imports so I tried this:

```js
it('handles errors gracefully', async () => {
    const version = '4.0.0';
    jest.doMock('file-system-cache', () => ({
      __esModule: true,
      default: jest.fn(() => ({})),
    }));

    return import('file-system-cache').then(async () => {
      expect(await getReleaseNotesData(version)).toEqual({
        currentVersion: version,
        showOnFirstLaunch: false,
        success: false,
      });
    });
  });
```

But `cache` doesn't end up having the mock value -- it is just the module's normal FileSystemCache object.

So I just tried manual mocks as suggested by @lifeiscontent:

```js
// __mocks__/file-system-cache.js
module.exports = jest.fn(() => ({ a: 'b' }));

// build-dev.test.js
it('handles errors gracefully', async () => {
  Cache.mockReturnValue(() => ({}));
  // The rest of the test
});
```

And the value is `({ a: 'b' })`.

Haha. So I donno -- maybe I am missing something but I have tried quite a lot.

lib/api/src/modules/stories.ts
kylesuss: Yeah that makes sense 👍  I just blindly copied the logic from the other function.

lib/api/src/modules/stories.ts
kylesuss: https://github.com/storybookjs/storybook/pull/11360/commits/5e1e0908668e0f4d50ecfe11afc7e29e1b7d86d3

lib/api/src/modules/stories.ts
kylesuss: https://github.com/storybookjs/storybook/pull/11360/commits/5e1e0908668e0f4d50ecfe11afc7e29e1b7d86d3

lib/api/src/modules/release-notes.ts
kylesuss: Hmm. Good point. I guess not. Moving it there means I have no logic inside of `initModule`, which feels kind of weird but maybe that's fine. I was trying to follow a pattern of defining the state inside of `initModule` but only because I thought that was how the other modules were doing it. I went ahead and moved it to `showReleaseNotesOnLaunch` 👌

lib/api/src/modules/stories.ts
tmeasday: Heh yeah I figured :)

tmeasday (approved):
LGTM once the timeouts pr is merged!

lib/ui/src/settings/about_page.tsx
```diff
@@ -19,16 +19,18 @@ class NotificationClearer extends Component<{ api: API; notificationId: string }
...
           latest={api.getLatestVersion()}
-          onClose={() => history.back()}
+          onClose={onClose}
         />
       </NotificationClearer>
     )}
   </Consumer>
 );
+
+export { AboutPage as default };
```
shilman, line 36: since this was written we've switched to named exports (as opposed to default exports), so please export it as AboutPage and update the reference.

lib/ui/src/settings/shortcuts_page.tsx
```diff
@@ -1,20 +1,22 @@
...
         shortcutKeys={getShortcutKeys()}
         {...{ setShortcut, restoreDefaultShortcut, restoreAllDefaultShortcuts }}
-        onClose={() => history.back()}
+        onClose={onClose}
       />
     )}
   </Consumer>
 );
+
+export { ShortcutsPage as default };
```
shilman, line 22: use named exports

lib/api/src/modules/url.ts
tmeasday: My PR is merged now!

lib/api/src/modules/url.ts
kylesuss: No longer relevant!

lib/api/src/modules/url.ts
kylesuss: Nice. I merged that in 👌

kylesuss:
@tmeasday @shilman ok this is up to date w/ @tmeasday 's changes for the URL module & the named export stuff has been addressed. Let me know what else needs addressing & I can handle tomorrow.

tmeasday (approved):
LGTM @kylesuss 

I guess you should approve the Chromatic changes.

Also I am not sure what's up with CI, you might want to try merging latest `next`.

kylesuss:
@tmeasday @shilman this seems good to go, pending UI Review. Feel free to merge once https://github.com/storybookjs/frontpage/pull/107 is merged.

PR #11364, opened 2020-06-30
Add/composition auto disable

ndelangen:
Issue: -auto refs are **always added** users might not want that-

## What I did

I added some code that would deal with:
```js
module.exports = {
  refs: {
    'my-auto-loaded-ref-id': null,
  },
};
```

This would remove the ref that was auto-injected

### DIFF
--- a/lib/core/src/server/manager/manager-config.js
+++ b/lib/core/src/server/manager/manager-config.js
@@ -70,7 +70,7 @@ async function getManagerWebpackConfig(options, presets) {
 
   if (definedRefs) {
     Object.entries(definedRefs).forEach(([key, value]) => {
-      if (value === null) {
+      if (value?.disabled) {
         delete refs[key];
         return;
       }
--- a/lib/core/src/server/manager/manager-config.js
+++ b/lib/core/src/server/manager/manager-config.js
@@ -70,6 +70,10 @@ async function getManagerWebpackConfig(options, presets) {
 
   if (definedRefs) {
     Object.entries(definedRefs).forEach(([key, value]) => {
+      if (value === null) {
+        delete refs[key];
+      }
+
       const url = typeof value === 'string' ? value : value.url;
       const rest =
         typeof value === 'string'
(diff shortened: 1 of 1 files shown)


tmeasday (changes requested):
Let’s use `{disabled: true}` for consistency with addons and to avoid confusion with accidental falsey values (<- this is something you originally recommended with addons ;))

ndelangen:
Good call!

tmeasday:
Uggh sorry this should have been `disable: true`. Will fix

lib/core/src/server/manager/manager-config.js
```diff
@@ -70,6 +70,11 @@ async function getManagerWebpackConfig(options, presets) {
 
   if (definedRefs) {
     Object.entries(definedRefs).forEach(([key, value]) => {
+      if (value?.disabled) {
```
ndelangen, line 73:
```
	      if (value?.disabled === true) {
```

oops

hems:
> Uggh sorry this should have been `disable: true`. Will fix

I'm using `"@storybook/react": "^6.0.26"` and when i set `disable: true` i'm getting the following error:

````bash
ERR! TypeError: Cannot read property 'replace' of undefined
ERR!     at stripTrailingSlash (.../node_modules/@storybook/core/dist/server/manager/manager-config.js:93:14)
ERR!     at .../node_modules/@storybook/core/dist/server/manager/manager-config.js:147:14
ERR!     at Array.forEach (<anonymous>)
ERR!     at getManagerWebpackConfig (.../node_modules/@storybook/core/dist/server/manager/manager-config.js:132:33)
ERR!  TypeError: Cannot read property 'replace' of undefined
ERR!     at stripTrailingSlash (.../node_modules/@storybook/core/dist/server/manager/manager-config.js:93:14)
ERR!     at .../node_modules/@storybook/core/dist/server/manager/manager-config.js:147:14
ERR!     at Array.forEach (<anonymous>)
ERR!     at getManagerWebpackConfig (/home/hz/Documents/git/aluna/aluna-next/node_modules/@storybook/core/dist/server/manager/manager-config.js:132:33)
````

this is my main.js file:

```js
module.exports = {
  stories: ['../src/**/*.stories.@(js|ts|jsx|tsx|mdx)'],
  addons: [
    '@storybook/addon-essentials',
    'storybook-addon-styled-component-theme/dist/register'
  ],
  refs: {
    '@chakra-ui/core': {
      disable: true
    } 
  }
}
````

Any ideas?

PR #11392, opened 2020-07-02
Composition: Change event source to ref

ndelangen:
Issue: #11371

## What I did

- [x] Render iframes with `&refId=$refId`.
- [x] Pass refId over the channel when emitting events
- [x] Match events to refs first via refId, falling back to postMessage event origin.

@tmeasday I removed a least code as possible.

I could delete more of the compatibility code dealing with `source`.


lib/channel-postmessage/src/index.ts
```diff
@@ -14,6 +14,16 @@ interface BufferedEvent {
...
+const getQueryObject = (): Record<string, string & { refId?: string }> => {
+  return (window.document.location.search as string)
+    .substr(1)
+    .split('&')
+    .reduce((acc, s) => {
+      const [k, v] = s.split('=');
+      return Object.assign(acc, { [k]: v || true });
+    }, {});
+};
+
```
tmeasday, line 26: You don't want to use `querystring` for this? It is a built-in.

lib/api/src/modules/stories.ts
```diff
@@ -72,6 +71,61 @@ export const init: ModuleFn = ({
...
+    function handle(data: T) {
+      const { source, refId, type }: Meta = this;
+      const [sourceType, sourceLocation] = getSourceType(source, refId);
+      // @ts-ignore
+      const v = data.v || this.v;
+
+      if (v > 2) {
+        // eslint-disable-next-line no-console
+        console.warn(`Received ${this.type} event with version ${v}, we'll try and handle it`);
+      }
```
tmeasday, line 95: This bit is specific to `SET_STORIES`

lib/api/src/modules/stories.ts
```diff
@@ -268,105 +322,63 @@ export const init: ModuleFn = ({
...
-
-      if (sourceType === 'local') {
-        const options = fullAPI.getCurrentParameter('options');
-
-        if (options) {
-          fullAPI.setOptions(options);
+    fullAPI.on(
+      CURRENT_STORY_WAS_SET,
+      createEventHandlerWithRefSupport<{ storyId: string; viewMode: ViewMode; [k: string]: any }>(
+        ({ storyId, viewMode }, { sourceType }) => {
```
tmeasday, line 328: This feels like an unnecessary level of indirection/complexity. Couldn't we just do this:
```suggestion
    fullAPI.on(CURRENT_STORY_WAS_SET, ({ storyId, viewMode }) => {
      const { sourceType } = getRefFromEvent(this);
```

tmeasday (changes requested):
Good stuff. I think we can simplify slightly.

Also I strongly suggest removing the `source` thing. What does it gain us?

lib/channel-postmessage/src/index.ts
ndelangen: This code doesn't run in Node?

lib/api/src/modules/stories.ts
ndelangen: Ow is it? I thought ALL events were going to be versioned?

lib/api/src/modules/stories.ts
ndelangen: `this` can only be used in non-arrow-functions; Sure, I can make that change everywhere.

lib/channel-postmessage/src/index.ts
tmeasday: You can use node libraries in webpack mate ;)

lib/api/src/modules/stories.ts
tmeasday: One day maybe.. but the version we support for a given event may change.. this is very much a event specific check.

lib/api/src/modules/stories.ts
tmeasday: Oh, yep, but you get the idea.

lib/channel-postmessage/src/index.ts
ndelangen: I switched to `qs` since that's what we use everwhere for this purpose.

I tried to import `querystring` but it is missing typescript definitions...

lib/api/src/modules/stories.ts
ndelangen: I do 👍

ndelangen: done, WDYT?

tmeasday (approved):
Looks great @ndelangen !

I think maybe we need to update the stories module tests so they have a ref example?

ndelangen:
@tmeasday do you mean these test?

https://github.com/storybookjs/storybook/blob/436bbeab83adcde8ab7913d1f1ba7a92f3ee65c8/lib/api/src/tests/stories.test.js#L693-L729

tmeasday:
@ndelangen I meant this test: https://github.com/storybookjs/storybook/blob/75306dad722b0b1fd1a12d20fd31e2db215c7810/lib/api/src/tests/stories.test.js#L718

Which uses `source` rather than `refId` to identify the event source.

lib/api/src/lib/events.ts
```diff
@@ -0,0 +1,45 @@
...
+      return meta;
+    }
+
+    // if we couldn't find the source, something risky happened, we ignore the input, and log a warning
+    default: {
+      logger.warn(`Received a ${type} frame that was not configured as a ref`);
+      return null;
+    }
+  }
+};
```
ndelangen, line 45: Write test for this, for likely scenarios

shilman (approved):
And we're green! Great job guys!

PR #11401, opened 2020-07-03
Composition: Ensure args get sent to the right ref

tmeasday:
Issue: #10115

## What I did

Made it so that args get sent to and from refs

## How to test

yes

You can also try adding this to the main.js of official-sb:

```
  refs: {
    'cra-ts-essentials': 'http://localhost:9009',
  },
```

(and running that SB too). Then you can try using args stories in both SBs.


lib/api/src/modules/stories.ts
```diff
@@ -368,10 +375,36 @@ export const init: ModuleFn = ({
...
+          (storiesHash[storyId] as Story).args = args;
+          store.setState({ storiesHash });
+          break;
+        }
+        case 'external': {
+          const refs = fullAPI.getRefs();
+          const ref = fullAPI.findRef(sourceLocation);
+          (ref.stories[storyId] as Story).args = args;
+          store.setState({ refs: { ...refs, [ref.id]: ref } });
+          break;
```
tmeasday, line 401: @ndelangen it feels a bit weird to be changing `refs` inside `modules/stories`. But then moving this logic to refs would be spreading the exact same logic (updating args) into two places 🤔 

I think we need to make refs/stories the same thing already!

ndelangen: You want to do that before the RC?

I absolutely agree that's where we want to be at some point. I was just being careful of when we do that.

lib/api/src/index.tsx
```diff
@@ -415,7 +415,14 @@ export function useArgs(): [Args, (newArgs: Args) => void] {
   const data = getCurrentStoryData();
   const args = isStory(data) ? data.args : {};
 
-  return [args, (newArgs: Args) => updateStoryArgs(data.id, newArgs)];
+  console.log('in use args');
```
ndelangen, line 418: remove

lib/api/src/modules/stories.ts
```diff
@@ -368,10 +375,36 @@ export const init: ModuleFn = ({
...
+        }
+        default: {
+          logger.warn('received a STORY_ARGS_UPDATED frame that was not configured as a ref');
+          break;
+        }
```
ndelangen, line 406: I updated the code here:
https://github.com/storybookjs/storybook/pull/11392

So there's much less boilerplate and duplicated code

lib/api/src/modules/stories.ts
tmeasday: No I don't really want to do it now. Just wondering what to do about the above code.

lib/api/src/modules/stories.ts
tmeasday: Yeah I know we can update this afterwards.

lib/api/src/modules/stories.ts
ndelangen: You could consider using fullAPI.setRef
https://github.com/storybookjs/storybook/blob/9a186bae80359c244a45ed94df47dcb566a51707/lib/api/src/modules/refs.ts#L26-L27
?

tmeasday: @ndelangen the implementation of that function expects `stories` to be passed in directly from `SET_STORIES`. Can I refactor it?

tmeasday: @ndelangen https://github.com/storybookjs/storybook/pull/11401/commits/38e66ad7cb59885eb1b6da7fd47aae7c3376a314 -- is that good for you?

lib/api/src/modules/refs.ts
```diff
@@ -213,12 +223,15 @@ export const init: ModuleFn = ({ store, provider, fullAPI }, { runCheck = true }
...
 
+    updateRef: (id, data) => {
+      const { [id]: ref, ...refs } = api.getRefs();
       store.setState({
         refs: {
-          ...api.getRefs(),
-          [id]: result,
+          ...refs,
+          [id]: { ...ref, ...data },
         },
```
ndelangen, line 235: fancy

lib/api/src/modules/stories.ts
```diff
@@ -283,8 +283,15 @@ export const init: ModuleFn = ({
...
-      fullAPI.emit(UPDATE_STORY_ARGS, id, newArgs);
+    updateStoryArgs: (story, updatedArgs) => {
+      const { id: storyId, refId } = story;
+      fullAPI.emit(UPDATE_STORY_ARGS, {
+        storyId,
+        updatedArgs,
+        options: {
+          target: refId ? `storybook-ref-${refId}` : 'storybook-preview-iframe',
+        },
+      });
```
ndelangen, line 294: 👍

lib/api/src/modules/stories.ts
```diff
@@ -397,10 +404,35 @@ export const init: ModuleFn = ({
...
+    fullAPI.on(STORY_ARGS_UPDATED, function handleStoryArgsUpdated({
+      storyId,
+      args,
+    }: {
+      storyId: StoryId;
+      args: Args;
+    }) {
+      // the event originates from an iframe, event.source is the iframe's location origin + pathname
+      const { source }: { source: string } = this;
+      const [sourceType, sourceLocation] = getSourceType(source);
```
ndelangen, line 416: this will be cleaned up a bit in my PR 👍

tmeasday: It's a race to see who gets merged first

ndelangen (approved):

shilman (approved):
Looks great

PR #11403, opened 2020-07-03
Addon-docs: Support absolute anchors when deployed at non-root

amannn:
Issue: #11404

## What I did

When Storybook is deployed at an URL like `https://example.com/storybook/component-library/v48.3.2/`, the anchors within docs content will remove the base url.

E.g. when you're on `https://example.com/storybook/component-library/v48.3.2/?path=/docs/components-icon--default` and you have a link to `/docs/components-button--default` in the documentation, the resulting link will point to `https://example.com/?path=/docs/components-button--default`.

## How to test

Not sure if there's a good way to add a test for this, as it depends on a special kind of deployment?


lib/components/src/typography/DocumentFormatting.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { FunctionComponent } from 'react';
 import { styled, CSSObject } from '@storybook/theming';
+import { window } from 'global';
```
amannn, line 3: Is this correct? I didn't know this pattern.

lib/components/src/typography/DocumentFormatting.tsx
```diff
@@ -71,10 +72,16 @@ export const Pre = styled.pre<{}>(withReset, withMargin, ({ theme }) => ({
 }));
 
 const Link: FunctionComponent<any> = ({ href: input, children, ...props }) => {
+  // If storybook is hosted at a non-root path (e.g. `/storybook/`),
+  // the base url needs to be prefixed to storybook paths.
+  let storybookBaseUrl =
+    typeof window !== 'undefined' ? window.parent.document.location.pathname : '/';
```
amannn, line 78: This access should be safe, even if the content is rendered outside of an iframe.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent):

> If a window does not have a parent, its parent property is a reference to itself.

amannn:
I'm not sure why some tests failed. Are they related to my change? Can you help me with them?

lib/components/src/typography/DocumentFormatting.tsx
ndelangen: That's fine.

shilman (approved):
Nice one @amannn!

Were you able to test it on your dev machine somehow? How about:

```
yarn bootstrap --core
yarn build-storybook
npx http-server
open localhost:9011/storybook-static
```

amannn:
@shilman Thanks for the review and your snippet!

I've tested it previously in a project of mine by editing `node_modules` 😬 , but I've just confirmed that this also works in this repo.

I've added:

```
 * [Go to color palette](/docs/basics-colorpalette--page)
```

to `lib/components/src/blocks/Description.tsx:14` and opened Storybook with:

```
open "http://localhost:8080/storybook-static/?path=/docs/docs-description--text"
```

(note that `http-server` uses port `8080` by default).

The anchor is rendered correctly and the link takes you to the color palette 🙂 .

I wasn't sure if there's any reasonable place in the code base where such an anchor would make sense, so we have a story to test this behaviour?

amannn:
@shilman is there anything I can do to get this merged?

ndelangen (approved):

PR #11407, opened 2020-07-03
Ignore globals from non-local refs

tmeasday:
Issue: #10115

## What I did

- Ignore global args messages from non-local refs
- Only send global changes to the local ref

## How to test

- Tests
- Official storybook if you add per https://github.com/storybookjs/storybook/pull/11401


shilman (approved):
Nice!! LGTM

PR #11410, opened 2020-07-03
Build: Fix CI breaks

ndelangen:
Issue: next has 2 failing tests

## What I did
- I removed a symlink to public inside cra-ts-kitchen-sink
- I added a delay when e2e testing knobs on official storybook


shilman (approved):
LGTM

PR #11431, opened 2020-07-06
Core: Deprecate configure and clearDecorators

tmeasday:
Issue: for #10005

## What I did

 - Deprecated configure
 - Stopped HMR for configure to be consistent with `main.js:stories`

## How to test

Try adding a `configure()` call to `preview.js` in official SB


shilman (approved):
Awesome!!! ❤️

PR #11444, opened 2020-07-07
Fix Yarn 2 compatibility in CLI and Essentials

gaetanmaisse:
Issue: Yarn 2 E2E tests are broken on next for a few days, let's fix them. And then I will think a bit about how to make them faster or add a Yarn 2 fail-fast test.

## What I did
 - CLI: add `@storybook/node-logger` as peerDep in Storybook for CRA generator, as of `@storybook/preset-create-react-app@3.1.3` this dep must be added to the base project directly.
 - Essentials: use `require.resolve` to ensure Yarn PnP compatibility. Files of various addons should be resolved in the context of `addon-essentials` and not in the one of SB user projects. File to load can be `preset.js`, `register.js`, or the package entry point, so we need to check all these cases as it's done in `lib/core/src/server/presets.js`.

--

Outdated, see https://github.com/storybookjs/storybook/pull/11444#issuecomment-654823786, will be done in another PR:
 - E2E: CLI now generates project with `addon-essential` so it enables `addon-viewport` by default. I added a simple test to check that viewport is properly loaded:
   - Go to Welcome Story
   - Click on Viewport button in the toolbar
   - Select "Small Mobile"
   - Check that Welcome Story is still displayed

Note: commit is here https://github.com/storybookjs/storybook/commit/3643404b9e07bbdee9d13d23f43006232e3584bc

## How to test

- CI should be 🟢

### DIFF
--- a/addons/essentials/src/index.ts
+++ b/addons/essentials/src/index.ts
@@ -1,4 +1,4 @@
-import path from 'path';
+import path, { join } from 'path';
 import { logger } from '@storybook/node-logger';
 
 interface PresetOptions {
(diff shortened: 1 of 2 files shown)


gaetanmaisse:
Arf 😞 `addon-essential` is not added yet when running `sb init` in an angular project so viewport test is failing. I will remove it, keep it in a branch and reapply it asap

ndelangen:
all green?! @gaetanmaisse

ndelangen (approved):

shilman (approved):
Looks great!!!

PR #11447, opened 2020-07-07
Composition: Add version query to composed ref

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/11321

## What I did
- When a ref is loaded from package.json, also load the version and set is as a property on ComposedRef.
- Add this version (if it exists) to the query when checking refs
- Add this version (if it exists) to the url for the iframe


shilman (approved):
LGTM

PR #11462, opened 2020-07-08
Deploy the `next` branch of frontpage too

tmeasday:
I want the `next` branch of frontpage to auto-deploy also, based on `next` SB (currently we are using `6.0-docs` while we refactor it).

I added a new build hook on netlify and a new env var to the circle ci config.


shilman (approved):
LGTM!

PR #11464, opened 2020-07-08
Remove gatsby and move everything to the top-level

tmeasday:
We are moving docs to `frontpage`. We don't need this gatsby site any more.


shilman (approved):
Didn’t look closely but structure WFM

PR #11491, opened 2020-07-10
Composition: Fix syntax on `no-cors`

tmeasday:
Issue: it's `mode: 'no-cors'` not `cors: 'no-cors'` - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch


shilman (approved):
LGTM

PR #11497, opened 2020-07-10
Composition: Don't show versions dropdown if there are no versions

ndelangen:
Issue: When metadata.json returns version, but it's an empty object, this causes a dropdown to render with 0 items

## What I did
Screenshot 2020-07-10 at 16 45 23
Screenshot 2020-07-10 at 16 45 33

This adds a check so see if there is more then 0 items in the list of versions


tmeasday (approved):
LGTM

PR #11531, opened 2020-07-13
Core: Improve translation of globs for main.js stories

daniel-ac-martin:
~~This PR should probably not be merged.~~ I've raised it to point out a failure in the globbing used by Storybook. However, it would seem that this is due to a bug in the underlying 'micromatch' library.

**Update:** A work-around has been added to fix the tests.

It adds a test case (that fails) for conservative globbing that matches
with wildcards, `*` rather than globstars, `**`.

An item is erroneously matched due what appears to be a bug in the
underlying micromatch package.

For convenience, here are the test failures:
```
Summary of all failing tests
 FAIL  lib/core/src/server/preview/to-require-context.test.js
  ● toRequireContext › matches only suitable paths - ../components/*/*.stories.js

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 3

    - Array []
    + Array [
    +   "../components/icon/node_modules/icon/Icon.stories.js",
    + ]

      169 | 
      170 |       expect(isNotMatchedForValidPaths).toEqual([]);
    > 171 |       expect(isMatchedForInvalidPaths).toEqual([]);
          |                                        ^
      172 |     });
      173 |   });
      174 | });

      at Object.<anonymous> (lib/core/src/server/preview/to-require-context.test.js:171:40)

  ● toRequireContext › matches only suitable paths - ../components/*/stories/*.js

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 3

    - Array []
    + Array [
    +   "../components/icon/node_modules/icon/stories/Icon.js",
    + ]

      169 | 
      170 |       expect(isNotMatchedForValidPaths).toEqual([]);
    > 171 |       expect(isMatchedForInvalidPaths).toEqual([]);
          |                                        ^
      172 |     });
      173 |   });
      174 | });

      at Object.<anonymous> (lib/core/src/server/preview/to-require-context.test.js:171:40)
```

### DIFF
--- a/lib/core/src/server/preview/to-require-context.test.js
+++ b/lib/core/src/server/preview/to-require-context.test.js
@@ -120,6 +120,32 @@ const testCases = [
       '../src/stories/components/Icon/Icon.tsx',
     ],
   },
+  {
+    glob: '../components/*/*.stories.js',
+    validPaths: ['../components/icon/Icon.stories.js'],
+    invalidPaths: [
+      '../components/icon/node_modules/icon/Icon.stories.js',
+      './stories.js',
+      './src/stories/Icon.stories.js',
+      './Icon.stories.js',
+      '../src/Icon.stories.mdx',
+      '../src/stories/components/Icon/Icon.stories.ts',
+      '../src/stories/components/Icon/Icon.mdx',
+    ],
+  },
+  {
+    glob: '../components/*/stories/*.js',
+    validPaths: ['../components/icon/stories/Icon.js'],
+    invalidPaths: [
+      '../components/icon/node_modules/icon/stories/Icon.js',
+      './stories.js',
+      './src/stories/Icon.stories.js',
+      './Icon.stories.js',
+      '../src/Icon.stories.mdx',
+      '../src/stories/components/Icon/Icon.stories.ts',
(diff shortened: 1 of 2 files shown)


daniel-ac-martin:
I've raised, what I believe to be a bug in micromatch's 'bash' mode here: https://github.com/micromatch/micromatch/pull/207

daniel-ac-martin:
I've added a work-around that makes the tests pass. It also probably makes the regex a bit stricter (as it reinstates `^` at the start). (Partially addressing #11530)

I've only run the unit tests though. I can't provide assurance beyond that and the CI that is running.

If you are happy with it should be merge-able.

daniel-ac-martin:
I should probably add that it's not clear to me why bash mode was enabled in the first place. It seems to me that bash mode would turn `**` into `*` which I don't think is what you want.

ndelangen:
> I should probably add that it's not clear to me why bash mode was enabled in the first place. It seems to me that bash mode would turn ** into * which I don't think is what you want.

I'd agree with that, seems like we should disable it then?

ndelangen (approved):
I think this looks great, @shilman @tmeasday what are your thoughts on this?

shilman (approved):
LGTM!

PR #11541, opened 2020-07-14
FIX missing refId on getData calls

ndelangen:
Issue: if getData is called without refId, it can fail

## What I did
- ensure getData is called with RefId from state
- ensure refId is always lowercase just like storyId
- order of refs was indeterministic, I fixed it so it's the order they are defined in.


lib/router/src/utils.ts
```diff
@@ -18,7 +18,7 @@ export const parsePath: (path: string | undefined) => StoryData = memoize(1000)(
     };
 
     if (path) {
-      const [, viewMode, refId, storyId] = path.toLowerCase().match(splitPathRegex) || [];
+      const [, viewMode, refId, storyId] = path.match(splitPathRegex) || [];
```
tmeasday, line 21: Does this mean URLs can include uppercase chars? This seems oddly inconsistent with our storyIds

ndelangen: Do you recon it's better to enforce lowercasing refId's everywhere?

tmeasday: I think so, if it's not too complicated

tmeasday: Like we could just make it a requirement from the beginning

ndelangen:
@tmeasday WDYT?

tmeasday (approved):
LGTM :+1:

PR #11549, opened 2020-07-15
Update webpack config example to include module sass files

nad182:
This change will let users know how to change storybook config to allow using both `.scss` and `.module.scss` files.

Issue:
Nowhere in the docs it explains how to configure webpack in main.js to allow users style their components with modular SASS (or modular CSS).

## What I did
Updated example in the docs.


shilman (approved):
LGTM! Thanks!

PR #11564, opened 2020-07-15
FIX the color of the menu separator

ndelangen:
The separator in the menu in dark mode was styled in the light-mode @domyen 

Screenshot 2020-07-15 at 15 12 00


domyen:
Nice find Norbert! 

While you're here, could you break the menu out into its own story? 

The tooltip has a `startOpen` prop ([example](https://github.com/storybookjs/storybook/blob/next/lib/components/src/tooltip/TooltipNote.stories.tsx)) that starts the tooltip open so we can see its contents. I think that would help prevent future regressions here especially when it comes to themes.

domyen:
The `menu:real` story doesn't seem to render the menu. [Video here](https://www.loom.com/share/78b667f643e24ebd8d9be0c7ae13e9d9).

ndelangen:
yeah, kinda intentional really :)

Since the other story already covers the expanded state?

domyen:
Heads up, [this story](https://www.chromatic.com/snapshot?appId=5a375b97f4b14f0020b0cda3&id=5f1071f0c18cba0022823352) gets captured strangely in Chromatic. Add more height?

lib/ui/src/containers/menu.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useMemo } from 'react';
 
 import { Badge } from '@storybook/components';
 import { API } from '@storybook/api';
-import { useTheme, Theme, darken } from '@storybook/theming';
+import { useTheme, Theme } from '@storybook/theming';
```
domyen, line 5: Oops I forgot to remove this sorry. Embarrassing 😅

PR #11576, opened 2020-07-16
Improve settings pages

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/11483

## What I did
- I refactored the code so the fullAPI has the minimal amount of additions to make this work, in an effort to keep the fullAPI somewhat lean.
- removed inline code
- fixed some theming issues


lib/api/src/modules/url.ts
```diff
@@ -154,10 +154,6 @@ export const init: ModuleFn = ({ store, navigate, state, provider, fullAPI, ...r
     fullAPI.on(NAVIGATE_URL, (url: string, options: { [k: string]: any }) => {
       fullAPI.navigateUrl(url, options);
     });
-
-    if (fullAPI.showReleaseNotesOnLaunch()) {
-      navigate('/settings/release-notes');
-    }
```
ndelangen, line 160: moved to the releases module (not deleted)

kylesuss: It does feel a bit odd to me that _any_ module can `navigate` when they are initialized. That is why I put this logic here -- to consolidate all of the navigation logic on startup in a place that made sense (the URL module). It is quite messy to jump around and figure out where the navigation is coming from & potentially bug prone since anyone could introduce another navigate call in another module that overrides the release module (as a result of having `init` called afterward). In fact, I spent a lot of time in this PR just figuring out where & why navigation was taking place. What are your thoughts on this topic?

lib/ui/src/containers/menu.tsx
```diff
@@ -50,10 +51,14 @@ export const useMenu = (
...
-      onClick: () => api.navigateToSettingsPage('/settings/shortcuts'),
+      onClick: () => api.navigate('/settings/shortcuts'),
       right: shortcutToHumanStringIfEnabled(shortcutKeys.shortcutsPage, enableShortcuts),
       left: <MenuItemIcon />,
-      style: { borderBottom: `4px solid ${color.mediumlight}` },
+      style: {
+        borderBottom: `4px solid ${
+          theme.base === 'light' ? theme.color.mediumlight : darken(theme.color.darkest)
+        }`,
+      },
```
kylesuss, line 61: 👍

lib/ui/src/settings/index.tsx
```diff
@@ -1,84 +1,94 @@
...
+      <Route path="release-notes">
+        <ReleaseNotesPage key="release-notes" />
+      </Route>
+      <Route path="shortcuts">
+        <ShortcutsPage key="shortcuts" />
+      </Route>
+    </Content>
+    <Route path="settings">
+      <GlobalHotKeys handlers={{ CLOSE: onClose }} keyMap={keyMap} />
+    </Route>
```
kylesuss, line 78: Why wrap this in a `Route`?

lib/ui/src/settings/index.tsx
```diff
@@ -1,84 +1,94 @@
...
-        <IconButton
-          onClick={(e: SyntheticEvent) => {
+const TabBarButton: FunctionComponent<{ id: string; title: string }> = ({ id, title }) => (
+  <Location>
+    {({ navigate, path }) => {
+      const active = path.includes(`settings/${id}`);
+      return (
+        <TabButton
+          id={`tabbutton-${id}`}
+          className={['tabbutton'].concat(active ? ['tabbutton-active'] : []).join(' ')}
```
kylesuss, line 18: Seems like we should be using the `selected` prop here? This feels brittle.

lib/ui/src/settings/release_notes_page.tsx
```diff
@@ -1,16 +1,22 @@
...
     api.setDidViewReleaseNotes();
   }, []);
 
-  return <ReleaseNotesScreen onClose={onClose} version={api.releaseNotesVersion()} />;
+  const version = api.releaseNotesVersion();
+
+  return version ? (
+    <ReleaseNotesScreen version={version} />
+  ) : (
+    <Fragment>Sorry, we don't know what version to load</Fragment>
```
kylesuss, line 18: I'm not sure this is the solution to this problem. I would guess we instead hide the release notes menu/tab/page altogether if we don't have the version.

lib/ui/src/settings/index.tsx
```diff
@@ -1,84 +1,94 @@
...
-      </div>
-    </Tabs>
-  </Wrapper>
+const Pages: FunctionComponent<{ onClose: API['closeSettings'] }> = ({ onClose }) => (
+  <Fragment>
+    <FlexBar border>
+      <TabBar role="tablist">
+        <TabBarButton id="about" title="About" />
+        <TabBarButton id="shortcuts" title="Keyboard shortcuts" />
+        <TabBarButton id="release-notes" title="Release notes" />
```
kylesuss, line 54: Release notes were intentionally put in the 2nd position according to the design.

kylesuss:
The settings pages are meant to have a white background. Previous:
Screen Shot 2020-07-22 at 10 55 42 AM

This PR:
Screen Shot 2020-07-22 at 10 57 07 AM

lib/api/src/modules/url.ts
ndelangen: There are navigate calls in the stories module as well.

lib/ui/src/settings/index.tsx
ndelangen: to optionally render that section based on location / route

ndelangen: Right, I will remove this, good call!

lib/ui/src/settings/index.tsx
ndelangen: We don't style this, the community wanted more theming options, and we wanted our theming system to stay relatively simple. this is the compromise we made

lib/ui/src/settings/release_notes_page.tsx
ndelangen: yeah I'll address this in higher up the component tree

lib/ui/src/settings/index.tsx
ndelangen: will move it!

kylesuss (commented):
LGTM for the most part. @ndelangen I added two commits to try and help close this out:

Prevent navigation to release notes on launch for non-dev builds:
https://github.com/storybookjs/storybook/pull/11576/commits/0a3a27b0bab70175adeb14737e82ca67ea5af98b

Hide release notes from dropdown menu for non-dev builds:
https://github.com/storybookjs/storybook/pull/11576/commits/62bf23492232c35f6fde028e98e3a68ecfb9e603

Can you verify these @ndelangen? Once you have I am 👍  on your work here!

lib/ui/src/containers/menu.tsx
```diff
@@ -41,7 +41,7 @@ export const useMenu = (
     () => ({
       id: 'release-notes',
       title: 'Release notes',
-      onClick: () => api.navigateToSettingsPage('/settings/release-notes'),
+      onClick: () => api.navigate('/settings/release-notes'),
       left: <MenuItemIcon />,
     }),
```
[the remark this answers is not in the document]
kylesuss: I added this logic here: https://github.com/storybookjs/storybook/commit/62bf23492232c35f6fde028e98e3a68ecfb9e603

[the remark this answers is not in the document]
ndelangen: right, think so too

ndelangen:
Is this PR salvageable?

kylesuss:
I think so, @ndelangen . The intent was to get much of it into a release after 6.0. There were some bugs w/ the `lastSuccessfulStoryPath` implementation that were the main cause for it not being pulled in (mainly that the release notes weren't opening when Storybook launched). Those are still likely an issue, so I would say most things are salvageable but that still needs a bit of work (or can be left out).

tmeasday:
There's the related issue (I think) I created about changing the router to deal in something more abstract (`path`/`subPath`) and storing only *real* `storyId`/`viewMode` in the state. Then we could leave the "last successful" one in the state without too much trouble.

PR #11579, opened 2020-07-16
FIX scrollbars in flexbar

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/11283 https://github.com/storybookjs/storybook/issues/11071

## What I did
I fixed the missing custom scrollbar on the FlexBar component

### DIFF
--- a/lib/components/src/ScrollArea/ScrollArea.tsx
+++ b/lib/components/src/ScrollArea/ScrollArea.tsx
@@ -11,7 +11,7 @@ export interface ScrollProps {
 }
 
 const Scroll = styled(({ vertical, horizontal, ...rest }: ScrollProps) => (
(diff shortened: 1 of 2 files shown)


domyen:
Nice work @ndelangen! 

The last thing is that the handle is missing when I hover over the bar. That makes it impossible for me to pan horizontally if I don't activate the scrolling first. Video below.

https://www.loom.com/share/4ddc5f68d9614a75b37bc645fad07f7c

ndelangen:
Does any other place we use the custom scrollbars exhibit that described behavior?

ndelangen:
K, I found the option in the overlayscrollbars library we use

domyen:
nice lgtm

Issue #11581, opened 2020-07-16
Jest Results not showing in HTML example


GiraffeCoding:
Hi,

I'm trying to get a working example of the `@storybook/addon-jest` working within the HTML example. The JSON results file exists in the file directory but the addon doesn't appear to ever find it. 

You can see a live example of this in storybooks own demo for the HTML Kitchen sink

![image](https://user-images.githubusercontent.com/12425975/87680622-8f72fc00-c775-11ea-9390-c44733d74221.png)

https://storybookjs.netlify.app/html-kitchen-sink/?path=/story/addons-jest--with-tests

Has anyone gotten this feature to work in the HTML framework?

yannbf:
Hey @GiraffeCoding, by opening this issue you helped us detect that our example file was actually wrong, so thanks a lot for that!

There's nothing wrong in the addon itself, it's just the way the parameters are passed in our example files. I already opened a PR to fix that, but in the meantime what you have to do is to pass the parameters in the following way:

```jsx
export const WithTests = () => 'This story shows test results';
WithTests.story = {
  parameters: { jest: 'addon-jest' },
};
```

Note that this is if you're using Storybook 5. If you're already in the Storybook 6 bandwagon, you should use `WithTests.parameters = { jest: 'addon-jest' };`

GiraffeCoding:
Hi @yannbf,

thank you for your response :)

I also managed to get the mdx version working as well which is shown below:

```
import { Story, Meta } from '@storybook/addon-docs/blocks';
import { withKnobs } from '@storybook/addon-knobs';
import { withTests } from '@storybook/addon-jest';
import results from './addon-jest.testresults.json';

<Meta title="Components/Sectors" decorators={[withKnobs, withTests({results})]} parameters={{options: {showPanel: true, isToolshown: true }}} />

<Story name="designOne" parameters={{jest: 'designOne' }}>
```

yannbf:
Really happy that you managed to get that solved! I'll be closing this issue as it has been resolved. 
Have a great day 🕺

FacePalmDev:
Where does WithTests.story go? in what file? Could you give a little more context?

yannbf:
> Where does WithTests.story go? in what file? Could you give a little more context?

I'm not sure I understand your question, but if it's because you want  to know how to use the addon, you can check this link: https://github.com/storybookjs/storybook/tree/main/addons/jest

FacePalmDev:
[...]

PR #11596, opened 2020-07-17
fix(html-examples): use correct parameters syntax

yannbf:
Issue: #11581

## What I did
In 5.3 we use the story syntax rather than `.parameters` as a static property, thus I fixed its usage in the html-kitchensink.


shilman (approved):
thanks!

PR #11643, opened 2020-07-22
Core: Remove duplicate decorators and warn

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/11623

## What I did

Ensure the same decorator is never added twice (to the global decorators array)

## How to test

Test should pass


shilman (approved):
LGTM

PR #11647, opened 2020-07-22
Core: Fix webpack recursion in mainjs glob processing logic

daniel-ac-martin:
`../base-dir/*/index.stories.jsx` will lead to `glob` being set to `*/index.stories.jsx` and so `glob.split('/').length` = 2.

With the existing code the `recursive` flag would _not_ be set and yet webpack will need to recurse into the directories in order to find any `index.stories.jsx` files that might exist within them.

Therefore, we should compare to 1 and not 2.

See: https://github.com/storybookjs/storybook/issues/11530#issuecomment-662465689

### DIFF
--- a/lib/core/src/server/preview/to-require-context.test.js
+++ b/lib/core/src/server/preview/to-require-context.test.js
@@ -4,6 +4,7 @@ import { toRequireContext } from './to-require-context';
 const testCases = [
   {
     glob: '**/*.stories.tsx',
+    recursive: true,
     validPaths: [
       './Icon.stories.tsx',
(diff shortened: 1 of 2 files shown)


daniel-ac-martin:
Unfortunately, the `recursive` flag is not covered by the unit tests and I'm not sure it could be covered in a meaningful way.

daniel-ac-martin:
I had a think about this and I think it's possible to have some tests for this, even if they are quite crude test. So I've added that.

shilman (approved):
Looks great! ❤️

PR #11660, opened 2020-07-23
Core: Fix existing behavior with story prefixes

tmeasday:
Issue: #11571

## What I did

Made sure if the prefix isn't an exact match it does the previous behaviour (chooses the first matching story)

## How to test

- Is this testable with Jest or Chromatic screenshots? Test


shilman (approved):
🙈

PR #11673, opened 2020-07-23
Composition: Fix composition of older storybooks

ndelangen:
Issue: -telejson v4.0.1 is now compatible with v3.x-

In some cases of composition, if the composed storybook uses telejson v3, the composed storybook would fail to load.

## What I did
This upgrade to the latest version of telejson that has backwards compatibility to deal with the v3 data.


shilman (approved):
LGTM!

PR #11679, opened 2020-07-24
Addon-storysource: Fix broken source when there's no story

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/11672

## What I did
I made rendering of storysource only happen when storyData is available.


shilman (approved):
LGTM 🙏

PR #11690, opened 2020-07-25
Addon-knobs: Move `@types/react-color` to devDeps

### LINKED ISSUE #11315: The `@storybook/addon-knobs` addon package, contains `@types/react-color` as a dependency.
**Describe the bug**
The `@storybook/addon-knobs` addon package, contains `@types/react-color` as a dependency.

Disregarding the fact that JS users will see no benefit from this additional dependency. I encountered a TS error due to it's inclusion. The dependency creates a conflict between  `@types/react-color`'s `@types/react` and the `@types/react` of my project. I am assuming that TypeScript has no idea how to handle duplicate typings and thus throws an error for every type in `@types/react` (see screenshot) marking that it's _duplicated_.

**Expected behavior**
 `@types/react-color` should not be a dependency of `@storybook/addon-knobs`, if it is indeed required it should be a devDependency.

**Screenshots**
image

**System:**
image

gaetanmaisse:
Issue: #11315

## What I did

See https://github.com/storybookjs/storybook/issues/11315#issuecomment-650012060

**TLDR;**

It looks like some function of components from `addon-knob` (`handleChange` and `handleClick`) should be private functions as they are used to handle internal stuff. Making then private leads to a generated `Color.d.ts` that doesn't import things from `@types/react-color` anymore so we can move this dependency from regular dep to devDep 🎉 .

## How to test

- `addon-knobs` should compile
- CI should be 🟢 (except `example-v2`, but for other reasons: https://github.com/storybookjs/storybook/pull/11354 fixes that)

### DIFF
--- a/addons/knobs/package.json
+++ b/addons/knobs/package.json
@@ -36,7 +36,6 @@
     "@storybook/components": "6.0.0-rc.14",
     "@storybook/core-events": "6.0.0-rc.14",
     "@storybook/theming": "6.0.0-rc.14",
-    "@types/react-color": "^3.0.1",
     "copy-to-clipboard": "^3.0.8",
     "core-js": "^3.0.1",
     "escape-html": "^1.0.3",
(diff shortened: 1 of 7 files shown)


shilman (approved):
LGTM! Do we need to do something similar for Controls? It uses similar patterns @gaetanmaisse

PR #11700, opened 2020-07-27
Cherry pick #11576, fix bug for non-dev builds of SB for release notes

kylesuss:
Issue: #11483 

## What I did

This PR cherry picks a few things from PR #11576 that are low risk for the 6.0 release (aka not changing the mechanism for how we determine if the settings page is active). There are some bugs in #11576 in how that mechanism was changed & so it is going to be handled in the 6.1 release instead.

## How to test

The main bug that #11576 set out to address is #11483 . The way to test that this bug is fixed is to run a non-dev build of SB & make sure you can load up a settings page w/out error.

### DIFF
--- a/lib/ui/src/components/sidebar/Menu.stories.tsx
+++ b/lib/ui/src/components/sidebar/Menu.stories.tsx
@@ -37,28 +37,50 @@ const DoubleThemeRenderingHack = styled.div({
   },
 });
 
+const ExpandedMenu: FunctionComponent<{ menu: any }> = ({ menu }) => (
+  <DoubleThemeRenderingHack>
+    <WithTooltip
+      placement="top"
+      trigger="click"
+      closeOnClick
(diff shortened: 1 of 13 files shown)


lib/ui/src/components/sidebar/Menu.stories.tsx
```diff
@@ -37,28 +37,50 @@ const DoubleThemeRenderingHack = styled.div({
   },
 });
 
+const ExpandedMenu: FunctionComponent<{ menu: any }> = ({ menu }) => (
```
kylesuss, line 40: Not quite sure of the best way to do this `menu` type. Seems like `menu` is a memoized array of memoized functions, which sounds a bit complicated. My typescript is not quite up to par on that one. Open to suggestions.

ndelangen (approved):

PR #11704, opened 2020-07-28
Addon-controls: Fix update logic for argTypes with custom names

jonspalmer:
Controls with a name don't update when changed. Failing test case against Official Storybook.

## What I did
 * changed the Controls props to pass row.key rather row.name as the `name` prop. 
 * changed the Controls API so that onChange just provides the changed value not the arg name/key
 * fixed all the stories

## How to test
 * load official example
 * Test the controls under "Controls" ie `/?path=/story/controls-boolean--basic`
 * Test the Args Table under "Core/Args" ie `/?path=/story/core-args--passed-to-story`

### DIFF
--- a/lib/components/src/blocks/ArgsTable/ArgControl.tsx
+++ b/lib/components/src/blocks/ArgsTable/ArgControl.tsx
@@ -21,7 +21,7 @@ export interface ArgControlProps {
 const NoControl = () => <>-</>;
 
 export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
-  const { name, control } = row;
+  const { key, control } = row;
 
   const [isFocused, setFocused] = useState(false);
   // box because arg can be a fn (e.g. actions) and useState calls fn's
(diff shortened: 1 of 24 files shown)


jonspalmer:
My sense is that `ArgControl` should be using `row.key` throughout rather than `row.name` That has a slightly scary blast radius as each of the Control components take the 'name' and use if for setting DOM ids and names.

lib/components/src/blocks/ArgsTable/ArgControl.tsx
```diff
@@ -32,20 +32,20 @@ export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
   }, [isFocused, arg]);
 
   const onChange = useCallback(
-    (argName: string, argVal: any) => {
+    (argVal: any) => {
```
jonspalmer, line 35: I don't see any reason for the callback to have the key. Before we were ignoring it anyway. The Controls shouldn't need to know this detail.

lib/components/src/blocks/ArgsTable/ArgControl.tsx
```diff
@@ -32,20 +32,22 @@ export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
...
   );
 
   const onBlur = useCallback(() => setFocused(false), []);
   const onFocus = useCallback(() => setFocused(true), []);
 
   if (!control || control.disable) return <NoControl />;
 
-  const props = { name, argType: row, value: boxedValue.value, onChange, onBlur, onFocus };
+  // row.name is a display name and not a suitable DOM input id or name - i might contain whitespace etc.
+  // row.key is a hash key and therefore a much safer choice
```
jonspalmer, line 49: @shilman and I discussed adding a `key` prop (which would actually have to be `argKey` as `key` is a React reserved property). However, with the onChange adjustment its mostly unnecessary. We're really intending to hand in something that can be used as a DOM Input id/name. This makes for a much smaller change

shilman (approved):
LGTM! Thanks for the fix @jonspalmer!!

PR #11718, opened 2020-07-29
Revert "feat: add skip to content on panel and story iframe"

ndelangen:
Reverts storybookjs/storybook#11066


ndelangen:
self merge

PR #11724, opened 2020-07-29
docs(controls): Fix example story of Controls + MDX section

gaetanmaisse:
Issue: The sample code provided in `addons/controls/README.md > How do controls work with MDX?` is not working out of the box so I fixed it.

## What I did

It looks like the extra `;` after a `Story` tag makes `@mdx-js/loader` unhappy, it throws `SyntaxError: Unexpected token (3:8)`. 
Removing that `;` fixes the issue. I also removed the one after `Meta` as it is not needed.

## How to test

- Copy/paste the updated code and... it should just work


shilman (approved):
🙈

PR #11733, opened 2020-07-30
Add example of wiring up an import decorator

tmeasday:
I added an example of how you might write a decorator for mocked imports.

Here's some code if you want to try running it: https://github.com/tmeasday/fetch-mock

### DIFF
--- a/docs/workflows/build-pages-with-storybook.md
+++ b/docs/workflows/build-pages-with-storybook.md
@@ -108,7 +108,7 @@ LoggedOut.parameters: {
 
 #### Mocking imports
 
-It is also possible to mock imports directly, similar to Jest, using webpack’s aliasing. This is useful if your component makes network requests directly.
+It is also possible to mock imports directly, similar to Jest, using webpack’s aliasing. This is useful if your component makes network requests directly via a library, such as [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch).
 
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/workflows/build-pages-with-storybook.md
```diff
@@ -108,7 +108,7 @@ LoggedOut.parameters: {
 
 #### Mocking imports
 
-It is also possible to mock imports directly, similar to Jest, using webpack’s aliasing. This is useful if your component makes network requests directly.
+It is also possible to mock imports directly, similar to Jest, using webpack’s aliasing. This is useful if your component makes network requests directly via a library, such as [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch).
```
jonniebigodes, line 111: @tmeasday we probably can improve this a bit. We could go with:

"It is also possible to mock imports directly, similar to Jest, using webpack´s aliasing. This is extremely useful if your component makes network requests directly with third-party libraries

We're going to use [isomorphic-fetch](ttps://www.npmjs.com/package/isomorphic-fetch) as a example"

docs/workflows/build-pages-with-storybook.md
```diff
@@ -117,13 +117,75 @@ module.exports = {
...
+  nextJson = null;
+}
+
+export function decorator(story, { parameters }) {
+  if (parameters && parameters.fetch) {
+    nextJson = parameters.fetch.json;
+  }
+  return story();
+}
+```
```
jonniebigodes, line 146: @tmeasday makes a bit more sense to move this up before the code snippet in the `.storybook/main`. And we change this:

"You would still need to write the fetch mock and wire up a decorator to provide results to it based on the current story. An example of how that might look:"

To:

"Let's start by creating our own mock, which we'll use later with a [decorator](../writing-stories/decorators#global-decorators]. Create a new file called `isomorphic-fetch.js` inside a directory called `__mocks__` (we'll leave the location to you, don't forget to adjust the imports to your needs) and add the following code inside:"

Something like this. WDYT?

docs/workflows/build-pages-with-storybook.md
```diff
@@ -117,13 +117,75 @@ module.exports = {
...
+
+You would then wire it up like so:
+```js
+// .storybook/preview.js
+import { decorator } from '../__mocks/isomorphic-fetch';
+
+// Add the decorator to all stories
+export const decorators = [decorator];
+```
```
jonniebigodes, line 155: @tmeasday we could probably improve this a bit we could say:

"Add the mock you've just implemented to your `storybook/preview.js` (if you don't have it already, you'll need to create the file)."

WDYT?

docs/workflows/build-pages-with-storybook.md
```diff
@@ -117,13 +117,75 @@ module.exports = {
...
+
+Then to set the value for a specific story you could do (to borrow an example from this [blog post](https://medium.com/@edogc/visual-unit-testing-with-react-storybook-and-fetch-mock-4594d3a281e6)):
```
jonniebigodes, line 157: @tmeasday we probably could improve this a bit and probably say:

" Once you got all the necessary configurations done, you can set the values in any specific story like so (you can borrow from this [blog post](https://medium.com/@edogc/visual-unit-testing-with-react-storybook-and-fetch-mock-4594d3a281e6)"

WDYT?

Also as a side note, probably being a bit picky and sorry for that. Can you add the following to the code snippet below:

```js
// App.story.js

import React from 'react'
.....
```

So that the reader can get a point of reference.

jonniebigodes (changes requested):
@tmeasday i've checked it and we're in good shape. Left a few comments, so that we can fine tune it a bit and we get it merged.

Feel free to provide feedback

tmeasday:
@jonniebigodes I implemented your changes more or less, WDYT?

jonniebigodes:
@tmeasday i've checked and like the improvements you made and i like them a lot. Merge it when you're able.

PR #11743, opened 2020-07-31
Core: Pass normalized parameters to the story sort function

tmeasday:
Issue: #11721

Don't denormalize params before sorting, as this has a perf cost that we don't want to pay. This is a breaking change but hopefully not a huge deal.

Notes:

 - I didn't update the docs because that is happening in the `6.0-docs` branch. We should probably just merge that to `next` now?
 - Do you think we should type / restrict the second entry in the array, as hinted at by the comment?


shilman (approved):
💯

PR #11745, opened 2020-07-31
Composition: Fix missing version property in autoref

ndelangen:
Issue: -autorefs have a version that's appended to the ref url-

## What I did

Ensure the version is passed from the package.json into the refs config


tmeasday (approved):
LGTM

PR #11750, opened 2020-07-31
Update Docs docs to replace existing <Props /> references

mdchristians:
Issue: References to the `<Props>` component still existed in the docs

## What I did

Replaced a few instances of the `<Props>` component I noticed were left in - if I replaced something that shouldn't have been let me know and I can fix it


shilman (approved):
Thanks!! 😂

PR #11757, opened 2020-08-01
Update TOC format, setup with currently ready pages

kylesuss:
Issue: n/a

## What I did

Updated the toc.js file for the 6.0 docs so that it reflects the current state of the MD files we have for the docs as well as the expected format of the frontpage for consumption.

### DIFF
--- a/docs/toc.js
+++ b/docs/toc.js
@@ -2,66 +2,79 @@ module.exports = {
   toc: [
     {
       title: 'Get Started',
+      pathSegment: 'get-started',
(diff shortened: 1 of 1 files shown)


tmeasday (approved):
Sure, WFM. @jonniebigodes would probably be the one with the most thoughts about this.

kylesuss:
@jonniebigodes I need this to merge to get builds on frontpage to work, but let me know your thoughts. Happy to make changes.

PR #11758, opened 2020-08-02
Add documentation on using context to avoid mocking container dependencies

dburles:
Added a section to the documentation on mocking which describes a method to avoid mocking dependencies of connected "container" components.

I've had to adjust the heading sizes as I ran out of sizes that adequately represent the hierarchy of the document. I'm not sure if this would need to be carried over elsewhere.


tmeasday (approved):
Looks great. I think maybe instead of nesting so deeply, let's just make it a separate section.

dburles:
Yeah that would work, though I don't want it to be overlooked as it's definitely worth users having an awareness of it as an option. Did you want me to take care of separating it out? Did you have an approach in mind?

domyen:
Thanks @dburles 🙏 

You don't need to do anything else. I'll take care of getting this merged from here.

jonniebigodes (approved):

PR #11766, opened 2020-08-03
Core: Add `STORY_SPECIFIED` event for initial sidebar selection/URL

tmeasday:
Issue: #11764 

## What I did

No longer respond to `CURRENT_STORY_WAS_SET` in manager as it is too hard to tell if it is out of date.

Instead, add a new event `STORY_SPECIFIED` which only applies to the first story selected.

## How to test

Try things in OfficialSB


shilman (approved):
LGTM!

PR #11770, opened 2020-08-03
Addon-docs: Fix link font size to inherit

ajkl2533:
Issue: #11393 

## What I did
My problem was that when I included a link in heading (in Markdown) the font size of the link was smaller than the heading itself because links had font size hardcoded and set to (14px). I changed the font-size property to `inherit` so now the links should have the same size as the parent block.

### Before
Screenshot 2020-08-03 at 14 03 36

### After
Screenshot 2020-08-03 at 14 02 40

## How to test

- Is this testable with Jest or Chromatic screenshots? yes, Chromatic


shilman (approved):
LGTM thanks for the fix!! 🙏

PR #11775, opened 2020-08-03
 Addon-docs: add escape hatch css classes in docpages

Jeremie-Chauvel:
Issue: #8502 

## What I did
Implemented escape hatch to style docpages
## How to test

- Is this testable with Jest or Chromatic screenshots? yes -> update WIP
- Does this need a new example in the kitchen sink apps? don't think so
- Does this need an update to the documentation? probably WIP

### DIFF
--- a/addons/docs/src/blocks/Anchor.tsx
+++ b/addons/docs/src/blocks/Anchor.tsx
@@ -4,8 +4,11 @@ export const anchorBlockIdFromId = (storyId: string) => `anchor--${storyId}`;
 
 export interface AnchorProps {
   storyId: string;
+  className?: string;
 }
 
-export const Anchor: FC<AnchorProps> = ({ storyId, children }) => (
-  <div id={anchorBlockIdFromId(storyId)}>{children}</div>
+export const Anchor: FC<AnchorProps> = ({ storyId, children, className }) => (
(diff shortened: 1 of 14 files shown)


Jeremie-Chauvel:
I'm a little lost concerning where I should add the documentation of this change ?

Concerning the tests, I'm surprised that my change didn't break any ? Should I create an additionnal story with overriden style to test the change ? (In chromatic ?)

ndelangen:
@shilman This is something for us to consider... 

I feel like we should -at some point- have a structured method for automatically adding these escape-hatch-css-classes to our components, rather then do it manually all over the place. We'll eventually forget them when we refactor.

I personally don't feel like adding a ton of tests for them either.
Maybe a few, to ensure the classes are there.

shilman (commented):
Thanks for for fixing this @Jeremie-Chauvel!

Generally looks good. I'm wondering, though, in some cases you are passing `className` through various layers of indirection. Why can't we just set the `className="sbdocs sbdocs-argstable"` directly where its used rather than adding the extra complexity of passing it around as a variable?

Jeremie-Chauvel:
It's been long enough, I'm not quite confident I remember such decision reasons, I will investigate it on occasion.

PR #11783, opened 2020-08-04
fix: cannot get iframe.html contnet correctly

billyyyyy3320:
Issue: Currently Composition only works when `stories.json` exists.

I've tracked down the history, it failed since `v6.0.0-rc.2`.

## What I did

> if stories.json 404's storybook should fallback to loading the iframe.html

It did fallback! However, somehow using `fetch` with `mode: 'no-cors'`,  it doesn't get the content correctly. 

## How to test

Leverage the composition feature without generating `stories.json`.

- Is this testable with Jest or Chromatic screenshots? N
- Does this need a new example in the kitchen sink apps? N
- Does this need an update to the documentation? N

### DIFF
--- a/lib/api/src/modules/refs.ts
+++ b/lib/api/src/modules/refs.ts
@@ -154,7 +154,6 @@ export const init: ModuleFn = ({ store, provider, fullAPI }, { runCheck = true }
           credentials: 'omit',
         }),
         fetch(`${url}/iframe.html${query}`, {
-          mode: 'no-cors',
           credentials: 'omit',
         }),
       ]);


ndelangen:
This fetch with no-cors isn't supposed to get any content.. It's there to check if the iframe will 404 / error.

The expected behavior is:
- A: try to load stories.json and parse
- B: try to load iframe.html without looking at the content

If both fail we show an error message

if A succeeds we use the data received to populate the sidebar

if A fails, but B is successful we inject an extra iframe and await the SET_STORIES event from it, to populate the sidebar

we additionally load metadata.json to pull non-cachable information like a the list of latest releases/versions etc.
This should fail gracefully, if it doesn't exist.

ndelangen:
This PR is superseded by https://github.com/storybookjs/storybook/pull/12085

Thank you for your contribution @billyyyyy3320 truly appreciated 🙏 
The problem required a different solution however.

PR #11805, opened 2020-08-05
Addon-controls: Fix ArgsTable bugs and styling

domyen:
Issue: #11362, #11698

## What I did
- Fixed width of ArgsTable columns when not `compact`
- Allow really long code `strings` to flow to the next line
- Prevent Boolean (true|false pill button) control from stacking when it's parent width is constrained
- Some code font-size tweaks 
- Fix missing border style in Safari only #11698

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, look at Chromatic changeset [here](https://www.chromatic.com/pullrequest?appId=5a375b97f4b14f0020b0cda3&number=11805&view=changes)

### DIFF
--- a/lib/components/src/blocks/ArgsTable/ArgJsDoc.tsx
+++ b/lib/components/src/blocks/ArgsTable/ArgJsDoc.tsx
@@ -57,6 +57,7 @@ export const Table = styled.table(({ theme }) => ({
     '& code': {
       margin: 0,
       display: 'inline-block',
+      fontSize: theme.typography.size.s1,
     },
   },
 }));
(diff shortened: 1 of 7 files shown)


shilman (approved):
LGTM. Only question is why we are hardcoding font size rather than inheriting from the parent element? This seems brittle and in the past we've had to fix it e.g. https://github.com/storybookjs/storybook/pull/11770

domyen:
@shilman I acknowledge that setting absolute values negates the ability to set one font-size at the top level and have it get inherited everywhere. But relative units can make it tough to apply precise sizing and debug what the resultant font-size is. You end up with values like `font-size: 92.8%;` and `font-size:85.7%` to represent 13px and 12px respectively. 

This came up in another issue with [relative line-heights](https://github.com/storybookjs/storybook/issues/11230). I guess there's a tradeoff here.

PR #11807, opened 2020-08-05
6.0 docs snippets initial setup

kylesuss:
## What I did

Updated the MD files to validate for MDX. Added some dummy code snippets to be used in the frontpage.

## How to test

### DIFF
--- a/docs/api/argtypes.md
+++ b/docs/api/argtypes.md
@@ -6,8 +6,7 @@ title: 'ArgTypes'
 
 NOTE: This API is experimental and may change outside of the typical semver release cycle
 
-<div>
-
+</div>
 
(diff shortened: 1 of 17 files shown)


kylesuss:
@jonniebigodes cross-posting https://github.com/storybookjs/frontpage/pull/129#issuecomment-669999518 so it doesn't get lost:

> @jonniebigodes this is important -- what that means is that when you are writing snippets for MDX files, write them like this:

` ```md ` rather than ` ```mdx `

> But you will define the filename like: `/framework/thing-snippet.mdx.mdx` so that we can show the user it is "MDX".

> There isn't a syntax highlighter for MDX in Prism right now, so we have to make do w/ the MD highlighter for now. An alternative that I have seen is some README's using jsx to highlight MDX. That isn't perfect either, but if need be, we can switch to that later.

PR #11818, opened 2020-08-05
ArgsTable: Fix styles to allow long text to wrap

cfhull:
Issue: #11362 

## What I did
[This PR](https://github.com/storybookjs/storybook/pull/11805), while doing some awesome things, doesn't resolve the word wrapping issue.

Specifically, the new `word-break: break-all` property conflicts with the existing `word-wrap: nowrap` property that is inherited from `styles.span`. I also added a min-width to keep it from breaking on short words.

Current behavior: 
image

New behavior:
image

## How to test

- Is this testable with Jest or Chromatic screenshots? Chromatic (but apparently I can't do that from a fork)

### DIFF
--- a/lib/components/src/blocks/ArgsTable/ArgValue.tsx
+++ b/lib/components/src/blocks/ArgsTable/ArgValue.tsx
@@ -30,10 +30,11 @@ const Summary = styled.div<{ isExpanded?: boolean }>(({ isExpanded }) => ({
   flexWrap: 'wrap',
   alignItems: 'flex-start',
   marginBottom: '-4px',
+  minWidth: 100,
 }));
 
 const Text = styled.span<{}>(codeCommon, ({ theme }) => ({
-  flex: 0,
+  flex: '0 0 auto',
   fontFamily: theme.typography.fonts.mono,
   fontSize: theme.typography.size.s1,
   wordBreak: 'break-word',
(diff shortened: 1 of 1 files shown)


lib/components/src/blocks/ArgsTable/ArgValue.tsx
```diff
@@ -30,10 +30,11 @@ const Summary = styled.div<{ isExpanded?: boolean }>(({ isExpanded }) => ({
   flexWrap: 'wrap',
   alignItems: 'flex-start',
   marginBottom: '-4px',
+  minWidth: 100,
```
cfhull, line 33: This value is a bit arbitrary. Without it, a short string like `default` would wrap the `lt` to the next line. 100px min width seemed small enough to me that the column wouldn't be overly big if there were no defaults, but would keep relatively short values from breaking to a new line.

ajkl2533:
@cfhull thank you I hit the same issue

lib/components/src/blocks/ArgsTable/ArgValue.tsx
ndelangen: Fair

ndelangen (approved):

ndelangen:
@shilman review?

shilman (approved):
Thanks @cfhull !!

jimmyandrade (commented):
LGTM

PR #11825, opened 2020-08-06
CLI: Add CSF types for Angular

tooppaaa:
Issue: No typings for Angular

## What I did

Add typings for angular same way we did for react

## How to test

- E2E


shilman (approved):
LGTM!

PR #11834, opened 2020-08-07
Chore: add snippets for writing stories section

jonniebigodes:
This pull request adds the snippets for the writing stories section:

What was done:

- Updated the markdown files to include the snippets based on the file naming and structure decided.
- Created the snippets (react,react ts, angular)

Some caveats. I need some help with the decorators page, for the angular snippets. I tried a couple of things and none of them worked. So that the documentation examples can match the implementation.

Feel free to provide feedback


docs/snippets/angular/button-group-story.ts.mdx
```diff
@@ -0,0 +1,32 @@
...
+  decorators: [
+    moduleMetadata({
+      declarations: [Button],
+      imports: [CommonModule],
+    }),
+  ],
+};
+
+const Template = (args: ButtonGroup) => ({
+  component:ButtonGroup,
```
tmeasday, line 23:
```suggestion
  component :ButtonGroup,
```

Do we need ensure we run prettier on these files?

docs/snippets/angular/button-story-with-blue-args.ts.mdx
```diff
@@ -0,0 +1,18 @@
...
+  parameters: {
+    backgrounds: {
+      values: [
+         { name: 'red', value: '#f00', },
+         { name: 'green', value: '#0f0', },
+         { name: 'blue', value: '#00f', },
+      ]
+    }
+  }
+}
```
tmeasday, line 17: Apart from the `./button.component` import (rather than `./Button`), this is identical to the react snippet, correct?

Do you think we can just change the name of the import and reuse the snippets--or do you think that'd be too weird for angular folk @jonniebigodes?

I just feel like the less snippets the better for maintainability.

docs/snippets/angular/list-story-expanded.ts.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+// Always an empty list, not super interesting
+const Template = (args: List) => ({
+  component: List,
+  props: args,
+});
+
+export const OneItem = Template.bind({});
+OneItem.args = {
+    items:[{}]
```
tmeasday, line 30: Should this be the `ListItem` rather than an empty object?

tmeasday (approved):
Really great stuff @jonniebigodes 

A couple of questions, and maybe can you try and get prettier to run over your snippets (I can take a look if you have trouble).

domyen (approved):

docs/snippets/angular/button-story-with-blue-args.ts.mdx
jonniebigodes: Factoring this particular case, angular has it's own ways of naming components, which will then imported into storybook. React works in a different way, ember and vue have their own ways as well. 

Reasoning behind this was the following:
- React (js), we're good in that department, imports are ok.
- React (ts), some additional items are added to the story file, and if we're being accurate then i went with the "carbon copy" (pardon the bad pun), to safe guard us from issues being raised, mentioning that the snippets available don't match the documentation.
- Angular once again was based on the actual implementation as the way the stories are defined when Storybook is added to a project/app.

With this in mind and factoring in your feedback, as this is a pattern that is starting to appear in the documentation, i'm ok with having one platform agnostic snippet, that focus on the information, leaving out any component import references. 

Should any issues be created regarding this case in the repo. I'm more than glad to respond to them.

Let me know if you're ok with it and i'll adjust accordingly.

docs/snippets/angular/button-group-story.ts.mdx
jonniebigodes: I'm going to check the snippets and see if any are left not properly indented. Still trying to wrap my head around why MDX and prettier aren't playing nice on my end.

docs/snippets/angular/list-story-expanded.ts.mdx
jonniebigodes: Fixed with the latest commit

docs/snippets/angular/button-group-story.ts.mdx
tmeasday: Yeah, maybe just something to figure out later, let's just get the snippets right for now.

docs/snippets/angular/button-story-with-blue-args.ts.mdx
tmeasday: I think that sounds OK, at least to start with. Let's see what folks say when they see the docs for real.

PR #11857, opened 2020-08-10
Essentials readmes

tmeasday:
- Remove all duplicate content from essential READMEs (this is in the docs now).
- Standardize on a common README format
- Create `ADVANCED.md` for actions.

I left the FAQs in for controls + toolbars. Not sure if that should be deleted/in a separate file/what?


shilman (approved):
💯

PR #11869, opened 2020-08-10
Reorder essentials

tmeasday:
Documentation


domyen (approved):

BPScott:
Yo @tmeasday, It doesn't look [like toolbars is part of the essentials addon](https://github.com/storybookjs/storybook/blob/next/addons/essentials/package.json), but these docs suggest that it is. Which is right? (Hopefully the docs are right as having toolbars be in essentials would be great)

tmeasday:
Well that would explain some confusion! 

@shilman it should be right?

PR #11880, opened 2020-08-11
Fix malformed markdown in 6.0 documentation that was causing bugs

domyen:
Issue:na

## What I did
Removed erroneous spaces which caused bugs.

image

## How to test


shilman (approved):
LGTM

PR #11921, opened 2020-08-12
fix(withA11y docs): Fix function call code example

andi1984:
Fix `addParameters` function call code example by adding missing closing parenthesis.

Issue:

Closing parenthesis was missing in doc code example.

## What I did

Fix code example in docs

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -623,7 +623,7 @@ addParameters({
     options: {},
     manual: true,
   }
-};
+});
 ```
 
 #### Essentials addon disables differently


andi1984:
Sorry, I forgot to tag it with `documentation`. Can some team member take care of it?

shilman (approved):
LGTM

@tmeasday do we want to use `export const parameters` instead for stuff like this?

tmeasday:
@shilman would have thought so. Might be worth a quick check through `MIGRATION.md`

PR #11932, opened 2020-08-12
Update widget-story-controls-enum.js.mdx

jldmn:
This is the only way I got the custom control types working. Following the argTypes-Docs from https://storybook.js.org/docs/react/api/argtypes the `type` field is for setting the property-type, not the control type.

Issue: Custom Controls not working as expected. Or to be more specific, the control type doesn't change.

## What I did

Updated the example code-snippet to match the argTypes api-docs by using `control` property instead of `type` property which is used for the actual property type.

## How to test

Try to run a story with custom-controls set via `argTypes` in the default export.

  I'm not sure.
  Probably not.
  This actually is an update to the documentation :)

### DIFF
--- a/docs/snippets/common/widget-story-controls-enum.js.mdx
+++ b/docs/snippets/common/widget-story-controls-enum.js.mdx
@@ -6,8 +6,10 @@ export default {
   component: Widget,
   argTypes: {
     loadingState: {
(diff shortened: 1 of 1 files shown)


shilman (approved):
Updated slightly. Thanks for the fix!!! 🙏

PR #11934, opened 2020-08-12
UI: update upgrade command in about section

yannbf:
## What I did
Now that Storybook is released, users that have older versions will have this button at the bottom:
image

That leads to a page showing how to upgrade. The commands in the current page are outdated. I assume we want to promote the usage of the CLI so I have updated the command:
image


shilman (approved):
Great catch ❤️

PR #11936, opened 2020-08-12
Vue: Add support for tsx

blake-newman:
Previous to this change vue file paths get `ts` suffix. However; if using `<script lang="tsx">` we want `tsx` to be the suffix

Screenshot from 2020-08-12 13-31-30

## What I did

This change adds `tsx` as a test rule and applies `appendTsxSuffixTo`. 

## How to test


shilman (approved):
Merging w/ help from @Aaron-Pool. Thanks for your patience!! 🙈

PR #11937, opened 2020-08-12
Use options parameter from first story on init.

tmeasday:
Issue: #11924 

## What I did

Use the full parameters of the first story (which is what we did on story change anyway), on init.

## How to test

Yes, we have tests


tmeasday:
Whoops, this accidentally included a different branch. Reopening

PR #11938, opened 2020-08-12
Use the denormed params on the first story for initial options

tmeasday:
Issue: #11924 

## What I did

Use the full parameters of the first story (which is what we did on story change anyway), on init.

## How to test

Yes, we have tests


shilman (approved):
LGTM!

PR #11939, opened 2020-08-12
fix: bump jest-specific-snapshot to v4

SimenB:
Issue: Reduces number of Jest deps in the tree

## What I did

Updated the dependency, removed `jest` from the lockfile to update it locally as well and ran `yarn`

## How to test

Dunno 😀


SimenB:
`tagName.includes is not a functionTypeError: tagName.includes is not a function` ergh, found a bug in Jest 😅

PR #11948, opened 2020-08-12
Remove deprecated addParameters from docs snippet

rafaelalmeidatk:
## What I did

Replaced the deprecated `addParameters` with the named export


shilman (approved):
LGTM thanks!

PR #11995, opened 2020-08-13
Change semi-colons in args object to commas

wahidshafique:
## What I did

Fixed a tiny typo with object definition


shilman (approved):
🙈

PR #11999, opened 2020-08-13
Chore: updates to the writing docs section

jonniebigodes:
With this pull request, some mentions to props tables are now removed.

What was done:
- Updated the docs files to mention the nomenclature used.
- Adjusted the snippet accordingly.

@shilman can you vet if that's the use case for the snippet? If that's want we want to document?

Feel free to provide feedback

### DIFF
--- a/docs/api/argtypes.md
+++ b/docs/api/argtypes.md
@@ -59,7 +59,7 @@ In this ArgTypes data structure, name, type, defaultValue, and description are s
 
 #### Manual specification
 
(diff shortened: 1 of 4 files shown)


ndelangen:
LGTM, if this is the intended direction, @shilman ?

shilman (approved):
LGTM. Thanks @jonniebigodes! I'm going to take another pass at these docs soon.

PR #12004, opened 2020-08-13
file extension changed from .ts.mdx to .tsx.mdx

virat21:
Issue:

## What I did
fixed file extension

Yes
No
Yes

`[ "documentation","patch"]`

### DIFF
--- a/docs/writing-stories/introduction.md
+++ b/docs/writing-stories/introduction.md
@@ -46,7 +46,7 @@ Use the named exports of a CSF file to define your componentâ€™s stories. Hereâ€
 <CodeSnippets
   paths={[
     'react/button-story.js.mdx',
-    'react/button-story.ts.mdx',
+    'react/button-story.tsx.mdx',
     'angular/button-story.ts.mdx',
   ]}
 />


jonniebigodes:
@virat21 can you give some context why you're changing the name? Probably fill out the pull request information to give me some more insights?

So that we're clear, from where i'm at, if you change the extension, the snippet will not be displayed and we don't to do that.

Also when i was creating the snippets for the documentation i added the `.ts` extension to them as an abstract. So that we could parse them and display them correctly. I know that certain developers/frameworks embrace the `.tsx` extension, but for this case it's something that we can leave it as is as to prevent further issues down the line.

Don't you agree?

Stay safe

shilman:
@jonniebigodes i think the file extension controls what extension shows up in the tab. i think having a .ts extension makes sense as you've described, even tho technically the file is TSX cc @kylesuss @tmeasday

virat21:
We can keep file extension as `.ts.mdx` that we using internally only so. but in docs, we should go with `.tsx` extension for stories and that's is the right way else IDE will yell the TypeScript warning, `.ts` extension might make trouble! for new learners at least our docs should be crystal clear for others!

kylesuss:
Yeah that part of the extension creates the "badge" in the UI:

Screen Shot 2020-08-14 at 1 48 51 PM

Also, importantly, we don't support `tsx`yet in the design system so if any of the actual snippets change their syntax to `tsx`, then we need an update to the design system to support that. All possible, if necessary, but requires changes so keep me posted.

PR #12026, opened 2020-08-14
add web-component story types

### LINKED ISSUE #11974: Define `Story` for web-components in typings
We currently define a `Story` in the react types:

https://github.com/storybookjs/storybook/blob/3fe7d96fd8d289f40bd8a38ec836ed95e6afe865/app/react/src/client/preview/types-6-0.ts#L23

But we don't have the equivalent for web components, even though it looks like it should be possible.

Is it as simple as defining the following in the typings?

```ts
export type WebComponentReturnType = TemplateResult|string|Node;
export type Story<Args = DefaultArgs> = BaseStory<Args, WebComponentReturnType> &
  Annotations<Args, WebComponentReturnType>;
```

im happy to PR it if that's right, and if you can let me know where it belongs if so.

43081j:
Fixes #11974 .

@shilman @tooppaaa can one of you let me know where the tests for this should live so i can add them?

ill leave it as a draft until then


tooppaaa:
Hey, you can check siblings PR for other frameworks:
https://github.com/storybookjs/storybook/pull/11811/files
https://github.com/storybookjs/storybook/pull/11825

I'm about to do the one for Vue

43081j:
@tooppaaa i had a look. seems this is one of the few (or the only one) that isn't yet written in typescript (the examples). so should that be moved in this PR too?

as it seems to "test" it, we basically need the examples to use it

43081j:
i did try in the last commit but im struggling with local setup at the min. maybe you can take a look and let me know if i was going down the right path / what needs doing?

lib/cli/src/frameworks/web-components/Button.stories.ts
```diff
@@ -1,14 +1,15 @@
-import { Button } from './Button';
+import { Story, Meta } from '@storybook/web-components';
```
tooppaaa, line 1: `frameworks/web-components` JS files should remain untouched as people are not forced to use typescript.

lib/cli/src/frameworks/web-components/Button.ts
```diff
@@ -1,10 +1,23 @@
...
  * Primary UI component for user interaction
  */
-export const Button = ({ primary, backgroundColor, size, label, onClick }) => {
+export const Button = ({
+  primary,
+  backgroundColor,
+  size,
+  label,
+  onClick,
+}: Partial<ButtonProps>) => {
```
tooppaaa, line 20: Best to use optional properties than Partial, you can reuse types from React framework

app/web-components/src/client/preview/types.ts
```diff
@@ -19,3 +21,18 @@ export interface ShowErrorArgs {
   title: string;
   description: string;
 }
+
+/**
```
tooppaaa, line 25: Could this be added to a types-6-0.ts file instead ? This might change in the 6.1 release

tooppaaa (changes requested):

tooppaaa:
I suggest you leave out typescript version of web components cli for now and focus on types that work in the monorepo.
Once this is done and released, it will be easier to add CLI for this using or e2e scripts.

Happy to walk through this at this time 😃

43081j:
@tooppaaa would you mind catching up on this. i just need to know if you want me to get rid of the typescript, have both, or what

tooppaaa:
sorry i will catch up tonigth

lib/cli/src/frameworks/web-components/Button.ts
tooppaaa: The reason is that it is not the same.
image
I'm curious to see the reasons of having everything Partial. 
Not sure I understand the "both/duplicated" but we do want all types to be the same in every framework.

package.json
```diff
@@ -299,4 +299,4 @@
       ]
     ]
   }
-}
```
tooppaaa, line 302: revert ?

tooppaaa (changes requested):

tooppaaa:
Your typescript template is something nice to have and wanted of course, I'm just thinking this can be done in a separate PR where we can add e2e config as well. 
Also, we should not touch the js file. We need to move the original files to a `js` folder and move what you need to a `ts` folder (as it's done for react)

Do you want me to take over that last point ?

lib/cli/src/frameworks/web-components/Button.ts
43081j: i don't mind either way, here we can use an optional interface for consistency.

in reality a `Partial<T>` is probably just a `[K in keyof T]?: T[K]` anyway which i suppose is the same as a fully optional interface. 

i'm struggling to recall where and why i've seen them frowned upon now that i think about it. they are equivalent pretty much, and i can't find any such lint rules anymore against them. so i have no clue where i've picked this up from! maybe i should start using fully optional interfaces again :joy:

43081j:
if you wouldn't mind helping out with that, that'd be much appreciated as it sounds like you already have in your head how you want it to look/be structured.

tooppaaa:
Sure, will do tomorrow / over weekend / soon and ask for your review !

43081j:
@tooppaaa could you have a look this week? just don't want the PR going stale is all. if you need any help just let me know

tooppaaa:
Hello ! Sorry been busy. I took some time to finish your PR, I coudn't add you as a reviewer, not sure why.
Could you please review I didn't miss any of your awesome work on this ?

Thanks

43081j:
Think this was fixed in #12395, closing.

PR #12036, opened 2020-08-14
Update instructions to configure toolbar menus

fgaleano:
Add instructions to install and register @storybook/addon-toolbars before configuring custom toolbar menus.

Issue:

The instructions to add custom toolbar menus did not mention the fact that `@storybook/addon-toolbars` needs to be installed and registered before a custom toolbar icon can be added. Without this step, the icon would never appear on the toolbar.

## What I did

Updated the text to include instructions to install and register `@storybook/addon-toolbars` before declaring `globalTypes`.

## How to test

n/a

### DIFF
--- a/docs/essentials/toolbars-and-globals.md
+++ b/docs/essentials/toolbars-and-globals.md
@@ -12,7 +12,25 @@ When the globals change, the story re-renders and the decorators rerun with the 
 
 ## Global types and the toolbar annotation
 
-Storybook has a simple, declarative syntax for configuring toolbar menus. In your [`.storybook/preview.js`](../configure/overview.md#configure-story-rendering), you can add your own toolbars by creating `globalTypes` with a `toolbar` annotation:
+Storybook has a simple, declarative syntax for configuring toolbar menus. First, install `@storybook/addon-toolbars`:
+
+```
+npm i -D @storybook/addon-toolbars
+```
+
+And register the addon in your `.storybook/main.js` file:
+
+```
+module.exports = {
+  stories: [...],
+  addons: [
+    '@storybook/addon-toolbars'
+  ],
+};
+
+```
+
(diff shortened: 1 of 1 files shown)


fgaleano:
I would also like to add a mention regarding available icon keywords for each toolbar menu. I'd just need to know if there's existing icon documentation to reference on this page.

jonniebigodes:
@fgaleano thank you for the contribution. Do you mind waiting a bit while i take a look at it in more detail?

Stay safe.

imdongchen:
Not sure if they have a "developer documentation", but I use [their storybook](https://next--storybookjs.netlify.app/official-storybook/?path=/story/basics-icon--labels) for builtin components

> I would also like to add a mention regarding available icon keywords for each toolbar menu. I'd just need to know if there's existing icon documentation to reference on this page.

lucalves (approved):

domyen:
Toolbars is installed automatically with the Essentials addons. If users are on 6.0 they will have gotten it out of the box. Is this for pre 6.0 users?

fgaleano:
I'm on @storybook/react 6.0.0-rc.27

The existing instructions to add a Toolbar menu produce nothing in this
version when just installing @storybook/react and just following these
steps.

It was clear I had to install or at least register `addon-toolbars` before
the menu item configuration took effect.

I can attempt to create a separate repo to reproduce this outside of my
project.

domyen:
I believe toolbars addon was a late addition to the release candidates. Do you mind trying it again on 6.0.10+ or later? I'm pretty sure you don't need to register the addon anymore.

ndelangen (approved):

shilman:
Thanks for the documentation fix, but this shouldn't be necessary. If it is, it's a bug in the 6.0 release version of essentials. Closing for now.

fgaleano:
Confirming my issue was because I was still on a RC version. I wasn't aware the final v6 release was out and had changed this functionality. Thank you all for your patience.

PR #12043, opened 2020-08-14
Doc: Update toolbars-and-globals.md

imdongchen:
Issue:
In storybook 6,  if you set the global option `passArgsFirst: false`, context should be passed as the **first** argument

## What I did
change typo `second argument` to `first argument`

## How to test


shilman (approved):
Right! 😄

PR #12085, opened 2020-08-17
Composition: Verify refs in node

ndelangen:
Issue: #11892

## What I did

- ADD the type property from node
- ADD check in node to fetch the `iframe.html`
- REMOVE check for `iframe.html` in `lib/api/modules/refs`
- CHANGE `lib/api/modules/refs` to use the set type of a ref as a replacement for runtime checking `iframe.html`
- CHANGE `lib/api/modules/refs` to not request `stories.json` with credentials when the server-side fetch succeeded, because that means it's public.

### DIFF
--- a/lib/api/src/modules/refs.ts
+++ b/lib/api/src/modules/refs.ts
@@ -62,16 +62,6 @@ export type RefUrl = string;
 // eslint-disable-next-line no-useless-escape
 const findFilename = /(\/((?:[^\/]+?)\.[^\/]+?)|\/)$/;
 
-const allSettled = (promises: Promise<Response>[]): Promise<(Response | false)[]> =>
-  Promise.all(
-    promises.map((promise) =>
-      promise.then(
-        (r) => (r.ok ? r : (false as const)),
-        () => false as const
-      )
-    )
-  );
-
 export const getSourceType = (source: string, refId: string) => {
   const { origin: localOrigin, pathname: localPathname } = location;
   const { origin: sourceOrigin, pathname: sourcePathname } = new URL(source);
@@ -96,6 +86,15 @@ const addRefIds = (input: StoriesHash, ref: ComposedRef): StoriesHash => {
   return Object.entries(input).reduce((acc, [id, item]) => {
     return { ...acc, [id]: { ...item, refId: ref.id } };
   }, {} as StoriesHash);
+};
+
+const handle = async (request: Response | false): Promise<SetRefData> => {
+  if (request) {
+    return Promise.resolve(request)
+      .then((response) => (response.ok ? response.json() : {}))
+      .catch((error) => ({ error }));
+  }
+  return {};
 };
 
 const map = (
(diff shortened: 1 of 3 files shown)


lib/core/src/server/manager/manager-config.js
```diff
@@ -91,6 +92,17 @@ async function getManagerWebpackConfig(options, presets) {
   }
   if (autoRefs || definedRefs) {
     entries.push(path.resolve(path.join(options.configDir, `generated-refs.js`)));
+
+    // verify the refs are publicly reachable, if they are not we'll require stories.json at runtime, otherwise the ref won't work
+    await Promise.all(
+      Object.entries(refs).map(async ([k, value]) => {
+        const { ok } = await fetch(`${value.url}/iframe.html`).catch((e) => ({
+          ok: false,
+        }));
```
tmeasday, line 101: I'm not sure about mixing promises and await. How about:

```suggestion
        let type = 'unknown';
        try {
          const { ok } = await fetch(`${value.url}/iframe.html`);
          if (ok) {
             type = 'server-checked';
          }
        } catch (err) {
          // pass
        }
```

lib/api/src/modules/refs.ts
```diff
@@ -36,7 +36,7 @@ export interface ComposedRef {
   id: string;
   title?: string;
   url: string;
-  type?: 'auto-inject' | 'unknown' | 'lazy';
+  type?: 'auto-inject' | 'unknown' | 'lazy' | 'server-checked';
```
tmeasday, line 39: How do the `auto-inject` and `lazy` types get set?

The "server-checkedness" feels sort of orthogonal to the type. But maybe i don't know what the type means.

lib/api/src/modules/refs.ts
```diff
@@ -168,7 +164,7 @@ export const init: ModuleFn = ({ store, provider, fullAPI }, { runCheck = true }
         return {};
       };
 
-      if (!included && !omitted && !iframe) {
+      if (!included && !omitted && type !== 'server-checked') {
```
tmeasday, line 167: I would expect to see a test for this scenario.

tmeasday (commented):
Not sure if this is done, but some comments

lib/core/src/server/manager/manager-config.js
ndelangen: I'd rather have the above personally. I don't think mixing async/await with promises is bad.

But maybe if I extract the fetch and catch into a function you wouldn't mind the mixing of async/await and promises?

lib/api/src/modules/refs.ts
ndelangen: added a test!

ndelangen:
Ready for review @rannyeli @TechieQian @basher @tmeasday

lib/api/src/modules/refs.ts
```diff
@@ -46,7 +46,7 @@ export interface ComposedRef {
 }
 export interface ComposedRefUpdate {
   title?: string;
-  type?: 'auto-inject' | 'unknown' | 'lazy';
+  type?: 'auto-inject' | 'unknown' | 'lazy' | 'server-checked';
```
tmeasday, line 49: Should this at least be an enum we reuse?

lib/api/src/modules/refs.ts
```diff
@@ -135,28 +135,27 @@ export const init: ModuleFn = ({ store, provider, fullAPI }, { runCheck = true }
...
+                Accept: 'application/json',
+              },
+              credentials: 'include',
+            }),
         fetch(`${url}/stories.json${query}`, {
           headers: {
             Accept: 'application/json',
           },
           credentials: 'omit',
         }),
```
tmeasday, line 158: If we know `iframe.html` fails without credentials (i.e. `isPublic == false`), the second request is definitely going to fail.

I recommend simplifying to:
```suggestion
      // In theory the `/iframe.html` could be private and the `stories.json` could still exist, but in practice
      // the only private servers we know about (Chromatic) always include `stories.json`. So we can tell
      // if the ref actually exists by simply checking `stories.json` w/ credentials.
      const storiesJsonAvailable = 
        await fetch(`${url}/stories.json${query}`, {
          headers: {
            Accept: 'application/json',
          },
          // If we include credentials the server needs proper CORS headers, so don't include them
          // unless we know we need them (i.e. initial fetch of `/iframe.html` failed).
          credentials: isPublic ? 'omit' : 'include',
        }));
```

tmeasday (changes requested):
Looking good, some tweaks

lib/api/src/modules/refs.ts
ndelangen: I'll refactor so it's re-used. All sub-types are re-usable even if not individually exported.

tmeasday (approved):
Looking good!

shilman (approved):
Clever!

PR #12087, opened 2020-08-17
Core: Add frameworkPath to options to support custom frameworks

ThaNarie:
Issue: #12075

## What I did
This addition allows an override of the hardcoded `@storybook/${framework}` import path in the (generated) preview code.

I tested these changes first in a separate project that would need this feature to work, by changing the node_modules.

I'm not sure if any of the automated tests cover the creation of custom frameworks, or running normal framework webpack builds at all, so I haven't run or included any tests.

## How to test

- Is this testable with Jest or Chromatic screenshots? - Don't think so
- Does this need an update to the documentation? - Yes, which I tried to provide in this PR

If your answer is yes to any of these, please make sure to include it in your PR. - Done


shilman (approved):
This looks awesome @ThaNarie !

Are you OK if we release this in a 6.1-alpha next week?

@ndelangen @tmeasday Do you think we can call this a bug and patch it to 6.0? I think it was an accident that we don't allow non-storybook framework packages.

tmeasday (approved):
Yeah I can see the argument this is a bugfix

ndelangen (approved):
👏

PR #12089, opened 2020-08-17
HTML: Add script tag support

adamonsoon:
I copied the page load simulation functionality from sb/aem to support components with `<script>` tags in sb/html.
This solution was suggested in [Discord](https://discordapp.com/channels/486522875931656193/486522876388704258/714311246471626842) several months ago but was never added.
## How to test

- I added an example named "Script" to the HTML kitchen sink under "Demo" (`button.stories.js`) that should trigger an `alert`

### DIFF
--- a/app/html/src/client/preview/helpers/simulate-pageload.ts
+++ b/app/html/src/client/preview/helpers/simulate-pageload.ts
@@ -54,7 +54,7 @@ function insertScript($script: any, callback: any, $scriptRoot: any) {
 
 // runs an array of async functions in sequential order
 /* eslint-disable no-param-reassign, no-plusplus */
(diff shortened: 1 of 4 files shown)


shilman (approved):
LGTM. Nice!

app/html/src/client/preview/helpers/simulate-pageload.ts
```diff
@@ -0,0 +1,97 @@
+import { document } from 'global';
+
+// https://html.spec.whatwg.org/multipage/scripting.html
+const runScriptTypes = [
+  'application/javascript',
```
hainguyengravityglobal, line 5: please add "application/json"

PR #12038, opened 2020-08-14
Chore: Minor tweaks to the documentation and snippets

jonniebigodes:
With this pull request some updates were made to the snippets to make them more clear and some pages in the documentation were updated.

What was done:
- Wrong naming of story files are fixed.
- Added missing component/stories mentions in the snippets.
- Snippets for React typescript files are now properly named, instead of being mentioned as `Button.stories.ts`, they are now called `Button.stories.tsx` in the commented block under each one.
- Added necessary imports to make the snippets more clear.
- Removed `common/button-story-disable-docspage-component.ts.mdx` file and mention from the Docs page in the Writing Docs section, as it was actually unnecessary to get the message across in the documentation.
- Properly named the snippet used in Story layout page in the Configure section and fixed the reference. 

Feel free to provide feedback

PR #12094, opened 2020-08-17
Fix snippet path in story layout docs

kylesuss:
Issue:

https://github.com/storybookjs/storybook/blob/next/docs/configure/story-layout.md

This Markdown file references:

`common/storybook-preview-layout-param.js.mdx`

but the file was called:

 `common/storybook-preview-layout.param.js.mdx `

## What I did

Renamed the file to `common/storybook-preview-layout-param.js.mdx`.


jonniebigodes:
@kylesuss thanks for the pull request, but that issue is addressed in #12038, i still have to update the per @domyen feedback

PR #12118, opened 2020-08-18
Updating and optimizing circleCI Config

PK1312:
## What I did

Made some changes (along with @ KyleTryon ) to the config file in order to increase the performance and decrease credit usage, primarily via the usage of different size executors. Also wrote a framework for more easily specifying the executor and size in the config file. These changes, though small, should add up to fewer consumed credits.

## How to test

Passing CI tests should be all that is required

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,14 +1,21 @@
 version: 2.1
 
-aliases:
-  - &defaults
+executors:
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM!

ndelangen (approved):

PR #12121, opened 2020-08-19
React: Improve ArgTypes type definition

MaximSagan:
Issue:

The `argTypes` property in the `Annotations<Args>` type definition is not checked such that it matches the `Args` type parameter.

## What I did

Updated the `ArgTypes` type definition such that it is generic and its keys are restricted to the keys of the object specified in its type parameter.

## How to test

- Does this need an update to the documentation?  **No**

### DIFF
--- a/lib/addons/src/types.ts
+++ b/lib/addons/src/types.ts
@@ -42,9 +42,9 @@ export interface ArgType {
   [key: string]: any;
 }
 
-export interface ArgTypes {
-  [key: string]: ArgType;
-}
+export type ArgTypes<TArgs = any> = {
+  [K in keyof Partial<TArgs>]: ArgType;
+};
 
 export interface StoryIdentifier {
   id: StoryId;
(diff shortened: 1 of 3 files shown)


lib/addons/src/types.ts
```diff
@@ -169,7 +169,7 @@ export interface Annotations<Args, StoryFnReturnType> {
    * ArgTypes encode basic metadata for args, such as `name`, `description`, `defaultValue` for an arg. These get automatically filled in by Storybook Docs.
    * @see [Control annotations](https://github.com/storybookjs/storybook/blob/91e9dee33faa8eff0b342a366845de7100415367/addons/controls/README.md#control-annotations)
    */
-  argTypes?: ArgTypes;
+  argTypes?: ArgTypes<Args>;
```
tooppaaa, line 172: I'm not sure this is the right direction to set the genericity.
In my mind, ArgTypes can be used to infer **additional** keys / typesof of Args

MaximSagan: @tooppaaa Additional on top of what? On top of the controls automatically generated from the component's props? Because I believe that's already what `Args` is supposed to be: a superset of props. I don't think the keys passed to `argTypes` should be a further superset of that. Both `Annotation.args` and `Annotation.argTypes` define the same set of items so should be based on the same keys. It's also worth noting that with this change, both are defined using `Partial<Args>` (`.args` already was), so there's no need to explicitly set either of them for every key of `Args`.

NamHuynhHoang: Các đường truyền tải rất đa dạng theo kiểu chi tiết nhỏ hoặc lớn có cả âm thanh

PR #12123, opened 2020-08-19
Update doc-blocks documentation to reference ArgsTable

mattdarveniza:
Issue:

## What I did

Replace usage of the `Props` mdx component with `ArgsTable`, which I'm assuming was the intention here.


shilman (approved):
thanks! 🙈

PR #12132, opened 2020-08-19
Fix erroneous code snippet

robbertkl:
Came across this in the documentation (on [this page](https://storybook.js.org/docs/react/essentials/controls)) and noticed 2 issues with it:

1. it had a semicolon instead of a comma within the `argTypes` object
2. it was missing the `control` object, like in the code snippet before it

I've corrected both in this PR.


shilman (approved):
LGTM! 💯

PR #12134, opened 2020-08-19
Addon-docs: Fix missing line-height on TypeSet block

jpzwarte:
Add `line-height` to `<TypeSet>` `<Sample>`. Fixes screenshot:
Screen Shot 2020-08-19 at 13 26 16


jpzwarte:
To be clear: I'm changing the line-height of the preview in `preview-head.html`. If you don't do that, then `line-height` will use the default value and this bug does not appear.

shilman (approved):
LGTM. @domyen?

ndelangen (approved):

ndelangen:
The new situation is at least better than the old? Can we merge and refine later if @domyen has remarks?

PR #12136, opened 2020-08-19
Fix link to the root configuration

stof:
this fixes a broken link in the doc.


jonniebigodes:
@stof sorry for that mishap, just checked and you're correct. Going to merge it 

Stay safe

PR #12148, opened 2020-08-19
Addon-viewport: Add preset to fix windows import

tooppaaa:
Issue: #12024

For windows users, addons that are missing preset are not loaded through essentials

## What I did

Add preset to viewport

## How to test

- https://www.microsoft.com/en-us/store/b/windows

### DIFF
--- a/addons/viewport/preset.js
+++ b/addons/viewport/preset.js
@@ -1,0 +1,1 @@
+module.exports = require('./dist/preset');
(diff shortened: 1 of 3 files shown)


tooppaaa:
@gaetanmaisse  I do not explain this behavior...

My feeling would telll me that require.resolve a path without export would do nothing on windows.
This is very weird but only explanation I've for so far 😞

shilman (approved):
LGTM tho technically this is a workaround and it would be better if we can find the underlying cause

PR #12159, opened 2020-08-20
Fix the migration guide to disable addons through parameters

stof:
I followed the migration guide, and the disabling was not working. Then I diged in the code to find the implementation, and I found out that the doc uses the wrong name:
https://github.com/storybookjs/storybook/blob/1c972a2bc2b5bfd70c035da89fda8a4cd1b4899c/lib/api/src/modules/addons.ts#L107

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -367,7 +367,7 @@ If you wish to selectively disable `knobs` checks for a subset of stories, you c
 export const MyNonCheckedStory = () => <SomeComponent />;
 MyNonCheckedStory.story = {
   parameters: {
-    knobs: { disable: true },
+    knobs: { disabled: true },
   },
 };
 ```
@@ -474,7 +474,7 @@ It's unclear and confusing what would happened if you did. If you want to disabl
 
 ```js
 export StoryOne = ...;
-StoryOne.story = { parameters: { addon: { disable: true } } };
+StoryOne.story = { parameters: { addon: { disabled: true } } };
 ```
 
 If you want to use a parameter for a subset of stories in a kind, simply use a variable to do so:


shilman:
Hmm it looks like there is an inconsistency here since `disable: true` is the preferred method for some addons. @yannbf didn't you take a look at this earlier?

ndelangen:
shall we turn 

https://github.com/storybookjs/storybook/blob/1c972a2bc2b5bfd70c035da89fda8a4cd1b4899c/lib/api/src/modules/addons.ts#L107

into:

```js
 if (paramKey && parameters && parameters[paramKey] && (parameters[paramKey].disabled || parameters[paramKey].disable) { 
```

shilman:
Great idea @ndelangen

stof:
Well, any addon using `@storybook/addons` actually relies on `disabled` to hide the panel, not on `disable`, due to the code I linked.

Can you tell me which addon actually work with `disable`  ?

shilman:
storyshots, chromatic, docs, controls, viewports, knobs .. to name a few

stof:
I assure you that knobs does not disable itself based on `disable` but based on `disabled`. That's precisely the addon for which I debugged things

shilman:
ok then we should apply @ndelangen 's proposed fix. it's `disable` in many other places in the code.

shilman:
addArgs_test_ts_—_storybook

stof:
@shilman the fact that examples use `disable: true` does not mean that the addon actually gets disabled when running storybook.

shilman:
Do you propose we change it in the 120 other places in the code where it uses disabled or the one place where it would match the documentation & enforce consistency in the code?

stof:
@shilman from my search in the code, it seems like `@storybook/ui` is pretty consistent  on using `disabled`. Usages of `disable` seem to be in documentation and examples.
Btw, working stories covering this API are also using `disabled`: https://github.com/storybookjs/storybook/blob/18e0738b46355f5e4de3f20b944d08aa2f204c60/lib/ui/src/containers/panel.stories.tsx

Searching for `disabled`, there are other places to change than only the one in `getStoryPanels`.
And if this is changed entirely, removing support for `disabled`, it will actually break BC, as it will break things for projects actually discovering that the working setting was `disabled` (maybe by looking at the stories of the ui component, or by digging in the code like me)

shilman:
Here's where `disable` is used: 

https://github.com/storybookjs/storybook/blob/master/lib/addons/src/make-decorator.ts#L21

stof:
oh, so it actually depends which feature the addon is using...

shilman:
Fixed in #11742 12171

PR #12165, opened 2020-08-20
Addon-docs: Fix Vue ArgsTable sanitizing of item.type.elements to item.type.value

HerrBertling:
Issue #11944

which now only happens if the elements entry is present
and also is applied at the correct nesting level of the
item object.

## What I did

I swear the fix in https://github.com/storybookjs/storybook/pull/12158 worked on my machine. Should have tested it with more components, sorry 😓  This definitely works now, because my fully fledged project storybook with > 50 stories works with this correct nesting.

@shilman I guess you were right to catch this case via `if` after all 😅


shilman (approved):
OK!

PR #12181, opened 2020-08-21
CLI: Add aurelia detection

tooppaaa:
Issue: 
Aurelia is not part of the detected framework
Preparatory work for e2e on this framework

## What I did
Add detection

## How to test

- `npx aurelia-cli@ new aurelia-latest --unattended`
- sb init (locally of course)

### DIFF
--- a/lib/cli/src/detect.test.ts
+++ b/lib/cli/src/detect.test.ts
@@ -209,6 +209,16 @@ const MOCK_FRAMEWORK_FILES = [
       },
     },
   },
+  {
+    name: ProjectType.AURELIA,
+    files: {
+      'package.json': {
+        dependencies: {
+          'aurelia-bootstrapper': '1.0.0',
+        },
+      },
+    },
+  },
 ];
 
 describe('Detect', () => {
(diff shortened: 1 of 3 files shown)


shilman (approved):
LGTM but let's merge when we start 6.1

yannbf (commented):
Nice work!
I believe there are different ways of setting up Aurelia projects, and one of them includes using jspm (check [here](https://aurelia.io/docs/build-systems/jspm/setup#running-the-unit-tests) and [here](https://github.com/aurelia/dialog/blob/master/package.json#L47)) and such format is not in our detection, BUT I have no experience with Aurelia, so let's ask the expert!

Hey @brandonseydel, what are the possible ways to detect if a project is an Aurelia project? Are the generated templates via storybook CLI compliant with older projects or only new ones? We need to figure out if a user has an Aurelia project and runs `sb init` if they will get a properly set Storybook for them, and even if we support all versions. Thanks!

ndelangen (approved):

brandonseydel:
@tooppaaa this is aurelia 2 and such this will not work.

tooppaaa:
@brandonseydel you mean this `dependencies: ['aurelia-bootstrapper'],` ? I'm not sure what to put here, I'm happy to use your suggestions :)

brandonseydel:
'aurelia' || '@aurelia/runtime'

PR #12203, opened 2020-08-23
Correct globs in documentation

rtivital:
Issue:
Globs are not correct as it is listed here – https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#correct-globs-in-mainjs

## What I did
Updated documentation


shilman (approved):
Thanks!

PR #12223, opened 2020-08-24
UI: Remove scrolling attribute from iFrame

albert-schilling:
Issue: #11799

The iframe that renders the stories has a scrolling attribute that is both unnecessary and deprecated.

## What I did
Remove the deprecated scrolling attribute from the iframe.

## How to test

- Is this testable with Jest or Chromatic screenshots? One could test the iframe for the absence of the attribute, but this would not make much sense.


shilman (approved):
LGTM. @ndelangen ?

PR #12271, opened 2020-08-26
CLI: Add Aurelia template

tooppaaa:
Issue: #12173

## What I did
Start to setup aurelia template

## What's left

- [ ] fix generator prompting something...
- [ ] css is not loaded in stories
- [ ] events are not triggered in Header and Page, something with .trigger / .delegate / .call

## How to test

- E2E


tooppaaa:
@brandonseydel Looks like you implemented the Aurelia app in storybook, are you able to assist on this ?

PR #12274, opened 2020-08-26
Components: Add additionalActions prop to Preview block

phated:
Issue: N/A

## What I did

This adds an `additionalActions` prop to the `Preview` component to specify additional items in the action bar. I also added a story to show how this can be used to add something like a button to "Open in GitHub". I was going to take a screenshot to attach, but I couldn't figure out the building for the official storybook.

## How to test

- Is this testable with Jest or Chromatic screenshots? I'm not sure.
- Does this need a new example in the kitchen sink apps? Added to the official storybook
- Does this need an update to the documentation? Don't think so.


jimmyandrade (commented):
LGTM

PR #12282, opened 2020-08-27
[docs] Fix single quote instead of backtick

ar4s:
## What I did
I replaced backticks by single quote.

## How to test
No backticks in the snippet.


shilman (approved):
LGTM. thanks!

PR #12298, opened 2020-08-28
UI: Fix the p > div nesting issue

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/12139

## What I did
- I forced the `p` into a `div`


lib/ui/src/components/sidebar/RefBlocks.tsx
```diff
@@ -229,7 +234,7 @@ export const AuthBlock: FunctionComponent<{ loginUrl: string; id: string }> = ({
 export const ErrorBlock: FunctionComponent<{ error: Error }> = ({ error }) => (
   <Contained>
     <Spaced>
-      <Text>
+      <TextStyle>
```
yannbf, line 237: I believe you can just add a `as="div"` rather than duplicating the styled in a new component:
https://emotion.sh/docs/styled#as-prop

ndelangen: That causes typescript issues.

If you want to work out how to fix that? I didn't want to put in any more time into this, and just get rid of the issue asap.

yannbf: Seems like the type [was only added in version 11](https://github.com/emotion-js/emotion/issues/1137#issuecomment-656372155) of emotion, which still isn't released. Ignore my comment and ship it!

yannbf (approved):

PR #12320, opened 2020-08-30
UI: Add close button to version update notification

Tomastomaslol:
Issue: #12281 

## What I did
- Added state that contains ids of dismissed notification 
- Added new button to that dismisses notifications on click

Kapture 2020-08-30 at 20 07 47


shilman:
@Tomastomaslol this looks awesome! @domyen @ndelangen can you take a look?

domyen:
I'll design a dedicated UI for this and future notifications. We didn't get to that when this feature was first launched. Hang tight!

domyen:
@shilman @Tomastomaslol 👋 

**Strawman flash notifications UI strawman** 
The original notification UI was created for Storybook 5.0 to provide a rudimentary upgrade experience for Storybook users. Fast forward a year and we're encountering rough edges in the UX. My goal here is to design a flash notification UI that:
- Covers more use cases and is more expressive 
- Includes a space for CTAs (dismiss, view, etc)

image


What do you think?

Tomastomaslol:
@domyen Looks great! 🙂 

Should we use a pre-defined list of CTAs? like: `'Close' | 'Open' | 'Dismiss'`

Or should the UI be flexible enough to support very long CTAs? like: `If you would like to close the notification you can click this link. thank you!`

Are you able to share your designs through Zeplin or a similar tool that would allow me to extract the exact CSS properties from your designs?

Have a good day!

domyen:
Updated the design to simplify the UI more. 

- No custom action buttons (that was confusing and opens the door to multiple actions in one notification, an over optimization for now)
- Use an `X` icon to dismiss, this better indicates that the whole notification can be clicked vs a close button
- UI affordances for `:hover` `:active` `:focus` states
- More examples for long titles and descriptions

@Tomastomaslol InVision [here](https://invis.io/B3YQ3Y54A86)(you should be able to use the Inspect feature). I might have to invite you via email if the Inpsect link isn't there.

image

Tomastomaslol:
Hi @domyen! it looks like you have to invite me for me to be able to access the inspect mode in inVision. Could you send me an invite to tomasnygren.854@gmail.com?

Tomastomaslol:
Hi @domyen!

I hope you are well! 🙂 

I'm not sure if I'm doing something strange when I export the "book" icon from inVision but I can't figure out how I can get rid of the background in the SVG. If you got a few spare minutes could you please help me export the book icon without the background? 

Other than that, I think I got all scenarios covered now:

Screen Shot 2020-09-19 at 5 57 05 pm
Screen Shot 2020-09-19 at 6 00 20 pm

Please let me if something looks wrong or if you would like me to cover any other scenarios!

Thanks!

domyen:
Apologies for the delay. Appreciate the work. Will review this tomorrow!

Tomastomaslol:
Hi @domyen!

No worries!  Would it be possible to get the book icon as a single string? so I could add it to https://github.com/storybookjs/storybook/blob/next/lib/components/src/icon/icons.tsx#L3 and implement in a similar way to other icons  🙂

domyen:
Hey Tomas, apologies again for the delay. Thank you for taking this on, we're super grateful! 🙏 

**CI**
Usually we'd do UI Review in Chromatic but the CI seems to be failing for some reason. Do you mind taking a look at getting CI running again? Maybe there's a fix on `next` that we can merge into this branch? cc @shilman. 

I'll review here instead, but we'll need to get CI green to QA the behavior in the published SB and to merge the code.  

**Styling**
- Book icon is in the SB icons (go [here](https://next--storybookjs.netlify.app/official-storybook/?path=/story/basics-icon--labels) and cmd + f `book`)
- Use the `closealt` icon for the close button
- Subheadline should be regular font weight
- Remove emojis from the stories (we're moving away from this pattern)
- When there's headline only it should be vertically centered.

**Implementation**
- Keep `itemContent` and `itemDismiss` inside `item`. It's fine if `item` is one big file. Maintaining intertwined functionality across multiple files can be a headache, especially when `itemDismiss` and `itemContent` will only be used with the `item`.

**Stories**
- For visual testing purposes, create a story for each discrete state in `item`

I'm also on Discord if it helps to chat! Find me in the #design channel or `@domyen`.

shilman:
@Tomastomaslol apologies i was messing around with your PR a bit. i think there's some circular dependency that gets created when `@storybook/api` depends on `@storybook/components`. i think the solution is to type it as string and then cast it ... WDYT?

Tomastomaslol:
@shilman thanks for taking your time helping me with this PR. I wasn't sure what the best solution would be but your suggestion makes sense.

There are still some test failures but I'm pretty confident they are not related to the code I wrote and will be fixed with https://github.com/storybookjs/storybook/pull/12746. Please let me know if I have drawn the wrong conclusion and need to spend some more time figuring out why tests are failing.

@domyen I have implemented your feedback. Please let me know if there is anything you would like me to change.

Screen Shot 2020-10-14 at 9 19 10 am
Screen Shot 2020-10-14 at 9 14 43 am

I'm pretty happy with the code and would appreciate a review. 🙂

lib/ui/src/components/notifications/item.tsx
```diff
@@ -1,34 +1,141 @@
...
+const SubHeadline = styled.div(({ theme }) => ({
+  color: theme.base === 'light' ? 'rgba(255,255,255,0.7)' : ' #999999',
+  fontSize: 11,
+  fontWeight: 'normal',
+  marginTop: 3,
+  marginRight: 5,
+  lineHeight: 1,
+}));
+
+const truncateLongHeadlines = (headline: string, length = 30) =>
```
domyen, line 73: Use CSS text-overflow ellipsis here instead of a function.

lib/ui/src/components/notifications/notifications.tsx
```diff
@@ -1,7 +1,6 @@
 import React, { FunctionComponent } from 'react';
```
domyen, line 1: Rename this component and filename NotificationList

lib/ui/src/components/notifications/item.tsx
```diff
@@ -1,34 +1,141 @@
-import React, { FunctionComponent } from 'react';
+import React, { FunctionComponent, SyntheticEvent } from 'react';
```
domyen, line 1: Rename this component and filename NotificationItem

lib/ui/src/components/notifications/item.tsx
```diff
@@ -1,34 +1,141 @@
...
-}> = ({ notification: { content, link } }) => {
+  setDismissedNotification: (id: string) => void;
+}> = ({ notification: { content, link, onClear, id, icon }, setDismissedNotification }) => {
+  const dismissNotificationItem = () => {
+    setDismissedNotification(id);
+    onClear();
+  };
   return link ? (
-    <NotificationLink to={link}>{content}</NotificationLink>
+    <NotificationLink to={link}>
```
domyen, line 132: When there is a link present the NotificationItem should have a hover effect. Right now the hover effect is blanket applied to all NotificationItems even if they aren't interactive.

lib/ui/src/components/notifications/item.stories.js
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-
 import NotificationItem from './item';
```
domyen, line 3: Gentle request: Can we add an action here so that we can confirm onClick in the Actions addon tab?

If you use [Args](https://storybook.js.org/docs/react/writing-stories/args) we should get this automatically.

lib/ui/src/components/notifications/item.tsx
```diff
@@ -1,34 +1,141 @@
...
+
 export const NotificationItemSpacer = styled.div({
   height: 48,
 });
 
 const NotificationItem: FunctionComponent<{
   notification: State['notifications'][0];
-}> = ({ notification: { content, link } }) => {
+  setDismissedNotification: (id: string) => void;
+}> = ({ notification: { content, link, onClear, id, icon }, setDismissedNotification }) => {
```
domyen, line 126: `onDismiss` for consistency with button name?

domyen (changes requested):

domyen:
Thanks for the refinements @Tomastomaslol. 

- I committed some style refinements and reviewed the UI code. See the comments/questions above.
- Let's get a deeper code review from another maintainer (maybe @ndelangen or @ghengeveld)
- Once that's done, we're good to go!

lib/ui/src/components/notifications/item.tsx
RichardVoyles09: The best app

RichardVoyles09 (approved):

lib/api/src/modules/notifications.ts
```diff
@@ -3,7 +3,15 @@ import { ModuleFn } from '../index';
 export interface Notification {
   id: string;
   link: string;
-  content: string;
+  content: {
+    headline: string;
+    subHeadline?: string;
```
ndelangen, line 8: Should we allow ReactNode here? So `span`, `strong` etc is possible?

lib/api/src/modules/notifications.ts
```diff
@@ -3,7 +3,15 @@ import { ModuleFn } from '../index';
...
+  };
+
+  icon?: {
+    name: string;
```
ndelangen, line 12: Should this be a enum?

lib/ui/src/components/notifications/notifications.tsx
```diff
@@ -28,11 +27,22 @@ const NotificationList: FunctionComponent<{
...
+        dismissedNotificationsIds.indexOf(notification.id) === -1 ? (
+          <NotificationItem
+            key={notification.id}
+            setDismissedNotification={(id: string) =>
+              setDismissedNotificationsWithIds([...dismissedNotificationsIds, ...[id]])
+            }
+            notification={notification}
+          />
+        ) : null
+      )}
```
ndelangen, line 45: What's the reason behind having a list of dismissed notifications id's rather then removing the notification from the state?

https://github.com/storybookjs/storybook/blob/341fc2d90d4f3ae42fd9b68173b9f7b1856917f0/lib/api/src/modules/notifications.ts#L30-L41

ndelangen (changes requested):

lib/api/src/modules/notifications.ts
shilman: We had that before but it caused a weird circular dependency, so we loosened it to get around that.

ndelangen: ok

lib/api/src/modules/notifications.ts
Tomastomaslol: Good point. I can't see a good reason to limit this to just string. Will add ReactNode too subHeadline and add an example for it.

lib/ui/src/components/notifications/notifications.tsx
Tomastomaslol: I must have missed this. I  changed the implementation to use `clearNotification` instead of hiding the notifications.

lib/api/src/modules/notifications.ts
Tomastomaslol: Quick update: added `ReactNode` to `subHeadline` but kept `headline` as `string` to be able to have a title attribute for the headline: https://github.com/storybookjs/storybook/pull/12320/files#diff-f0193139de5892a0c79f1452e9f8bc49391f4aab7f7d947d83f05d53a13e9426R92

Tomastomaslol:
@ndelangen @domyen Thank you for the review. I have now addressed all the points raised in the review. Please let me know if there is something else you would like me to take a look at.

ndelangen (approved):

lib/api/src/modules/notifications.ts
ndelangen: 👍

ndelangen:
I say it's good to merge!

domyen (approved):

shilman:
Thanks for your hard work and patience on this @Tomastomaslol 🙏  Merging!

PR #12329, opened 2020-08-31
Adapt text from README for hide controls section

novellac:
Issue: #12260

This PR addresses #12260. The [docs](https://storybook.js.org/docs/vue/essentials/controls) previously did not include instructions on how to disable controls for specified fields, but the information was available on a [README file](https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-can-i-disable-controls-for-certain-fields-on-a-particular-story).

## What I did
Adapted the [section of the README](https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-can-i-disable-controls-for-certain-fields-on-a-particular-story) file into a new section in the [Controls](https://storybook.js.org/docs/vue/essentials/controls) documentation.

I would also like to include documentation on how to disable controls for an entire story file, but I don't know how to do that! (And I figure it can be it's own PR.)

## How to test
This is an update to the docs. It should be reviewed for clarity, but I doubt any automated tests are applicable.

### DIFF
--- a/docs/essentials/controls.md
+++ b/docs/essentials/controls.md
@@ -188,6 +188,29 @@ And here's what the resulting UI looks like:
 
 ![Controls addon expanded](./addon-controls-expanded.png)
 
+## Hide controls for certain fields on a particular story
+
+The `argTypes` annotation can be used to hide controls for a particular row, or even hide rows.
+
+Suppose you have a `Button` component with `borderWidth` and `label` properties (auto-generated or otherwise) and you want to hide the `borderWidth` row completely and disable controls for the `label` row on a specific story. Here's how you'd do that:
+
+```js
+import { Button } from 'button';
+
+export default {
+  title: 'Button',
+  component: Button,
+};
+
+export const CustomControls = (args) => <Button {...args} />;
+CustomControls.argTypes = {
+  borderWidth: { table: { disable: true } },
+  label: { control: { disable: true } },
+};
+```
+
+Like [story parameters](https://storybook.js.org/docs/react/writing-stories/parameters), `args` and `argTypes` annotations are hierarchically merged, so story-level annotations overwrite component-level annotations.
+
 ## Hide NoControls warning
 
 If you don't plan to handle the control args inside your Story, you can remove the warning with:


docs/essentials/controls.md
```diff
@@ -188,6 +188,29 @@ And here's what the resulting UI looks like:
 
 ![Controls addon expanded](./addon-controls-expanded.png)
 
+## Hide controls for certain fields on a particular story
+
+The `argTypes` annotation can be used to hide controls for a particular row, or even hide rows.
+
+Suppose you have a `Button` component with `borderWidth` and `label` properties (auto-generated or otherwise) and you want to hide the `borderWidth` row completely and disable controls for the `label` row on a specific story. Here's how you'd do that:
```
jonniebigodes, line 195: @novellac we have room for improvement here and we can avoid having a "carbon copy" (pardon the bad pun) from the readme.

We probably can go with:
```md

## Customize stories with argTypes

Let's assume we have a `Button` component that contains the following properties, `borderWidth` and `label` and we need to create a story that hides the `borderWidth` and disable any controls for the `label`.  

How could we implement such a story? 

We can use the `argTypes` annotation in our stories to customize how they will be rendered in the UI:
```

Probably something like this.

Let me know if you're ok with the wording or we can always fine tune it.

docs/essentials/controls.md
```diff
@@ -188,6 +188,29 @@ And here's what the resulting UI looks like:
...
+  title: 'Button',
+  component: Button,
+};
+
+export const CustomControls = (args) => <Button {...args} />;
+CustomControls.argTypes = {
+  borderWidth: { table: { disable: true } },
+  label: { control: { disable: true } },
+};
+```
```
jonniebigodes, line 210: @novellac if you don't mind a couple of things regarding the snippet:
- This would be probably better suited for a snippet file, i'll leave the naming to you. And as this is based on example in the README. It's for React, so if you don't mind place it inside that folder.
- We can make a adjustment to the snippet and go with:
```js
// Button.stories.js

import React from 'react';
import { Button } from './Button';

export default {
  title: 'Button',
  component: Button,
};

const Template = (args) => <Button {...args}/>;

export const CustomControls= Template.bind({});
CustomControls.argTypes = {
 // hides the property in the UI.
 borderWidth: { table: { disable: true } },
  // disables controls for the label in the UI
  label: { control: { disable: true } },
};
```

jonniebigodes (changes requested):
@novellac sorry for not getting back at you sooner, but i had to shift my attention elsewhere and this pull request was "left hanging" (pardon the bad pun). I left some comments, let me know if you're ok with or we can always adjust it.

Stay safe

docs/essentials/controls.md
novellac: That's probably fine for now, but! If you're ok with improvements (famous last words):
- I'd like to add an additional bit about how to disable controls for all stories in the `export default` area (I think this is more explicit).
- Since hiding the entire `borderWidth` row in the example is done with `table.disable`, I'd like to add a line in the table under the [Customizing](https://storybook.js.org/docs/react/writing-docs/doc-blocks#customizing) section on the [Doc Blocks](https://storybook.js.org/docs/react/writing-docs/doc-blocks) page (I can make a PR for this). We could then leave off that part of the example and simplify the snippet to just address the disabled controls.
- I'd like to have the heading specify that this section addresses how to hide controls, since (hopefully) the section would just address hiding controls.

docs/essentials/controls.md
novellac: Sure I can make this a snippet. I'll also try to figure out how to make a Vue snippet for this (though I might need some help).

docs/essentials/controls.md
jonniebigodes: @novellac no need for the famous last words, i'm more than ok with improvements to the documentation if they add add value to it and help out not only you but the rest of the community :)

Regarding your list:

Looking at the the pull request and the current state of the documentation, factoring in the first item, we could do better, we can break this into smaller parts, go from component to story. This will improve the documentation and help out the community understand that how they can use the annotation more effectively and that they are not strictly bound to use it in a story. 

Regarding your second item, i'm ok with adding a entry to the table and we could probably link it to this page and we present the example. 

Regarding your third item, i went with that phrasing based of the fact that in future releases the behavior can change and probably more options will be added. So it turned out a bit more agnostic.

Let me know if you're ok with this and we can make the adjustments and merge it :)

docs/essentials/controls.md
jonniebigodes: Let me know once the snippets are up and we go from there.

jonniebigodes: @novellac just a friendly reminder if you're willing to address the feedback that i've mentioned. We could do without the vue snippet for now and work on it later with a follow up pull request. 

Let me know when you've addressed this.

Stay safe

jonniebigodes:
@novellac with the merge of #12917 the controls documentation was updated addressing the issue you attached to this pull request (#12260). I went with a simplified approach on how to use the feature. 

Thank you for the time and effort you put into the pull request and willingness to help fine tune the documentation. But i'm going to close this pull request as it's now a bit outdated. 

Stay safe

PR #12375, opened 2020-09-03
Addon-docs: Apply transformSource to any SourceType

phated:
Issue: N/A

## What I did

I made the `transformSource` function (if specified) apply to all `SourceType` snippets.

While trying to modify the snippets produced for `SourceType.DYNAMIC`, I noticed that the `transformSource` option was only being applied to `SourceType.CODE` snippets. This seems to be due to it being called inside `enhanceSource`, but the `SourceType` refactoring came after, so I think it got missed in the shuffle.

## How to test

- Is this testable with Jest or Chromatic screenshots? I don't think so, I see some other refactors that can happen with enhanceSource and then things can be tested together.
- Does this need a new example in the kitchen sink apps? I believe the kitchen sink app will get these through it's `transformSource` config.
- Does this need an update to the documentation? Nope, the recipe actually makes it sound like this is globally called already.

### DIFF
--- a/examples/official-storybook/stories/addon-docs/transform-source.stories.js
+++ b/examples/official-storybook/stories/addon-docs/transform-source.stories.js
@@ -11,15 +11,15 @@ export default {
 
 export const code = () => 'StoryType "CODE" story which has source transformed';
 code.parameters = {
-  docs: { source: { type: 'CODE' } },
+  docs: { source: { type: 'code' } },
 };
 
 export const dynamic = () => 'StoryType "DYNAMIC" story which has source transformed';
 dynamic.parameters = {
-  docs: { source: { type: 'DYMANIC' } },
+  docs: { source: { type: 'dynamic' } },
 };
 
 export const auto = () => 'StoryType "AUTO" story which has source transformed';
 dynamic.parameters = {
-  docs: { source: { type: 'AUTO' } },
+  docs: { source: { type: 'auto' } },
 };
--- a/examples/official-storybook/stories/addon-docs/transform-source.stories.js
+++ b/examples/official-storybook/stories/addon-docs/transform-source.stories.js
@@ -21,5 +21,5 @@ dynamic.parameters = {
 
 export const auto = () => 'StoryType "AUTO" story which has source transformed';
 dynamic.parameters = {
-  docs: { storySource: 'AUTO', source: { type: 'AUTO' } },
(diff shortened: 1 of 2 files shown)


shilman (changes requested):
Looks great! Can you add an `official-storybook` example in `Addons/Docs` to demo this?

phated:
Yep! I'll put one together now 😄

phated:
Here's a screenshot of what this looks like in the `official-storybook`!
Screen Shot 2020-09-09 at 5 49 15 PM

examples/official-storybook/stories/addon-docs/transform-source.stories.js
```diff
@@ -0,0 +1,25 @@
...
+      transformSource(src, ctx) {
+        return `// We transformed this!\n const example = ${src};`;
+      },
+    },
+  },
+};
+
+export const code = () => 'StoryType "CODE" story which has source transformed';
+code.parameters = {
+  docs: { source: { type: 'CODE' } },
```
shilman, line 14:
```suggestion
  docs: { source: { type: 'code' } },
```

examples/official-storybook/stories/addon-docs/transform-source.stories.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const dynamic = () => 'StoryType "DYNAMIC" story which has source transformed';
+dynamic.parameters = {
+  docs: { source: { type: 'DYMANIC' } },
```
shilman, line 19:
```suggestion
  docs: { source: { type: 'dynamic' } },
```

examples/official-storybook/stories/addon-docs/transform-source.stories.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const auto = () => 'StoryType "AUTO" story which has source transformed';
+dynamic.parameters = {
+  docs: { source: { type: 'AUTO' } },
```
shilman, line 24:
```suggestion
  docs: { source: { type: 'auto' } },
```

shilman (changes requested):
the string values are not capitalized 😭

phated:
Whoops, sorry about that. I've applied your suggestions!

shilman (approved):
🙌

PR #12376, opened 2020-09-03
Addon-docs: Prefer flow's union elements over raw values

phated:
Issue: N/A

## What I did

I changed the generation of Flow's union type to prefer the `elements` over `raw` values because the `raw` value doesn't include fully resolved types (as demonstrated in the new tests I added).

## How to test

- Is this testable with Jest or Chromatic screenshots? Yep! Added Jest tests.


shilman (approved):
LGTM

PR #12380, opened 2020-09-04
docs: remove reference to `configure()`

diedu89:
Issue:

## What I did

Removed reference to deprecated function `configure()`

## How to test

- Is this testable with Jest or Chromatic screenshots? :man_shrugging: 
- Does this need an update to the documentation? it is an update to documentation

### DIFF
--- a/docs/writing-stories/naming-components-and-hierarchy.md
+++ b/docs/writing-stories/naming-components-and-hierarchy.md
@@ -111,4 +111,4 @@ Which would result in this story ordering:
 7. `Components` and `Components/*` stories
 8. All other stories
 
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@diedu89 first and foremost sorry for taking a bit longer to check this, but i wanted to make sure what you're proposing would lead to inconsistencies. And this is the case, upon checking, technically the `configure` method is not deprecated, it's used in the react-native version of Storybook, to allow configuration and sorting stories. With that i'm going to close this pull request to avoid introducing misleadings in the documentation.

It goes without saying that despite this pull request isn't going to be merged and make it's way into the documentation. We're more than welcome for any further pull requests from you.

Stay safe.

PR #12383, opened 2020-09-04
Core: require.resolve loaders and add missing dependencies

merceyz:
## Issue

Storybook is passing webpack loaders and babel plugins/presets as strings instead of their absolute location causing it to rely on hoisting to bring the dependencies to the top level so webpack/babel can find them which isn't guaranteed. Some of the dependencies are also undeclared which causes the same problem

## What I did

- `require.resolve` plugins and loaders
- Add missing dependencies

## How to test

Run the tests and e2e integrations


gaetanmaisse (approved):

shilman (approved):
LGTM. thanks so much for all the work on this! 🙏

PR #12436, opened 2020-09-09
Addon-controls: Default to radio control for small enums

phated:
Issue: N/A

## What I did

I wanted to propose this and see what others thought.

I changed the default inferred control for an enum to be a radio input for any enum that is 5 elements or less. I think radio inputs are a more natural default, but they get unwieldy with large enums. I just picked 5 randomly but could see it being adjusted.

Anyway, what do you think? If this makes sense, I can update the docs and stuff.

Edit: Oh, I forgot my screenshot from the `cra-ts-essentials` storybook!

Screen Shot 2020-09-09 at 3 29 45 PM

## How to test

### DIFF
--- a/docs/essentials/controls.md
+++ b/docs/essentials/controls.md
@@ -96,7 +96,7 @@ Up until now, we only used auto-generated controls based on the component we're 
 
 By default, Storybook will add controls for all args that:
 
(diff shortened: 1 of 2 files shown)


shilman (approved):
Nice idea, I like it! Is there documentation that needs updating?

PR #12440, opened 2020-09-10
Ember: Add `emberOptions` to `main.js` config ✨

dbendaou:
This PR aims to make *ember polyfills* work with storybook.
Because Storybook use a custom pipeline to make ember precompile, it lake for a step that is not present in storybook.
This missing step check if any polyfills are available/used with the app and precompile them before the render of storybook.

As it was too difficult to reproduce this step on Storybook, I've instead had a way to pass some `frameworkOptions` from the `.storybook/main.js`

Issue:

## What I did
As it was too difficult to reproduce this step on Storybook, I've instead had a way to pass some `frameworkOptions` from the `.storybook/main.js`

Here's what my `main.js` look like (this example use the [ember-named-block-polyfill](https://github.com/ember-polyfills/ember-named-blocks-polyfill))
```javascript 
const namedBlockPolyfill = require('ember-named-blocks-polyfill/lib/named-blocks-polyfill-plugin');

module.exports = {
  frameworkOptions: {
    polyfills: [namedBlockPolyfill],
  },
[...]
};
```

With this, I was able to retrieve the `frameworkOptions` object within the `options` parameter passed to `babel` function.

## How to test

This can be tested by building the ember example by adding a polyfill to it, you can use either the same as me or another one.

Once this will be merged, I'll submit another PR to teach how to use this.

**Here's an example** of a [component using the ember-named-block-polyfill feature](https://gist.github.com/dbendaou/9290db0f27845c2b84d26d2ea942c9cf) and the [story that use it](https://gist.github.com/dbendaou/9290db0f27845c2b84d26d2ea942c9cf)

### DIFF
--- a/app/ember/README.md
+++ b/app/ember/README.md
@@ -27,3 +27,22 @@ You can also build a [static version](https://storybook.js.org/docs/ember/workfl
 - [Basics](https://storybook.js.org/docs/ember/get-started/introduction)
 - [Configurations](https://storybook.js.org/docs/ember/configure/overview)
 - [Addons](https://storybook.js.org/docs/ember/configure/storybook-addons)
+
+## Working with polyfills
+If you need to use a polyfill that is already in use in our Ember application,
+you will need to add some options to have Storybook working with them.
+
+In this example we'll use the [named-block-polyfill](https://github.com/ember-polyfills/ember-named-blocks-polyfill).
+This example also assume that you already have the package in your `package.json`.
+
+In your `.storybook/main.js` you can add the following lines: 
+```javascript
+const namedBlockPolyfill = require('ember-named-blocks-polyfill/lib/named-blocks-polyfill-plugin');
+
+module.exports = {
+  emberOptions: {
+    polyfills: [namedBlockPolyfill],
+  },
+  [...]
+};
+```
--- a/app/ember/src/server/framework-preset-babel-ember.ts
+++ b/app/ember/src/server/framework-preset-babel-ember.ts
@@ -17,6 +17,8 @@ export function babel(config: Configuration, options: any) {
     options.presetsList.forEach((e: any, index: number) => {
       if (e.preset && e.preset.emberOptions) {
         emberOptions = e.preset.emberOptions;
+        // eslint-disable-next-line no-param-reassign
+        delete options.presetsList[index].preset.emberOptions;
       }
     });
   }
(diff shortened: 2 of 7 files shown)


dbendaou:
For some unknown reason, letting the `frameworkOptions` key in the `options.presetsList` did break the render of Storybook.
Screenshot 2020-09-10 at 10 58 52

the `delete options.presetsList[index].preset.frameworkOptions;` fixed the issue.

shilman:
@dbendaou FYI, @tooppaaa has implemented `reactOptions` for react based on our conversation about `emberOptions`. you might want to take a look at how he did it: https://github.com/storybookjs/storybook/pull/12470

shilman (changes requested):
Nice @dbendaou -- a compact set of changes!

One thing from #12470 is the ability for a preset to return `<framework>Options` ... I'm wondering whether we should merge that first and then implement this on top of that? @tooppaaa do you have any thoughts about this?

A couple requests:
- can you call it `emberOptions` instead of `frameworkOptions`? That way it won't be a breaking change if we try to support multiple frameworks at once in the future.
- Can you add an example in the monorepo that shows this in action? Maybe the primary use case you had in mind?

dbendaou:
> One thing from #12470 is the ability for a preset to return <framework>Options ... I'm wondering whether we should merge that first and then implement this on top of that? @tooppaaa do you have any thoughts about this?

Yeah I saw, I think that there's some interesting things that we could leverage for Ember. 
I would however having this merge and then work on the refact if that's ok for you.

And for the requested changes about `frameworkOptions` and adding some example of it in the `examples/ember-cli` I'll make them in the day and push the changes :)

dbendaou:
Added some documentation in the `ember/README`. 
And I've also update the ember example repo to add a story of a polyfill example. 

Let me know if that sounds good to you @shilman :) 
✨

shilman:
@dbendaou looking great! just waiting for the CI checks 👀

tooppaaa:
> @tooppaaa do you have any thoughts about this?

This is taking the same direction so all good to me !

We just need to make sure my changes will not break this one. Let's sync after the first one is merged to make sure everything is working as expected.

dbendaou:
Chromatic circle ci job seems to have timeout :/

dbendaou:
Chromatic circle ci job seems to have timeout, may we restart it ? I don't have permissions

shilman:
Restarted & passing now

shilman:
@gossi can you review?

gossi (approved):
Amazing stuff here!

PR #12451, opened 2020-09-10
dependencies: use `dequal` for equality checks

lukeed:
Issue: https://github.com/storybookjs/storybook/issues/12433

## What I did

Replaced `fast-deep-equal` with `dequal`, which is smaller (>100B) and [faster](https://github.com/lukeed/dequal#benchmarks). 
This `dequal/lite` variant has an identical featureset.

## How to test

No code changes. All tests (`yarn test --core`) pass locally.

I was asked to do a performance comparison before/after, but I've no idea how to set that up within this project.


shilman:
Storybook_Benchmark_›_Explorer__Install__Start__Build_

Interestingly install size seems to have gone up a bit. I'm guessing this is because some other library already depends on `fast-deep-equal` so we're not benefitting in that way. However, manager render time seems to have dropped.

We're doing a lot of perf work right now, so I'm going to leave this PR open for awhile and update it from `next` to see how the perf holds up over time.

lukeed:
I guess it would depend how you're measuring install size. Is it raw `node_module` weight? If that's the case, `dequal` has 6 total files (3 formats for 2 variants), but you're only loading one here.

Also, what module is this charting? Checked out the dep graph (https://npm.anvaka.com/#/view/2d/%2540storybook%252Fapi) for `@storybook/api` and there's only 1 direct link to fde

shilman:
The benchmark code is here: https://github.com/storybookjs/bench

The runner is `cra_bench` here: https://github.com/storybookjs/storybook/blob/next/scripts/run-e2e-config.ts

It runs as part of our e2e tests on every push. I don't expect much performance change one way or another, but since this is purely a performance-related change AFAIK I'd like to see if it makes a difference.

lukeed:
Hey there, is there something I can do for this?

ndelangen:
I've replaced this PR with this one:
https://github.com/storybookjs/storybook/pull/18608

Because I was not able to get this one updated with `next` (I do not have the authorisation)

PR #12454, opened 2020-09-11
Core: Fix Yarn monorepo compatibility 

gaetanmaisse:
Follows https://github.com/storybookjs/storybook/pull/11753, I created a new branch as I can't push on @merceyz's fork.

## Issue

> Storybook doesn't work correctly in monorepos with both node_modules (Yarn 1/2) and PnP (Yarn 2) because it's trying to require the addons declared in .storybook/main.js from @storybook/core instead of requiring them on behalf of the config.

> The e2e tests currently pass because Yarn has a compatibility feature, called fallbackMode, that allows dependencies to require top-level dependencies without declaring them as their own. In a monorepo this fails as the top level no longer declares these dependencies, the workspace does, and under the node_modules linker, the addons are not guaranteed to be hoisted to the same level as @storybook/core.

## What I did

All credits go to @merceyz! And a big thanks to @ndelangen for the debugging hour we spend together 😃 

 - Disabled the fallbackMode to highlight the issue in the e2e tests: https://yarnpkg.com/configuration/yarnrc#pnpFallbackMode
 - Resolve presets relatively to configDir
 - Fix issue with preset name checks as they are now resolved path and no more simple names.

## How to test
 - Disable the fallback mode and try to use storybook in a PnP project
 - `example-v2-yarn2` has been updated to not use fallback mode, it should still be 🟢

### DIFF
--- a/app/react/src/server/framework-preset-cra.ts
+++ b/app/react/src/server/framework-preset-cra.ts
@@ -7,7 +7,7 @@ type Preset = string | { name: string };
 const checkForNewPreset = (presetsList: Preset[]) => {
   const hasNewPreset = presetsList.some((preset: Preset) => {
     const presetName = typeof preset === 'string' ? preset : preset.name;
-    return presetName === '@storybook/preset-create-react-app';
+    return presetName.includes('@storybook/preset-create-react-app');
   });
 
   if (!hasNewPreset) {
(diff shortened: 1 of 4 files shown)


merceyz:
> I created a new branch as I can't push on @merceyz's fork.

Even with `Allow edits by maintainers` checked... invited you so you can edit directly

merceyz:
I cherry-picked your changes into #11753 and fixed the check so it handles Windows paths

gaetanmaisse:
Closed in favor of base PR: https://github.com/storybookjs/storybook/pull/11753

PR #12460, opened 2020-09-11
Storyshots: Allow taking a screenshot of just a specific element

StephanBijzitter:
I have a lot of very small React components, but the screenshots taken are of the entire viewport, which makes the files a lot larger than they need to be. Furthermore, it's much more difficult to see a 1 or 2 pixel difference on a file that's 600px wide, when the component itself is 50px wide.

## What I did
I added an optional return value to the `beforeScreenshot` function; whenever an ElementHandle is returned, it is used as the basis for the screenshot. This is _not_ a breaking change.

## How to test

Probably, but there were not any tests yet.

Don't think so.

Yes, and done.


shilman (approved):
Nice feature!

PR #12461, opened 2020-09-11
Addon-docs: Resolve vue-docgen-loader from @storybook/vue

merceyz:
## Issue

`@storybook/addon-docs` is declaring `vue-docgen-api` and `vue-docgen-loader` as peer dependencies which means `@storybook/addon-essentials` needs to declare them as well, but this means that the user needs to provide them while what we're actually after is to load them from `@storybook/vue`.

Continuation of https://github.com/storybookjs/storybook/pull/11753 and https://github.com/storybookjs/storybook/pull/12383

## What I did

Changed the resolution to get the `vue-docgen-loader` from `@storybook/vue`

## How to test

The PnP e2e test has been updated and should be green

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
           command: yarn info @storybook/core
       - run:
           name: run e2e tests
-          command: yarn test:e2e-framework --use-yarn-2 cra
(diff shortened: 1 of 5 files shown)


shilman (approved):
LGTM!

PR #12471, opened 2020-09-14
Add layout option "none" to remove inline styles from body

johnnybenson:
Issue: https://github.com/storybookjs/storybook/issues/12434

## What I did

Added "none" as a Layout option in StoryRenderer.tsx and updated Official Storybook stories to include None in layouts

I looked at this PR to try to understand why so many inline styles are added in the first place: https://github.com/storybookjs/storybook/pull/10643

This "none" option may not be sufficient as-is to unset other layouts if styles are always merged somewhere else downstream.

### DIFF
--- a/examples/official-storybook/stories/core/layout.stories.js
+++ b/examples/official-storybook/stories/core/layout.stories.js
@@ -29,5 +29,8 @@ CenteredBlock.parameters = { layout: 'centered' };
 export const CenteredInline = () => <Box display="inline-block">centered</Box>;
 CenteredInline.parameters = { layout: 'centered' };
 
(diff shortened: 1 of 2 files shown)


ghengeveld:
I think this is a valid option we should provide. However as pointed out in the PR description, the current implementation will not work, because if you apply a different layout elsewhere and switch from such story to a story with `layout: "none"`, the styles won't be removed.

The solution we'll have to implement is replace inline styles with classnames, as pointed out in #12041. That way we can simply swap classnames (or change a data attribute value). That would resolve both #12041 and fix this PR. So until #12041 is resolved, this PR is on hold.

PR #12492, opened 2020-09-15
React: Fix reactDocgen option when false

AndrewLeedham:
Issue: https://github.com/storybookjs/storybook/issues/12491

## What I did
- Stopped `babel-plugin-react-docgen` being added when `typescript.reactDocgen` option is set to `false`.
- Updated the typescript docs to reflect change (no more `"none"` option, which did not seem to do anything).
- Added tests for the `false` case as well as the `"react-docgen-typescript"` case.


shilman (approved):
LGTM. Thank you for the fix! 🙏

PR #12509, opened 2020-09-17
Web Components: Add script tag support

zolk:
Following the example of #12089 for HTML, this adds `<script>` tag support for Web Components.

Not sure if we want to move `simulate-pageload.ts` into a shared location rather including a separate copy with each app?

## How to test

Example stories have been added to `web-components-kitchen-sink` in  `script.stories.js`, one rendered with `lit-html` and another as a basic string.

### DIFF
--- a/app/web-components/package.json
+++ b/app/web-components/package.json
@@ -40,6 +40,7 @@
     "@babel/plugin-syntax-import-meta": "^7.10.4",
     "@babel/preset-env": "^7.11.5",
     "@storybook/addons": "6.1.0-alpha.12",
+    "@storybook/client-api": "6.1.0-alpha.12",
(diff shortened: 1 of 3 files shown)


jonspalmer:
I extracted the simulate-pageload to client-api in #12522. Would be great to take advantage of that for this PR.

shilman (changes requested):
Thanks @zolk, looks great!! Can you please update to take advantage of @jonspalmer 's refactor? 🙏

zolk:
> Thanks @zolk, looks great!! Can you please update to take advantage of @jonspalmer 's refactor? 🙏

Updated!

shilman (approved):
Super! 💯

PR #12511, opened 2020-09-18
Point users towards docs for getting started

tmeasday:
The getting started info here was out of date, and it is simpler to keep it in one spot.

Also fix up some links into the docs that were broken.


shilman (approved):
LGTM!

PR #12522, opened 2020-09-20
Server: Add support for script tags

jonspalmer:
Issue: Script tags included in server story html don't get included in the preview pane

## What I did

* moved the simulate-pageload added in #12089 from html to client-api
* leverage new util methods in server preview
* added examples to server-kitchen-sink


shilman (approved):
LGTM. Thanks for the refactor!

PR #12533, opened 2020-09-21
CLI: Change suggested upgrade command to sb@latest

nnnnoel:
Issue:

## What I did
Fix upgradeCommant from @next to @latest
## How to test

I found the upgrade warning, but the next version is in alpha. so i could think i cannot install the pre-release version. we need to upgrade to stable / latest release.


shilman (approved):
I can get behind this. The original rationale was to always use the most up-to-date version of the CLI, but I suppose in the future it could contain breaking changes.. Thanks!

PR #12535, opened 2020-09-21
React: Fix fast refresh

yannbf:
## What I did
In 6.1 alpha, support for React fast refresh was added. Unfortunately the fast refresh config ended up adding a new babel-loader rather than integrating the rules with the existing loader. This PR fixes that by separating the webpack and babel work in the react preset.

The fast refresh example component + story was moved from official-storybook to cra-kitchen-sink, given that the official-storybook is deliberately removing the babel loader and adding a custom one.

## How to test
`yarn bootstrap`
`yarn --cwd examples/cra-kitchen-sink storybook` and tweak with either the reactOptions in main.js or add REACT_REFRESH=true then change to false in the `.env` file.


ndelangen:
Awesome!

ndelangen (approved):

shilman (approved):
Looks great @yannbf

PR #12549, opened 2020-09-23
Addon-docs: Remove leading pipe if using raw value for Flow union

phated:
Issue: N/A

## What I did

I ran into this issue prior to preferring `elements` in Flow's union type. In Flow, a Union can be specified with a leading `|` character, which results in an extra empty string value once processed by storybook. This removes that extra character if we don't have `elements` (for some reason) and must use the `raw` value.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yep, added a couple jest tests.
- Does this need an update to the documentation? Nope


shilman (approved):
LGTM!

PR #12587, opened 2020-09-26
ArgsTable: Remove the "simple" detection for enum types

phated:
Issue: Partial reversion of #11868

## What I did

While working on a very large storybook, I noticed that a subset of types weren't formatted correctly. For example, they had no background. They were really standing out from the rest of the system. These types often looked like `(arg1: string) => 'return1' | 'return2'`.

I traced it back to the `simple` option added in #11868 which added a check that is too simplistic for what it is trying to do. From what I understand, it was trying to avoid adding styling to complex unions, but it didn't consider a union-type value anywhere else in a signature.

My suggestion is to remove this check for now until a more sophisticated check is applied.

## How to test

- Is this testable with Jest or Chromatic screenshots? The previous PR said it added these?


phated:
cc @ajkl2533 as well.

shilman (approved):
WFM cc @ajkl2533

PR #12588, opened 2020-09-26
Addon-docs: Move summary & detail equality check to createSummaryValue

phated:
Issue: N/A

## What I did

I moved the `summary` and `detail` equality comparison into the `createSummaryValue` utility function and cleaned up the call sites that were checking equality themselves. I also added Jest tests for this util method.

This also lays some groundwork for a bigger Flow related PR that I've been working on.

## How to test

- Is this testable with Jest or Chromatic screenshots? Jest, added a new test file!
- Does this need a new example in the kitchen sink apps? Nope
- Does this need an update to the documentation? Nope

If your answer is yes to any of these, please make sure to include it in your PR.

PR #12589, opened 2020-09-27
Addon-docs: Change the way PropType used by ArgTable are generated for Flow types

phated:
Issue: N/A
Requires: #12588 

## What I did

I changed the way Flow types were converted into `PropType` because ArgsTable expects a string for `summary` and `details`, but Flow types passed to the converter will contain rich type information that is often lost when just using the `raw` value (as was done previously).

The way the generation is done, all the previous tests that just used `raw` values are still passing. Additionally, I added a bunch of tests around more specific types of generation.

It might also make sense to have a discussion around whether `ArgsTable` should receive the rich type information produced by react-docgen instead of only operating on string values. I think that would resolve issues like #12587 in a cleaner manner.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yep, added a bunch of jest snapshots.


shilman:
Seeing what looks like a formatting regression in this PR:

ZoomAnnotationContentWnd

Versus what I'm seeing in `next`:

ZoomAnnotationContentWnd

(NOTE: I updated this PR from `next` just in case it was something in the meantime)

@phated can you possibly:
1) comment on whether this was intended and if not, fix it
2) add a story that captures this change in chromatic

Thanks and sorry for the slow slow response time on this 🙈

phated:
Very curious! There was no formatting when I originally worked on this. I probably need some time to investigate where that is being tripped up.

shilman:
closing this for now. please feel free to reopen if you need it!

PR #12597, opened 2020-09-28
Make link to fork-ts-checker-webpack-plugin documentation point to v4

HJain13:
Issue:#11804
> The "See docs" link in Storybook v6's TypeScript configuration guide points to the wrong version of `fork-ts-checker-Storybook v6's package.json declares "fork-ts-checker-webpack-plugin": "^4.1.4" [here](https://github.com/storybookjs/storybook/blob/next/lib/core/package.json#L94), but the link points to the most recent version, which is v5.0.14. Fork-TS-Checker's configuration changed from v4 -> v5, so the docs in question provide a configuration shape that is not compatible with the version used by Storybook.

## What I did
Updated documentation to point to the correct version of a dependency documentation

## How to test

### DIFF
--- a/docs/configure/typescript.md
+++ b/docs/configure/typescript.md
@@ -6,7 +6,7 @@ Storybook has built-in Typescript support, so your Typescript project should wor
 
 ### Default configuration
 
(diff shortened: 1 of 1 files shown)


shilman (approved):
Thank you @HJain13 ! 🙏

PR #12629, opened 2020-09-30
Relative link to docs broken

whoisryosuke:
The link to Storybook docs was relative, which on [the production site](https://storybook.js.org/docs/react/writing-docs/docs-page) links to the **next** branch, instead of the **master**.

Old link using relative:
`https://github.com/storybookjs/storybook/blob/next/addons/docs/README`

New link:
`https://github.com/storybookjs/storybook/blob/master/addons/docs/README.md`

Not sure if there's a way to use the relative link and have it point to master instead of next? But for now, hard-coded link will definitely work.

Issue:

## How to test

### DIFF
--- a/docs/writing-docs/docs-page.md
+++ b/docs/writing-docs/docs-page.md
@@ -2,7 +2,7 @@
 title: 'DocsPage'
 ---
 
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM. thanks!

PR #12638, opened 2020-10-01
Addon-docs: Fix exotic React components in Source block

andezzat:
## The Problem

React Memo & ForwardRef exotic components would show up in the code snippet of storybook as `<[object Object]>` because it didn't know where to lookup the inner render function's name.

This is a bug in the library Storybook uses to render JSX into a string: https://github.com/algolia/react-element-to-jsx-string/issues/352

## What I did

I stole the default `displayNameFn` from the original library [here](https://github.com/algolia/react-element-to-jsx-string/blob/4974512273c86c72415376fea89f9d6e07e2b2e5/src/parser/parseReactElement.js#L15) 
And added what it needed to fetch Memo & ForwardRef component's names.

## How to test

Maybe? I have no idea
No
No


ndelangen (approved):

addons/docs/src/frameworks/react/jsxDecorator.test.tsx
```diff
@@ -103,7 +103,7 @@ describe('renderJsx', () => {
           'item 18',
           'item 19'
         ]}
-       />	
+       />
```
andezzat, line 106: 😂 
For some reason linter worked first time, but not when I forked 🤷 
Sorry!

shilman (changes requested):
Hey @andezzat !! Thanks so much for putting this together -- will make a ton of people happy. Unfortunately, I added a test case to `official-storybook` and it doesn't seem to be working end-to-end. You can find that in the latest commit on this branch.

Would you mind:
- looking into this specific case
- adding more test cases to official-storybook
- reconciling the official-storybook behavior and the jest test cases (look great, but somehow don't line up)

examples/official-storybook/stories/addon-docs/react-forward-ref.stories.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { DocgenButton } from '../../components/DocgenButton';
+
+const ButtonWithForwardRef = React.forwardRef(DocgenButton);
+
+export default {
```
shilman, line 6: can we delete this file now?

shilman (approved):
LGTM! 💯 💯 💯

examples/official-storybook/stories/addon-docs/react-forward-ref.stories.js
andezzat: Woopsie! 😅

ghengeveld:
This appears to be broken:

Screenshot 2020-10-09 at 14 05 32

It seems to originate from `ForwardRefButtonInnerPropTypes` which wraps a button component directly in `forwardRef`, which is not allowed.

shilman:
cc @andezzat can you take a look ☝️ \

andezzat:
Ah yes, I should've made a new component w/ proper `forwardRef` fn args than passing `BaseButton` as-is 😂
PR: #12733

PR #12639, opened 2020-10-01
Core: Add apng support

roblan:
Issue: storybook file-loader wasn't configured to load apng files

## What I did

Fixed typo (anpg -> apng) and added apng extensions to file-loaders

## How to test

no
no
no


shilman (approved):
LGTM. @ndelangen? How common is this format? I think it could also be something for users to configure on their own rather than built into core.

roblan:
@shilman I don't think is very common, but I thought that storybook is kind of 'just works'

It was in design-assets addon though :)

PR #12679, opened 2020-10-06
Fix comment in vue global decorator snippet

philippmunzert:
## What I did

I fixed a wrong comment in the global decorator snippet for vue.


shilman (approved):
LGTM. thanks!

PR #12690, opened 2020-10-07
Core: Log which CSF file is failing to load

mathieudutour:
## What I did

When a story fails to load, Storybook spits out an error but there is no information about it (we don't know which the story is failing).

This PR adds the filename to that error to make it easier to track.


shilman (approved):
LGTM! 💯

PR #12707, opened 2020-10-08
UI: Instant-on manager

ghengeveld:
Issue: #12702

## What I did

This implements an instant-on manager UI by bundling a prebuilt manager with the Storybook package, as well as caching a locally built manager. The cache gets busted whenever a change is made to `manager.js` or the webpack config (typically via `main.js`). There's also a `--no-manager-cache` flag which disables caching and ignores the prebuilt manager. The prebuilt manager is also used when building a static Storybook, assuming all criteria are met.

[▶️ Watch the demo](https://www.youtube.com/watch?v=lQLo4GpSPqw)

## How to test

- Run `start-storybook` against a vanilla, newly installed Storybook which only has `addon-links` and `addon-essentials` installed. It will say "Using prebuilt manager" and skip building the manager (see the startup message, it won't mention the manager build time).
- Run `build-storybook` against the same Storybook, and notice it will use the prebuilt manager here as well.
- Run `start-storybook` against any other Storybook. The first time it will build the manager as normal. Then kill and run `start-storybook` again. Now it will say "Using cached manager" and work the same as above.

The `--no-manager-cache` flag should be added to the docs.

## To do

- [x] Refactor the dev server for readability / maintainability
- [x] Start the browser as soon as the manager is ready
- [x] Show an appealing and useful UI as the preview builds
- [x] Use prebuilt when creating a static build
- [x] Setup our publish pipeline to prebuild the manager before publishing


lib/core/package.json
```diff
@@ -20,8 +20,8 @@
   "files": [
     "dist/**/*",
     "dll/**/*",
+    "prebuilt/**/*",
     "types/**/*",
-    "README.md",
```
lifeiscontent, line 24: Why the removal of the README.md? this is a pretty common practice for npm packages.

shilman (approved):
This is looking great. How do we publish this as part of the release? Is that a separate PR or an addition to this one?

lib/components/src/Loader/Loader.stories.tsx
```diff
@@ -1,33 +1,46 @@
...
+    {storyFn()}
+  </div>
+);
+
+export default {
+  title: 'Basics/Loader',
+};
+
+export const InfiniteState = () => <Loader role="progressbar" />;
+InfiniteState.storyName = 'infinite state';
```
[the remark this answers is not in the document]
ghengeveld: Do you mean the storyName or the story altogether?

ndelangen:
@ghengeveld need pairing?

ghengeveld:
> @ghengeveld need pairing?

Yes, tomorrow.

lib/components/src/Loader/Loader.stories.tsx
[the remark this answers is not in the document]
ndelangen: storyName

lib/core/package.json
```diff
@@ -121,6 +121,7 @@
     "shelljs": "^0.8.4",
     "stable": "^0.1.8",
     "style-loader": "^1.2.1",
+    "telejson": "^5.0.2",
```
ndelangen, line 124: do we still need this? are we still stringifying anything?

lib/core/src/server/build-static.js
```diff
@@ -164,7 +164,7 @@ async function buildPreview(configType, outputDir, packageJson, options) {
 
 function prepareFilesStructure(outputDir, defaultFavIcon) {
   // clear the output dir
-  logger.info('clean outputDir..');
+  logger.info('=> Cleaning outputDir..');
   shelljs.rm('-rf', outputDir);
```
ndelangen, line 168: We should use something safer here IMHO

lib/core/src/server/build-static.js
```diff
@@ -192,7 +192,12 @@ export async function buildStaticStandalone(options) {
   shelljs.cp('-r', dllPath, path.join(outputDir, 'sb_dll'));
 
   await buildManager(configType, outputDir, configDir, options);
```
ndelangen, line 194: does buildManager re-use prebuild / cache?

should we add this? if people don't add custom addons, this has the potential to make publishing storybook a lot faster!

ndelangen: maybe we can add a --prebuild-manager="{{path_to_prebuilt}}", so users can pull their custom prebuilt manager from some directory?

lib/core/package.json
ghengeveld: Yes we are

lib/core/src/server/build-static.js
ghengeveld: Done

ghengeveld:
🚀

PR #12723, opened 2020-10-10
Build: Improve angular e2e tests

gaetanmaisse:
## What I did
  
 - Use yarn instead of `npx` to bootstrap angular e2e tests 
 - Display yarn install/add command output when running the E2E tests

## How to test

- CI should be 🟢  except Angular tests that are currently failing
- We should be able to see the cause of the issue in Angular tests:
> error @aduh95/viz.js@3.1.0: The engine "node" is incompatible with this module. Expected version "^12.17.0 || >=13.2.0". Got "12.16.2"
error Found incompatible module.


shilman (approved):
LGTM -- good catch!

shilman (approved):
LGTM -- good catch!

PR #12731, opened 2020-10-10
Addon-a11y: Add blurred vision effect

yannbf:
Issue: #12730

## What I did
Added support for blurred vision effect as part of the color blindless emulation in addon-a11y. Also, because the filters do more than color blindness now, I renamed the file from ColorBlindness to VisionDeficiency and the tooltip title to "Vision Deficiency Emulation". Also the tooltip option from "Reset color filter" to "Reset filter" (this was done after I recorded the video below).

You can see it in action by checking this video: https://streamable.com/tty7bk
image

## How to test

Not applicable I believe
Nope
Would be nice to have a better section on the README explaining the existing options, unfortunately there's nothing there yet. I could assist on improving it.

## Attention!

If by any chance someone somewhere is using cypress to test a component in storybook with addon-a11y, which might be a really slight chance, and they happen to use this selector: `cy.get('[title="Color Blindness Emulation"]')` it will break because now the tooltip has a different title. I don't think this will happen but it's important to note in the PR :)

### DIFF
--- a/addons/a11y/src/components/VisionDeficiency.tsx
+++ b/addons/a11y/src/components/VisionDeficiency.tsx
@@ -7,6 +7,7 @@ import { Filters } from './ColorFilters';
 const iframeId = 'storybook-preview-iframe';
 
 const baseList = [
+  'blurred vision',
   'protanopia',
   'protanomaly',
   'deuteranopia',
@@ -23,6 +24,9 @@ type Filter = typeof baseList[number] | null;
 const getFilter = (filter: Filter) => {
   if (!filter) {
     return 'none';
+  }
+  if (filter === 'blurred vision') {
+    return 'blur(2px)';
   }
   if (filter === 'mono') {
     return 'grayscale(100%)';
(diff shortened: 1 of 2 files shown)


shilman (approved):
Great feature @yannbf! ❤️ 

What does it do in non-chrome browsers?

shilman:
Looks like it's breaking in IE:

https://www.chromatic.com/build?appId=5def629a9aec2800201cd626&number=5432

yannbf:
> Great feature @yannbf! ❤️
> 
> What does it do in non-chrome browsers?

I tested in Chrome, Firefox and Safari and the filter works great on all of them. I haven't tested on IE11 but I think `filter` is not supported. I can add a gaussianblur svg filter instead, and I believe it would work for IE as well. However, when I tried that, the effect was not quite interesting.


with css filter 2px intensity:
image

with svg filter standard deviation of 2:
image

Notice that with the svg filter the image is not only blurred, but there's actually a decrease in the image quality. Zoom in at the bottom to see it better

shilman:
Thanks for testing in those browsers. Let's not add any extra complexity for IE11 👍

Issue #12735, opened 2020-10-11
Angular E2E tests are broken due to Node requirement of Compodoc


gaetanmaisse:
**Describe the bug**

Angular E2E tests are broken for a few days causing CI to fails on `next`, see: https://app.circleci.com/pipelines/github/storybookjs/storybook?branch=next

**Investigation**

This error is caused by a dependency change in `@compodoc/ngd-transformer` between `2.0.0` and `2.1.0`:
> Move viz.js to @aduh95/viz.js

See https://github.com/compodoc/ngd/commit/60dbb2ac17e1d87f75b2378975607e578ae2a1fd. 
The node requirement of the new dependency (`@aduh95/viz.js`) is stricter than `viz.js` as it needs `^12.17.0 || >=13.2.0`.

As `@compodoc/compodoc` has a dependency on `"@compodoc/ngd-transformer": "^2.0.0"` (^2.1.0 for a few hours) it causes the following error when running `yarn install` with Node <12.17.0 : 

> error @aduh95/viz.js@3.1.0: The engine "node" is incompatible with this module. Expected version "^12.17.0 || >=13.2.0". Got "12.16.2"
error Found incompatible module.

I created an issue on  `@compodoc/ngd`: https://github.com/compodoc/ngd/issues/68. Let's see what happens.

At the same time, we can try to run Angular E2E tests on special executors with the latest Node 12 and keep Node 10 for all other tests.

_Note: Compodoc is used by `@storybook/docs` when running `@storybook/angular` to extract properties of components._

gaetanmaisse:
Closing as everything is back to 🟢   with the `3.2.0` release of `@aduh95/viz.js`

PR #12746, opened 2020-10-12
Build: fix E2E tests and rework related CircleCI jobs

gaetanmaisse:
## What I did

 - Run Angular E2E tests on executors with the latest version of Node 12, workaround for https://github.com/storybookjs/storybook/issues/12735
 - Manually install Cypress to avoid using a docker image shipping its version of Node: We want all E2E tests to run against Node 10 but `cypress/included:4.7.0` docker image is based on Node 12.16.0.
 - Use Webpack 4 in `webpack_react` and `sfcVue` E2E tests as Webpack 5 is officially out for a few days but it looks like Storybook is not 100% compatible with it yet.
 - Extract `cra_bench` in a dedicated job, CRA Bench requires Node 12 (see https://github.com/storybookjs/bench/pull/2) and isn't a "true" E2E test so it makes sense to move it in a dedicated job.

## How to test

 - CI should now contain an `example-v2-angular` step
 - `example-v2-angular` step should be 🟢 
 - CI should be 🟢

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,10 +172,6 @@ jobs:
     executor:
       class: large
       name: sb_node
-    docker:
-      - image: cypress/included:4.7.0
-        environment:
-          TERM: xterm
     working_directory: /tmp/storybook
     parallelism: 10
     steps:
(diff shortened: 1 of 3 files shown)


gaetanmaisse:
image
:youhou:

shilman (approved):
you da man @gaetanmaisse! ❤️ ❤️ ❤️

PR #12765, opened 2020-10-14
Search UX improvements

ghengeveld:
Issue: #12714

## What I did

1. Preserve the search query when selecting a search result, so that re-focusing the search will restore the previous list of results so one can easily select the next result.
2. Updated search to show 200 results before the "Show more" appears. This makes it unlikely to ever appear in a sensible query, but still guards against performance issues when searching for a single letter.
3. Hide the history panel (i.e. show the tree) when clearing history.

## How to test

- Search for something that yields multiple results.
- Select a result. This should open the story and close the search pane.
- Focus the search again (or press `/`). This should restore the previous search query and results list.
- Press down + enter to quickly select the next result.
- Focus the search and press Escape to clear the query and results. You can also click the (x) in the search field.

### DIFF
--- a/lib/ui/src/components/sidebar/Search.stories.tsx
+++ b/lib/ui/src/components/sidebar/Search.stories.tsx
@@ -37,9 +37,9 @@ export const FilledIn = () => (
 
 export const LastViewed = () => (
   <Search {...baseProps} lastViewed={lastViewed}>
-    {({ inputValue, results, getMenuProps, getItemProps, highlightedIndex }) => (
+    {({ query, results, getMenuProps, getItemProps, highlightedIndex }) => (
       <SearchResults
-        isSearching={!!inputValue}
+        query={query}
         results={results}
         getMenuProps={getMenuProps}
         getItemProps={getItemProps}
--- a/lib/ui/src/components/sidebar/Search.tsx
+++ b/lib/ui/src/components/sidebar/Search.tsx
@@ -222,34 +222,64 @@ export const Search: FunctionComponent<{
 
   const stateReducer = useCallback(
     (state: DownshiftState<DownshiftItem>, changes: StateChangeOptions<DownshiftItem>) => {
-      const { blurInput, clickItem, keyDownEnter, keyDownEscape } = Downshift.stateChangeTypes;
-      const { type, inputValue, selectedItem = {} } = changes;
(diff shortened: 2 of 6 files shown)


domyen:
Thanks Gert! 

👍 Nice work on highlighting the row of the selected story in the search results. This all looks good to me. I found one interaction bug below.

**Bug: Escape key interaction**
Pressing the escape key removes the query and blurs the input. But the recently opened UI persists until you focus and blur the input again. I suggest tweaking this behavior so that:

- Pressing the `esc` once when a value is present clears the value but continues to focus the input. That way recently opened list remains easily accessible.
- Pressing the `esc` again, blurs the input and show the sidebar again.

yannbf:
This goes for both of you @ghengeveld @domyen I was wondering if what I'm showing in the video makes sense or not. If it doesn't feel free to ignore!

Situation: given that clicking outside of the search closes the search and goes back to the normal sidebar state, it could be frustrating if the user clicks close but not directly in the results or the "clear history" button, and as a result they go back to the list of stories. Given the input is cleared when toggling its state, the user just lost what they were searching for thus have to type again.

Please check the video for more clarity on the situation: https://streamable.com/cyipn9

yannbf:
I also noticed that the input is cleared on focus. Here is an example:
Kapture 2020-10-18 at 23 16 36

And how it used to work in the previous version:
Kapture 2020-10-18 at 23 18 10

domyen:
Thanks for pointing this out @yannbf. I agree, we shouldn't clear the input in cases where you click out of it. These are the interactions where it happens:

**Click outside the search results**
1. Type `bu` to see search results
2. Click on the blank area below the search results
3. 🙅‍♂️ See that the search is cleared

**Refocus search**
1. Type `bu` to see search results
2. Click on the canvas to blur search
3. 🙅‍♂️ Refocus search and see that search is cleared

ghengeveld:
@domyen @yannbf I've made some changes to the input field behavior so it won't clear the query on focus, and continue to display the results until you either select one or explicitly clear the search. The escape key is now two-step. If there's a value it'll clear it but keep the recently viewed items open. If there's no value (the 2nd time) it'll return to the tree view.

ndelangen:
Congrats on this amazing contribution @ghengeveld

PR #12774, opened 2020-10-14
Chore: Update documentation for issue #12513

jonniebigodes:
With this pull request the issue #12513 is addressed.

What was done:
- Updated the `@storybook/addon-storyshots` to include information about the customization of the snapshot directory.
- Updated the Snapshot (workflows/snapshot-testing) page to address the issue with an example.
- Included a snippet for the configuration

@GasimGasimzada and @latviancoder if you both also could take a look at this as you are involved parties i would greatly appreciate any feedback you're willing to offer to fine tune it.

@shilman let me know as well if you're ok with it


shilman (commented):
LGTM!

PR #12790, opened 2020-10-16
Storyshots: Add missing vue peer dependencies

merceyz:
`@storybook/addon-storyshots` is trying to use `vue` and `@storybook/vue` without declaring them as dependencies

## What I did

Added them as optional peer dependencies

## How to test

Try to use `@storybook/addon-storyshots` with vue in a Yarn PnP project


merceyz:
This section should probably be updated as this is not the case anymore
https://github.com/storybookjs/storybook/blob/9da7db78d0241a466a01445131faa759753f679a/addons/storyshots/storyshots-core/README.md#L253-L255

PR #12800, opened 2020-10-16
Components: Minor update checking color in Button

Connormiha:
## What I did
Just minor refactoring of checking color in `<Button />`.

### DIFF
--- a/lib/components/src/Button/Button.tsx
+++ b/lib/components/src/Button/Button.tsx
@@ -83,14 +83,12 @@ const ButtonWrapper = styled.button<ButtonWrapperProps>(
   ({ theme, primary, secondary, gray }) => {
     let color;
 
-    if (primary) {
-      color = theme.color.primary;
-    }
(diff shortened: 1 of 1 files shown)


lib/components/src/Button/Button.tsx
```diff
@@ -83,14 +83,12 @@ const ButtonWrapper = styled.button<ButtonWrapperProps>(
...
-    }
-    if (secondary) {
-      color = theme.color.secondary;
-    }
     if (gray) {
       color = theme.color.medium;
+    } else if (secondary) {
+      color = theme.color.secondary;
+    } else if (primary) {
+      color = theme.color.primary;
```
ndelangen, line 91: is there a reason why secondary is above primary?

Connormiha: Make the assignment more explicit. Without reassign on next line.

shilman (approved):
LGTM

Issue #12797, opened 2020-10-16
Lots of broken links on addon-docs npm page


kaiyoma:
Not sure if this affects all the npm pages for all the plugins, but every github.com link on this page is broken:

https://www.npmjs.com/package/@storybook/addon-docs

jonniebigodes:
@kaiyoma I would like to thank you for bringing this to our attention and i've just made a pull request that addresses the issue you've encountered. Hopefully once it's merged the links should be working as intended. 


Stay safe

darleendenno:
Closing as this issue is resolved.

Issue #12317, opened 2020-08-29
"MDX Reference" link in storybook DocPage README is broken


bbugh:
**Describe the bug**
The title is descriptive of the issue. 

**To Reproduce**

https://www.npmjs.com/package/@storybook/addon-docs#mdx

Click the MDX reference link in the paragraph:

> For more information on MDX, see the MDX reference

Results in a GitHub 404. 

**Expected behavior**
Links to somewhere valid with docs!

jonniebigodes:
@bbugh and sorry for the delay in answering. I've been sidetracked with other aspects of the documentation and your issue was left behind a bit. Nonetheless i would like to thank you for bringing this to our attention and i've just made a pull request that addresses the issue you've encountered. Hopefully once it's merged the links should be working as intended.

Stay safe

jonniebigodes:
Closing this as a pull request was made to address the issue. Feel free to let me know if the issue persists and I'll gladly re-open the issue.

Stay safe

PR #12801, opened 2020-10-16
Chore: Absolute links for addon-docs README

jonniebigodes:
With this pull request the issue #12317 and #12797 are addressed and can be closed.

What was done:
- Update the urls from relative to more absolute urls to prevent the urls from resolving to 404 in npm. 

Despite configuration it seems that urls are being resolved to https://github.com/storybookjs/storybook/blob/HEAD/docs/filename.md when they should be resolved as https://github.com/storybookjs/storybook/blob/HEAD/addons/docs/docs/filename.md

It seems that despite configuration of the package  npm will not include the `addons/docs` in the path. 

I've considered the `directories.docs` as an alternative to solve the issue, but discarded it as a possibility as it would add unnecessary weight to the package and also most users will not read the documentation from their `node_modules` folder.

Feel free to provide feedback

### DIFF
--- a/addons/docs/README.md
+++ b/addons/docs/README.md
@@ -37,7 +37,7 @@ Click on the `Docs` tab to see it:
   <img src="https://raw.githubusercontent.com/storybookjs/storybook/master/addons/docs/docs/media/docs-tab.png" width="100%" />
 </center>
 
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM!

PR #12803, opened 2020-10-16
Fix stray smart quotes on React snippet

msakrejda:
Issue: React code example uses smart quotes instead of apostrophes

This changes the example to use apostrophes so it can be copied and used directly.


shilman (commented):
LGTM. Thanks!

PR #12810, opened 2020-10-18
Core: Pass framework options as global

tooppaaa:
Issue: #12734 global vars not set

## What I did

use globals over define plugin


shilman (approved):
LGTM!

PR #12812, opened 2020-10-18
Addon-docs: Dynamic source rendering for Vue

pocka:
Issue: #11400 

## What I did

Added a decorator takes Vue story and emits source code for that.

NOTE: I didn't test directives (e.g. `v-model`, `v-on`, custom directives)

## How to test

- Is this testable with Jest or Chromatic screenshots? ... yes
- Does this need a new example in the kitchen sink apps? ... no
- Does this need an update to the documentation? ... no?

If your answer is yes to any of these, please make sure to include it in your PR. ... done (`addons/docs/src/frameworks/vue/sourceDecorator.test.ts`)

To see, run `cd examples/vue-kitchen-sink && yarn storybook`


addons/docs/src/frameworks/vue/sourceDecorator.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    channel.emit(
+      SNIPPET_RENDERED,
+      (context || {}).id,
+      prettier.format(`<template>${code}</template>`, {
+        parser: 'vue',
+        plugins: [prettierHtml],
+        // Because the parsed vnode missing spaces right before/after the surround tag,
+        // we always get weird wrapped code without this option.
+        htmlWhitespaceSensitivity: 'ignore',
```
pocka, line 66: I'm using Prettier here just because "I know how to format code with Prettier". If another lib/approach is preferable, please tell me.

addons/docs/src/frameworks/vue/sourceDecorator.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+  return matched;
+}
+
+interface VueInternal {
+  // We need to access this private property, in order to grab the vnode of the
+  // component instead of the "vnode of the parent of the component".
+  // Probably it's safe to rely on this because vm.$vnode is a reference for this.
+  // https://github.com/vuejs/vue/issues/6070#issuecomment-314389883
+  _vnode: Vue.VNode;
```
pocka, line 174: I think it's safe to use this "private" property. Without this, we can't access the VNode for stories having no children such like this:

```js
// Instance for the story has zero-length `$children` and `$vnode` describes itself, not content
export const myStory = () => ({
  template: `
    <button>Button</button>
  `
})
```

addons/docs/src/frameworks/vue/sourceDecorator.ts
```diff
@@ -0,0 +1,198 @@
...
+  if (sourceParams?.type === SourceType.DYNAMIC) {
+    return false;
+  }
+
+  // never render if the user is forcing the block to render code, or
+  // if the user provides code, or if it's not an args story.
+  return !isArgsStory || sourceParams?.code || sourceParams?.type === SourceType.CODE;
+};
+
+export const sourceDecorator = (storyFn: any, context: StoryContext) => {
```
pocka, line 25: Named this `sourceDecorator` instead of `jsxDecorator` because it's not a JSX... if it's a requirement I'll change this.

addons/docs/src/frameworks/vue/sourceDecorator.ts
```diff
@@ -0,0 +1,198 @@
...
+
+import { addons, StoryContext } from '@storybook/addons';
+import { logger } from '@storybook/client-logger';
+import prettier from 'prettier/standalone';
+import prettierHtml from 'prettier/parser-html';
+import Vue from 'vue';
+
+import { SourceType, SNIPPET_RENDERED } from '../../shared';
+
+export const skipJsxRender = (context: StoryContext) => {
```
shilman, line 11: skipSourceRender?

addons/docs/src/frameworks/vue/sourceDecorator.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        getVNode({
+          template: `
+          <div>
+            <form>
+              <button>Button</button>
+            </form>
+          </div>`,
+        })
+      )
+    ).toMatchInlineSnapshot(`<div ><form ><button >Button</button></form></div>`);
```
shilman, line 75: Anyway to get rid of the extra spaces? E.g. `<div >` => `<div>`? Or is that standard in Vue?

shilman (changes requested):
This is looking awesome--great work @pocka ! A few minor questions and/or suggestions below.

Mouse_Highlight_Overlay

Any idea why it's `:rounded` and not `:color`?` Seems inconsistent but I don't know Vue well enough to say.. 🙈

addons/docs/src/frameworks/vue/sourceDecorator.test.ts
pocka: Yeah, it's ugly but will be erased before emitting by Prettier. Should we remove it before formatting?

pocka:
@shilman 
> Any idea why it's :rounded and not :color?` Seems inconsistent but I don't know Vue well enough to say.. 🙈

Yes, you can use the v-bind shorthand syntax (`:`) for string props as well. But, for me, it feels not straight-forward and is hard to read... I guess opinions from Vue users would help since I don't write Vue app often :neckbeard: 

```vue
<template>
  <my-component
    :bool-prop="true"
    :number-prop="1"
    :string-prop="'extra quotes needed since this is a JS expression'"
  />
</template>
```

addons/docs/src/frameworks/vue/sourceDecorator.test.ts
shilman: Aha didn't realize this was the pre-prettier output. If prettier fixes, that's fine by me.

addons/docs/src/frameworks/vue/sourceDecorator.test.ts
```diff
@@ -0,0 +1,77 @@
...
+                  foo: 'bar',
+                },
+              },
+            };
+          },
+          template: `<my-component v-bind="props"/>`,
+        })
+      )
+    ).toMatchInlineSnapshot(
+      `<my-component :propD='{"foo":"bar"}' :propC="null" :propB="1" propA="propA"/>`
```
shilman, line 59: @pocka I still don't understand why some of these props get a colon in front and others don't. It seems inconsistent to me.

pocka: The React equivalent is:

```jsx
<MyComponent propA="propA" propB={1} propC={null} propD={{"foo": "bar"}} />
```

. And,

```jsx
<MyComponent propA={"propA"} propB={1} propC={null} propD={{"foo": "bar"}} />
```

Vue equivalent of the above (`propA` changed from string literal to expression) is:

```vue
<my-component :propA='"propA"' :propB="1" :propC="null" :propD='{"foo":"bar"}' />
```

https://vuejs.org/v2/guide/syntax.html#Attributes (a single colon is a shorthand for `v-bind:`)

shilman (approved):
LGTM! ❤️

dallinja:
@pocka 
> I guess opinions from Vue users would help since I don't write Vue app often

I would prefer to see it like this...
```
<MyComponent
  :number-prop="1"
  string-prop="no need to bind strings"
  bool-prop
/>
```
My goal would be to just click copy and then paste it into my code, without any editing. This means:

1. Remove `<template>` tag. I don't want to have to delete the `<template>` tag every time. It should be extremely obvious to any Vue developer that the code goes inside a `<template>` tag. And never would the component be the _only_ thing in a `<template>` tag, so leaving them in the source here would required editing 100% of the time. It would be the same as doing `return (<MyComponent />)` in React. No need to show the `return`.
2. Remove v-bind for strings. You're right, binding them is weird.
3. No need to v-bind booleans. Just extra code that I'd want to delete. If it's `false`, don't include it.

Lastly, the Vue style guide recommends using PascalCase for Components.

I was gonna jump on this request, but really excited someone already did! Awesome work @pocka!

pocka:
The `<template>` tag is just for Prettier to work, I think we can add a strip-templates feature (an option like `noTemplateTag` or do it by default idk).
That `v-bind` shortcut for boolean is for demonstrating purpose, not show up in an actual result 👍  

> Lastly, the Vue style guide recommends using PascalCase for Components.

Probably the casing transform is easy to implement (https://github.com/storybookjs/storybook/pull/12812/files#diff-83d08076a6b73dab376e42608b87b8720952063d14015e3fbf5d38ac3a23fa94R76) and a good-to-have-feature. Anyone interests?

shilman:
Thanks @dallinja and @pocka for pushing this forward 🙏

zhangjiangqiu:
> This is looking awesome--great work @pocka ! A few minor questions and/or suggestions below.
> 
> Mouse_Highlight_Overlay
> 
> Any idea why it's `:rounded` and not `:color`?` Seems inconsistent but I don't know Vue well enough to say..

I want to know how can I write a story to make this happen！

addons/docs/src/frameworks/vue/sourceDecorator.test.ts
zhangjiangqiu: I wonder how I can write a story to achieve this effect!

MariaNicolosi:
This looks so much better!
However, the source code isn't displaying named slots.

Screen Shot 2020-12-16 at 9 47 00 AM

Screen Shot 2020-12-16 at 9 44 42 AM

litan1106:
I am using vue 3 with storybook 6.2.0-alpha.23 but it didnt render the dynamic code.

image

PR #12824, opened 2020-10-20
Addon-controls: Add JSON tree editor for Object/Array Type args

hydrosquall:
Issue: #12599 , reimplements https://github.com/storybookjs/storybook/pull/12685

## What I did

- Replaced the existing text form control with the `react-editable-json` lib (public demo here: https://oxyno-zeta.github.io/react-editable-json-tree/ ). I also added typings for the library to the closest typings file so we can safely use it in typescript.
- Made a first pass at customizing the styles (spacing and colors) based on @domyen's feedback [here](https://github.com/storybookjs/storybook/pull/12685#issuecomment-705017673) and looking at the colors available in Storybook [here](https://next--storybookjs.netlify.app/official-storybook/?path=/story/basics-colorpalette--page) . The "dark/light mode toggles" aren't toggling the styles in the control pane (just inside the main component body area), so I'm not exactly how this will look in Dark Mode yet, but I set up the style customization function such that it should be easy to replace things
- Updated the `official-storybook` story about using `addon-controls` to make use of the fact that we now handle nested objects.

- Demo GIF: https://a.cl.ly/geuqpZrq

## How to test

It's possible, but I'm not sure that it's necessary.

We could, but I think updating the main storybook story may be sufficient, as this change should benefit most users without needing them to customize anything. 

We could update [this page](https://storybook.js.org/docs/react/essentials/controls#annotation) to indicate that tree view is used instead of a text view.

## FAQ

- Why not some other JSON tree viewer library XYZ?
  - We explored several different alternatives, tradeoffs are documented [here](https://github.com/storybookjs/storybook/pull/12685#issuecomment-708699165), [here](https://github.com/storybookjs/storybook/pull/12685#issuecomment-706739298), and [here](https://github.com/storybookjs/storybook/issues/12599#issuecomment-701097502) 

- What terminal commands did you use that you might use again next time?

```bash
# to add a dependency scoped to 1 package 
yarn lerna add react-editable-json-tree --scope=@storybook/components

# to test locally
yarn build components --watch
yarn start
```

- Any modifications useful in dev that you didn't commit?

 1 change in `examples/official-storybook/main.ts` to speed up build time:

```js
{
stories: [
    './stories/**/addon-controls.stories.@(js|ts|tsx|mdx)',
  ]
 // ... rest
}
```


hydrosquall:
I'm not sure how to debug the CI failures because I don't have access to any of the TeamCity URLs in the Github status checks. I can look into any issues, but I wasn't able to pick up any glaring issues in my diff, since this runs without error when I build locally. Is there anything else I should do before this can be reviewed?

cc @shilman

shilman:
Don't worry about it. Sorry i missed this PR first time around. Looking amazing! I'll test and review later today 🙏

shilman:
@hydrosquall Thanks for your patience on this. We ended up shipping 6.1 a couple months ahead of schedule to deal with the React 17 compatibility issue, so everything else got bumped to 6.2. I finally had a chance to take a look at this PR and it's looking pretty great. 

So far there isn't any significant performance penalty that comes with this change. In the following graph, the pink line is `next` and the purple dot is this PR:

It does have a dependency on `react-hot-keys` which is fairly large. That doesn't show up on this graph because Storybook already has a dependency on that library, tho I'd like to remove it and this makes that harder. We're converting over to ESM #13013 and hopefully that will help prune down react-hot-keys.

I do have a couple other minor concerns:
1) the interaction is not smooth. for example, if you add a new item and hit "return" it would be nice if the behavior was to open a new item by default.
2) the form for adding new items is not styled for storybook.

I'm sure @domyen will have small requests here.

Finally, the build is currently brokenon some `babel-runtime`-related error:

```
ERR! ModuleNotFoundError: Module not found: Error: react-editable-json-tree tried to access babel-runtime, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
ERR! 
ERR! Required package: babel-runtime (via "babel-runtime/helpers/possibleConstructorReturn")
ERR! Required by: react-editable-json-tree@virtual:26021639ee2fec27411f43115adbcd4c7cc8f21d5ccd21278d3f995acaf42cc28f0295e4ccd9ef6147f32767fc3fc9993e7676cccb478c71d904953b19cbce69#npm:2.2.1 (via /tmp/storybook-e2e-testing/cra-latest/.yarn/$$virtual/react-editable-json-tree-virtual-f589331f00/0/cache/react-editable-json-tree-npm-2.2.1-a0c4107a5e-945a18bc73.zip/node_modules/react-editable-json-tree/dist/JsonTree.js)
```

This could be related to #13055 ... @43081j can you take a quick look?

43081j:
I had a look but I'm struggling to get X server sorted in wsl2 so can't run the test to debug. I'll try on my MacBook once I'm on it. 

Seems strange that the dependency is requiring something it hasn't declared.

hydrosquall:
Hi @shilman , thanks for taking a close look at this and taking care of the merges to the latest `next` branch! Neat to see these performance regression check graphs, I didn't know the project had them. Is there a way that I can access the live dashboard that produces these?

> react-hotkeys dependency is large

Good to know that there's an ongoing project to remove this - for other parts of storybook, what library (if any) are people replacing it with? An option to reduce the impact could be to open a PR to replace its [3 usages](https://github.com/oxyno-zeta/react-editable-json-tree/search?q=react-hotkeys) if there's a smaller alternative with the same api. 

> the interaction is not smooth. for example, if you add a new item and hit "return" it would be nice if the behavior was to open a new item by default.

I can see the value of this if someone is adding several new objects to an array, but not necessarily for adding keys to an object. But I also feel like that "auto open a new object" could be unexpected if someone was only expecting enter to save. I think this library already has a slight improvement over the UX of the [react cosmos](https://cosmos.flatris.space/?fixtureId=%7B%22path%22%3A%22components%2Fshared%2FGamePanel%2FGamePanel.fixture.js%22%2C%22name%22%3Anull%7D) opens up a newline rather than saving.

 On a practical level, I looked through the library API, and don't currently see a way to trigger the opening of a new item from above the JsonTree component, so if this behavior ends up being a key part of the interaction, we'd either need someone to modify/fork the underlying library to let parent components trigger "add" actions at arbitrary levels of the tree, or implement a custom JSON tree component (since I don't remember any of the options I surveyed [previously](https://github.com/storybookjs/storybook/issues/12599#issuecomment-701097502) having this capability by default.

Are there any other aspects of smoothness you'd like to see improved in addition to the "save and create new entry on pressing enter" idea?

> the form for adding new items is not styled for storybook.

I wasn't sure where to go for all the storybook specific padding/font/etc requirements, so I started with trying to use the storybook theme colors in the custom styling function `getCustomStyleFunction` in `Object.tsx`. If there's a reference form input you'd suggest looking at I can try to match with that, in addition to reviewing any additional specific suggestions from @domyen .

> babel-runtime-error

@43081j thanks for looking into this! Let me know if there's any additional debugging input from me that would help.

hydrosquall:
@shilman Hi, just wanted to check if the UX request regarding the behavior of the `enter` button is still a blocking concern - we'd love to be able to use this improved object tree component in our codebase, and it wasn't clear to me what is needed from my side to continue moving this PR forward.

lib/components/src/controls/Object.tsx
```diff
@@ -1,72 +1,133 @@
...
-  return trimmed ? JSON.parse(trimmed) : {};
-};
-
-const validate = (value: any, argType: ArgType) => {
-  if (argType && argType.type.name === 'array') {
-    return Array.isArray(value);
-  }
-  return true;
-};
+import { JsonTree, JsonTreeProps } from 'react-editable-json-tree';
```
ndelangen, line 4: Perhaps we should lazy-load these components?

shilman:
@hydrosquall We're going to get this into 6.2. @ghengeveld is working on #13803 and after that we plan to pick up this PR and figure out the UX polish needed to get it over the line, hopefully this week or next. Thanks for your patience!

ghengeveld:
Hey @hydrosquall I'm going to work on getting this ready and merged soon. I'm making a few tweaks and get @domyen to review it, I hope you don't mind. Great work so far!

ghengeveld:
I did a ton of tweaking and hacking on this thing. Here's the result:

https://user-images.githubusercontent.com/321738/108496507-6fc32980-72aa-11eb-9fef-5b1bdfc84421.mov

I had to use some very ugly code to get things to work just right, given the very limited customization options provided by `react-editable-json-tree`. For now I think this is good enough.

@hydrosquall WDYT?

domyen:
This is looking amazing @ghengeveld! Super impressed with how much customization you were able to do within the constraints of their library.

## UX feedback
When I add an item the form appears _outside_ the place that I'm trying to add it.
image

Can we move the add button inside the brackets somewhere? In the current location, the add button seems to imply that it will add another attribute outside the object.
image

Remove offset. The visual vertical alignment with the rest of the controls is broken up by the horizontal borders, so the offset for the down arrow isn't needed.
image

Change color of the input from violet to `darkest`
image

Use `gray` button color for the Cancel button. 
image

hydrosquall:
> Hey @hydrosquall I'm going to work on getting this ready and merged soon. I'm making a few tweaks and get @domyen to review it, I hope you don't mind. Great work so far!

Sounds great, thanks so much for picking this up! I have limited bandwidth in the next few weeks for dev on open source, but I'm very much looking forward to being able to use this officially :)

shilman:
@ghengeveld unfortunately `react-editable-json-tree` is breaking our yarn pnp tests because it requires `babel-loader` but does not actually depend on it.

Can you please:
- Fork the package into the storybook org & publish as `@storybook/react-editable-json-tree`
- Add a babel-loader dependency (or peer dep?)
- Get this PR working on CI with the fork
- Contribute a PR back to `react-editable-json-tree`
- Update the dependency once your PR is merged/published

ghengeveld:
> When I add an item the form appears _outside_ the place that I'm trying to add it.

Believe me, I tried. I hate it just as much as you do. But all the CSS wizardry in the world isn't going to fix a crappy DOM structure. It could be solved with a JS hack, but I don't really want to go there. The same limitation applies to the add icon (it's literally the same `span` element).

ghengeveld:
Alright so I had to pull in the code from `react-editable-json-tree` into the Controls addon directly, because doing so was infinitely easier than trying to get that library up to date with everything that's changed since it was last released 3 years ago. 

That also means that we now have full control over how it's rendered, so I was able to lay it out as @domyen requested.

https://user-images.githubusercontent.com/321738/108739940-6b5e7100-7535-11eb-86af-e5abc22cf9e3.mov

shilman (approved):
Looking great @ghengeveld @hydrosquall 💯

shilman:
Hmm, the UI is kind of strange for undefined values. @ghengeveld WDYT?

Reviewing_Build_21108_•_storybookjs_storybook

PR #12839, opened 2020-10-20
Core: Migrate core to TypeScript

gaetanmaisse:
Issue: Part of #5030 and supersede #9870

## What I did

I migrated `@storybook/core` to TypeScript, some types still need to be refined but the base is here.
A PR had been started in early 2020 about that (https://github.com/storybookjs/storybook/pull/9870) but never gets merged and has too many conflicts now so I just started from scratch here.

I think this is the final step of the TS migration for `lib` packages 🎉  

## How to test

- CI should be 🟢


gaetanmaisse:
I will push new commits improving typing in the few next days 😉

yannbf:
> The types still need some refinement and refactoring but as discussed during the [monthly meeting](https://github.com/storybookjs/community/blob/master/meetings/2020/monthly-2020-10-02.md#ts-migration-of-storybookcore) we will merge as it to avoid many conflicts resolution.
> 
> I added a big warning to avoid exporting these WIP types and I will work to improve them in the next few days/weeks.
> 
> @yannbf was the typing bug you had in mind yesterday the one I fixed in the last commit?

Yes, exactly! I was also wondering if options could be `Record<string, any>` instead of any.. Or do you think options could receive a primitive value rather than an object?

gaetanmaisse:
>I was also wondering if options could be Record<string, any> instead of any.. Or do you think options could receive a primitive value rather than an object?

I would say no but I'm really not sure so I didn't take the risk 🙈  Let's see what the experts are saying 😄

lib/core/src/server/types.ts
```diff
@@ -0,0 +1,120 @@
...
+  outputDir?: string;
+  cache: boolean;
+  previewUrl?: string;
+  versionCheck: VersionCheck;
+  releaseNotesData: ReleaseNotesData;
+  presets: any;
+}
+
+export interface Presets {
+  apply(
```
tooppaaa, line 25: I see the point but isn't too limitative (and potentially incomplete) to list all presets ?

lib/core/src/server/types.ts
```diff
@@ -0,0 +1,120 @@
+import { Configuration, Stats } from 'webpack';
+import { TransformOptions } from '@babel/core';
+import { typeScriptDefaults } from './config/defaults';
+
+/**
+ * ⚠️ This file contains internal WIP types they MUST NOT be exported outside this package for now!
```
gaetanmaisse, line 6: This is the most important line of code in this file 😁

lib/core/src/server/types.ts
gaetanmaisse: I somehow need these types to avoid type inference to "return" `any` on all these calls: https://github.com/storybookjs/storybook/pull/12839/files#diff-d0f0c842d74746ce0b792a604e01fb7ded59c01898321fb1b482ba3ca016ab7dR12-R28

It's not directly the presets but the functions a preset can expose and so for instance `webpack` function should/must always return a webpack `Configuration` whatever the preset. Do you agree? 🤔 

But it's not 100% clear for me yet, maybe I will get rid of that when improving the typing.

I'm surely missing some parameters that why I introduced a fallback on [L45](https://github.com/storybookjs/storybook/pull/12839/files#diff-672824f9073da83026877803f53d07dc680263367ef9da36b14a016c72bcf2bbR45)

shilman (approved):
Great work @gaetanmaisse. 💯 

Probably needs refinement, but will merge for now and we can pick that up between now and release, or down the road.

PR #12863, opened 2020-10-22
Composition: Filter out disabled refs in getAutoRefs

g1eny0ung:
Issue:

If some packages don‘t export `package.json`, start storybook will warn:

image

## What I did

I think to solve this problem we can define these packages in the `refs` and set it as disabled in `.storybook/main.js`, like:

```js
refs: {
  rollup: { disable: true },
  '@rollup/plugin-node-resolve': { disable: true },
}
```

Then filter out the disabled refs in `getAutoRefs` func. Also, the changed code will reduce the extra disabled refs check than the previous code.

I'm not sure if this is correct or reasonable, please help me point out if there is an error, thx. 😃

## How to test


g1eny0ung:
Hi @shilman, is there any progress on this PR?

shilman (approved):
LGTM!

ndelangen:
it's possible for an npm package to not contain a package.json?!?!

g1eny0ung:
> it's possible for an npm package to not contain a package.json?!?!

@ndelangen They actually have the package.json. The reason I guess is both `rollup` and `@rollup/plugin-node-resolve` defined the `exports` field to limit the scope of resolve.

ndelangen (approved):

g1eny0ung:
@shilman @ndelangen I noticed the latest changes use typescript and I have already resolved the conflicts. Please check again.

PR #12870, opened 2020-10-22
Addon-viewport: Fill entire iframe width with drop shadow

yannbf:
Issue: When using addon-viewports in high resolution monitors, you can notice a weird looking background, like this:

(2048 x 1152)
image

(3840 x 2160)
image

This can be seen in any [live storybook](https://next--storybookjs.netlify.app/official-storybook/?path=/story/addons-a11y-basebutton--default).

## What I did
Set the drop shadow to take full width instead of 1000px, so it will always fill everything:
image

## How to test

Run Storybook, enable viewports and either emulate a higher resolution monitor or check it in a high resolution monitor like 4k.


shilman (commented):
LGTM!

PR #12891, opened 2020-10-23
Add missing peerDependencies.

brandonbloom:
Fixes module resolution under Yarn 2 (berry).

Issue:

## What I did

Added peerDependencies.

## How to test

Create an example project that uses storyshots and do:

```
yarn set version berry
yarn
```

See if your app builds/runs/works.

### DIFF
--- a/addons/storyshots/storyshots-core/package.json
+++ b/addons/storyshots/storyshots-core/package.json
@@ -66,7 +66,9 @@
     "@storybook/vue": "*",
     "jest-preset-angular": "*",
     "jest-vue-preprocessor": "*",
-    "vue": "*"
+    "vue": "*",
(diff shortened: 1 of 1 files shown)


merceyz (commented):
Both of these are already normal dependencies
https://github.com/storybookjs/storybook/blob/902b366dfbfe8e4a46d1e772ddac635630bec5e9/addons/storyshots/storyshots-core/package.json#L48-L50

brandonbloom:
My bad, this was fixed in 5e5b0a0647cd54d895874f4419a246aadb5b44f8. I was fixing peerDependencies in a lot of libraries and got overzealous with this one. Apologies for the noise.

Issue #12819, opened 2020-10-19
Shortcuts menu is not keyboard accessible


micahgodbolt:
**Describe the bug**
When opening the shortcuts menu via mouse or keyboard, the items inside of the shortcuts menu are not accessible via keyboard.

**To Reproduce**

1. Go to https://next--storybookjs.netlify.app/official-storybook/
2. Click on "Shortcuts"
3. Press down arrow, or attempt to tab to "About your Storybook" link
4. Down arrow only scrolls down the page, and tabbing moves focus to search. There is no way to get to that (or any of the shortcut links) without using mouse

**Expected behavior**
As a dropdown, I would expect arrow keys to move up and down the menu

tylerbutler:
Is the work on this still in progress? Anything we can do to raise priority or help?

yannbf:
> Is the work on this still in progress? Anything we can do to raise priority or help?

There's a PR with a solution, hopefully will be done soon!

https://github.com/storybookjs/storybook/pull/12904#issuecomment-845886817

vanessayuenn:
Closing as duplicate of #21578

PR #12904, opened 2020-10-24
UI: Add keyboard navigation on menu tooltip

tooppaaa:
Issue: #12819

## What I did
- Add right roles on menu 
- Add keyboard accessibility for Tooltip Menu

## How to test

- use basics-tooltip-tooltiplinklist--links story
- Chromatic
- NO UI CHANGE


lib/components/src/tooltip/TooltipLinkList.tsx
```diff
@@ -50,16 +53,88 @@ const Item: FunctionComponent<TooltipLinkListProps['links'][number]> = (props) =
...
+      event.preventDefault();
+      moveFocus(list, currentFocus, previousItem);
+    } else if (key === 'Home') {
+      event.preventDefault();
+      moveFocus(list, null, nextItem);
+    } else if (key === 'End') {
+      event.preventDefault();
+      moveFocus(list, null, previousItem);
+    }
+  };
```
ndelangen, line 123: Is this going to conflict with @ghengeveld work on the new sidebar?

tooppaaa: Nop, that's the all point of preventDefaut.
I'll do more testing though

micahgodbolt: escape to close the menu and return focus to button would be nice

lib/components/src/tooltip/WithTooltip.tsx
```diff
@@ -48,15 +48,28 @@ const WithTooltipPure: FunctionComponent<WithTooltipPureProps> = ({
...
+  const handleKeyDown: React.KeyboardEventHandler<HTMLDivElement> = (event) => {
     if (event.key === 'Tab' || event.key === 'Escape') {
       event.preventDefault();
       onVisibilityChange(false);
     }
   };
 
+  const prevOpen = React.useRef(tooltipShown);
+  React.useEffect(() => {
+    console.log('tooltipOpen changed', tooltipShown);
```
shilman, line 61: console.log?

tooppaaa: Oopsie

lib/components/src/tooltip/TooltipLinkList.tsx
```diff
@@ -50,16 +53,88 @@ const Item: FunctionComponent<TooltipLinkListProps['links'][number]> = (props) =
...
-    ))}
-  </List>
-);
+}) => {
+  const listRef = React.useRef<HTMLUListElement>(null);
+
+  const handleKeyDown: React.KeyboardEventHandler<HTMLUListElement> = (event) => {
+    const list = listRef.current;
+    const { key } = event;
+    const currentFocus = ((list && list.ownerDocument) || document).activeElement;
```
ghengeveld, line 107: You'll have to test that there's no modifier keys active. Otherwise they'll conflict with other key listeners. Example:

https://github.com/storybookjs/storybook/blob/adae4e5011d756927d60ff8139cb01e7b07cb89e/lib/ui/src/components/sidebar/useHighlighted.ts#L83

ghengeveld (changes requested):
Good stuff. Needs QA to verify it doesn't conflict with the new sidebar, but I think it'll be fine besides the comment I made.

lib/components/src/tooltip/TooltipLinkList.tsx
tooppaaa: I don't think this is right / needed, I've tested with both and didn't changed anything for me.
Keyboard navigation is working in menus.
Keyboard navigation is working between stories with / without a menu open

micahgodbolt:
One more item to note is that the button itself needs to inform the user that it invokes a menu. In FluentUI we currently do that via `aria-expanded="false" aria-haspopup="true"`

This should probably be added in this PR.

tooppaaa:
TODO: Missing selection from return key or space key (sorry for the notif, I'm slow on this PR, I'll fix later)

tooppaaa:
@shilman @ghengeveld Actually I'm not sure what is the right way to handle return key. This can be done as a second step of this accessibility work.
Good to review / test / merge ? :)

lib/components/src/tooltip/TooltipLinkList.tsx
```diff
@@ -50,16 +53,89 @@ const Item: FunctionComponent<TooltipLinkListProps['links'][number]> = (props) =
...
+    if (event.shiftKey || event.metaKey || event.ctrlKey || event.altKey) return;
+
+    if (key === 'ArrowDown') {
+      // Prevent scroll of the page
+      event.preventDefault();
```
shilman, line 112: Can you move this into `moveFocus` and simplify the code a bit?

ghengeveld:
I did a bit of cleanup and noticed these Deepscan warnings in `TooltipLinkList.tsx`:

Screenshot 2021-05-21 at 13 28 04

Screenshot 2021-05-21 at 13 28 22

@tooppaaa I know it's been a long time, but can you explain what you were trying to do here?

tylerbutler:
@tooppaaa @ghengeveld Is this PR still being worked on? Is there anything we could do to help? We're keen to roll this fix out on fluidframework.com. :)

darleendenno:
@tooppaaa @tylerbutler do either of you have availability to help get this over the finish line? This would be a remarkable improvement ❤️ ❤️ ❤️

darleendenno:
Hi there 👋 I'm extremely excited at the prospect of making this component accessible and navigable via keyboard! 

However, as I've been taking a closer look, I have a few concerns:
1. Do we want to be responsible for maintaining this functionality? There are a number of libraries that have accessibility baked in (like [reachui](https://reach.tech/)), off of which we could build.
2. Hard-coding this type functionality strips the web of doing what it does natively; semantic html is the foundation of accessibility. 
3. I do not think we should dictate how users can interact with the UI. For example, once I've hit ENTER to be in the list context, I should be able to use the arrows _or_ TAB, SHIFT+TAB to navigate through the list. If we hard-code these options, we lose opportunities.

What do others think? I eagerly want to see this cross the finish line, but I want to ensure we are doing this the right way.

MichaelArestad:
> Hard-coding this type functionality strips the web of doing what it does natively; semantic html is the foundation of accessibility.

💯 Let's let native elements and browser vendors do the hard work for us.

> In my opinion, we could update TooltipMenu to being a dropdown menu. I don't see the value in a "tooltip" menu.

Right. We could use semantic elements to compose the tooltip, which will simplify making it accessible.

> I do not think we should dictate how users can interact with the UI.

+1

ndelangen:
@tooppaaa Do you feel like continuing working on this? I've heard @MichaelArestad talk about a storybook UI refresh, so perhaps it doesn't make much sense?

ndelangen:
@MichaelArestad Could you look at the chromatic diffs?

MichaelArestad:
@ndelangen I don't think we should merge this as is. I would like to see a solution that isn't hard-coded like this one is.

ndelangen:
@tooppaaa I'm going to close this. We can't merge it as-is, we have a list of merge conflicts which make it impracticle for anyone to work on.

You're of course absolutely welcome to re-open this if you think you can get it in a merge-able state 🙏

danielmarcano:
Hi there @MichaelArestad @darleendenno, I would love to work on this enhancement!

I have checked [ReachUI](https://reach.tech/), and it looks like a great solution.

I have my doubts as to what you mean by "hard-coded" in this case. Are you referring to the `handleKeyDown` function?

Something like ReachUI would indeed apply similar code, the only advantage (and not small) being that Storybook would not need to take care of the maintenance of this component, as mentioned by @darleendenno.

I have noticed that this issue occurs with all "dropdowns" from Storybook's UI, as mentioned in [this open issue](https://github.com/storybookjs/storybook/issues/21578). There's someone who has taken the issue, but I see no progress from his end.

Could I contribute with this, or is someone already taking care of this enhancement?

ndelangen:
@danielmarcano Your contribution would be most welcome!

danielmarcano:
Awesome, @ndelangen! I will begin working on this.

If I end up using something like [https://reach.tech/](ReachUI), what would I need to consider when it comes to Storybook? In terms of: package.json dependencies, any necessary credits to the used open source library...

It would be my first contribution to Storybook.

MichaelArestad:
@danielmarcano I would write a proposal of the benefits of ReachUI first and potential pitfalls. We also were looking a bit at Radix.

danielmarcano:
Thank you, @MichaelArestad! I will take this into account, and take a look at Redix as well. Should I write this proposal as a new PR?

danielmarcano:
@MichaelArestad should I open a new RFC for this, in the [Discussions](https://github.com/storybookjs/storybook/discussions) tab?

PR #12915, opened 2020-10-26
Correct typo on naming-components and hierarchy

matharden:
Issue: When referring to stories in the past tense, "show" should be "shown".

## What I did

Changed "show" to "shown".

## How to test

No test required. Docs issue.


shilman (approved):
Good catch!

PR #12916, opened 2020-10-26
Core: Fix async keyword that breaks hermes

reimertz:
We have been having weird issues witch react-native once we enabled Hermes.

We finally managed to figure out what it was by running:

```
 curl 'http://localhost:8081/index.bundle?platform=android&dev=true&minify=false' > bundle.js
 node_modules/hermes-engine/osx-bin/hermesc bundle.js -emit-binary -out bundle.hbc
```
Thanks you @kyle-ssg (https://github.com/facebook/hermes/issues/298#issuecomment-663999190)

This builds a bundle for android and then runs it through the hermes compiler.

We can then see that we get syntax errors because of the [websocket lib](https://github.com/storybookjs/storybook/blob/next/lib/channel-websocket/src/index.ts#L78)

I think it's because of the `async`keyword..

By renaming async to useAsync, we mitigated these issues and could properly run our app on android with hermes enabled.

have tested storybook + hermes locally and works great with this PR applied.

### DIFF
--- a/lib/channel-websocket/src/index.ts
+++ b/lib/channel-websocket/src/index.ts
@@ -11,7 +11,7 @@ interface WebsocketTransportArgs {
 
 interface CreateChannelArgs {
   url: string;
-  async: boolean;
+  useAsync: boolean;
   onError: OnError;
 }
 
@@ -73,7 +73,7 @@ export class WebsocketTransport {
   }
 }
 
-export default function createChannel({ url, async, onError }: CreateChannelArgs) {
+export default function createChannel({ url, useAsync, onError }: CreateChannelArgs) {
   const transport = new WebsocketTransport({ url, onError });
-  return new Channel({ transport, async });
+  return new Channel({ transport, useAsync });
 }
--- a/lib/channels/src/index.test.ts
+++ b/lib/channels/src/index.test.ts
@@ -33,7 +33,7 @@ describe('Channel', () => {
     });
 
     it('should set isAsync to true if passed as an argument', () => {
-      channel = new Channel({ async: true });
+      channel = new Channel({ useAsync: true });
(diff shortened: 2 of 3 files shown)


kyle-ssg:
I would just point out that in the end updating storybook fixed my particular issue:

```
    "@storybook/addon-actions": "^6.0.0-rc.14",
    "@storybook/addon-knobs": "^6.0.0-rc.14",
    "@storybook/addon-links": "^6.0.0-rc.14",
    "@storybook/addon-ondevice-knobs": "^5.3.19",
    "@storybook/addons": "^6.0.0-rc.14",
    "@storybook/react-native": "^5.3.19",
```

reimertz:
Thanks @kyle-ssg for letting us know! Sadly, updating Storybook wasn't enough to get around hermes failing. 😞

shilman (approved):
Thanks, this seems reasonable to me! @ndelangen ?

kyle-ssg:
> Thanks @kyle-ssg for letting us know! Sadly, updating Storybook wasn't enough to get around hermes failing. 😞

In that case thank you very much for saving me a future headache! :D

ndelangen (approved):

reimertz:
After some exploration and some [chatting with the hermes team,](https://github.com/facebook/hermes/issues/398) it seems this was fixed in `hermes-engline@0.5.1`. Sadly, react-native still have `hermes-engine@0.5.0` as the dependency which results in others probably having these issues but just adding

```
"resolutions": {
    "hermes-engine": "0.5.1"
  }
```

in your package.json file fixed it for us.

Considering all this - I don't feel like this is on Storybook to be honest. 
So feel free to close this PR if you want. :)

PR #12936, opened 2020-10-28
CLI: use Jest to test CLI commands and remove outdated fixtures

jamesgeorge007:
# What I did

- Removed CLI test fixtures.
- Leverage Jest for e2e tests.

closes #13318


jamesgeorge007:
Running smoke-test in `sfc_vue` directory exhibits an inconsistent behavior.

Starting storybook manually works fine (process completes with exit code `0`):-
Screenshot 2020-10-29 at 1 37 50 PM

It fails when invoked via `jest`:-
Screenshot 2020-10-29 at 1 32 57 PM

/cc @shilman @ndelangen

lib/cli/package.json
```diff
@@ -68,7 +68,8 @@
     "@types/puppeteer-core": "^2.0.0",
     "@types/semver": "^7.2.0",
     "@types/shelljs": "^0.8.7",
-    "@types/update-notifier": "^0.0.30"
+    "@types/update-notifier": "^0.0.30",
+    "rimraf": "^3.0.2"
```
gaetanmaisse, line 72: I think you can use [`fs-extra`](https://github.com/jprichardson/node-fs-extra#methods) (which is already a dep) instead of adding `rimraf`

lib/cli/test/commands/init.test.js
```diff
@@ -0,0 +1,49 @@
...
+
+beforeAll(() => {
+  fs.mkdirSync(runDirPath);
+  // Copy all files from fixtures directory to `run`
+  const dirs = fs.readdirSync(fixturesDirPath);
+  dirs.forEach((dir) => copyDirSync(path.join(fixturesDirPath, dir), runDirPath));
+});
+
+afterAll(() => {
+  rimraf.sync(runDirPath);
```
gaetanmaisse, line 20: We should find a way to clean things if users exit command in the middle of the tests. Or maybe just clean the directory in the beforeAll if it exists?

lib/cli/test/helpers.js
```diff
@@ -0,0 +1,64 @@
...
+
+const CLI_PATH = path.join(__dirname, '..', 'bin');
+
+/**
+ * Copy files
+ * @param {String} source - source directory path
+ * @param {String} target - path to the destination directory
+ * @returns {Void}
+ */
+const copyFileSync = (source, target) => {
```
gaetanmaisse, line 13: As for `rimraf` I think you can find methods doing that in [`fs-extra`](https://github.com/jprichardson/node-fs-extra#methods)

gaetanmaisse: Maybe https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy-sync.md?

gaetanmaisse:
Thanks for the PR @jamesgeorge007 💪🏻 

I just pushed a small commit to fix a path, broken after a directory move, and left a few comments. 
Can you take a look at them? Let me know if something is unclear :)

I will take a look at the `sfc_vue` case 🤔

gaetanmaisse:
@jamesgeorge007 I'm focusing on another PR but will take a new look at this one soon ;)

gaetanmaisse:
Found the issue with some fixture! I will add commits to fix that in the next few days :)

gaetanmaisse:
https://github.com/storybookjs/storybook/pull/13231 should fix things but a new error appeared, I'm investigating a bit more 🔍

gaetanmaisse:
`CLI tests` are back to 🟢  on `next` so it should be here too. Can you update the branch?

To get the full result of the tests and not stop at the first failing one is really great 🎉 ! 

I think one more thing we need to improve is maybe how the underlying error/logs can be catch and display clearly in Jest test result. To have something more meaningful than:
```sh

    Expected: 0
    Received: 1
```

jamesgeorge007:
Running smoke-test in the `sfc_vue` directory is still failing.

gaetanmaisse:
I discussed with @ndelangen and @shilman and it looks like all the fixtures are outdated. And so the result of these tests aren't really relevant nor reflect if Storybook properly works or not. As keeping up to date the fixtures is very time-consuming and as we have more robust e2e tests, it looks like we will remove these ones: https://github.com/storybookjs/storybook/pull/13318. 

I'm a bit disappointed to tell you that although you have spent time to them 😕

gaetanmaisse (approved):
🚀

gaetanmaisse:
@shilman can I merge this one or are we still merging only PRs stabilizing 6.1?

ndelangen:
should be good to merge to next IMHO

PR #12948, opened 2020-10-29
Components: react-syntaxt-highlighter optimization

ndelangen:
storybook perf, remove the dependency on highlight.js

see https://github.com/storybookjs/storybook/pull/12811


ndelangen:
before above

after below

Screenshot 2020-10-29 at 12 31 20

ndelangen:
@shilman do you think the visual regressions are worth it?

shilman (approved):
LGTM! 💯

PR #12966, opened 2020-10-31
UI: Don't listen for progress updates in static builds

ghengeveld:
Fix for the prebuilt manager. This ensures we don't try to fetch /progress on static builds, because we won't have a backend there.


shilman (approved):
LGTM! 💯

PR #12980, opened 2020-11-02
UI: Add support for expand/collapse keyboard shortcuts

teimurjan:
## What I did

Added `STORIES_EXPAND_ALL` and `STORIES_COLLAPSE_ALL` listeners to `useExpanded` hook. By doing that I fixed broken collapse/expand shortcuts and made `api.collapseAll`, `api.expandAll` working.


lib/ui/src/components/sidebar/useExpanded.ts
```diff
@@ -82,6 +87,25 @@ export const useExpanded = ({
     setExpanded({ ids: getAncestorIds(data, selectedStoryId), value: true });
   }, [data, selectedStoryId]);
 
+  const collapseAll = useCallback(() => {
+    setExpanded({ ids: Object.keys(data), value: false });
+  }, [data]);
```
ghengeveld, line 92: You'll have to omit the `rootIds` from this set, because those can never be collapsed.
This should probably do the trick:

```suggestion
  const collapseAll = useCallback(() => {
    const ids = Object.keys(data).filter(id => !rootIds.includes(id));
    setExpanded({ ids, value: false });
  }, [data, rootIds]);

```

For `expandAll` this doesn't really matter, because the rootIds will already be set to `true`.

ghengeveld (changes requested):
Thanks for this! I completely forgot about it. I have only one small suggestion. Other than that this LGTM.

teimurjan:
@ghengeveld ✅

teimurjan:
@ghengeveld done

teimurjan:
@ghengeveld it still fails. do you have an idea why?

PR #12981, opened 2020-11-02
CLI: Fix deprecation check

g1eny0ung:
Issue:

After `console.log` options, seems that there are no `dlls` option but have a `dll` option. @shilman PTAL.

## What I did

`!options.dlls` -> `!options.dll`

## How to test


lib/core/src/server/cli/utils.ts
```diff
@@ -25,11 +25,11 @@ const warnDLLsDeprecated = deprecate(
 );
 
 export function checkDeprecatedFlags(options: {
-  dlls?: boolean;
+  dll?: boolean;
   uiDll?: boolean;
   docsDll?: boolean;
 }) {
-  if (!options.dlls || options.uiDll || options.docsDll) {
+  if (options.dll || options.uiDll || options.docsDll) {
```
shilman, line 32: shouldn't this be `!options.dll`?

g1eny0ung: I tested it again, after passing `--no-dll`, the `dll` will become `false`, so it should be `option.dll`.

image

image

shilman: `--no-dll` is no longer needed because the DLLs have been entirely removed, so it should warn the user. if `.dll` is set, it means that the user did not pass `--no-dll` so, so no warning is needed.

g1eny0ung: Got it! I'm already updated.

g1eny0ung: Wait, is it should be `option.noDll`. As u said, the DLLs have been entirely removed. So, in the future, there will not have `dll` option?

shilman: Thanks! 👍

shilman (approved):
LGTM! Thanks @g1eny0ung 🙏

PR #12990, opened 2020-11-02
Build: Replace enzyme with react-test-library

tooppaaa:
Issue: #12979

## What I did
Setup RTL userEvent, jest-dom 
Convert css resource addon tests

## How to test

- CI should be greener


scripts/jest.init.ts
```diff
@@ -1,15 +1,18 @@
 import 'jest-enzyme/lib/index';
+import '@testing-library/jest-dom';
 
 // setup file
 import { configure } from 'enzyme';
```
tooppaaa, line 5: going to remove all enzyme at some point, I don't want to deal with typings now

shilman (approved):
LGTM! Great stuff @tooppaaa

PR #12994, opened 2020-11-03
Addon-docs: Angular empty string now infers to "string" instead of "void"

activenode:
Issue: #12986

## What I did
- I added a test for an empty string to make sure it's covered: `['', { name: 'string' }]` 
- I fixed the `extractTypeFromValue` function to verify that when `valueType == string` it will infer a string

## How to test

Test is added. Should be covered now. For testing if the issue is fixed please refer to the description in the issue.

### DIFF
--- a/addons/docs/src/frameworks/angular/compodoc.test.ts
+++ b/addons/docs/src/frameworks/angular/compodoc.test.ts
@@ -12,6 +12,7 @@ describe('extractType', () => {
   describe('with compodoc type', () => {
     it.each([
       ['string', { name: 'string' }],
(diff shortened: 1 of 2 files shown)


shilman (approved):
Looks good. Merging for now, but are there more types that need improved inference? Any chance you can put together a followup @activenode ?

activenode:
I can do a follow-up.

activenode:
I also found that the test was put in the wrong place and it should've failed (seems to have slipped in).
Will immediately create a follow-up PR to fix this.

PR #13003, opened 2020-11-03
Addon-docs: Hide React default props in source block

zeckdude:
Issue: https://github.com/storybookjs/storybook/issues/11510

## What I did

Added option to not show default props on code preview when a component isn't overriding those values. In `jsxDecorator.tsx`, it is using the `reactElementToString` method from the [react-element-to-jsx-string](https://www.npmjs.com/package/react-element-to-jsx-string#reactelementtojsxstringreactelement-options) package. If you check their options, they accept a `showDefaultProps` property, which when set to `false` omits the default props in the JSX. This sets that to `false` so it doesn't show the default props by default.

## How to test

  Yes. I added a test case, but it looks like it is failing due to a snapshot. I'm not familiar with snapshots, so I could use some help there please.
  No
  Yeah I think so, but I'm struggling determining where that would be added the best. Any advice?

### DIFF
--- a/addons/docs/src/frameworks/react/jsxDecorator.test.tsx
+++ b/addons/docs/src/frameworks/react/jsxDecorator.test.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */
 import React from 'react';
 import range from 'lodash/range';
+import PropTypes from 'prop-types';
 import addons, { StoryContext } from '@storybook/addons';
 import { renderJsx, jsxDecorator } from './jsxDecorator';
 import { SNIPPET_RENDERED } from '../../shared';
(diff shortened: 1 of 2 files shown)


shilman (approved):
LGTM. Thanks for tracking this down @zeckdude

Greenheart:
Thank you for fixing this @zeckdude! This has been one of the few issues my team have had with Storybook since we upgraded to v6, and thanks to your PR we'll get a really smooth DX with relevant code samples in docs.

Keep up the good work with the 6.1.0 release! ✨

shleewhite:
@zeckdude does this work with typescript? instead of setting Component.defaultProps, we just depend on the type.

zeckdude:
@shleewhite It currently has nothing to do with TS but I'm sure that could be included. Which type would it be looking for to determine that it is a component?

shleewhite:
@zeckdude our components are written like this:

```
export type componentProps = {
  propA?: string,
  propB: bool
};

export const myComponent = React.functionComponent<componentProps>(
  ({
       propA = 'Sample Text',
       propB,
  })
)
```

So I'd want the story for myComponent to ignore propA if it hasn't changed.

victor-grajski:
+1 to @shleewhite !

PR #13005, opened 2020-11-03
Build: Remove more enzyme

tooppaaa:
Issue:

## What I did
Move some to enzyme

## How to test

- CI


addons/links/src/preview.test.ts
```diff
@@ -32,11 +32,13 @@ jest.mock('global', () => ({
...
 
+const mockAddons = (addons as unknown) as jest.Mocked<typeof addons>;
+
 describe('preview', () => {
   describe('linkTo()', () => {
     it('should select the kind and story provided', () => {
-      const channel = { emit: jest.fn() };
-      addons.getChannel.mockReturnValue(channel);
+      const channel = { emit: jest.fn() } as any;
+      mockAddons.getChannel.mockReturnValue(channel);
```
gaetanmaisse, line 41: Maybe we can move that in a `beforeEach` as we are doing it in every test?

addons/links/src/react/components/link.test.tsx
```diff
@@ -28,17 +28,21 @@ const mockChannel = () => {
...
+
+      const { container } = render(<LinkTo kind="foo" story="bar" />);
 
-      const wrapper = shallow(<LinkTo kind="foo" story="bar" />);
-      await wrapper.instance().updateHref(wrapper.props());
-      wrapper.update();
-      expect(wrapper).toMatchSnapshot();
+      expect(container.firstChild).toMatchInlineSnapshot(`
+        <a
+          href="/"
```
gaetanmaisse, line 44: Shouldn't we have a `foo--bar ` somewhere because we are setting `kind="foo" story="bar"` on L39?

addons/links/src/preview.test.ts
```diff
@@ -48,9 +50,9 @@ describe('preview', () => {
     });
 
     it('should select the kind (only) provided', () => {
-      const channel = { emit: jest.fn() };
-      addons.getChannel.mockReturnValue(channel);
-      __STORYBOOK_STORY_STORE__.fromId.mockImplementation((input) => null);
+      const channel = { emit: jest.fn() } as any;
+      mockAddons.getChannel.mockReturnValue(channel);
+      __STORYBOOK_STORY_STORE__.fromId.mockImplementation((): any => null);
```
gaetanmaisse, line 55: This is not directly related to this PR but all the tests feel weird to me, we are always expecting the same thing i.e. `channel.emit` to have been called with `{ kind: 'kind', story: 'name' }` (which are the mocked values a the beginning of the file) whatever the input 🙃

tooppaaa: Before I saw your comment I saw this as well and went: "I must have forgot to reset mock, it doesn't make sense".
Definitely something weird here

addons/links/src/react/components/link.test.tsx
tooppaaa: Fixed, href is set asynchronously (componentDidMount)

shilman (approved):
LGTM!

PR #13007, opened 2020-11-04
Core: Fix possible "write after end" exception for response stream

matheuspiment:
Issue: #13006 

## What I did

I removed the code that closes the response to keep the "buffer / stream" open, making the hot-reload work again.

Screenshot from 2020-11-03 23-00-07

> I'm not sure if it is the best solution. Feel free to propose changes.

## How to test

1. Pull this branch and run `yarn bootstrap`
2. Go to `examples/react-ts` and run `yarn storybook`
3. Make any changes to the `examples/react-ts/src/button.tsx` component and save the file
4. See the hot-reload working again.


ghengeveld:
I think rather than keeping the connection open, we should make sure we don't write anything after it's closed. I'll look into it.

ghengeveld:
I [updated](https://github.com/matheuspiment/storybook/commit/c406dfa4ca1a4c341fe2cbdc7f1296396429d2aa) your PR with a different solution. Could you verify that this resolves the issue for you?

By the way I was not able to reproduce the issue based on your description. Maybe it's due to a Node.js version? I'm running `v12.13.1`.

matheuspiment:
That fixed the issue here. My Node.js version is `v10.23.0`.

ghengeveld (approved):
LGTM

Tests are failing but those don't originate from this branch, they're already failing on `next` as well.

tmeasday:
Deployed in Release 51.

PR #13009, opened 2020-11-04
Addon-docs: Fix test for Angular type inference

activenode:
Issue:
For the #12986 issue the PR https://github.com/storybookjs/storybook/pull/12994 was opened to fix it.

However the test was in the wrong array ('with compdoc type') but it fails on this array since it needs to be in 'without compodoc type'.

## What I did
I fixed the test - simply that.

## How to test

Run the tests.


shilman (approved):
Good catch

PR #13012, opened 2020-11-04
Core: Make sure cache is available before trying to use it

ho1234c:
I encountered error as below. so I could not start story book dev server. I try remove and install node_moudles many time but still broken.

```
info => Loading presets
info => Loading presets
info => Loading 1 config file in "libs/design-system/.storybook"
info => Loading 6 other files in "libs/design-system/.storybook"
ERR! TypeError: Cannot read property 'remove' of undefined
ERR!     at startManager (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/dev-server.js:288:21)
ERR!     at async Promise.all (index 1)
ERR!     at async storybookDevServer (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/dev-server.js:426:28)
ERR!     at async Object.buildDevStandalone (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/build-dev.js:306:33)
ERR!  TypeError: Cannot read property 'remove' of undefined
ERR!     at startManager (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/dev-server.js:288:21)
ERR!     at async Promise.all (index 1)
ERR!     at async storybookDevServer (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/dev-server.js:426:28)
ERR!     at async Object.buildDevStandalone (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/build-dev.js:306:33) {
ERR!   stack: "TypeError: Cannot read property 'remove' of undefined\n" +
ERR!     '    at startManager (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/dev-server.js:288:21)\n' +
ERR!     '    at async Promise.all (index 1)\n' +
ERR!     '    at async storybookDevServer (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/dev-server.js:426:28)\n' +
ERR!     '    at async Object.buildDevStandalone (/Users/jonghojeong/source/dano-frontend/node_modules/@storybook/core/dist/server/build-dev.js:306:33)'
ERR! }
```

## What I did

I think file system cache is fragile. in my case, don't know why, but the cache is broken.
so I add logic to prevent non-execution in the event of cache problem.
It is nit change but since I spent a lot of time to find solution, I hope it helps someone.

and I modified eslint rule to pass commit lint.


ghengeveld:
Good call @ho1234c! What's up with the changes to `.eslintrc.js`? Those seem completely unrelated.

lib/core/src/server/dev-server.ts
```diff
@@ -213,14 +214,14 @@ const startManager = async ({
 
     if (options.managerCache) {
```
ghengeveld, line 215: Please use `if (options.managerCache && options.cache)` instead of optional chaining here. If will avoid needlessly stringifying the managerConfig.

.eslintrc.js
```diff
@@ -39,6 +39,8 @@ module.exports = {
       rules: {
         'react/prop-types': 'off', // we should use types
         'no-dupe-class-members': 'off', // this is called overloads in typescript
+        'no-unused-expressions': 'off',
+        '@typescript-eslint/no-unused-expressions': 'error',
```
ghengeveld, line 43: I think these changes don't belong in this PR.

ghengeveld (changes requested):
Good call. I left some remarks.

lib/core/src/server/dev-server.ts
ho1234c: better way. fix it!

lib/core/src/server/dev-server.ts
```diff
@@ -174,17 +174,23 @@ const useProgressReporting = async (
...
     }
     reportProgress(progress);
   };
 
-  const modulesCount = (await options.cache.get('modulesCount')) || 1000;
-  new ProgressPlugin({ handler, modulesCount }).apply(compiler);
+  if (options.cache) {
+    const modulesCount = (await options.cache.get('modulesCount')) || 1000;
+    new ProgressPlugin({ handler, modulesCount }).apply(compiler);
+  }
```
ghengeveld, line 193: This one was actually fine. We do still want the ProgressPlugin even if you don't have cache, because it'll be used by the prebuilt manager.

lib/core/src/server/dev-server.ts
ghengeveld: 👍

lib/core/src/server/dev-server.ts
ghengeveld: I actually prefer the optional chaining, it's just that in the situation below it wasn't the best solution.

ho1234c:
```ts
let modulesCount = 1000;
if (options.cache) {
  modulesCount = await options.cache.get('modulesCount');
}
new ProgressPlugin({ handler, modulesCount }).apply(compiler);
```

how about it!?

ho1234c:
I see. optional chain is best..!

ghengeveld (approved):
LGTM

Thanks for contributing and pulling it through!

lib/core/src/server/dev-server.ts
ghengeveld: See other comment. Looks good now 👍

PR #13014, opened 2020-11-04
UI: Restrict layout styles to only apply to a rendered preview area

ghengeveld:
Issue: -

## What I did

I've bumped the specificity for the layout classes to only apply when the `sb-show-main` class is also present. This avoids an issue with the layout class somehow getting applied to the manager itself. I haven't seen that happen with the new class approach, but I've seen it happen with the old styles approach, with `body.style` getting set on the outer (manager) body.

## How to test

Not really testable, because I can't reproduce. This is a "better safe than sorry" tweak.

For those with access, this happened [here](https://5e5fcc0d6aa8bd00220faff4-vaajmccdlc.chromatic.com/?id=buttons--routing-choice-disabled&viewMode=story).

### DIFF
--- a/lib/core/src/server/templates/base-preview-head.html
+++ b/lib/core/src/server/templates/base-preview-head.html
@@ -7,7 +7,7 @@
     display: none;
   }
 
-  .sb-main-centered {
+  .sb-show-main.sb-main-centered {
     margin: 0;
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM Gert!

PR #13069, opened 2020-11-10
React: Don't create a new story function on every render

tmeasday:
This trips up React's reconciliation for re-using components.

Related to #12255


shilman (approved):
LGTM .. great stuff. Getting a bit complicated tho ... 🤔

PR #13073, opened 2020-11-10
UI: Fix Escape key handling perf

ghengeveld:
# What I did

While I was profiling, I discovered this `keydown` listener isn't actually targeted specifically at `Escape` but executes for any key. It's not apparent but this causes the sidebar to return to the tree view when you click _any_ button while the search input doesn't have focus.

I recommend hiding whitespace changes when looking at the [diff](https://github.com/storybookjs/storybook/pull/13073/files?diff=split&w=1).

## How to test


shilman (approved):
LGTM 💯

PR #13084, opened 2020-11-11
Disable instant-on manager for now

ghengeveld:
Issue: #13077

## What I did

Instead of opening the browser as soon as the manager is done, open it when both the preview and manager are done.

## How to test


shilman (approved):
Thanks @ghengeveld 🙏

PR #13095, opened 2020-11-12
UI: Fix tree highlight performance

ghengeveld:
## What I did

Various performance tweaks for how items in the sidebar are highlighted. Specifically this avoids rerendering nodes in the tree by using global CSS rather than passing props down. Also we've replaced `throttle` with `requestAnimationFrame` and implemented some other efficiency improvements.

In particular this is noticeable when holding down `ArrowDown` or `ArrowUp` to navigate the tree view. Here's a benchmark in `official-storybook` with `Chromatic` added as a ref (for a total of 2706 nodes), on a 2019 MacBook Pro:

### Before
Screenshot 2020-11-12 at 15 38 18

### After
Screenshot 2020-11-12 at 15 32 53

Where a single keydown event used to take ~205ms and ends up colliding with the next event, it now takes less than 20ms, which is fast enough to not cause any issues.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, we've updated the stories, but the snapshots should not change.


shilman (approved):
Looks good to me!

lib/ui/src/components/sidebar/Tree.tsx
```diff
@@ -261,14 +254,21 @@ export const Tree = React.memo<{
       );
     }, [data]);
 
+    const ancestry = useMemo(() => {
+      return collapsedItems.reduce(
+        (acc, id) => Object.assign(acc, { [id]: getAncestorIds(collapsedData, id) }),
+        {} as { [key: string]: string[] }
+      );
+    }, [collapsedItems, collapsedData]);
```
tmeasday, line 262: 👌

PR #13096, opened 2020-11-12
Angular: Support Angular 11

gaetanmaisse:
Issue: #13093

## What I did

Some part of the Angular CLI webpack config has moved between Angular 10- and Angular 11. 
So we now dynamically require them from Angular 11 paths and fallback to the ones of Angular 10-.

I also: 
 - added an E2E config for Angular 10 as `latest` is now Angular 11
 - move back all Angular E2E tests to classic CircleCI executors as they rework with Node 10 (https://github.com/storybookjs/storybook/issues/12735)

## How to test

- Angular E2E tests should be 🟢  on CI and no error should be thrown in the logs anymore

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,8 +41,8 @@ jobs:
       - restore_cache:
           name: Restore Yarn cache
           keys:
-            - build-yarn-cache-v1--{{ checksum "yarn.lock" }}
-            - build-yarn-cache-v1--
+            - build-yarn-cache-v2--{{ checksum "yarn.lock" }}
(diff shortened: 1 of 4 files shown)


shilman (changes requested):
LGTM. are we testing angular9 / 10 / 11 now? if so, can we remove angular9 because our circle bill is going out of control?

rhutchison:
Thanks... definitely a blocker for us to upgrade.

shilman (approved):
Awesome @gaetanmaisse 💯

PR #13115, opened 2020-11-14
Build: Remove some dependencies and improve E2E workflow

gaetanmaisse:
## What I did

- Replace `del` calls by `fs-extra` ones as it was only used in one place
- Remove some unused dependencies
- Remove `setResolutions` step in the E2E as it appears to not be needed anymore. This step was causing Yarn to install all the SB packages for each E2E test so the whole `example-v2` job should be faster.
- Reorder E2E steps to be sure `yarn install` is done before `sb init`, this ensures that `sb init` will use Yarn and its cache under the hood instead of NPM.

## How to test

- CI should be 🟢


scripts/run-e2e.ts
```diff
@@ -137,30 +135,6 @@ const initStorybook = async ({ cwd, autoDetect = true, name }: Options) => {
   }
 };
 
-// Verdaccio doesn't resolve *
-// So we set resolutions manually in package.json
-const setResolutions = async ({ cwd }: Options) => {
```
tooppaaa, line 142: I was going to ask about this one. Does it make sense to resolve to sibling path in order to use local sb when working on e2e ?
This would also help getting examples out of the monorepo.
Does this even work !?

tooppaaa (approved):
🚀

shilman (approved):
LGTM!

PR #13129, opened 2020-11-16
Build: Fix Yarn 2 E2E tests

gaetanmaisse:
## What I did

- use an unreleased version of Yarn 2 to fix the issue with fsevents patch
- update missing package extensions for Yarn 2 tests
- add react as additionalDeps in sfcVue E2E tests to tmp fix Yarn 2 E2E tests

## How to test

- CI job `example-v2-yarn-2` should be 🟢


shilman (approved):
LGTM. thanks!

PR #13140, opened 2020-11-16
Build: reduce job parallelism to 4 

gaetanmaisse:
Issue: We are currently facing some billing issues for CircleCI so we are back on the "classic" open source plan. The latter doesn't allow us to use massive parallelism as we are doing in `examples`, `examples-v2` etc jobs. 

These jobs are not started with less parallelism, they are just not ran at all:

image

And doesn't appear in PR checks 😱 

## What I did

Reduce job parallelism to 4 as a workaround to still have all CI jobs ran.

## How to test
 - All CI jobs should be back in the game

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ jobs:
             - node_modules
   chromatic:
     executor: sb_node
-    parallelism: 11
+    parallelism: 4
     steps:
       - checkout
       - attach_workspace:
@@ -140,7 +140,7 @@ jobs:
     executor:
       class: medium
       name: sb_node
-    parallelism: 11
+    parallelism: 4
     steps:
       - checkout
       - attach_workspace:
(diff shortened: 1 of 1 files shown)


gaetanmaisse:
CI takes forever to complete... not sure what we should do until having feedback from the guys from CircleCI:
 - On the one hand, we currently have "half" of our CI jobs that are just not ran.
 - On the other hand, setting the parallelism to 4 makes CI run for ~45min.

⚖️   @ndelangen @shilman I let you decide whether we should merge or close this PR.

ndelangen:
Let's merge for now, until we get feedback from CircleCI @gaetanmaisse @shilman ?

gaetanmaisse:
👌🏻  Ok for me @ndelangen

shilman (approved):
💯

PR #13158, opened 2020-11-18
Server: Forward globals in fetchStoryHtml

### LINKED ISSUE #13133: @storybook/server inlcude global parameters when requesting story to the server
I want to use storybook/server to show stories of emails generated with mjml.

Each of the emails has an English and a Spanish version and also an HTML and a text version. For this reason I created a couple globalTypes with the toolbar-addon.
However the selected values in these dropdowns are not passed as parameters to the rendering server.

Is there a way of passing these global parameters to the server as query parameters?

santialbo:
Issue: #13133

This PR allows `@storybook/server` to pass global params to the server for rendering the story.

### DIFF
--- a/app/server/src/client/preview/render.ts
+++ b/app/server/src/client/preview/render.ts
@@ -6,9 +6,9 @@ import { RenderContext, FetchStoryHtmlType } from './types';
 
 const rootElement = document.getElementById('root');
 
-const defaultFetchStoryHtml: FetchStoryHtmlType = async (url, path, params) => {
+const defaultFetchStoryHtml: FetchStoryHtmlType = async (url, path, params, storyContext) => {
   const fetchUrl = new URL(`${url}/${path}`);
-  fetchUrl.search = new URLSearchParams(params).toString();
+  fetchUrl.search = new URLSearchParams({ ...params, ...storyContext.globals }).toString();
 
   const response = await fetch(fetchUrl);
   return response.text();
@@ -52,6 +52,7 @@ export async function renderMain({
   showError,
   forceRender,
   parameters,
+  storyContext,
   storyFn,
   args,
   argTypes,
@@ -65,8 +66,8 @@ export async function renderMain({
   } = parameters;
 
   const fetchId = storyId || id;
-  const fetchParams = { ...params, ...storyArgs };
-  const element = await fetchStoryHtml(url, fetchId, fetchParams);
+  const storyParams = { ...params, ...storyArgs };
+  const element = await fetchStoryHtml(url, fetchId, storyParams, storyContext);
 
   showMain();
   if (typeof element === 'string') {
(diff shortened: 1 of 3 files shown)


app/server/src/client/preview/types.ts
```diff
@@ -1,8 +1,10 @@
+import { StoryContext } from '@storybook/addons';
+
 export { RenderContext } from '@storybook/core';
 
 export type StoryFnServerReturnType = any;
 
-export type FetchStoryHtmlType = (url: string, id: string, params: any) => Promise<string | Node>;
+export type FetchStoryHtmlType = (url: string, id: string, params: any, globals: StoryContext) => Promise<string | Node>;
```
jonspalmer, line 7:
```suggestion
export type FetchStoryHtmlType = (url: string, id: string, params: any, context: StoryContext) => Promise<string | Node>;
```

app/server/src/client/preview/render.ts
```diff
@@ -6,9 +6,9 @@ import { RenderContext, FetchStoryHtmlType } from './types';
 
 const rootElement = document.getElementById('root');
 
-const defaultFetchStoryHtml: FetchStoryHtmlType = async (url, path, params) => {
+const defaultFetchStoryHtml: FetchStoryHtmlType = async (url, path, params, storyContext) => {
   const fetchUrl = new URL(`${url}/${path}`);
-  fetchUrl.search = new URLSearchParams(params).toString();
+  fetchUrl.search = new URLSearchParams({ ...params, ...storyContext.globals }).toString();
```
jonspalmer, line 11: I still wonder if we should have a configuration setting to control this behavior. Should it be opt in?

santialbo: Opt-in as in configuration parameter to merge the globals? Another option is to not merge globals and let the user write a custom `fetchStoryHtml` if they need globals.

jonspalmer: Yes opt-in to merge. We could force the user to write a custom `fetchStoryHtml` but as seen here in the file there is enough going on that its a pretty heavy lift it all you want is globals. 

Maybe I'm over thinking the automatic merge of globals. It's unlikely that a few extra query params would break the  server implementation, they'd get ignored. Perhaps we stick with this implementation and wait for someone make the feature request for something more complicated and a real world usecase to solve against?

jonspalmer: Do we worry about conflicts between globals and params keys? Should we combine in the opposite order so that params would take precedence in any conflict?

santialbo: Yes swaping the order makes sense. I'll change that.

I think it makes sense to have them by default. I would say that, as a user, if you use globalTypes you would expect to receive them in the query (at least I did).

jonspalmer (approved):
Looks great

santialbo:
This is not stale, this is awaiting to be merged.

PR #13159, opened 2020-11-18
UI: Fix CSS for IE11

ghengeveld:
Issue: #12386

## What I did

This fixes vertical centering in IE11 when using `layout: "centered"`. It also replaces [8-digit hex color values](https://www.w3.org/TR/css-color-4/#hex-notation) with `transparentize`, because IE11 doesn't support this notation.

I used Browserstack to test this.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, but we don't test against IE right now.


shilman (approved):
LGTM!

PR #13162, opened 2020-11-18
Fix incorrect component reference in angular-cli example

S1ngS1ng:
## What I did
Fixed a typo

## How to test

- Is this testable with Jest or Chromatic screenshots? Not sure

### The online version
image

### Local with fix
image


shilman (approved):
Thanks for the fix!!

PR #13165, opened 2020-11-18
Core: Fix using cached manager on the 2nd run

ghengeveld:
Issue: #13156

## What I did

This fixes the cache key matching to also work on the 2nd run, when we have a cached manager. It wasn't working because the cache key contained a `cache` prop which is different on the first run (before we have a cache), causing the cached manager to get ignored on the 2nd run.

## How to test


shilman (approved):
LGTM. Thanks for the quick turnaround @ghengeveld !

PR #13195, opened 2020-11-20
Core: Resolve react and react-dom from core

merceyz:
Issue: Fixes the runtime part of https://github.com/storybookjs/storybook/issues/13194

`@storybook/core` attempts to resolve `react` and `react-dom` from the users project to set some webpack aliases but it should resolve them from itself. This avoids relying on hoisting which makes it work under PnP when the user doesn't have `react` and  `react-dom` as dependencies (i.e. when using `@storybook/vue`)

## What I did

Use `require.resolve` to resolve them from `@storybook/core`, this will get it from the parent which is one of the apps, and in the case of `@storybook/react` it will get it from its parent which is usually the users project

## How to test

CI should still be green


lib/core/src/server/preview/iframe-webpack.config.ts
```diff
@@ -22,10 +20,7 @@ import { getPreviewHeadHtml, getPreviewBodyHtml } from '../utils/template';
 import { toRequireContextString } from './to-require-context';
 import { useBaseTsSupport } from '../config/useBaseTsSupport';
 
-const reactPaths: Record<string, string> = {};
-
 const storybookPaths: Record<string, string> = [
-  'addons',
```
merceyz, line 28: Presumably doesn't need to be resolved twice

lib/core/src/server/preview/iframe-webpack.config.ts
```diff
@@ -41,19 +36,12 @@ const storybookPaths: Record<string, string> = [
   (acc, sbPackage) => ({
     ...acc,
     [`@storybook/${sbPackage}`]: path.dirname(
-      resolveFrom(__dirname, `@storybook/${sbPackage}/package.json`)
+      require.resolve(`@storybook/${sbPackage}/package.json`)
```
merceyz, line 39: Has the same result

gaetanmaisse (approved):

PR #13197, opened 2020-11-20
Correctly reference managerWebpack in preset docs

koop:
## What I did

Searched the codebase for webpackManager and realized it was managerWebpack.


shilman (commented):
LGTM. thanks!

PR #13205, opened 2020-11-21
UI: Fix lineheight in typeset component

domyen:
Issue: #12643 

## What I did
Set a reasonable line-height for the Typeset component. This fixes the regression that makes typeset look like this 👇
image

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes. Go to Typeset component.


shilman (approved):
Thanks @domyen 🙏

yannbf (approved):
LGTM! If we were to leave lineHeight unset maybe the component would be affected by global css that is injected by the users, so I think fixing it is a good idea.

Issue #9727, opened 2020-02-03
[addon:google-analytics] Clarify where to add environment variables


liamross:
**Is your feature request related to a problem? Please describe.**
It's unclear as to where to add the environment variables. According to the README:

> Then, set an environment variable
>
> ```
> window.STORYBOOK_GA_ID = UA-000000-01
> window.STORYBOOK_REACT_GA_OPTIONS = {}
> ```

This doesn't specify where they could be added. If I add the following to the beginning of my preview.js file, `react-ga` throws a warning into console

```js
// In preview.js
window.STORYBOOK_GA_ID = 'UA-0000000-0';
```

```bash
# In console
[react-ga] gaTrackingID is required in initialize()
```

**Describe the solution you'd like**
Some clearer documentation around where to add the ENV variables.

**Describe alternatives you've considered**
I guess if an explanation on this issue is enough then that would work too.

**Are you able to assist bring the feature to reality?**
Yes, I'd be happy to add the clarification to the README

liamross:
It seems to work when added to the top of `manager.js`, but based on the documentation I don't know if that's the preferred location.

urish:
Thanks for the manager.js tip, worked perfectly for me!

PR #13234, opened 2020-11-24
Update Google Analytics instructions

mattfelten:
Based on what we found, as well as what was shown in #9727


shilman (approved):
Makes sense. Thanks!

shilman (approved):
Makes sense. Thanks!

PR #13239, opened 2020-11-24
Core: Fix webpack5 compatibility check for ProgressPlugin

FabianSellmann:
Issue:
Webpack 5 ProgressPlugin only supplies 2 arguments to the handler. See https://github.com/webpack/webpack/blob/6394e2d1d87e82871e16945822b14fbfab954b71/lib/ProgressPlugin.js#L491-L494

This would make storybook compatible with webpack 5 again (at least from what I have seen so far).


shilman (approved):
Looks safe to me! Merging even if it doesn't fix the whole compatibility story. 😄

PR #13260, opened 2020-11-25
Addon-backgrounds: Fix grid offset always using default value

cone56:
## What I did

- Fix falsy value check to allow zero values in offsetX and offsetY parameters
- Fix default value fallback conditional logic

## How to test


yannbf (approved):
Great catch on the missing parenthesis! Thanks for your contribution!

PR #13266, opened 2020-11-25
Core: Don't use prebuilt or cached manager when running smoke test

ghengeveld:
Issue: #13262

## What I did

Ignore any prebuilt or cached manager when running with `--smoke-test`.
Additionally I improved some logging and made the related code a little more resilient.

## How to test

Run with `--smoke-test`. Note it will not use the prebuilt or cached manager. It will also not clear the cache, so running without `--smoke-test` afterwards will use a cached version as normal.


shilman (approved):
Good stuff 💯

PR #13275, opened 2020-11-25
Core: Ensure node `name` does not contain leading/trailing whitespace

ghengeveld:
Issue: #13251

## What I did

Primarily, this trims each part of the `title` (aka `kind`), after splitting it on `/`:

```js
const groups = kind.split('/').map((part) => part.trim());
```

The `.map(part => part.trim())` is what's added. Everything else in this PR is refactoring. I'll leave comments to clarify.

Pro tip: ignore whitespace changes when reviewing this 👍 

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, a unit test should be added.

### DIFF
--- a/lib/api/src/lib/stories.ts
+++ b/lib/api/src/lib/stories.ts
@@ -117,14 +117,6 @@ const warnChangedDefaultHierarchySeparators = deprecate(
   `
 );
 
-const toKey = (input: string) =>
-  input.replace(/[^a-z0-9]+([a-z0-9])/gi, (...params) => params[1].toUpperCase());
-
-const toGroup = (name: string) => ({
-  name,
-  id: toKey(name),
-});
-
 export const denormalizeStoryParameters = ({
   globalParameters,
   kindParameters,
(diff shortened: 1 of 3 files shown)


lib/api/src/lib/stories.ts
```diff
@@ -117,14 +117,6 @@ const warnChangedDefaultHierarchySeparators = deprecate(
   `
 );
 
-const toKey = (input: string) =>
-  input.replace(/[^a-z0-9]+([a-z0-9])/gi, (...params) => params[1].toUpperCase());
-
-const toGroup = (name: string) => ({
-  name,
-  id: toKey(name),
-});
```
ghengeveld, line 126: `toGroup` only adds an `id` prop, but this prop was never really used because it got overwritten by the `id` created through `sanitize`. So we don't need this mapping at all.

lib/api/src/lib/stories.ts
```diff
@@ -144,103 +136,89 @@ export const transformStoriesRawToStoriesHash = (
   input: StoriesRaw,
   { provider }: { provider: Provider }
 ): StoriesHash => {
-  const anyKindMatchesOldHierarchySeparators = Object.values(input)
-    .filter(Boolean)
-    .some(({ kind }) => kind.match(/\.|\|/));
+  const values = Object.values(input).filter(Boolean);
```
ghengeveld, line 139: This avoids doing the conversion and filtering twice.

lib/api/src/lib/stories.ts
```diff
@@ -144,103 +136,89 @@ export const transformStoriesRawToStoriesHash = (
...
-      let groups: string[];
-      const parts: string[] = kind.split('/');
-      // Default showRoots to true if they didn't set it.
-      if ((!setShowRoots || showRoots) && parts.length > 1) {
-        [root, ...groups] = parts;
-      } else {
-        groups = parts;
-      }
+    const groups = kind.split('/').map((part) => part.trim());
+    const root = (!setShowRoots || showRoots) && groups.length > 1 ? [groups.shift()] : [];
```
ghengeveld, line 152: Using `array.shift()` allows us to avoid `let` and greatly simplify the way we assign these two variables. `root` gets wrapped in an array so we can directly spread it below, avoiding the odd-looking `concat(root || [])` logic.

lib/api/src/lib/stories.ts
```diff
@@ -144,103 +136,89 @@ export const transformStoriesRawToStoriesHash = (
...
+
+    const rootAndGroups = [...root, ...groups].reduce((list, name, index) => {
```
ghengeveld, line 154: We now just get the `name` string, rather than a group object, because that's all we ever really used.

lib/api/src/lib/stories.ts
```diff
@@ -144,103 +136,89 @@ export const transformStoriesRawToStoriesHash = (
...
               Did you create a path that uses the separator char accidentally, such as 'Vue <docs/>' where '/' is a separator char? See https://github.com/storybookjs/storybook/issues/6128
             `
-            );
-          }
+        );
+      }
 
-          if (!!root && index === 0) {
-            const result: Root = {
-              ...group,
```
ghengeveld, line 193: Note `group` had only two properties, `name` and `id`. But `id` is overwritten immediately after, so effectively we're only assigning `name` here. Might as well do that directly.

lib/api/src/lib/stories.ts
```diff
@@ -144,103 +136,89 @@ export const transformStoriesRawToStoriesHash = (
...
+          id,
+          name,
+          depth: index,
+          children: [],
+          isComponent: false,
+          isLeaf: false,
+          isRoot: true,
+        });
+      } else {
+        list.push({
```
ghengeveld, line 179: Pushing to the accumulator (`list`) is more efficient than concatting, because that would yield a new array with each iteration.

lib/client-api/src/story_store.ts
```diff
@@ -343,10 +343,8 @@ export default class StoryStore {
         'Cannot add a story when not configuring, see https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#story-store-immutable-outside-of-configuration'
       );
 
-    if (storyParameters) {
-      checkGlobals(storyParameters);
-      checkStorySort(storyParameters);
-    }
```
ghengeveld, line 349: This is technically unrelated, but DeepScan warned me about it. `storyParameters` has a default value, so checking if it exists is useless. Later in this file, this variable is also used without existence check, so we can be sure it indeed is never null. I did a quick reference check to be sure.

tmeasday:
I'm surprised this didn't involve any test changes. Do we need a test for this case?

ghengeveld:
> I'm surprised this didn't involve any test changes. Do we need a test for this case?

Yes we do, but I'm not sure where to add it. This whole lib is untested.

tmeasday:
> This whole lib is untested

Untrue! 

https://github.com/storybookjs/storybook/blob/0f7a3a07bedf44a62df6303b5f969e87b76e2471/lib/api/src/tests/stories.test.js

ghengeveld:
@tmeasday I added a test.

tmeasday (approved):
LGTM

PR #13281, opened 2020-11-26
Remove unused dependency @svgr/webpack

cmrn:
## What I did
While slimming down the `node_modules` folder of one of our large projects at work, I noticed a transient dependency on `@svgr/webpack` coming from storybook which seems to be unused. The dependency was added in #4836, but the only remaining reference seems to be [this commented out story](https://github.com/storybookjs/storybook/blame/99223b147fb19f46b341dc7627a4ed13f6f5c561/examples/cra-kitchen-sink/src/stories/cra-svgr.stories.js).

Let me know if I'm missing something here :)

### DIFF
--- a/app/react/package.json
+++ b/app/react/package.json
@@ -41,7 +41,6 @@
     "@storybook/core": "6.1.6",
     "@storybook/node-logger": "6.1.6",
     "@storybook/semver": "^7.3.2",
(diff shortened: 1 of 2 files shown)


shilman (approved):
Nice catch

PR #13283, opened 2020-11-26
cleanup teamcity

ndelangen:
teamcity was running a few tasks that are no longer needed (docs is now deployed from frontpage) (eslint is ran both in CircleCI & deepscan)


ndelangen:
@shilman self merge, since this only touches teamcity CI config, nothing else

PR #13284, opened 2020-11-26
sort package.json & fix incorrect ts3.5 ignore

ndelangen:
package.json's should be sorted correctly && there was an incorrect gitignore entry regarding ts3.5, should be 3.4


ndelangen:
@shilman self-merge since this only reorders some keys in package.json

PR #13285, opened 2020-11-26
cleanup run-registry script

ndelangen:
just removed a function in a script that's no longer used (internal script for testing)


ndelangen:
@shilman self-merge because this removed unused code in an internal testing script

PR #13288, opened 2020-11-26
Addon-docs: Fix Vue source snippets for function attributes

leettaylor:
Followup to the excellent work done in #12812 by @pocka 

# What I did

The `sourceDecorator` currently throws an error if for some reason the value of an attribute is a function so I've added a simple check for the typeof the value:

```
Failed to generate dynamic story source: TypeError: Cannot read property 'includes' of undefined
```

I'm not sure what the underlying issue is that causes the type to be a function but it seems to be related to event handlers.

Here's some example code for a component/story (Not committed to the repo as it breaks storyshot tests) that causes the source to not be generated.

Example component:
```vue
<template>
  <div class="pagination">
    <button :disabled="isPreviousButtonDisabled" @click="previousPage">
      « Previous
    </button>

    <button :disabled="isNextButtonDisabled" @click="nextPage">
      Next »
    </button>
  </div>
</template>

<script>
export default {
  name: 'simple-pagination',
  props: {
    currentPage: {
      type: Number,
      required: true,
    },
    pageCount: {
      type: Number,
      required: true,
    },
  },
  computed: {
    isPreviousButtonDisabled() {
      return this.currentPage === 1;
    },
    isNextButtonDisabled() {
      return this.currentPage >= this.pageCount;
    },
  },
  methods: {
    nextPage() {
      this.$emit('next-page');
    },
    previousPage() {
      this.$emit('previous-page');
    },
  },
};
</script>
```

```js
import SimplePagination from './SimplePagination.vue';

export default {
  title: 'Components/Pagination/SimplePagination',
  component: SimplePagination,
  argTypes: {
    nextPage: { action: 'Next page' },
    previousPage: { action: 'Previous page' },
  },
  args: {
    currentPage: 2,
    pageCount: 3,
  },
};

  props: Object.keys(argTypes),
  components: { SimplePagination },
  template:
    '<simple-pagination @next-page="nextPage" @previous-page="previousPage" v-bind="$props" />',
});

```

For some reason the value of the `@previous-page` attribute is being passed through as a function instead of just being undefined like `@next-page`:

```
 actionHandler() {
    var channel = _addons.addons.getChannel();

    var id = (0, _uuid.v4)();
    var minDepth = 5; // anything less is really just storybook internals

    for (var _len = argument…
```

## How to test

If your answer is yes to any of these, please make sure to include it in your PR. ... done (`addons/docs/src/frameworks/vue/sourceDecorator.test.ts`)

### DIFF
--- a/addons/docs/src/frameworks/vue/sourceDecorator.test.ts
+++ b/addons/docs/src/frameworks/vue/sourceDecorator.test.ts
@@ -32,7 +32,7 @@ describe('vnodeToString', () => {
 
   it('attributes', () => {
     const MyComponent: ComponentOptions<any, any, any> = {
-      props: ['propA', 'propB', 'propC', 'propD', 'propE', 'propF', 'propG', 'propH'],
+      props: ['propA', 'propB', 'propC', 'propD', 'propE', 'propF', 'propG'],
       template: '<div/>',
     };
 
@@ -48,16 +48,14 @@ describe('vnodeToString', () => {
                 propC: null,
                 propD: {
                   foo: 'bar',
-                  bar: 'foo',
                 },
-                propE: 'propE',
-                propF: true,
-                propG() {
+                propE: true,
+                propF() {
                   const foo = 'bar';
 
                   return foo;
                 },
-                propH: undefined,
+                propG: undefined,
               },
             };
           },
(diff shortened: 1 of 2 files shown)


pocka (commented):
@leettaylor 
Thank you for the fix! Could you add a test at `sourceDecorator.test.ts`?

leettaylor:
> Thank you for the fix! Could you add a test at `sourceDecorator.test.ts`?

@pocka I've now updated the `attributes` test to have coverage for a prop being set to a `function`, `undefined`, or `true`.

pocka (approved):
LGTM 👍

PR #13294, opened 2020-11-26
CLI: Fix not printing managerTotalTime when using cached manager

ghengeveld:
Issue: #13292

## What I did

I simply made this data optional, so it will be falsy again and thus not print the manager time. See issue for info.

Screenshot 2020-11-26 at 14 27 41

## How to test

### DIFF
--- a/lib/core/src/server/build-dev.ts
+++ b/lib/core/src/server/build-dev.ts
@@ -153,8 +153,8 @@ function outputStartupInformation(options: {
   version: string;
   address: string;
   networkAddress: string;
-  managerTotalTime: [number, number];
-  previewTotalTime: [number, number];
+  managerTotalTime?: [number, number];
+  previewTotalTime?: [number, number];
 }) {
   const {
     updateInfo,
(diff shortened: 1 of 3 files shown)


lib/core/src/server/build-dev.ts
```diff
@@ -283,10 +283,9 @@ export async function buildDevStandalone(
 
     if (options.smokeTest) {
       await outputStats(previewStats, managerStats);
-      const managerWarnings = (managerStats as any).toJson().warnings.length > 0;
-      const previewWarnings =
-        !options.ignorePreview && (previewStats as any).toJson().warnings.length > 0;
-      process.exit(managerWarnings || previewWarnings ? 1 : 0);
+      const hasManagerWarnings = managerStats && (managerStats as any).toJson().warnings.length > 0;
```
gaetanmaisse, line 286: As types are now correct and match the objects you should be able to remove this `as any` and the one from next line.

lib/core/src/server/dev-server.ts
```diff
@@ -258,7 +258,7 @@ const startManager = async ({
   }
 
   if (!managerConfig) {
-    return { managerStats: null, managerTotalTime: [0, 0] } as ManagerResult;
+    return {} as ManagerResult;
```
gaetanmaisse, line 261: Same here, you should be able to drop `as ManagerResult`

lib/core/src/server/dev-server.ts
```diff
@@ -311,7 +311,7 @@ const startPreview = async ({
   outputDir,
 }: any): Promise<PreviewResult> => {
   if (options.ignorePreview) {
-    return { previewStats: null, previewTotalTime: [0, 0] } as PreviewResult;
+    return {} as PreviewResult;
```
gaetanmaisse, line 314: And same here

gaetanmaisse (approved):

shilman (approved):
LGTM

PR #13295, opened 2020-11-26
Core: Only apply `express.json()` middleware to /runtime-error route

ghengeveld:
Issue: #13268

## What I did

This moves the `express.json()` middleware from application global to just the route handler for `/runtime-error`.

## How to test


shilman (approved):
💯

PR #13302, opened 2020-11-26
Components: Fix Zoom for IE 11

tooppaaa:
Issue: #13272

## What I did
Add args and try catch

## How to test

- IE 11


shilman (approved):
💯 @tooppaaa is VIP

PR #13318, opened 2020-11-27
remove the cli tests

ndelangen:
remove the cli tests

why?
These test were setup before we had the examples-v2 tests

The removed test tested whether the storybook CLI & resulting bootstrapped storybook works.
However these fixtures do not get updated, and so over time have come to reflect what user experience less and less.

Now that we have tests that run against the latest (and other versions) of things like CRA & angular-cli I think it's safe to remove these tests.


jamesgeorge007:
Isn't it a better idea to not remove the entire test suite, but only the test fixtures? #12936 aims at leveraging Jest for the purpose; we can keep the test case for default behavior.

ndelangen:
The fixtures ARE the only tests AFAIK?

jamesgeorge007:
> The fixtures ARE the only tests AFAIK?

#12936 introduces a test case that makes sure that the closest match shows up on supplying an unknown command.

ndelangen:
@jamesgeorge007 ah that seems really valuable, I like that. If this is merged, can you adjust your PR to add any requirements back for the above?

jamesgeorge007:
> @jamesgeorge007 ah that seems really valuable, I like that. If this is merged, can you adjust your PR to add any requirements back for the above?

I think a better idea would be to address this as part of #12936 if you're okay with it.

jamesgeorge007:
ca7649b1826eb096ac335034716fa4863198e41e

PR #13319, opened 2020-11-27
UI: Fix keybindings on non-US keyboard layouts

ghengeveld:
Issue: #13296

## What I did

This fixes the `/` keybinding on Hungarian keyboards (and probably other). Besides that, I've taken the opportunity to refactor some of this logic, without introducing a whole library dependency.

## How to test

Test the `/` keybinding to focus the search input using `Shift+6` on a Hungarian keyboard layout.


shilman (approved):
Nice one

PR #13329, opened 2020-11-29
React: Make @babel/core an optional peer dependency

arcanis:
It seems that `@babel/core` is only used in two locations in the `react` app:

```
storybook/app/react/src/server/framework-preset-react-docgen.ts:import type { TransformOptions } from '@babel/core';
storybook/app/react/src/server/framework-preset-react.ts:import { TransformOptions } from '@babel/core';
```

Both of those only import types from `@babel/core` (and only in the server files, which don't seem to be publicly exposed), making it at best an optional peer dep (I think you could remove it altogether, but I haven't investigated to that effect).

## What I did

Used `peerDependenciesMeta` to mark the peer dep as optional.

## How to test

n/a

### DIFF
--- a/app/react/package.json
+++ b/app/react/package.json
@@ -67,6 +67,11 @@
     "react": "^16.8.0 || ^17.0.0",
     "react-dom": "^16.8.0 || ^17.0.0"
   },
+  "peerDependenciesMeta": {
+    "@babel/core": {
+      "optional": true
+    }
+  },
   "engines": {
     "node": ">=8.0.0"
   },


gaetanmaisse (approved):
Note to myself (or to other SB members 😛 ): 
Usage of `@babel/core` needs to be checked and cleaned across the whole monorepo. Some packages have it as regular or peer dep but aren't importing anything from it (maybe to satisfy transitive peerDep? to check) and others (as `@storybook/react`) are only importing a type, which isn't used/reexported publicly.

shilman (approved):
LGTM. @gaetanmaisse can you do the follow-on work for this? 🙏

PR #13330, opened 2020-11-29
Core: Add typescript as optional peer dependency

arcanis:
Issue: TypeScript is a peer dependency for `fork-ts-checker-webpack-plugin` and `react-docgen-typescript-plugin`, but isn't provided by neither `@storybook/core` nor `@storybook/react`, and thus generates warnings:

```
➤ YN0002: │ @storybook/core@npm:6.1.0-beta.7 [7b568] doesn't provide typescript (pa4c6c), requested by fork-ts-checker-webpack-plugin
➤ YN0002: │ @storybook/react@npm:6.1.0-beta.7 [61f66] doesn't provide typescript (p56aea), requested by react-docgen-typescript-plugin
```

## What I did

Since Yarn requires peer dependencies to be listed at [every level](https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0) of dependency chains, I've added TypeScript as an optional peer dependency of both `@storybook/core` and `@storybook/react`, so that they can forward the package to their own dependencies.

## How to test

n/a


app/react/package.json
```diff
@@ -67,6 +67,11 @@
     "react": "^16.8.0 || ^17.0.0",
     "react-dom": "^16.8.0 || ^17.0.0"
   },
+  "peerDependenciesMeta": {
+    "typescript": {
```
gaetanmaisse, line 71: Shouldn't `typescript` be listed as `peerDependencies` too? Or having this `peerDependenciesMeta` info is enough and Yarn will "do it" for us under the hood?

arcanis: Technically it would be more correct to list it, yes. However when we introduced the feature some maintainers were concerned that it would print more warnings on older package managers (for example for their users running npm 5), so we improved it by making `peerDependenciesMeta` imply a corresponding peer dependency on `*` (this way older package managers just ignore the peer dependency). So in short, it's as you prefer, we support both 🙂

gaetanmaisse (approved):

shilman (approved):
LGTM. Thanks @arcanis 🙏

Issue #9640, opened 2020-01-26
Addon-docs: Syntax Highlighter style is inconsistant


patricklafrance:
**Describe the bug**

Currently, with the SB "light" theme at least, out of the box you get:

- MDX code blocks are "light"
- Code sample of the <Preview /> block are "dark"

**Expected behavior**

- All syntax highlighted use the same style

**Screenshots**

Code block:

![image](https://user-images.githubusercontent.com/794579/73140946-756d2200-404c-11ea-9a2e-72f03246d207.png)

Code sample of the <Preview /> block:

![image](https://user-images.githubusercontent.com/794579/73140955-83bb3e00-404c-11ea-9be0-86485e0b40a9.png)

PR #13337, opened 2020-11-30
Add support for custom prism themes

GuilleDF:
Issue: #9641, #9640

For simplicity, I didn’t integrate it with the theme as suggested in the issue, but instead added a parameter and a prop for the `Source` component. Please let me know if this implementation makes sense or if you prefer the theme-based approach.

This allows users to specify the syntax highlighting theme as follows:

```tsx
import { vs } from 'react-syntax-highlighter/dist/esm/styles/prism';

export default {
  title: 'Components/My Component',
  component: MyComponent,
  parameters: {
    docs: { 
      source: {
        language: 'tsx',
        prismTheme: vs
      },
  },
};
```

## What I did

- Added a `prismTheme` parameter to modify a story's syntax highlighting
- Allow setting language as a parameter using `parameters.docs.source.language`
- Removed hardcoded `dark={true}` prop being passed to `Source` component
- Added examples for testing

## How to test

- _Is this testable with Jest or Chromatic screenshots?_ Not 100% sure what's best here - help would be appreciated!
- I've added some examples in `examples/official-storybook/stories/addon-docs/source.stories.tsx`
- _Does this need an update to the documentation?_ Yes, I'll update it :)


shilman (commented):
Hey @GuilleDF ! Thanks so much for your contribution -- looks very useful. I've marked it as "do not merge" until we've ascertained the performance implications, since we are looking to dramatically reduce our bundle size and I'm not sure how this fits in yet. We will try to figure it out over the next few weeks.

GuilleDF:
@shilman Great! Let me know if you need me to change something when the time comes!

loberbichler:
@shilman is there any update on this PR? don't see anything that would increase the bundle size here

ndelangen:
@shilman if our goal is move docs (and rendering the source/code) into manager (out of the preview) this will make that harder in the long run.

This adds configurability in the preview for rendering docs. If we add this we will have to find a replacement or drop this feature when inevitably we separate docs & preview rendering.

ndelangen:
@GuilleDF Thank you for your time, efforts and skill going into this PR. We've neglected it for quite some time, and that's totally on us. I'm sorry for this. We weren't sure we were ready to commit to this API addition.

I've looked into this PR once again, and mostly still feel the same way about adding this. In addition the SyntaxHighlighter and surrounding code have change significantly, as well as the docs-addon. It's hard to proceed with this PR.

Now that we're looking into a new revision of docs and working towards 7.0 we can make some breaking changes.
Which means we might want to implement the feature this PR does, but very likely in a different way.

I'll ping @shilman on this as well as @tmeasday to maybe discuss this next time during one of our roadmap meeting (which you are welcome to join btw!) to see what are plans are to proceed with this feature request.

For now I'll close this PR, but as said, a lot has changed, and will quite possibly change a lot more.
I'm sorry for closing your PR that you worked hard on.

GuilleDF:
@ndelangen I completely understand, thank you for the explanation!

PR #13344, opened 2020-11-30
Core: Fix `--static-dir` with absolute path on Windows

ghengeveld:
Issue: #13336

## What I did

Primarily this fixes the handling of the `-s` (`--static-dir`) flag for Windows users. This flag accepts a string in the shape of `<dirname>:<endpoint>`. Currently it fails to handle absolute paths on Windows: for a value like `C:\path` it would assume `C` to be the dirname and `\path` the endpoint. That bug is addressed here.

Besides fixing the bug at hand, I've extracted this piece of logic into a util, because it previously existed in two separate places. The util is now under unit test and the surrounding code has been cleaned up with some nicer logging. I've also replaced `shelljs` with `copy` from `fs-extra`. One less dependency 📉 

## How to test


shilman (changes requested):
small thing: `.test.js` => `.test.ts`

ghengeveld:
> small thing: `.test.js` => `.test.ts`

Done

shilman (approved):
LGTM - nice tests 💯

PR #13355, opened 2020-12-01
UI: Fix display of custom brand image

kylegach:
Issue: #13313

When supplying `theme.brandImage`, the image renders very small, effectively invisible:

Screenshot of misrendered element

## What I did

The ([Brand](https://github.com/storybookjs/storybook/blob/next/lib/ui/src/components/sidebar/Heading.tsx#L45)) (which renders the selected element in the screenshot, above) is inside a flex container ([BrandArea](https://github.com/storybookjs/storybook/blob/next/lib/ui/src/components/sidebar/Heading.tsx#L17)), so I applied `flex: 1 1 auto` (and removed the now redundant `width: auto`) to BrandArea's immediate descendants to fill the available width. It was previously (implicitly) using the default of `flex: 0 1 auto`.

## How to test

- Is this testable with Jest or Chromatic screenshots? Not sure, probably

### DIFF
--- a/lib/ui/src/components/sidebar/Brand.tsx
+++ b/lib/ui/src/components/sidebar/Brand.tsx
@@ -31,41 +31,28 @@ export const LogoLink = styled.a(({ theme }) => ({
   },
 }));
 
-export const Brand = withTheme(
-  ({
-    theme: {
-      brand: { title = 'Storybook', url = './', image },
-    },
-  }) => {
-    const targetValue = url === './' ? '' : '_blank';
-    if (image === undefined && url === null) {
-      return <StorybookLogoStyled alt={title} />;
-    }
-    if (image === undefined && url) {
-      return (
-        <LogoLink title={title} href={url} target={targetValue}>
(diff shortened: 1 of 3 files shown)


ghengeveld:
👏 Great work, I didn't think it was possible. I added some stories to test the layout with an unsized SVG (i.e. one that doesn't have a width/height prop). Also I reduced some margin so the logo can be a bit bigger.

ghengeveld (approved):
LGTM. Do check the Chromatic tests.

kylegach:
Thanks for adding the stories, Gert!

Looks like we have one failing story, Standard No Link: https://www.chromatic.com/component?appId=5a375b97f4b14f0020b0cda3&name=UI%2FSidebar%2FHeading&mode=interactive&buildNumber=20261&specName=Standard+No+Link

Which seems related to whether or not an anchor is rendered around the image. I'll take a look asap.

ghengeveld:
@kylegach I fixed the misaligning image and also refactored the Brand component because it was pretty complicated.

PR #13370, opened 2020-12-03
UI: Fix brandImage visibility for unsized SVGs

ghengeveld:
Issue: #13313

## What I did

This restores the original `width: 100%` on the anchor tag, so that it won't collapse when it contains an unsized SVG. Unfortunately this means the link will be wider than the actual image, but this is a tradeoff.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, I added some.


ghengeveld:
Closing this in favor of #13355

PR #13397, opened 2020-12-08
Typo in addon-controls documentation

pascaloliv:
Issue: There is a number() typo instead of text()

## What I did
Replace the number() function call by text() in addon-controls documentation


shilman (approved):
👍 good catch. thanks!

PR #13411, opened 2020-12-09
Provide more precise usage of `npx sb extract`

philsawicki:
## Issue:
Documentation about `npx sb extract` can be misleading when attempting to provide a custom build directory.

### Details
From the current documentation, one is given to understand that `npx sb extract` supports 2 optional **directories** when attempting to build the _stories.json_ file from a custom build folder:
> If required, you can also add additional arguments to this command. You can use the following to generate the `stories.json` file to a custom directory:
>
> ```shell
> npx sb extract my-built-storybook-directory my-other-directory
> ```
>
> Once the command executes it will look for a built Storybook in the `my-built-storybook-directory` and create the file in `my-other-directory`.

However, providing directories from a previously-built Storybook output produces the following error:
```shell
# Example: npx sb extract build build

[Error: EISDIR: illegal operation on a directory, open 'build'] {
  errno: -21,
  code: 'EISDIR',
  syscall: 'open',
  path: 'build'
}
```

The reason for this error is the CLI script backing the feature attempts to open the second argument in order to write the `stories.json` file, which fails due to it being a directory. Indeed, from the source of the CLI script, it appears the `output` is expected to be the name of a JSON file (`stories.json` by default):
https://github.com/storybookjs/storybook/blob/880d7ee001f189ac24ca2e6aeea39c13b1a5bca7/lib/cli/src/generate.ts#L82-L90

## What I did

Updated the documentation to make it more precise that, when provided, the second optional argument of `npx sb extract` should be the path to a file.

Another option to address the issue, and perhaps a consideration for the future, would be to additionally offer a message to the User at runtime if the file cannot be created (file vs. directory path, permission issue, etc.).

## How to test

- Is this testable with Jest or Chromatic screenshots? **No** (documentation only)
- Does this need a new example in the kitchen sink apps? **No** (documentation only)
- Does this need an update to the documentation? **Yes** (current PR provides this very update)


philsawicki:
This is a great software project! Thank you for making development & collaboration between teams and disciplines so much easier!

Please let me know if there is any additional improvements I can provide -- maybe there could be better wordsmithing for the issue I am attempting to resolve.

ndelangen (approved):

philsawicki:
@ndelangen @shilman Thank you for considering the change!

Glad to have this change included in the next version. Hopefully this will benefit others! 😃 
Once again, thank you for the fantastic tool!

PR #13416, opened 2020-12-10
Add section for installing with React apps built without CRA

jwlms:
Issue: Playbook doesn't work in React projects without the ReactDom package listed as a dependency.

## What I did

In order to provide information on how to integrate the Playbook
module with a non-Create React App application I updated the 
React `README.md` document with the required steps.

It shouldn't be necessary to require `react-dom` as it isn't 
necessarily needed for a React project, but in most cases
will be used.

## How to test

This PR should not need to be tested, but the bug can be 
easily verified by spinning up a React project in webpack 
without ReactDom and then add Playbook to that project.

It will install but not run and throws errors about missing 
ReactDom from within the `playbook` project files.

## References
#9216

### DIFF
--- a/app/react/README.md
+++ b/app/react/README.md
@@ -27,6 +27,14 @@ Here are some featured storybooks that you can reference to see how Storybook wo
 - [Demo of React Dates](http://airbnb.io/react-dates/) - [source](https://github.com/airbnb/react-dates)
 - [Demo of React Native Web](https://necolas.github.io/react-native-web/docs/) - [source](https://github.com/necolas/react-native-web)
 
+## React Applications
+
+For React apps that are not built with Create React App, ReactDom is a required for Storybook to work with your project. Add it to your devDependencies if it is not a dependency in your project:
+
+```bash
+npm install react-dom --save-dev
+```
+
 ## Create React App
 
 Support for [Create React App](https://create-react-app.dev/) is handled by [`@storybook/preset-create-react-app`](https://github.com/storybookjs/presets/tree/master/packages/preset-create-react-app).
--- a/app/react/README.md
+++ b/app/react/README.md
@@ -27,19 +27,13 @@ Here are some featured storybooks that you can reference to see how Storybook wo
 - [Demo of React Dates](http://airbnb.io/react-dates/) - [source](https://github.com/airbnb/react-dates)
 - [Demo of React Native Web](https://necolas.github.io/react-native-web/docs/) - [source](https://github.com/necolas/react-native-web)
 
-## React Applications
-
-For React apps that are not built with Create React App, ReactDom is a required for Storybook to work with your project. Add it to your devDependencies if it is not a dependency in your project:
-
-```bash
-npm install react-dom --save-dev
-```
-
 ## Create React App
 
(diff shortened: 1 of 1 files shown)


app/react/README.md
```diff
@@ -27,6 +27,14 @@ Here are some featured storybooks that you can reference to see how Storybook wo
 - [Demo of React Dates](http://airbnb.io/react-dates/) - [source](https://github.com/airbnb/react-dates)
 - [Demo of React Native Web](https://necolas.github.io/react-native-web/docs/) - [source](https://github.com/necolas/react-native-web)
 
+## React Applications
+
+For React apps that are not built with Create React App, ReactDom is a required for Storybook to work with your project. Add it to your devDependencies if it is not a dependency in your project:
+
+```bash
+npm install react-dom --save-dev
+```
```
jonniebigodes, line 36: @jwlms I'm more ok with enhancing the documentation to help you and the other Storybook users. But here's the thing, with CRA it bootstraps all the necessary dependencies, and for instance ReactDom.

With non-Cra apps, as with other React apps, ReactDom is a requirement that you should factor in when initializing your project. Storybook has a peer dependency on it and warns you about it once you're adding it to the project.

With that in mind, we have a bit of room for improvement. We could probably go with:
" If you're working on an app that was initialized manually (i.e., without the use of Create React App), ensure that your app has [react-dom](https://www.npmjs.com/package/react-dom) included as a dependency. Failing to do so can lead to unforeseen issues with Storybook and your project."

Let me know and we'll go from here.

Stay safe

jonniebigodes (changes requested):
@jwlms left a small item that needs attention. Let me know once you've addressed it and let me know.

app/react/README.md
jwlms: I understand that CRA bootstraps all the necessary dependencies — and for my project I should have added `react-dom` as a peer dependency. However, it's not required as a dependency for my project as I'm only exporting components as a library for others to render and using storybook to pull these into for documentation purposes.

Anyway I went ahead and updated the copy to your recommended language, but I think the likelihood someone else runs into this is low so no hard feelings if you want to close it out!

Cheers

jonniebigodes: @jwlms thank you very much for addressing the feedback and sorry for not reporting back sooner. I understand your use case and I'm ok with this. As one thing this industry has taught me is that the "devil is in the details" (pardon the bad pun) and sooner or later someone will run into the same issue and to avoid having him/her open a new issue and wait for feedback from the maintainers or the community we have it covered and documented.

jonniebigodes:
Self merging this. @jwlms thank you very much once more for the time and effort you've put into this and for adjusting the pr accordingly. We really appreciate it. 

If you run into any other issues with the documentation, reach out to us so that we can help you and the rest of the community as well.

Looking forward to your next contribution.

Stay safe

app/react/README.md
jwlms: @jonniebigodes - thanks! No worries about the slow reply, my email has been filtering github notices as spam so I totally missed this one, but appreciate you merging it in nonetheless.

PR #13423, opened 2020-12-10
UI: Fix null ref in sidebar

ghengeveld:
Issue: -

## What I did

Ensure we have a ref value before try to invoke querySelector on it.

## How to test


shilman (approved):
LGTM but are there any tests? Repros? When does this happen?

ghengeveld:
React refs are always null on first render (unless given a default, which is uncommon for DOM element refs). I just never noticed the bug until we started showing runtime errors in the terminal.

PR #13424, opened 2020-12-10
Perf: Replace react-hotkeys with useEffect keybinding

ghengeveld:
Issue: -

## What I did

This gets rid of react-hotkeys, a 310kb dependency. It was used only once, for a global Escape key listener. That's not even react-hotkeys' intended use case, because it's mainly useful for local (focus-based) events.

I replaced it with a simple `document.addEventListener` wrapped in `useEffect`, just like the ones in the sidebar.

## How to test

Go to one of the settings pages (e.g. "About your Storybook") and press the Escape key. The page should close and you should return to the main preview UI.


shilman (approved):
Beautiful 😍

PR #13433, opened 2020-12-11
Addon: Webpack Bundle Analyzer

ndelangen:
telescoped off of https://github.com/storybookjs/storybook/pull/13013

## What I did

I added an addon that starts webpack-bundle-analyser and then storybook iframe's the running server in dev mode.
In static build mode, we have the bundle-analyser generate a static file and iframe that.


ndelangen:
see discussion here:
https://github.com/storybookjs/storybook/pull/13425

addons/bundle-analyzer/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "webpack"
+  ],
+  "homepage": "https://github.com/storybookjs/storybook/tree/master/addons/bundle-analyzer",
+  "bugs": {
+    "url": "https://github.com/storybookjs/storybook/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/storybookjs/storybook.git",
+    "directory": "addons/jest"
```
yannbf, line 18:
```suggestion
    "directory": "addons/bundle-analyzer"
```

addons/bundle-analyzer/preset.js
```diff
@@ -0,0 +1,33 @@
...
+const analyzerPort = detectFreePort();
+
+function managerEntries(entry = [], options) {
+  return [...entry, require.resolve('./dist/esm/register')];
+}
+
+async function managerGlobals(config, { configType }) {
+  return {
+    ...config,
+    ANALYSER_ADDRESS:
```
yannbf, line 12:
```suggestion
    ANALYZER_ADDRESS:
```

addons/bundle-analyzer/src/components/Panel.tsx
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable react/default-props-match-prop-types */
+import { ANALYSER_ADDRESS } from 'global';
```
yannbf, line 2:
```suggestion
import { ANALYZER_ADDRESS } from 'global';
```

addons/bundle-analyzer/src/components/Panel.tsx
```diff
@@ -0,0 +1,24 @@
...
+interface Props {
+  active: boolean;
+}
+
+const Panel = ({ active }: Props) =>
+  active ? (
+    <iframe
+      style={{ margin: 0, border: 0, width: '100%' }}
+      title="bundle-analyzer"
+      src={ANALYSER_ADDRESS}
```
yannbf, line 16:
```suggestion
      src={ANALYZER_ADDRESS}
```

addons/bundle-analyzer/src/shared.ts
```diff
@@ -0,0 +1,4 @@
+// addons, panels and events get unique names using a prefix
+export const PARAM_KEY = 'bundle-analyzer';
+export const ADDON_ID = 'storybookjs/bundele-analyzer';
```
yannbf, line 3:
```suggestion
export const ADDON_ID = 'storybookjs/bundle-analyzer';
```

lib/core/src/server/manager/manager-webpack.config.ts
```diff
@@ -88,6 +89,7 @@ export default async ({
             RELEASE_NOTES_DATA: JSON.stringify(releaseNotesData),
             DOCS_MODE: docsMode, // global docs mode
             PREVIEW_URL: previewUrl, // global preview URL
+            ...globals,
```
yannbf, line 92: Does this open more opportunities for other addons that couldn't set global configuration in their presets? If so, would it be a good idea to document this?

addons/bundle-analyzer/src/components/Panel.tsx
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable react/default-props-match-prop-types */
+import { ANALYSER_ADDRESS } from 'global';
+import React, { Fragment } from 'react';
+// import { styled } from '@storybook/theming';
```
yannbf, line 4: I guess either remove this line or create a `styledIframe` using it?

addons/bundle-analyzer/README.md
```diff
@@ -0,0 +1,3 @@
+# Webpack Bundle Analyzer for Storybook
+
+Exposes Webpack Bundle Analyzer results in a Storybook addon panel.
```
yannbf, line 3: What do you think of adding an explanation like this? So that people don't necessarily need to find the webpack bundle analyzer website to understand it better. I got the text from their repo.

```suggestion
Exposes Webpack Bundle Analyzer results in a Storybook addon panel.

This addon will help you:

- Realize what's really inside your application bundle
- Find out what modules make up the most of its size
- Find modules that got there by mistake
- Optimize it!

```

yannbf:
Awesome stuff!!

One question: In development, does this addon generate a report file? How would people differentiate the report that they're doing locally from the one that should be versioned?

Assuming that it does generate a report, see the situation: 
1 - I run Storybook locally and it generates a report. I stop running Storybook.
2 - I make changes in my components, I add a new lib, then I make a PR (but I didn't run Storybook again). 

The report will be wrong then, right? Is there a way to prevent such situation to happen? Is it worth mentioning in the docs?

PR #13447, opened 2020-12-14
Riot: migrate to TS

gaetanmaisse:
Issue: Part of #5030

## What I did

Fastly migrate `@storybook/riot` to TS as I want to close the big TS Migration Issue before moving to improve types of `@storybook/core`.

## How to test

- CI should be 🟢 , we rely only on examples as there is no E2E for Riot (it looks like there are some issues with Riot 4 + SB 🤷🏻 )


app/riot/src/client/preview/index.ts
```diff
@@ -1,6 +1,7 @@
 import { start } from '@storybook/core/client';
 
 import './globals';
+// @ts-ignore
```
gaetanmaisse, line 4: There are some miscellaneous things with `tag2` across many files, it's imported from `riot` (but isn't available is the type), sometimes it's just stored in an unused `const` but if you remove them everything breaks 🧙🏻

app/riot/src/client/preview/render-riot.test.ts
```diff
@@ -23,18 +24,21 @@ beforeEach(() => {
 
 describe('render a riot element', () => {
   it('should not work with nothing', () => {
+    // @ts-ignore
```
gaetanmaisse, line 27: Same here... `render`  should only have 1 argument but in every test, there is a magic `context` given too.

app/riot/src/client/preview/render-riot.test.ts
```diff
@@ -23,18 +23,22 @@ beforeEach(() => {
 
 describe('render a riot element', () => {
   it('should not work with nothing', () => {
+    // `render` should only have 1 argument but in every test, there is a magic context given too… 🤷
```
ThibaudAV, line 26: I think the 2nd argument can be deleted without any problem.

I think the test has been missed with this commit 
https://github.com/storybookjs/storybook/commit/f555ce0d6956a8fcb9cfd7d9c0fbe899de401bdc#diff-dc99c8ceca53cf99aca341cfaa97633689b0d7edbdaa26631026779cd0e723dc

ThibaudAV (approved):
LGTM

app/riot/src/client/preview/render-riot.test.ts
gaetanmaisse: Thanks, you're right, I thought I tried to remove them and everything was OK but I just retried, all tests are 🟢 . 🙇🏻

PR #13448, opened 2020-12-14
Marionette: migrate to TS

gaetanmaisse:
Issue: Part of #5030

## What I did

Fastly migrate `@storybook/marionette` to TS as I want to close the big TS Migration Issue before moving to improve types of `@storybook/core`.

## How to test

- CI should be 🟢 , we rely only on examples as there is no E2E for Marionette.


app/marionette/src/client/preview/element_check.ts
```diff
@@ -8,10 +8,10 @@ const allMarionetteViewConstructors = [
 ];
 const viewConstructorsSupportedByMarionette = allMarionetteViewConstructors
   .filter((constructorName) => constructorName in Marionette)
-  .map((constructorName) => Marionette[constructorName]);
+  .map((constructorName) => (Marionette as any)[constructorName]);
```
gaetanmaisse, line 11: I tried to find the correct type but it's too 🤯  for a non-marionette user.

app/marionette/src/client/preview/index.ts
```diff
@@ -15,8 +15,9 @@ export const {
 } = clientApi;
 
 const framework = 'marionette';
-export const storiesOf = (...args) => clientApi.storiesOf(...args).addParameters({ framework });
-export const load = (...args) => coreLoad(...args, framework);
+export const storiesOf = (...args: any) =>
+  clientApi.storiesOf(...args).addParameters({ framework });
+export const load = (...args: any) => coreLoad(...args, framework);
```
gaetanmaisse, line 20: Will be properly type as soon as I will rework some `@storybook/core` types

app/marionette/src/server/framework-preset-marionette.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { Configuration } from 'webpack';
+
+export function webpack(config: Configuration) {
+  return config;
+}
```
gaetanmaisse, line 6: Not sure we need this but I didn't want to make changes to production code other than a simple TS migration.

gaetanmaisse:
I think we are good to go with this one @ndelangen @shilman, not sure the diff in chromatic are relevant/related to this PR.

ThibaudAV (approved):
LGTM

IMHO take time to add robust typescript is interesting if users (+framework) are also adepts. if they are users of "js" this is not (or less) necessary 🙊

PR #13449, opened 2020-12-14
Marko: migrate to TS

gaetanmaisse:
Issue: Part of #5030

## What I did

Fastly migrate `@storybook/marko` to TS as I want to close the big TS Migration Issue before moving to improve types of `@storybook/core`.

## How to test

- CI should be 🟢 , we rely only on examples as there is no E2E for Marko. According to the comment in e2e-config: 
> // TODO: not working on start-storybook
//  - Marko CLI is failing with Node 12 and looks to work with Node 10
//  - Demo components of @storybook/marko must be updated
//  - Marko Story templates of @storybook/cli must be updated


ThibaudAV (approved):
We discover new ones every day 🤣

PR #13451, opened 2020-12-14
Update addon descriptions to match frontpage

winkerVSbecks:
## What I did

Updated descriptions for core addons to match frontpage, in preparation of the catalog launch

### DIFF
--- a/addons/a11y/package.json
+++ b/addons/a11y/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@storybook/addon-a11y",
   "version": "6.2.0-alpha.7",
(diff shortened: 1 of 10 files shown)


shilman (changes requested):
LGTM but can you revert the changes to the `typesVersions`. I think this due to an outdated version of the package.json sorting installed on your machine and it will just flip flop back and forth when somebody else edits the files next, generating more unnecessary changes. Thanks!

domyen (approved):

shilman (approved):
💯

PR #13466, opened 2020-12-16
Svelte: Statically load docgen info for svelte components

### LINKED ISSUE #13247: Storybook v6.1.5 with Svelte fails to start
**Describe the bug**

Storybook v6.1.5 doesn't work with Svelte if a story use a component with a mandatory property.

**To Reproduce**

Create a Story with a Svelte component which assume a property is set.

For example :

```
<script>
  export let myFunction;
  myFunction();
</script>
```

It's related to this PR by @BlackFenix2  : https://github.com/storybookjs/storybook/pull/12347
In this PR, every component are instanciated without properties, in order to extract the generated documentation here : https://github.com/storybookjs/storybook/blob/next/addons/docs/src/frameworks/svelte/extractArgTypes.ts#L37
It doesn't work if the component has mandatory properties.

More over, it assumes every component have a docgen here : https://github.com/storybookjs/storybook/blob/next/addons/docs/src/frameworks/svelte/extractArgTypes.ts#L48
This not always the case (if a component doesn't have any property, then docgen will be undefined but I am not 100% sure about that). If a component have docgen is not defined, then the Storybook doesn't starts.

j3rem1e:
Issue: #13247

## What I did

Every svelte component has now a static __docgen property.
I have upgraded the svelte kitchen sink to use controls in order to test it

## How to test

### DIFF
--- a/addons/docs/src/frameworks/svelte/extractArgTypes.ts
+++ b/addons/docs/src/frameworks/svelte/extractArgTypes.ts
@@ -35,10 +35,8 @@ type Docgen = {
 
 export const extractArgTypes: ArgTypesExtractor = (component) => {
   try {
-    // eslint-disable-next-line new-cap
-    const comp: ComponentWithDocgen = new component({ props: {} });
     // eslint-disable-next-line no-underscore-dangle
-    const docgen = comp.__docgen;
+    const docgen = component.__docgen;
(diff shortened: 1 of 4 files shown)


shilman (approved):
great work @j3rem1e. much better! 💯

PR #13501, opened 2020-12-22
Core: Fix typing of dev CLI options

gaetanmaisse:
## What I did

The mistake was spotted in https://github.com/storybookjs/storybook/commit/13947884387fec650bab8ebdcd7b11a00cf01607#r45325210

I fixed the type by setting `host` to `string`

## How to test

- Build should be ok


shilman (approved):
💯

PR #13508, opened 2020-12-22
faster levenshtein

ka-weihe:
## What I did
Changed the package used for levenshtein distance. [leven](https://github.com/sindresorhus/leven) is one of the slowest levenshtein packages out there and the author is trying to deceive people with shitty benchmarks where he is only comparing it slower packages. Please consider using  [fastest-levenshtein]( https://github.com/ka-weihe/fastest-levenshtein) instead which is 10x faster and much higher quality. 

## How to test

no
no
no


merceyz:
> and the author is trying to deceive people with shitty benchmarks where he is only comparing it slower packages

To be fair it was written 2 years ago so it was probably true at that point in time and `fastest-levenshtein` didn't exist back then so can't really compare against something that didn't exist

ka-weihe:
The reason I write that the benchmarks are "shitty" is because his benchmarks are a special case where his implementation skips the main part of the algorithm completely - you can have a look yourself. That is the same as benchmarking a big heavy function with a special case where it returns from an if-statement right in the beginning. That is just deceiving. He is also only comparing it to slower packages and he is ignoring PRs to update this.

gaetanmaisse:
Thanks for your contribution but after reviewing it I will close the PR for the following reasons:
 - `fastest-levenstein` is way bigger than `leven`, more than 2x, which is a big deal in our case as the CLI is mainly used via `npx` or `yarn dlx`
 - The fact that it's faster isn't relevant in the Storybook CLI context as we are using it with ~10 words and only if the user's command is invalid
 - It adds a new dependency in the already huge Storybook monorepo, and as `leven` is needed by another package it will make us pay the cost of `leven` + `fastest-levenstein` (as you can see in the `yarn.lock` update)

PR #13509, opened 2020-12-23
fix(docs): correct react viewport snippet's import

lwyj123:
Issue:
The `MINIMAL_VIEWPORTS` is no longer contain 'iphone6' and 'iphonex' in [code](https://github.com/storybookjs/storybook/blob/4f5ab9fe9e590da7b841ec37cb1bed8d6327ea4b/addons/viewport/src/defaults.ts). 

## What I did
1. Correct the value of `viewports`
2. add `import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport';` for more clear example

## How to test

Docs fix


shilman (approved):
LGTM. thanks!

PR #13518, opened 2020-12-26
Typescript: Fix qs import in @storybook/client-api

ryanml:
Issue: storybookjs/storybook#13496

## What I did
Explicitly importing `parse` from `qs` in `lib/client-api/src/queryparams.ts`

## How to test

Compile a typescript project using storybook as a dependency at this branch, with `allowSyntheticDefaultImports` set to false. You should not run in to the error described in the issue.


shilman (approved):
LGTM 👍

PR #13520, opened 2020-12-26
CLI: fix sb init prompt when framework type is undetected

yannbf:
Issue: When inquirer was replaced with prompts in #13226, `sb init` had a bug that wouldn't display a list of installable projects when no framework was detected.

## What I did

Set the correct prompt type in the prompt function.

Result of `sb init` in @next:
latest-version

Result of `sb init` with this PR:
with-fix

## How to test

1 - `yarn build cli`
1 - create a folder with just a package.json like with `npm init`
2 - run `node ../../storybook/lib/cli/dist/esm/generate init` (adjust the path to where you folder will be)


yannbf:
@shilman  @ndelangen after the work done in #13013 I believe the CLI can't install Storybook in typescript projects anymore. Would any of you care to help? Maybe the fix would suit in this PR?

The ESM/CJS build result won't contain the typescript files from the templates. When running `sb init` in a typescript based project (e.g. Angular, Aurelia, React + TS, etc.), the CLI will try to copy files from an [empty folder](https://github.com/storybookjs/storybook/blob/e8e37ac7ad87836c509ded3f41175d931330ebe5/lib/cli/src/helpers.ts#L216), and as a result the boilerplate will only include the files in `frameworks/common`:


What I mean by copying an empty folder you can see with the result of the esm build in the cli project:

yannbf:
@gaetanmaisse you're the best ❤️

scripts/run-e2e.ts
```diff
@@ -127,7 +127,7 @@ const initStorybook = async ({ cwd, autoDetect = true, name }: Options) => {
     const type = autoDetect ? '' : `--type ${name}`;
 
     const sbCLICommand = useLocalSbCli
-      ? 'node ../../storybook/lib/cli/dist/generate'
+      ? 'node ../../storybook/lib/cli/dist/esm/generate'
```
gaetanmaisse, line 130: I would have used `cjs` to ensure compatibility with Node10+ but I'm not sure about that.

🤔 I dived into `lib/cli/dist/esm` and it looks there are `require('...')`, is it normal @ndelangen? I thought there would be only `import` statements.

gaetanmaisse: Tested locally with Node 10.x, everything was ok.

gaetanmaisse (approved):

PR #13521, opened 2020-12-27
CLI: Ensure --host option changes the network host

ryanml:
Issue: storybookjs/storybook#12865

## What I did

- Made it so that the host passed via cl args modifies the network host, not localhost
- Added a new test file `dev-server.test.ts`

## How to test
The test plan is defined in the issue

### DIFF
--- a/lib/core/src/server/dev-server.ts
+++ b/lib/core/src/server/dev-server.ts
@@ -356,6 +356,13 @@ const startPreview = async ({
   return { previewStats, previewTotalTime: process.hrtime(startTime) };
 };
 
+export function getServerAddresses(port: number, host: string, proto: string) {
+  return {
+    address: `${proto}://localhost:${port}/`,
+    networkAddress: `${proto}://${host || ip.address()}:${port}/`,
+  };
+}
+
 export async function storybookDevServer(options: any) {
   const app = express();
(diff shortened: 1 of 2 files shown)


lib/core/src/server/dev-server.ts
```diff
@@ -356,6 +356,13 @@ const startPreview = async ({
   return { previewStats, previewTotalTime: process.hrtime(startTime) };
 };
 
+export function getServerAddresses(port: number, host: string, proto: string) {
```
ryanml, line 359: Moving this logic to its own function is useful for unit tests

lib/core/src/server/dev-server.ts
```diff
@@ -414,7 +420,7 @@ export async function storybookDevServer(options: any) {
   ]);
 
   // TODO #13083 Remove this when compiling the preview is fast enough
-  if (!options.ci) openInBrowser(address);
+  if (!options.ci) openInBrowser(networkAddress);
```
ryanml, line 423: @yannbf - I think it's the case now that we'll want the browser to open to `networkAddress`, given that's what the user is configuring with the `--host` option

shilman (approved):
💯

PR #13533, opened 2020-12-29
Addon-docs: Use theme text color header anchors

etlovett:
Issue: None that I could find.

## What I did

Without this change, the anchor images in Storybook Docs always render in `#000`, which means they can be very hard to see in dark themes. This PR changes them to render using the theme's text color, which is a reasonable guarantee of sufficient contrast.

Some before-and-after screenshots from `examples/cra-kitchen-sink`, with theme changes as indicated:

|                     | Before | After |
|---------------------|--------|-------|
| Default light theme | before-default-light | after-default-light |
| Custom light theme  | before-custom-light | after-custom-light |
| Default dark theme  | before-default-dark | after-default-dark |
| Custom dark theme 1 | before-custom-dark-1 | after-custom-dark-1 |
| Custom dark theme 2 | before-custom-dark-2 | after-custom-dark-2 |

## How to test

- Is this testable with Jest or Chromatic screenshots? I don't think so, since the icon only appears on hover, but I could be wrong. I'm happy to add a test for this if y'all would like, but I'll need some pointers since I was unable to run the image snapshot tests following the instructions at https://github.com/storybookjs/storybook/blob/next/CONTRIBUTING.md#cra-kitchen-sink---image-snapshots-using-storyshots.

### DIFF
--- a/addons/docs/src/blocks/mdx.tsx
+++ b/addons/docs/src/blocks/mdx.tsx
@@ -132,6 +132,8 @@ const OcticonAnchor = styled.a(() => ({
   float: 'left',
   paddingRight: '4px',
   marginLeft: '-20px',
+  // Allow the theme's text color to override the default link color.
+  color: 'inherit',
 }));
 
 interface HeaderWithOcticonAnchorProps {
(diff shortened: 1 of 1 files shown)


addons/docs/src/blocks/mdx.tsx
```diff
@@ -164,7 +166,14 @@ const HeaderWithOcticonAnchor: FC<HeaderWithOcticonAnchorProps> = ({
...
         }}
       >
-        <svg viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true">
+        <svg
+          viewBox="0 0 16 16"
+          version="1.1"
+          width="16"
+          height="16"
+          aria-hidden="true"
+          fill="currentColor"
```
etlovett, line 175: This prop is new; the rest is just linter formatting.

shilman (approved):
LGTM! Thanks so much for the fix! 💯

etlovett:
No problem, @shilman; thanks for accepting the fix and, more importantly, for the phenomenal tool!

PR #13534, opened 2020-12-29
Storyshots: Add `storyFilterFunction` option

leipert:
Issue: none

## What I did

Exposing a filter function is more flexible for users and they can
finetune the matching to their liking. Initially I thought it could
deprecate the `storyNameRegex` and `storyKindRegex`, because we could
just create a default filter function which applies those two.

Our use case for this filter function is splitting the story shots based
on an arbitrary count of machines doing the tests in CI. But it will
also make it possible for people just to implement their test matching
to their liking.

## How to test

-   Is this testable with Jest or Chromatic screenshots?
      
    > we could adjust the tests in `storyshots-puppeteer`
-   Does this need a new example in the kitchen sink apps?

    > I don't know
-   Does this need an update to the documentation?

    > Added to the README.


leipert:
Hey folks! This is my first contribution 🕺  I tried to keep the change minimal, we certainly could deprecate the existing filters in favor of this one, or could adjust certain tests/examples, e.g. (`examples/official-storybook/storyshots-puppeteer/axe-tests.runner.js`) to use it.

If you need any adjustments, just let me know 💃

shilman (changes requested):
This looks great -- thanks for putting it together and congratulations on your first contribution! 🎉 

Please create a minimal test case for this new option. I'd suggest taking one of the existing kitchen sink apps and exclude a single snapshot using a filter function. We'd expect to see the snapshot file updated with one less result. WDYT?

leipert:
@shilman Sorry for the long wait. I am a bit lost as to _which_ kitchen sink app I should change... 

So I just replaced an existing `storyKindRegex` with the same logic as a filter function in: `examples/official-storybook/storyshots-puppeteer/axe-tests.runner.js`

toughdeveloper:
Just stumbled upon this after suggesting a very similar feature - https://github.com/storybookjs/storybook/discussions/16932

Is there any appetite for this to be revived?

PR #13535, opened 2020-12-29
Addon-jest: Infer parameter from story filename if not provided

dominictwlee:
Issue: #12084

## What I did
Added functionality to infer jest parameter from file name if not provided explicitly.

## How to test

### DIFF
--- a/addons/jest/README.md
+++ b/addons/jest/README.md
@@ -128,6 +128,9 @@ defaultView.parameters = {
 };
 ```
 
+The jest parameter will default to inferring from your story file name if not provided. For example, if your story file is `MyComponent.stories.js`,
+then "MyComponent" will be used to find your test file results.
+
 ### Disabling
 
 You can disable the addon for a single story by setting the `jest` parameter to `{disable: true}`:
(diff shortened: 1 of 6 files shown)


shilman (changes requested):
Thanks for building this feature @dominictwlee !

Unfortunately, it does not seem to be working in the static build of Storybook:

Addons___Jest_-_With_Inferred_Tests_⋅_Storybook

The issue is that the file path is not available in the static build. We can try to fix that underlying issue, which I suspect is hard, although I don't know the full background. More likely, we should fix the bug in the current code and add documentation that this doesn't work in production builds.

WDYT?

dominictwlee:
> Thanks for building this feature @dominictwlee !
> 
> Unfortunately, it does not seem to be working in the static build of Storybook:
> 
> Addons___Jest_-_With_Inferred_Tests_⋅_Storybook
> 
> The issue is that the file path is not available in the static build. We can try to fix that underlying issue, which I suspect is hard, although I don't know the full background. More likely, we should fix the bug in the current code and add documentation that this doesn't work in production builds.
> 
> WDYT?

Oh odd, thought the falsy check would fix that.

I had a closer look and it seems that the `fileName` property ends up being a number, something like `1264`.
I'll do the extra check for that. Would you also like me to update the interface for Parameter? (currently it's defined as `fileName?: string`)

dominictwlee:
I'll maybe leave the Parameter type for a separate PR (if needed), since it could potentially break anything else that relies on it.

shilman (approved):
Excellent! 💯

PR #13544, opened 2021-01-02
Addon-controls: Files control

dominictwlee:
Issue: #13004

## What I did
Port over FilesControl component from knobs. Used `createObjectUrl()` instead of `FileReader.readAsDataURL()` as the latter was  slow and causing issues when dealing with larger files.

## How to test

### DIFF
--- a/docs/essentials/controls.md
+++ b/docs/essentials/controls.md
@@ -119,7 +119,6 @@ As they can be complex cases:
 
 <!-- prettier-ignore-end -->
 
-
 Or even with certain types of elements, such as icons:
 
 <!-- prettier-ignore-start -->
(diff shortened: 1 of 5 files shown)


shilman (approved):
Awesome -- looks great!!! 💯

rnarcos:
@shilman, has this been planned to work on the "full-screen" mode? Because I can't seem to add a file control URL to the iframe's `args` URL to force a specific file to be used from there. Am I missing something?

```javascript
export default {
  // ... commented for brevity
  argTypes: {
    myJSON: {
      control: {
        type: 'file',
        accept: '.json',
    },
  }
}

const MyTemplate = ({ myURL }) => {
  console.log(myURL);
  
  return null;
};

const MyStory = MyTemplate.bind({});
MyStory.args = {
  myURL: ['my-default-url/something.json'], // Array to be compatible with the File control API
};
```

On my browser, if I open **_MyStory_** it will use its args as default, but I can't seem to override this value from the URL:

Here's what my URL looks like: `http://localhost:6006/?path=/story/my-story&args=myURL:foooo` but all I can see logged when my story loads is: `my-default-url/something.json`.

Should I open a bug for this?

PR #13558, opened 2021-01-05
Angular: Use Nx function to read non-angularCli configs

mandarini:
Issue: #12232

## What I did
This is a fix for a behavior described [here](https://github.com/nrwl/nx/issues/4305) and [here](https://github.com/storybookjs/storybook/issues/12232).

A new function is created in the `@nrwl/workspace` package, which will take care of reading/parsing the workspace configuration, and returning it to Storybook in the `angularCli` format, which it expects.

The new functionality will work after this [Nx PR](https://github.com/nrwl/nx/pull/4440) is successfully merged.


shilman (approved):
LGTM @mandarini and thanks so much for the contribution. i'm OK with it except for the comments above. will get a few more eyes on it and then merge/release if nobody objects. after you've tested the alpha, we can patch it back into 6.1.x!

merceyz:
> @gaetanmaisse @merceyz should @nrwl/workspace be an optional peer dependency for yarn2 compat?

Based on its usage, yes, though not just for Yarn PnP (Yarn 2 != pnp as Yarn 2 also supports node_modules) it must be declared so package managers can account for it when they do hoisting

mandarini:
I just updated my PR, added `@nrwl/workspace` as an optional peer dependency!

app/angular/package.json
```diff
@@ -90,10 +90,16 @@
     "@angular/platform-browser": ">=6.0.0",
     "@angular/platform-browser-dynamic": ">=6.0.0",
     "@babel/core": "*",
+    "@nrwl/workspace": ">=11.1.0",
```
merceyz, line 93: Needs to be a devDependency as well so that it's accesible for the code

gaetanmaisse (approved):

gaetanmaisse (changes requested):

gaetanmaisse:
The failing unit test looks related to a caching/hoisting issue in the `node_modules`, I did an ssh connection to the CI container, `yarn install`, reran the tests and everything was ok. Maybe updating the `yarn.lock` will fix that.

app/angular/src/server/angular-cli_config.ts
```diff
@@ -47,17 +47,32 @@ function getTsConfigOptions(tsConfigPath: Path) {
...
-  const validIndex = possibleConfigPaths.findIndex((configPath) => fs.existsSync(configPath));
-
-  if (validIndex === -1) {
-    logger.error(`Could not find angular.json using ${possibleConfigPaths[0]}`);
-    return undefined;
+  let angularCliConfig;
+  try {
+    /**
+     * Apologies for the following line
+     * If there's a better way to do it, let's do it
```
ThibaudAV, line 54: I don't know if it's better (or if it's possible)🙈 . but an alternative without try/catch would be to let storybook choose if it should use nx or not 🤷‍♂️ 

Something like this: 
`const isNxWorkspace = fs.existsSync(path.join(dirToSearch, 'nx.json'))`

app/angular/package.json
```diff
@@ -73,6 +73,7 @@
     "@angular/forms": "^11.0.0",
     "@angular/platform-browser": "^11.0.0",
     "@angular/platform-browser-dynamic": "^11.0.0",
+    "@nrwl/workspace": ">=11.1.0",
```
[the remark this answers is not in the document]
mandarini: Ok, I'll use `11.1.0-beta.2`! How can I capture the yarn.lock changes, though?

app/angular/src/server/angular-cli_config.ts
mandarini: The thing is that it will need to access the `readWorkspaceConfig()` from `@nrwl/workspace`, so I apologized for using the `eslint-disable` to let me require it... :)

app/angular/package.json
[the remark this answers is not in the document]
mandarini: It still defaults to 
```"
@nrwl/workspace@>=11.1.0-beta.2":
  version "9999.0.0"
```
since `11.1.0-beta.2` is not an actual tag/release. We can wait until the end of the week, when that release will happen, maybe?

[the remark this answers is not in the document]
gaetanmaisse: Yes sure, I let you update that when the release will be done and then ping me back 😉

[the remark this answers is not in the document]
mandarini: @gaetanmaisse I committed the `yarn.lock` changes too!

Hasbiallah210012 (approved):
[]()

app/angular/src/server/angular-cli_config.ts
Kamarulzamansalehudin: Done

ThibaudAV: okay, 
my point is more about the try/catch. I find that the way we manage the nominal case (without nx) in a Catch is not pretty.
and I propose this solution if it works and suits you of course :) 

```js
const isNxWorkspace = fs.existsSync(path.join(dirToSearch, 'nx.json'))
if(isNxWorkspace) {
 return  require('@nrwl/workspace').readWorkspaceConfig({
      format: 'angularCli',
    });
}
... 

```

mandarini:
Hi @gaetanmaisse ! Let me know if I need to make any further changes! I see that the tests are still failing :(

gaetanmaisse:
We are currently working on fixing the tests (they are failing on `next` too), I will update your branch as soon as the fix will be merged 😉

mandarini:
Thank you very much! :)

app/angular/src/server/angular-cli_config.ts
gaetanmaisse: @ThibaudAV are you ok to go with the current state?

gaetanmaisse (approved):

app/angular/src/server/angular-cli_config.ts
ThibaudAV: Yes 🚀 , I have plans to improve this part. I will do it with a dedicated PR  ;)

PR #13559, opened 2021-01-05
Enables Yarn 2

arcanis:
## What I did

Upgrading the Yarn binary to the latest stable release (2.4), which should make them faster.

I've configured the project to stay on the node_modules linker since it makes the upgrade fairly painless. Tests pass locally, I'll see how it goes on CI and tweak the command line references accordingly (things like `--frozen-lockfile` being renamed to `--immutable`, etc).

Notable changes:

- The Yarn binary is checked within this repository. This mean that all your contributors will now be ensured to use the exact same version of Yarn, regardless what they have installed globally.

- The root package.json now lists all workspaces as dependencies (using the `workspace:` protocol). This is because in Yarn 2 the root workspace doesn't implicitly depend on all workspaces anymore, they have to be explicitly listed. In theory the dependencies should be listed in each workspace that depends on them (so for example `ui` should depend on `api`) but I'm worried it would lead to a gigantic diff, so for now the repo keeps relying on the hoisting to work.

## How to test

The existing CI pipeline should be enough to validate that the upgrade is safe. One potential risky part is the publish workflow; I've heard it's already covered by tests using Verdaccio so it should be fine, but if someone can double check it would be perfect!


arcanis:
What are those TeamCity tests? Can't see their content 🤔

wKich:
@arcanis, you can login as a guest

gaetanmaisse:
@arcanis I think you don't need to worry about the TeamCity tests for now, our "official" CI workflows are the ones from CircleCI. Plus, if I remember correctly I had some issue with them in a previous CI PR because there were always using workflow defined in `next` branch and not the one I updated in my PR.

gaetanmaisse:
I have just check out your branch locally, ran yarn install + build + test, and everything was ok 💪🏻  

Doing the same in the CI executor (with ssh connection) and it's stuck after the resolution step (in the fetch step? 🤷🏻 ), maybe the [current `medium` executor](https://github.com/storybookjs/storybook/blob/06fba1d51153829a7acd674346dff4a7c47b7125/.circleci/config.yml#L79) isn't powerful enough, you can try with a `large` or `xlarge` one. 

I'm really interested in this PR, feels free to ping me on discord if you want!

gaetanmaisse:
Note for myself: As everything is working with Yarn 2 now, the Yarn 2 e2e tests must be reworked and renamed to Yarn PnP

merceyz:
The e2e tests should avoid using npx as well, as that wont use the Yarn cache which will slow them down

gaetanmaisse:
Tested locally https://github.com/storybookjs/presets/pull/181 is fixing a lot of the current issues, I can't merge that PR as required CI checks are outdated and I'm not admin on that repo so I can't change them 🙈 . I have asked some people to take a look, hope it will be updated soon, in the meanwhile we are a bit stuck.

examples/angular-cli/package.json
```diff
@@ -5,8 +5,7 @@
   "license": "MIT",
   "scripts": {
     "build": "ng build",
-    "prebuild-storybook": "yarn storybook:prebuild",
-    "build-storybook": "build-storybook -s src/assets",
+    "build-storybook": "yarn storybook:prebuild && build-storybook -s src/assets",
```
merceyz, line 8: Do note that script names containing `:` (colon) are global and thus don't allow duplicates
https://yarnpkg.com/getting-started/qa/#how-to-share-scripts-between-workspaces

gaetanmaisse: Good to know! I think we will need to do a big clean around that then.

gaetanmaisse:
Closing in favour of smaller and more focused PRs:
 - https://github.com/storybookjs/storybook/pull/13871
 - https://github.com/storybookjs/storybook/pull/13872
 - https://github.com/storybookjs/storybook/pull/13876
 - https://github.com/storybookjs/storybook/pull/13878
 - https://github.com/storybookjs/storybook/pull/13907 <-- only this one is really related to Yarn 2 migration

PR #13567, opened 2021-01-06
Build: remove redundant checks for TS type declaration generation

unrevised6419:
`--declaration true` is always added no matter what app/addon/lib, moved it in `args` array initializer
And add `--emitDeclarationOnly` only is not angular and not storyshots


gaetanmaisse (approved):
LGTM 🚀

gaetanmaisse:
🚀  Merged, thanks for your contribution @iamandrewluca!

PR #13596, opened 2021-01-10
Angular: Properly handle empty tsconfig compilerOptions

matheo:
Issue:

## What I did

Considered the case when `tsconfig` is extending another file and has an empty `compilerOptions`.

This avoid an exception like:

image


gaetanmaisse:
I think we can merge this one but we need to plan some more work about the TS config, sum up of discussion I had with @ThibaudAV: we should follow the `extends` key to find the parent `tsconfig.json` and build the resulting one to be sure all the options/configs are taking into account.

gaetanmaisse (approved):

PR #13600, opened 2021-01-11
fix: Link to docs for param/decorator after story

maael:
Issue: The current link for adding a decorator/parameter after the first story is wrong, as GitHub makes the link as https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#can-no-longer-add-decoratorsparameters-after-stories, the `/` gets removed, not replaced with a `-` like a space.

## What I did

I replaced the [broken link](https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#can-no-longer-add-decorators-parameters-after-stories) that takes you just to the top of the migrations doc, with the [working link](https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#can-no-longer-add-decoratorsparameters-after-stories).

## How to test


shilman (approved):
LGTM. Thanks for fixing this! 🙏

PR #13618, opened 2021-01-12
Document public path splitting feature for static files

Josh-a-e:
Issue: None

## What I did

Documented a useful feature that I found here: https://github.com/storybookjs/storybook/blob/next/lib/core/src/server/utils/static-files.ts

## How to test

Not really relevant - check out the documentation?

### DIFF
--- a/docs/configure/images-and-assets.md
+++ b/docs/configure/images-and-assets.md
@@ -56,6 +56,16 @@ You can also pass a list of directories separated by commas without spaces inste
 }
 ```
 
+Static directories can be split using a `:` to specify the path where they are accessible.
+
+```json
+{
+    "scripts": {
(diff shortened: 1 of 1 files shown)


shilman (changes requested):
I don't think this is right. I think the `:` denotes a mapping from the file on disk to the path where the files are served by the dev server. @jonniebigodes can you please verify?

jonniebigodes:
@Josh-a-e sorry for the long delay on this, but it seems that the comment left above is actually correct. Adding this to the documentation as is may lead to some misconceptions, if you're willing to adjust accordingly based on the comment above I'll gladly go over it and hopefully get it merged and help not only you but the rest of the community.

Let me know and we'll go from there? Sounds good?

Stay safe

PR #13624, opened 2021-01-12
Addon-actions: Normalize args

guilhermewaess:
Issue: #7764

addon-action now will show results out of an array when it's not an array.

## What I did

I added treatment for the args received on addon-actions. Before, it was sending the args forward. Now, when it is only one argument, it will send the argument outside of an array.

Before:

image

After: 

image

## How to test

 I did update the tests
 No
 I don't think so.

Thanks @yannbf for the guidance :D

### DIFF
--- a/addons/actions/src/preview/__tests__/action.test.js
+++ b/addons/actions/src/preview/__tests__/action.test.js
@@ -16,7 +16,7 @@ describe('Action', () => {
 
     action('test-action')('one');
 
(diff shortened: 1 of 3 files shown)


gaetanmaisse (approved):

gaetanmaisse:
LGTM 💪🏻 , I will let @shilman make a double-check.

shilman (approved):
Looking great! 🚀  Thanks @guilhermewaess 🙏

PR #13625, opened 2021-01-13
Fix writing-addons docs

cleversprocket:
Issue:
The `src` path throws a webpack error because it's expecting CJS but it's using ESM.

## What I did
Changed `src` to `dist` in docs.

## How to test
No testing needed; just a documentation fix.

### DIFF
--- a/docs/snippets/common/my-addon-storybook-registration.js.mdx
+++ b/docs/snippets/common/my-addon-storybook-registration.js.mdx
@@ -7,7 +7,7 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/preset-create-react-app",
-    "../src/preset.js" //👈 Our addon registered here
+    "../dist/preset.js" //👈 Our addon registered here
   ]
 }
 ```


jonniebigodes:
@cleversprocket could you provide a little more context on this?  I was checking against my code base, which i would gladly push to a repo and share it here and the documentation checks out. The way it's currently documented we're covering the workflow you would go about to get a addon working locally.

Let me know. 

Stay safe

docs/snippets/common/my-addon-storybook-registration.js.mdx
```diff
@@ -7,7 +7,7 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/preset-create-react-app",
-    "../src/preset.js" //👈 Our addon registered here
+    "../dist/preset.js" //👈 Our addon registered here
```
Astrodevil, line 10: here dist is the minified/concatenated version  ( used on production sites)

src  is the raw code(used to read/edit the code)

i think this is the change made here because  its addon registered written .

jonniebigodes: @Astrodevil thank you for the comment on the pull request, but documentation flow wise at this stage the reader is still in "development mode" (pardon the bad pun), the addon is still early on. Applying the change here would involve having to build it first. And as the Storybook's  addon capabilities and possibilities are far and wide, we cannot know for sure if the case applied to the author is valid for other cases and including this change can lead to some misleadings. Also in all fairness it's best to reference the changed introduced with this pull request in a [root level preset](https://storybook.js.org/docs/react/addons/writing-addons#root-level-presetjs) , that will make the life easier for the addon consumers.

Stay safe

cleversprocket: Hi all,

This code change is in reference to the documentation for [writing addons](https://storybook.js.org/docs/react/addons/writing-addons#register-the-addon). I did a search through the codebase to find reference to this snippet so let me know if the change should be made somewhere else.

This snippet is referencing `/my-addon/.storybook/main.js` which is entirely local to the addon developer and not part of the published package (let me know if there is a case where it is part of the production package).

The [documentation example](https://storybook.js.org/docs/react/addons/writing-addons#build-system) uses ESM (`import/export`) in `src` but `./storybook/main.js` file cannot reference it and thus must use the transpiled CJS export (note how `main.js` is using `module.exports` and not `export default`).

cleversprocket:
@jonniebigodes I responded to the code comment thread—let me know if that helps clear it up. I'm surprised your codebase worked—I would like to see it if you wouldn't mind pushing it up. Thanks!

jonniebigodes:
@cleversprocket i have the repo [here](https://github.com/jonniebigodes/my-addon-example) it's a bit different in terms of "content" but achieves the same principle.

> This snippet is referencing /my-addon/.storybook/main.js which is entirely local to the addon developer and not part of the published package (let me know if there is a case where it is part of the production package).

That is correct, the snippet is targeting a local Storybook inside the package as a means to help during the development phase. 

As once you publish it you shouldn't be referencing that at all, like i mentioned that's for development only. If you follow along the rest of the documentation you'll find the key piece of information when you publish and later consume the addon, more specifically [here](https://storybook.js.org/docs/react/addons/writing-addons#root-level-presetjs). Including the root level preset helps you and addon consumers use it. One test that was done with the addon asides from the development phase of it, was to test it locally emulating a production environment. I used another project for it. As you can see below:

addon-vs-code

And running Storybook on the other project yields the following:
example-addon-storybook

Indeed my example code is based off node 14 and that could be part of the issue you're experiencing. With that in mind what i was thinking in terms of this pull request is to re-task it a bit and mention the fact that the code is based off node 14 and if the reader uses a different node version some adjustments might be required. 

Take the time to go over the repo code i've mentioned and let me know what you think of what i'm suggesting and we'll go from there. Sounds good?

Stay safe

cleversprocket:
@jonniebigodes thanks for sharing your repo. I was able to run storybook in your repo but only because you are referencing the `register.js` file and not the `preset.js` ([here](https://github.com/jonniebigodes/my-addon-example/blob/7d37820d79b019cd20206d2a5906a81e1ec52061/.storybook/main.js#L9)). Change that to `preset.js` and you'll see webpack blows up.

Following the stack trace it blows up because the core server is doing an [explicit require](https://github.com/storybookjs/storybook/blob/65fe593cd137cb6d1bf1c4b94085b4459d72b343/lib/core/src/server/presets.ts#L109). I haven't dug into the difference with how storybook internally handles `register.js` here instead of `preset.js`—If I had to guess its because preset can export webpack and babel configs which in most environments need to use `require`.

jonniebigodes:
Closing this pull request to avoid adding misconceptions. The documentation was vetted beforehand and the snippets were used as well. More even this documentation is intended for getting started and we have multiple sources that run through the entire workflow

Looking forward to your next contribution.

Stay safe

PR #13640, opened 2021-01-14
Dependencies: Update postcss-loader to ^4.1.0

phated:
Issue: Ref #12668

## What I did

I dug into the PostCSS stuff related to the v8 upgrades, specifically what happened in the breaking change to `postcss-loader`. It looks like the primary changes were dropping support for node <10.13, changing the way config options were specified and making PostCSS a peerDepedency.

Since the embedded `postcss-loader` in Storybook has control of all those things, it seems that this can be updated (in a non-breaking manner) to ^4.1.0 as long as a PostCSS@^7 is added as a direct dependency of `@storybook/core`.

I believe upgrading this loader is the first step in moving us towards supporting PostCSS v8.

## How to test

- Is this testable with Jest or Chromatic screenshots? Unsure, I checked the usage in `examples/html-kitchen-sink`.
- Does this need a new example in the kitchen sink apps? Already in the `examples/html-kitchen-sink`.


phated:
cc'ing @ThibaudAV since I updated the loader in the angular stuff, too.

shilman (approved):
LGTM! 🚀

PR #13653, opened 2021-01-15
Addon-docs: Add dynamic snippet support for Svelte

j3rem1e:
Issue: none

## What I did

This PR implements the dynamic snippet feature with Svelte.

It supports "wrapper" components, used in storybook to test slots : 

- A component can be annoted with @wrapper to be declared as a wrapper : in this case, the source generated will not reference the wrapper but the real component used by the wrapper
- A wrapper can annotate a property with @slot to be used as an alias to the default slot

The svelte kitchen sink has been updated to use this feature. 

## How to test

- Does this need an update to the documentation? I don't think

### DIFF
--- a/addons/docs/src/frameworks/svelte/config.ts
+++ b/addons/docs/src/frameworks/svelte/config.ts
@@ -1,6 +1,7 @@
 import { extractArgTypes } from './extractArgTypes';
 import { extractComponentDescription } from '../../lib/docgen';
 import { prepareForInline } from './prepareForInline';
+import { sourceDecorator } from './sourceDecorator';
 
 export const parameters = {
   docs: {
@@ -10,3 +11,5 @@ export const parameters = {
     extractComponentDescription,
   },
 };
+
+export const decorators = [sourceDecorator];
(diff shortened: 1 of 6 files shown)


shilman (commented):
I'm seeing some funny behavior in the `svelte-kitchen-sink`; can you check?

Button_-_Rounded_⋅_Storybook

Not sure where the text is coming from.

I'm also seeing some story rendering issues where clicking around sometimes stories don't render at all. It's also showing up in `next` so I don't think it's at all related to this PR. But I thought I'd bring it up just in case you can take a look:

Button_-_Square_⋅_Storybook

j3rem1e:
I have created another issue/PR for the rendering of the stories. It's unrelated of this PR, I see it on my own instance with the latest released version.

#13662 : the bug can be story duplicated, or not rendered, or rendered in the wrong preview canvas

j3rem1e:
The 'Squared text' come from here: https://github.com/storybookjs/storybook/blob/next/examples/svelte-kitchen-sink/src/stories/button.stories.js

It's not related to this PR, this property doesn't exist on the ButtonView component so it's not rendered. However you can see it now in the generated code. 

I'll try to improve this story.

shilman (approved):
💥 great job @j3rem1e 💥

PR #13666, opened 2021-01-18
Add catalog metadata to the addons

winkerVSbecks:
## What I did

Added catalog metadata for the addons


phated (approved):
@shilman anything else you'd want done before merging this?

shilman (changes requested):
@winkerVSbecks sorry for the slow reply -- totally missed this PR the first time around. can you run yarn install in the storybook repo and then re-save the files? i'm pretty sure you're running an old version of prettier, which is causing the `typesVersions` to move around: an unnecessary change that will just get reversed the next time somebody else edits the files. The rest of the curation looks great. Thanks!

winkerVSbecks:
@shilman updated

shilman (approved):
Thanks @winkerVSbecks 🙏

PR #13673, opened 2021-01-19
Core: Throw an error for invalid story format

yannbf:
Issue:
When mistakenly a user writes a story such as:
```jsx
export const Basic = <div>Hello</div>
```

rather than:
```jsx
export const Basic = () => <div>Hello</div>
```

Storybook fails with an error that is not very helpful:
image

## What I did

This PR adds a check in the client api that will break Storybook if the story received is not a function telling exactly which story is broken. It's a similar mechanism than the one that checks if `storyName` is not a string.

Here's how it looks like:
image

I personally find it quite useful that Storybook breaks entirely because it tells the users that there is a critical error given that their stories are broken and it forces them to fix the problem right away.

## How to test

1 - check the branch
2 - `yarn build client-api`
3 - add this to any of the example stories:
```jsx
export const Basic = <div>Hello</div>
```
4 - run Storybook (e.g. `yarn start`)

### DIFF
--- a/lib/client-api/src/client_api.test.ts
+++ b/lib/client-api/src/client_api.test.ts
@@ -658,6 +658,18 @@ describe('preview.client_api', () => {
           expect(entry.stories[0].render()).toBe('story2');
         }
       });
+
+      it('should throw an error if story is in wrong format', () => {
+        const {
+          clientApi: { storiesOf },
+        } = getContext();
+
+        expect(() => {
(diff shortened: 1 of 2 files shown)


gaetanmaisse (approved):
Useful and clean 💪🏻  👏🏻

shilman (approved):
Great fix Yann 👍 👍 👍

PR #13675, opened 2021-01-19
Addon-controls: Infer color and date controls

yannbf:
Issue:

Args like "iconColor, background, purchaseDate" shouldn't be inferred as control type "text" by default. 
Most of the time props that end with color or background will indeed be a color control, and props ending with `date` will be date.. so let's infer those to save people's time 👌 

## What I did

This PR adds more inference to controls, for `color` and `date` types.

Before:
image

After:
image

## How to test

1 - checkout the branch
2 - `yarn build client-api`
3 - run `yarn start`
4 - add this to any story in the example:
```jsx
MyStory.args = {
  iconColor: '#BADA55',
  purchaseDate: '10/11/2012', // could also be unix timestamp
};
```

- Does this need an update to the documentation? not that I know of


gaetanmaisse (approved):

shilman (commented):
Much better! Looking great 💯 

Do you want to:
- [ ] Add support for other types
- [x] Add documentation

yannbf:
Docs are updated! I think it would be awesome if @jonniebigodes could take a look at it <3

shilman (approved):
💯

tmeasday:
Is this technically a breaking change? (Nice feature though!)

yannbf:
Hey @tmeasday thinking about it, kind of.. It will change the behavior for users that used to see strings instead of the pickers and it's opt out rather than opt in 🤔. I guess we should mention about it on the migration!

tmeasday:
Yeah, probably easy enough to opt out if people really care and a nice automatic upgrade, so I guess we can not worry about it.

PR #13679, opened 2021-01-20
docs: fix args definition in How to write stories

livthomas:
## What I did

I changed the definition of arguments in How to write stories. Because [slots](https://v3.vuejs.org/guide/component-slots.html) is something completely different from [props](https://v3.vuejs.org/guide/component-props.html#prop-types) in Vue.js and it might confuse less experienced developers.

I also changed Angular's `@Input` to match the actual syntax in Angular components (title case).


shilman (approved):
LGTM! thanks @livthomas !

PR #13704, opened 2021-01-22
Angular: Add Angular 11.1 support

gaetanmaisse:
Issue: #13689, #13563 surely #13693

## What I did

Some internal plumbing about webpack config changed in the latest version of Angular, for details see angular/angular-cli@eb30a92. As Storybook is using private function of `@angular-devkit/build-angular` I added the new required properties in the webpack config.

I also updated the Angular example and the Angular dependencies in the app to Angular 11.1.

## How to test

- As Angular example has been updated to Angular 11.1 the CI will check if everything is OK


shilman (approved):
LGTM. Thanks for jumping on top of this @gaetanmaisse! Smoke test failures are unrelated to this PR and @j3rem1e and I are investigating separately. Merging so we can test this & patch to `master` if it's working

PR #13705, opened 2021-01-22
Fixes smoke-test on svelte-kitchen-sink

j3rem1e:
Issue: svelte-kitchen-sink smoke-test fails because of the warnings in the argstable story.

## What I did

I have fixed the warning in this stories.
It's probably a bug in the svelte compiler because this warning should be disabled when $$props is used, but i'll check that later.

## How to test


shilman (approved):
LGTM! Thanks for the quick fix @j3rem1e 🙏

PR #13714, opened 2021-01-23
Build: Exclude all test and story files from transpilation

phated:
Issue: #11220

## What I did

I split out the additional excludes from #11220 into a separate commit. One question that @shilman had was whether or not we could put these in the root config that the packages extend from instead of duplicating all the boilerplate.

## How to test

### DIFF
--- a/addons/a11y/tsconfig.json
+++ b/addons/a11y/tsconfig.json
@@ -11,5 +11,12 @@
     "noFallthroughCasesInSwitch": true
   },
   "include": ["src/**/*"],
-  "exclude": ["src/__tests__/**/*"]
+  "exclude": [
+    "src/**/*.test.*",
+    "src/**/tests/**/*",
+    "src/**/__tests__/**/*",
+    "src/**/*.stories.*",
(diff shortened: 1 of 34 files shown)


gaetanmaisse (dismissed):
> One question that @shilman had was whether or not we could put these in the root config that the packages extend from instead of duplicating all the boilerplate.

Yes definitely, I would suggest to set up a clean `exclude` property in the root `tsconfig.json` and then remove `exclude`  in all sub tsconfig.

phated:
@gaetanmaisse how does the exclude work in relation to the the working directory (if being used in extends)?

phated:
@gaetanmaisse it looks like it is relative to the file you are importing, so we can't put them at the top level without using a very liberal glob (which would cause us to lose all the performance gains).

From https://www.typescriptlang.org/tsconfig#extends
> All relative paths found in the configuration file will be resolved relative to the configuration file they originated in.

@shilman Given that limitation, are you okay merging this?

gaetanmaisse (approved):

PR #13715, opened 2021-01-23
Build: Generate version file with preval macro

phated:
Issue: #11220

## What I did

I split out the preval macro code into its own commit from #11220. Couple open questions:
1. Why does the e2e PnP tests fail?
2. Does compilation use the babel cache? If so, then https://github.com/kentcdodds/babel-plugin-preval/issues/19 says we can't use this method.

## How to test

- Is this testable with Jest or Chromatic screenshots? e2e tests seem to catch it


gaetanmaisse:
@phated the CI error is:

> ERR! Module not found: Error: Can't resolve '../version' in '/tmp/storybook-e2e-testing/react-latest/node_modules/@storybook/api/dist/cjs/modules'
> ERR! ModuleNotFoundError: Module not found: Error: Can't resolve '../version' in '/tmp/storybook-e2e-testing/react-latest/node_modules/@storybook/api/dist/cjs/modules'

But it doesn’t seem to be related to Yarn 2 or PnP as all the E2E are failing. You can try to reproduce locally with : 
`yarn test:e2e-framework react`

phated:
@gaetanmaisse I just ran that command locally at the react e2e suite passes. Any thoughts on what else I should look into??

gaetanmaisse:
🤔 Everything looks ok locally but part of `@storybook/api` is missing when the package is published in the CI workflow, will investigate a bit by SSHing the CI.

gaetanmaisse:
Should be fixed with https://github.com/storybookjs/storybook/pull/13786

gaetanmaisse (approved):

PR #13717, opened 2021-01-23
CLI: Add safe check for eslint overrides

yannbf:
Issue:
When running `npx sb@next init` in a CRA project and the project does not contain `eslintConfig.overrides` in package.json, the CLI throws an error:
image

## What I did

Add a safe check so that won't happen again

## How to test

- Checkout this branch
- Create a project without eslintConfig.overrides
- Run local cli


tooppaaa (approved):

phated (approved):
We should land this. @shilman any objects?

shilman (approved):
LGTM!

PR #13725, opened 2021-01-25
Chore: Increase node version minimums to 10.13

phated:
Issue: #8207

## What I did

In #8207, the storybook CI was changed to only be node 10+ but the engine fields weren't changed to reflect that supported version range. I've updated all the engine references to be >=10.13 (which is when recursive mkdir was added) to be inline with the CI test matrix.

Additionally, node 8 is no longer LTS after Jan 1, 2021.

I also updated the babel settings to target node 10 (plucked from #11220).

## How to test

### DIFF
--- a/app/angular/package.json
+++ b/app/angular/package.json
@@ -103,7 +103,7 @@
     }
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=10.13.0"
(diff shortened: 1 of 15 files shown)


shilman (approved):
LGTM 🎉

Issue #13718, opened 2021-01-23
Theming storybook


cherylcarpenter:
**Describe the bug**
I have implemented a theme for my storybook project following the exact instructions here
https://storybook.js.org/docs/vue/configure/theming

There is no observable change to any colors or logo. No errors in the console or terminal on build.

**Expected behavior**
on storybook serve I expected the colorPrimary in the navbar and storybook UI to be hotpinik, the logo to be changed etc. just like in the demo 

**System**
Environment Info:

  System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  Binaries:
    Node: 12.13.1 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.12.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 88.0.4324.96
    Safari: 14.0.1
  npmPackages:
    @storybook/addon-actions: ^6.1.15 => 6.1.15 
    @storybook/addon-essentials: ^6.1.15 => 6.1.15 
    @storybook/addon-links: ^6.1.15 => 6.1.15 
    @storybook/addons: ^6.1.15 => 6.1.15 
    @storybook/theming: ^6.1.15 => 6.1.15 
    @storybook/vue: ^6.1.15 => 6.1.15 

**Additional context**
Vue-cli  setup

jonniebigodes:
@cherylcarpenter could you check one thing for me? If you don't mind try running storybook with the `--no-manager-cache` flag. Turning your package.json script into `"storybook": "start-storybook -p 6006 --no-manager-cache"` . Let us know if that solved the issue.

Stay safe

cherylcarpenter:
@jonniebigodes  hi, thanks for reply! I had searched for any caching info so this is good to know.
That did indeed do something, but not exactly what we're going for based on the documentation I mentioned here:
https://storybook.js.org/docs/vue/configure/theming

The 'them' is based on 'light' and the colorPrimary is hotpink but now it looks like the dark theme? LOL

![image](https://user-images.githubusercontent.com/13105904/105612033-b8a4c280-5d7e-11eb-8557-d7e43a712971.png)

cherylcarpenter:
@jonniebigodes Thanks for reply! I had looked for caching info so this is good to know.

I'm getting a new error now in the build terminal.
WARNING in ./.storybook/manager.js 5:9-16
[...]

PR #13726, opened 2021-01-25
docs: import from theming module to support esm

phated:
Issue: #13718

## What I did

In #13718, some users were having issues with the theming docs because they tried to import from `@storybook/theming/create` but the `create` method is exported from the main file, too. I'm not sure of the reason for this "re-export", but I switched it to using the main file for better typescript and ESM support.

Does anyone know why there's that extra export file?

## How to test

- Does this need an update to the documentation? ✅ 


shilman (approved):
LGTM!

PR #13730, opened 2021-01-26
CLI: Add try/catch on readFileAsJson to improve error message

andrefarzat:
Issue:
When running `npx sb init` in an angular project, one json file was wrong (with a trailing comma) and it took me too much time to figure out which json file was wrong.
Screen Shot 2021-01-26 at 10 18 26

## What I did
Added try/catch wrapping the `JSON.parse` to improve error message and the user be aware which json file is wrong

## How to test
My test case is having a tsconfig.app.json with a wrong json in the file. In an angular project, the command `npx sb init` reads about 3 json files ( among `angular.json`, `tsconfig.json` and `tsconfig.app.json` )

### DIFF
--- a/lib/cli/src/helpers.ts
+++ b/lib/cli/src/helpers.ts
@@ -32,10 +32,10 @@ export function readFileAsJson(jsonPath: string, allowComments?: boolean) {
 
   try {
     return JSON.parse(jsonContent);
-  } catch(e) {
-    throw new Error('Invalid json on file: ' + filePath);
+  } catch (e) {
+    console.error(chalk.red(`Invalid json on file: ${filePath}`));
+    throw e;
   }
-
 }
 
 export const writeFileAsJson = (jsonPath: string, content: unknown) => {
--- a/lib/cli/src/helpers.ts
+++ b/lib/cli/src/helpers.ts
@@ -29,7 +29,13 @@ export function readFileAsJson(jsonPath: string, allowComments?: boolean) {
 
   const fileContent = fs.readFileSync(filePath, 'utf8');
   const jsonContent = allowComments ? stripJsonComments(fileContent) : fileContent;
-  return JSON.parse(jsonContent);
+
+  try {
+    return JSON.parse(jsonContent);
+  } catch(e) {
+    throw new Error('Invalid json on file: ' + filePath);
+  }
+
 }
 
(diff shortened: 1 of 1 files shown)


lib/cli/src/helpers.ts
```diff
@@ -29,7 +29,13 @@ export function readFileAsJson(jsonPath: string, allowComments?: boolean) {
 
   const fileContent = fs.readFileSync(filePath, 'utf8');
   const jsonContent = allowComments ? stripJsonComments(fileContent) : fileContent;
-  return JSON.parse(jsonContent);
+
+  try {
+    return JSON.parse(jsonContent);
+  } catch(e) {
+    throw new Error('Invalid json on file: ' + filePath);
```
phated, line 36: Maybe we should log the filename and still throw the original error? I would hate to lose the stack trace for debugging

phated (changes requested):
I think this is a good idea! I just don't want to lose the stack trace, instead we should add additional information somehow. 🤔

lib/cli/src/helpers.ts
andrefarzat: Changed as you suggested. Just added `chalk.red` on it:
```ts
try {
    return JSON.parse(jsonContent);
  } catch(e) {
    console.error(chalk.red('Invalid json on file: ' + filePath));
    throw e;
}
```

lib/cli/src/helpers.ts
```diff
@@ -29,7 +29,13 @@ export function readFileAsJson(jsonPath: string, allowComments?: boolean) {
 
   const fileContent = fs.readFileSync(filePath, 'utf8');
   const jsonContent = allowComments ? stripJsonComments(fileContent) : fileContent;
-  return JSON.parse(jsonContent);
+
+  try {
+    return JSON.parse(jsonContent);
+  } catch (e) {
+    console.error(chalk.red(`Invalid json on file: ${filePath}`));
```
phated, line 36: I believe the pattern in this file is to use `logger.error` (logger is just assigned as `const logger = console` above.
```suggestion
    logger.error(chalk.red(`Invalid json on file: ${filePath}`));
```

shilman (approved):
Yay. Thanks @andrefarzat !!!

PR #13752, opened 2021-01-27
CI: Upgrade cache GH Action & remove fallback caches

phated:
Issue: N/A

## What I did

I'm noticing a bunch of failures on the "Unit Tests" GH Action. I believe this is because the cache action is fetching stale caches. I removed the "restore-keys" option so it only ever pulls a cache that matches the yarn.lock file.

## How to test

### DIFF
--- a/.github/workflows/tests-unit.yml
+++ b/.github/workflows/tests-unit.yml
@@ -12,13 +12,10 @@ jobs:
         node-version: '10.x'
     - uses: actions/checkout@v2
     - name: Cache node modules
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
         path: node_modules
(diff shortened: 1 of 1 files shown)


gaetanmaisse:
I would go one step further and completely remove the unit tests workflow from the GitHub Actions.
I guess it was added to give it a try with GH Actions but, at the moment, I don't see any benefit to keep it as it is a pure duplicate of what we have on CircleCI, it takes so much time to run and is often failing. What's your opinion on that @ndelangen ?

phated:
@gaetanmaisse Can we merge this and make that decision later? It is causing people's PRs to be stalled.

gaetanmaisse (approved):

PR #13757, opened 2021-01-28
Update badge-story-starter-example.mdx.mdx

qoalu:
I stumbled on this while setting up storybook docs and it took me way too long to understand what the problem is.

Canvas does not exist anymore and should be called Preview

Issue:

Wrong instructions, import is for `Preview` but the component used is `<Canvas>`

## What I did

Updated documentation


qoalu:
this actually the other way around
https://github.com/storybookjs/storybook/blob/next/addons/docs/src/blocks/Preview.tsx

The import should be on Canvas from v6 (?)

PR #13777, opened 2021-01-30
UI: Fix copy to clipboard for insecure deployments

j3rem1e:
Issue: #13776

## What I did

If `navigator.clipboard` is not defined, then the copy-to-dashboard uses `execCommand('copy')`.

## How to test


shilman (approved):
LGTM 💯

ghengeveld:
Why don't we just use [`copy-to-clipboard`](https://www.npmjs.com/package/copy-to-clipboard)? It's already used in various other places, so it already gets bundled anyway.

j3rem1e:
AFAIK `copy-to-clipboard` use `execCommand` which has been removed in  #10526 because of jsdom failures.

PR #13784, opened 2021-02-01
Core: Generate manager cache in smoke test, but don't use/clear any cache

ghengeveld:
Issue: -
Discussion: https://github.com/storybookjs/storybook/discussions/13737

## What I did

When running with `--smoke-test`, currently no manager cache is used, created or cleared. However, sometimes it's beneficial to be able to prepare (generate) a cached manager without actually starting Storybook, for example when preparing a Docker image with embedded manager cache. This change makes `--smoke-test` cache the manager it built like normal, but won't use or clear any cache it finds as was intended in #13266.

## How to test

### DIFF
--- a/lib/core/src/server/dev-server.ts
+++ b/lib/core/src/server/dev-server.ts
@@ -413,14 +413,14 @@ export async function storybookDevServer(options: any) {
     startManager({ startTime, options, configType, outputDir, configDir, prebuiltDir })
       // TODO #13083 Restore this when compiling the preview is fast enough
       // .then((result) => {
-      //   if (!options.ci) openInBrowser(address);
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM 🙏

shilman (changes requested):
@ghengeveld Unit test are failing, can you fix?

phated:
@shilman I believe unit tests are failing because of https://github.com/storybookjs/storybook/pull/13752 (an old, bad cache is used when a new cache is not found).

phated:
@ghengeveld That PR was landed, but I think you need to rebase master to get that action base to run 🎉

PR #13785, opened 2021-02-01
Svelte: Improved decorators

j3rem1e:
Issue: None

## What I did

Today, writing decorators for svelte is hard:
- A decorator must be a svelte component and the decorated story must be injected "by hand" with properties and a <svelte:component/>
- It's not possible to use svelte context

In a "svelte world", data is passed to child through context, and a decorator has children with a `<slot/>`

This PR allow to use a syntax closest to Svelte.

A decorator component can be declared as:

```
<!-- MarginDecorator.svelte -->
<div>
  <slot/>
</div>

<style>
  div { margin: 3em; }
</style>
``` 
And a decorator for this component is just `() => MarginDecorator`.

It's possible to use Svelte context with `() => { setContext('myKey', 'myValue') }`

Moreover, this PR:
- Implements #8673 : If a story doesn't define a component, then the parameters.component is used
- Remove dependencies to internal svelte implementation: the rendering is done through a real component

## How to test

- Does this need an update to the documentation? Done

### DIFF
--- a/addons/docs/src/frameworks/svelte/HOC.svelte
+++ b/addons/docs/src/frameworks/svelte/HOC.svelte
@@ -1,7 +1,7 @@
 <script>
-  export let component;
-  export let props;
+  export let storyFn;
+
+  let { Component: component, props } = storyFn();
 </script>
 
-<svelte:options accessors={true} />
 <svelte:component this={component} {...props}/>
--- a/addons/docs/src/frameworks/svelte/prepareForInline.ts
+++ b/addons/docs/src/frameworks/svelte/prepareForInline.ts
@@ -1,22 +1,17 @@
-import { StoryFn, StoryContext } from '@storybook/addons';
+import { StoryFn } from '@storybook/addons';
 
 import React from 'react';
 
 // @ts-ignore
 import HOC from './HOC.svelte';
 
-export const prepareForInline = (storyFn: StoryFn, context: StoryContext) => {
-  // @ts-ignore
-  const story: { Component: any; props: any } = storyFn();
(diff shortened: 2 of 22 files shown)


shilman (approved):
@j3rem1e is this a breaking change?

Also, if the user specifies `Component` in the story output, does it still get used? Also can you add it to MIGRATION.md? And, can you add a deprecation warning that it's no longer needed and reference MIGRATION.md?

Example: https://github.com/storybookjs/storybook/pull/13858

shilman (changes requested):
Did not mean to approve this, just ask questions first.

j3rem1e:
> @j3rem1e is this a breaking change?

No, it works like before, at least in my tests. Most of this PR is a rewrite of the renderer to use svelte component instead of "hacking internals", and to call the story function inside a svelte context.

Storyshots generates the same snapshots, and decorators implemented like in the documentation continue to work as before.

> 
> Also, if the user specifies `Component` in the story output, does it still get used? 

Yes, it works like before.

> Also can you add it to MIGRATION.md? And, can you add a deprecation warning that it's no longer needed and reference MIGRATION.md?

I am note sure to understand 😅 In svelte/storybook, without the "native svelte format" PR, it's almost always necessary to use in a story a component which is not the component defined in meta - because a story can't define a custom template like in vue or react. What I have done is to make this Component optional, and if it's not set, then I use the component defined in meta. It's not really deprecated.

I can add a message if the component in a story is the same component as in meta, but I don't think it's really a deprecation warning, as it will not be removed.

shilman (approved):
Hi @j3rem1e thanks for the clarification. Merging!

PR #13786, opened 2021-02-01
Build: Do not recompile packages in publish step of the CI

gaetanmaisse:
## What I did

Do not recompile packages before publishing with the `run-registry` script. It will:
 - 🚓  ensure that the outputs of the build step are really used
 - 🚀  speed up the publish step on CI ~a lot~ a bit

## How to test
 
 - CI should still be 🟢  and the duration of the `publish` step should decrease ~a lot~ a bit


phated (approved):
We chatted about this and the reasoning makes sense! Do we need to label this with e2e-extended?

shilman (approved):
LGTM

PR #13787, opened 2021-02-01
CI: Remove generic cache key from Circle CI

phated:
Issue: N/A

## What I did

While investigating my CI failures, I noticed that an old cache was being pulled (in the same fashion as #13752) - we should only use a cache in CI if it matches the lockfile in the project, otherwise this can cause flakiness in the tests.

## How to test


shilman (approved):
LGTM!

Issue #13789, opened 2021-02-02
Support manager cache and force prebuilt manager with build-storybook


ghengeveld:
In order to speed up `build-storybook` in CI, it would be useful to be able to force the use of the prebuilt manager UI. That way the build process can be much faster, while the Storybook for local development can still be customized. Of course the limitation here is that the compiled Storybook will not include any customizations.

Tangent to this, we currently do not support manager cache in `build-storybook`, so even if there is a cached manager available, it will not be used when building a static Storybook. This makes it impossible to optimize a CI build with a cached manager, as was the goal of https://github.com/storybookjs/storybook/pull/13784.

zol:
>  Of course the limitation here is that the static Storybook will not include any customizations.

If we include the proposed `-force-prebuilt-manager` flag we should be extremely clear in our documentation that users may end up with what they perceive as 'broken' Storybooks because of missing customizations/add-ons.

ndelangen:
storybook 7.0 beta no longer has this flag/feature. I'll close this issue, as it's no longer relavant.

PR #13798, opened 2021-02-03
CLI: Add `--build-only` option to `start-storybook` (useful to warm up the cache)

kaelig:
### Issues:

Related to Discussion #13737 (where I initially described the use-case)
Related to #13266
Related to #13789
Related to #13784

---

I’d like a deterministic way to warm up a reusable cache for `start-storybook` without needing to start the server (nor opening the browser), which ideally would ensure that the next time somebody runs `start-storybook`, the cache is picked up and it runs faster.

For example, in the proprietary equivalent to GitHub Codespaces we run internally at Shopify, a warm cache means reducing Storybook's boot time by 30s (from 1m to 30s).

This is close to `--ci` but without the server. It's also close to `--smoke-test`, but without outputting stats, and with properly working caches.

## What I did

I added a new CLI option that intends to run everything a standard `start-storybook` would, but doesn't start the server.

## How to test

Assuming `yarn bootstrap` ran successfully,

1. `cd examples/cra-ts-examples`
2. `rm -rf node_modules/.cache/storybook`
2. `yarn start-storybook --build-only`: should exit without opening a browser
3. `yarn start-storybook`: should run faster

Screen_Shot_2021-02-02_at_6_12_11_PM


kaelig:
Before this goes into the mainline for the CLI, I'd love to hear feedback, I'm especially interested in thoughts on:

- whether this issue should be solved at another level of abstraction
- does the name `--build-only` convey the intent correctly?
- how to write integration tests for a new flag like this one?

ghengeveld:
Just today I discovered we actually have an undocumented flag called `--preview-url` which will skip building the preview. Combine that with `--smoke-test` and you'll have something that builds and caches (as of 6.2.0-alpha.20) the manager without bundling your stories. Can you try that and compare it to your approach?

kaelig:
I confirm that `yarn storybook --smoke-test --preview-url=""` works with the latest alpha. Thank you!

Happy to close this PR, but this undocumented solution feels quite hacky and easily breakable. I'd appreciate a solution that feels more solid and that would end up being documented!

ndelangen:
@kaelig if the use-case is to "prep the cache for faster starts" then I think the cli flag should reflect that.

my suggestion: `--prepare-manager-cache`

kaelig:
I'll close for now and revise when #13808 is out.

PR #13809, opened 2021-02-04
Addon-docs: Add support for Vue 3

phated:
Issue: Follow up to #13775 

## What I did

I added initial support for Vue 3 in `@storybook/addon-docs`. This was mostly a copy-paste of the `vue` implementation, but the `prepareForInline` became very streamlined since Vue 3 allows you to render a vnode directly into the DOM.

Notably missing: The `sourceDecorator` from Vue 2 won't work because vnodes no longer contain all the information to build up the templates. I explored calling `render(vnode, document.createElement('div'))` before looking at the props and it helped a bit, but I think the true way to build this would be a [custom renderer](https://v3.vuejs.org/api/global-api.html#createrenderer), similar to how they do SSR and their test utils. However, that is way out of my depth.

I also needed to fix the types in the html framework for some reason.

## How to test

I tested this in the vue 3 example storybook and it seems to be working.

### DIFF
--- a/addons/docs/package.json
+++ b/addons/docs/package.json
@@ -131,7 +131,7 @@
     "react": "^16.8.0 || ^17.0.0",
     "react-dom": "^16.8.0 || ^17.0.0",
     "sveltedoc-parser": "^3.0.4",
-    "vue": "^2.6.10",
+    "vue": "^2.6.10 || ^3.0.0",
     "webpack": ">=4"
   },
   "peerDependenciesMeta": {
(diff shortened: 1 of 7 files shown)


shilman (approved):
LGTM 🥇

Mootook:
@phated Perhaps I'm missing something but why is this still concerned with previous vue?
``` ks
// addons/docs/src/frameworks/vue3/preset.ts
loader: require.resolve('vue-docgen-loader', { paths: [require.resolve('@storybook/vue')] })
```
since I get ` Error: Cannot find module '@storybook/vue'` when running `start-storybook -p 6006` using @storybook/vue3.

phated:
@Mootook Just a bug. Good catch. A lot of this work was just copy-pasting the `@storybook/vue` code and adapting some parts to vue 3 APIs and I missed this reference. I'll submit a patch.

PR #13813, opened 2021-02-04
docs(vue-kitchen-sink): improve punctuation

bodograumann:
Previous punctuation was confusing. I think this is how it was meant.

I hope the `—` characters is allowed!?!


bodograumann:
Also what I just noticed:

```
import { stripIndents } from 'common-tags';
```

is not used explicitely in the document. Can it be removed or does it have side-effects?

bodograumann:
Furthermore I think this

```
<Meta title="Addon/Docs" />
```

should be

```
<Meta title="Addon/Docs" component={MyButton} />
```

shilman (approved):
😅 thanks!!

PR #13822, opened 2021-02-05
CLI: only kill other processes on fail

yannbf:
Issue: #13821

## What I did

The --watch command was using --kill-others with concurrently, which kills other processes in case any process fails **or succeeds**.

This PR changes it so that it will only kill other processes on failure. It also adds an extra filter to remove --watch in the lerna scope, as it should not be there.
 
## How to test

- yarn build addon-docs --watch (should not fail anymore)


shilman (approved):
LGTM! cc @phated

PR #13828, opened 2021-02-05
Addon-storyshots: Add support for Vue 3

phated:
Issue: Follow up to #13775

## What I did

I added support for Vue 3 in `addon-storyshots`. Additionally, I added some more stuff to the `vue-3-cli` example that makes it easier to try out the storyshots stuff.

As you can see, I was able to successfully generate the snapshots with storyshots, but those can't run during a normal test run because of all the troubles I've been encountering with having Vue 2 and Vue 3 in the dependency tree of this monorepo.

## How to test

- Is this testable with Jest or Chromatic screenshots? Generated but not run on normal test runs
- Does this need a new example in the kitchen sink apps? Added
- Does this need an update to the documentation? Updated the README for storyshots

### DIFF
--- a/addons/storyshots/storyshots-core/README.md
+++ b/addons/storyshots/storyshots-core/README.md
@@ -210,6 +210,29 @@ module.exports = {
 };
 ```
 
+### Configure Jest for Vue 3
+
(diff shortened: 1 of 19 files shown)


shilman (approved):
Great job--very nice!!! 💯

PR #13845, opened 2021-02-09
Addon-docs: Fix ArgsTable tab renamed to `Story` when using args

somenickname:
Issue: #12357

## What I did

Used component name for first tab of ArgsTable instead of hardcoded "Story" title

## How to test


shilman (approved):
LGTM!! Thanks for the fix!

PR #13869, opened 2021-02-10
Addon-docs: Configure syntax highlighter language by story parameter

yngvebn:
Issue: https://github.com/storybookjs/storybook/issues/13839

## What I did

I added some code to fetch `language` from `parameters.docs.source.language`. Previously it defaulted to jsx, and was not really configurable on a global level. For Angular, HTML is a better fit than JSX

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, Chromatic should catch the change on Angular-cli example


shilman (approved):
Good one! 💯

PR #13876, opened 2021-02-11
CLI: Don't try to add packages that are already installed

gaetanmaisse:

## What I did

Previously for some framework, the Storybook CLI was installing extra dependencies (listed in the generator of each framework, [for instance for rax](https://github.com/storybookjs/storybook/blob/3b6a06aaa10ae72e84215d1e706a093058fc5f67/lib/cli/src/generators/RAX/index.ts#L22)) but, in some case, some were already present in the user's project. 

It doesn't cause any issue with NPM/Yarn, I guess they are just updating the version 🤷🏻 . But, reasonably from my point of view, Yarn 2 is throwing an error: 
> Usage Error: Package "XXXXXX" is already listed as a regular dependency - remove the -D,-P flags or remove it from your dependencies first

To fix that I added a filter before adding the extra dependencies to remove all the already installed ones. 

## How to test

- CI should be green (especially the e2e extended test suite)

### DIFF
--- a/lib/cli/src/generators/baseGenerator.ts
+++ b/lib/cli/src/generators/baseGenerator.ts
@@ -2,7 +2,7 @@ import { NpmOptions } from '../NpmOptions';
 import { StoryFormat, SupportedLanguage, SupportedFrameworks } from '../project_types';
 import { getBabelDependencies, copyComponents } from '../helpers';
 import { configure } from './configure';
-import { JsPackageManager } from '../js-package-manager';
+import { getPackageDetails, JsPackageManager } from '../js-package-manager';
 
 export type GeneratorOptions = {
   language: SupportedLanguage;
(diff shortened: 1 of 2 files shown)


lib/cli/src/generators/baseGenerator.ts
```diff
@@ -64,21 +64,29 @@ export async function baseGenerator(
...
   const packages = [
     `@storybook/${framework}`,
     ...addonPackages,
     ...extraPackages,
     ...extraAddons,
     ...yarn2Dependencies,
-  ].filter(Boolean);
+  ]
+    .filter(Boolean)
+    .filter((packageToInstall) => !alreadyInstalledDependencies.includes(packageToInstall));
```
phated, line 81: As of a recent PR from me, these packages might be something like `vue-loader@^16` instead of just the package name. Do we need to account for that?

gaetanmaisse: Interesting! I will update the code split it and keep only the pre-`@` part. Thanks!

lib/cli/src/generators/baseGenerator.ts
```diff
@@ -64,21 +64,29 @@ export async function baseGenerator(
   const yarn2Dependencies =
     packageManager.type === 'yarn2' ? ['@storybook/addon-docs', '@mdx-js/react'] : [];
 
+  const packageJson = packageManager.retrievePackageJson();
+  const alreadyInstalledDependencies = [
```
merceyz, line 68: Should use a Set here as it will be faster

lib/cli/src/generators/baseGenerator.ts
phated: 👍 There is logic in CLI to do this that you can pull into a utility: https://github.com/storybookjs/storybook/blob/next/lib/cli/src/js-package-manager/JsPackageManager.ts#L11-L21

gaetanmaisse: I decided to check only the regular deps because if a package is already installed as devDep it will be updated without any issue (the problem is when you try to install regular dep as dev dep and dev dep as regular dep)

shilman (approved):
LGTM!

PR #13884, opened 2021-02-12
UI: Select search input value on /

ghengeveld:
Issue: -

## What I did

Besides setting focus to the search input, this also selects the input value so it can easily be replaced with a new query on subsequent searches.

## How to test


shilman (approved):
LGTM 👍

PR #13894, opened 2021-02-13
Addon-docs/Vue3: Attach app context from preview to inline stories

phated:
Issue: #13883

## What I did

I figured out how to attach an "app" context to a Vue 3 vnode that is rendered inline inside addon-docs. This solves the issue linked above. I'm unsure how this affects Vue 3 internals, but it seems to work in the example I added.

## How to test

- Does this need a new example in the kitchen sink apps? ✅ Added to the vue-3-cli example


shilman (approved):
LGTM!! 💯

PR #13912, opened 2021-02-15
Addon-actions: Change to override default values

tmeasday:
Issue: #12120

## What I did

Reverse current behaviour (to prefer the defaultValue) to instead prefer the action.

We think this can be considered a bugfix rather than a breaking change. Although the previous behaviour was intentional (all the way to having a test), it wasn't very well thought out, surprising to users, and AFAIK not documented anywhere.

## How to test

Yes see test


shilman (approved):
💯

PR #13916, opened 2021-02-15
Components: Add missing `regenerator-runtime` dependency

gaetanmaisse:
Issue: #13830

## What I did

- Add missing `regenerator-runtime` dependency
- Fix a typo in the E2E, for a few weeks the Yarn 2 PnP E2E tests were simply not running with Yarn 2.

## How to test

- CI should be 🟢 
- The logs in `e2e-tests-yarn-2-pnp` should match Yarn 2 ones


shilman (approved):
@gaetanmaisse LGTM! 🙏  does this need to get patched back to master? i don't think so, so removing `patch` label. feel free to correct me if i'm wrong.

PR #13937, opened 2021-02-17
Args: Prefer react runtime default values

tmeasday:
Issue: #12635

Telescoping on https://github.com/storybookjs/storybook/pull/13919

## What I did

Start by using the *actual* default prop as the default arg value. Only try unreliable eval-ing of the table summary if we don't have access to that.

## How to test

Yep, see affected stories

No

Not sure.


shilman (approved):
LGTM

PR #13954, opened 2021-02-18
fix: Vue 3 override args kitchen sink story

lee-chase:
Issue: #13861 #13920

## What I did

Changed the Vue 3 kitchen sink example Override Args, as the existing example could be mistaken for passing a reactive value back from setup. It also removes the need to adjust the template to receive a 2nd icon value (from `$props` and `icon`) and removes the possibility that Vue implementation could pass a string through with `$props.icon` when an object is expected.

## How to test

### DIFF
--- a/examples/vue-3-cli/src/stories/OverrideArgs.stories.js
+++ b/examples/vue-3-cli/src/stories/OverrideArgs.stories.js
@@ -30,11 +30,9 @@ const Template = (args, { argTypes }) => {
   return {
     props: Object.keys(argTypes),
     components: { OverrideArgs },
-    template: '<override-args v-bind="$props" :icon="icon" />',
-    setup(props) {
-      return {
-        icon: icons[props.icon],
-      };
+    template: '<override-args v-bind="$props" />',
+    setup(props, context) {
+      context.props.icon = icons[context.props.icon];
     },
   };
 };


phated:
Hmmm, I *think* `props` only exists on `context` because we `inject` it into a story component. We probably don't want to rely on this implementation detail to show this off.

phated:
@lee-chase can you explain when a story would be leveraging reactive-ness? Anytime an arg changes for a story, the Vue component is entirely re-rendered.

lee-chase:
My bad @phated, I saw what looked like a nail and hit it with a hammer.

Typically any value returned from the setup is to allow the template to access it. 

These are often reactive, and certainly, if I were computing a value based on a prop I would make it computed.

In the case of Storybook the Vue component gets re-rendered, which is not ideal from a test point of view, but it is how Storybook normally functions I believe. This often leads to story definitions and/or decorators being used to add functionality to a story so it can be interacted with without a full re-render. e.g. http://vue.carbondesignsystem.com/?path=/story/components-cvdropdown--vmodel

The above, aside, the code as-is does not look how I would expect to write a component using the OverrideArgs component. Passing icon back from setup also makes my brain do loops, thinking about the priority of 'props.icon' and the returned value

It may just be me, so feel free to close the issue.

phated:
I definitely don't want to close the issue and am very appreciative of your feedback! I don't write Vue myself, but brought the Vue 3 work for Storybook across the finish line. We are trying to engage Vue users, like you, to help us figure out what is the right/natural interface for Vue 3 users into Storybook. Thank you!

lee-chase:
Thanks for bearing with me.

In an ideal world Storybook would update props when controls change, instead of replacing the entire component. This would enable state in a Vue component to persist when a control changes and also enable calls to the onUpdated lifecycle hook. A bit beyond the scope for this discussion perhaps.

With regards to this PR. On reflection, returning something non-reactive is perfectly acceptable. I find it clearer if it is moved out of the return statement. Both of the following I find clearer, I would be more likely to write the latter.

```js
const Template = (args) => {
  const icon = icons[args.icon];
  return {
    props: Object.keys(OverrideArgs.props),
    components: { OverrideArgs },
    template: '<override-args v-bind="$props" :icon="icon"/>',
    setup() {
      return { icon };
    },
  };
};
```

```js
const Template = (args, { argTypes ) => {
  const newProps = { ...args, icon: icons[args.icon] };
  return {
    props: Object.keys(argTypes),
    components: { OverrideArgs },
    template: '<override-args v-bind="newProps"/>',
    setup() {
      return { newProps };
    },
  };
};
```

phated:
Aha! In your 2nd example, are the props bound to the "acceptable props" of the underlying component? In the Vue 3 docs, I read this: 
```
<!-- pass down parent props in common with a child component -->
<child-component v-bind="$props"></child-component>
```

So I assumed that `$props` had to be given to the `v-bind` directive for that "props mapping" to happen. Maybe that was incorrect.

lee-chase:
Any JSON object can be passed to v-bind. E.g. https://v3.vuejs.org/guide/migration/v-bind.html#_3-x-syntax

As you've found with icon, individual props can be passed after and merged.

The only reason for preferring 'newProps' over 'icon' is that all props and anything returned from setup are ordinarily available without any name spacing or value getter. So both versions of icon might be expected by an unwitting dev.

P.S. Also considered 'props: Object.keys(OverrideArgs.props)' but figured argTypes could have extra keys.

lee-chase:
@phated  Did you want me to update this PR with the 2nd example given above?

phated:
@lee-chase I wonder if it should ever be advised to use `v-bind="$props"` in a story. That actually came up in my original PR but we didn't have a good answer. Theoretically, we could have users do the `args` passing as you've shown, or maybe even inject it ourselves if they forget. Do you think that would be good/valuable?

lee-chase:
@phated Yes while use of $props is a good catch all, it is not one I have ever used. Probably as I'm a bit old school I prefer being verbose and would list out the props in the template.

From a clarity point of view I think having the user pass the args is much better in all cases, overridden or not. It makes the template less mysterious.

If props can be removed from context at the same time, it's a bonus.

phated:
If we change the style for writing Vue 3 stories so that `setup()` must always be specified, users can name the args anyway they want, including something like this:
```js
// override args story
const template = `<cv-button @click="onClick" v-bind="args">{{defaultSlot}}</cv-button>`;
const Template = (args, { argTypes }) => {
  args = { ...args, icon: icons[args.icon] }; // replace the local variable with the override (no mutation needed)
  return {
    props: Object.keys(argTypes),
    components: { CvButton },
    setup() {
      return { args, onClick: action('click') };
    },
    template,
  };
};
```

I think that is really cool, because then a template can be written very generically and it's up to the story author to do the mapping from story args to template "values".

This would also allow us to delete a **lot** of code that was added to inject props into the tree.

I'll check what the process is for making this change (maybe I need to write up a proposal, etc) and get back to you!

You've been a tremendous help 🙇 If you ever want to join the `#vue3` channel of the [Storybook discord](https://discord.gg/storybook), we'd love to have you as part of the ongoing discussion!

phated:
Closed by #13981 - Thank you tremendously for all the feedback @lee-chase! It feels much better to write Vue 3 stories now 🎊

PR #13973, opened 2021-02-20
Remind to add a new script for running storybook

GitCodeDestroyer:
Every time, all new users of storybook stuck at here. Please, add some instruction to adding a script for running npm run storybook.

Issue:

## How to test

### DIFF
--- a/docs/get-started/install.md
+++ b/docs/get-started/install.md
@@ -46,7 +46,19 @@ The command above will make the following changes to your local environment:
 - 🛠 Add the default Storybook configuration.
 - 📝 Add some boilerplate stories to get you started.
 
-Depending on your framework, first build your app and then check that everything worked by running:
+Depending on your framework, first build your app and then check that everything worked. Add a script into a package.json:
+
+```javascript
+{
+  "scripts": {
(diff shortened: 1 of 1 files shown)


shilman (changes requested):
If you run `npx sb init` which is the way we recommend installing Storybook, it will automatically add this to your `package.json`, so I'm not sure why we need this?

phated:
I tend to agree with @shilman that the documentation should be clear on the recommended path `npx sb init` which will already include the script.

jonniebigodes:
@GitCodeDestroyer thank you for the time and effort you put into this, but i'm going to close this. I'm with @shilman on this as well. We instruct users to use `npx sb init` for the optimal experience. As it will streamline the workflow for you. It adds Storybook and its dependencies, adjusts your project scripts accordingly and also some examples to get you started. If your specific case, or if you're using a framework /meta-framework (Gatsby for instance) that is not covered, or requires manual intervention feel free to open a discussion and we'll go from there.

Stay safe

GitCodeDestroyer:
OK, I got it.

PR #14010, opened 2021-02-22
FAQ: Make link to symbols more apparent

bodograumann:
I was first confused why there are no icon previews in the FAQ.
Only while thinking about opening a bug and looking at the raw source did I notice that there is a link to the icons below the table.

By moving it in front of the table it is more obvious that the table is not supposed to contain icon previews.

Also btw: The addon pages, e.g. https://storybook.js.org/addons/@storybook/addon-toolbars, contain the main storybook readme, not the addon specific documentation. You should probably take a look on that ;-)


docs/workflows/faq.md
```diff
@@ -188,6 +188,7 @@ With the release of version 5.3 we've updated how you can write your stories to
 ### What icons are available for my toolbar or my addon?
 
 With the [`@storybook/components`](https://www.npmjs.com/package/@storybook/components) package you get a set of icons that you can use to customize your own UI. Use the table below as a reference while writing your own addon or defining your own Storybook global types.
+Go through this [story](https://5a375b97f4b14f0020b0cda3-wbeulgbetj.chromatic.com/?path=/story/basics-icon--labels) to see how the icons look like.
```
saschwarz, line 191:
```suggestion
Go through this [story](https://5a375b97f4b14f0020b0cda3-wbeulgbetj.chromatic.com/?path=/story/basics-icon--labels) to see how the icons look.
```
While this is being edited... I think this is more idiomatic English.

phated (approved):
LGTM! Burying the link below the table is definitely weird.

jonniebigodes:
I'm good with it. Thanks for the time and effort you put into this pull request @bodograumann.

Self merging this @shilman

PR #14018, opened 2021-02-22
CLI/Vue 2: install vue-loader upon init of vue 2 storybook

phated:
Issue: #14011

## What I did

**I'm not sure if we should do this or not.**

Essentially, the `@storybook/vue` package has a peerDep on `vue-loader` and errors if someone doesn't install it (as was noted in the linked issue). It seems that some versions of the Vue CLI don't use webpack and don't install this dependency, so we could specify this as an `extraPackages` so it is always installed by `sb init`. Thoughts?

## How to test

### DIFF
--- a/lib/cli/src/generators/VUE/index.ts
+++ b/lib/cli/src/generators/VUE/index.ts
@@ -1,7 +1,9 @@
 import { baseGenerator, Generator } from '../baseGenerator';
 
 const generator: Generator = async (packageManager, npmOptions, options) => {
-  baseGenerator(packageManager, npmOptions, options, 'vue');
+  baseGenerator(packageManager, npmOptions, options, 'vue', {
+    extraPackages: ['vue-loader@^15.7.0'],
+  });
 };
 
 export default generator;


shilman (approved):
Makes sense to me. Thanks for grabbing this! 🙏

shilman (commented):
Hmm, then again by that logic why not install all the peer deps? Where to draw the line? 🤔 

```
  "peerDependencies": {
    "@babel/core": "*",
    "css-loader": "*",
    "ts-loader": "^6.2.2",
    "vue": "^2.6.8",
    "vue-loader": "^15.7.0",
    "vue-template-compiler": "^2.6.8"
  },
```

pocka (approved):
Looks good to me but I couldn't find

> It seems that some versions of the Vue CLI don't use webpack and don't install this dependency,

the changes or issues in https://github.com/vuejs/vue-cli...

phated:
> Hmm, then again by that logic why not install all the peer deps? Where to draw the line?

I don't have an answer for this, that's why I didn't know if we should do it. For Vue 3, I added the loader because you **have** to have v16 which is tagged `@next` on npm. It was the only extra dependency that I needed to layer onto a fresh "Vue CLI"-created project to get it to run.

PR #14049, opened 2021-02-25
Core: Support null and undefined in URL args

ghengeveld:
Issue: -

## What I did

This adds support for `null` and `undefined` arg values, so that removing an object property or array value in a JSON value will get properly synced to/from the URL.

Additionally, this fixes encoding of the space character as `+` rather than `%20`.

## How to test

### DIFF
--- a/lib/client-api/src/args.ts
+++ b/lib/client-api/src/args.ts
@@ -4,7 +4,7 @@ import { isPlainObject } from 'lodash';
 type ValueType = { name: string; value?: ObjectValueType | ValueType };
 type ObjectValueType = Record<string, ValueType>;
 
-export const INCOMPATIBLE = Symbol('incompatible');
+const INCOMPATIBLE = Symbol('incompatible');
 const map = (arg: unknown, type: ValueType): any => {
   if (arg === undefined || arg === null) return arg;
   switch (type?.name) {
--- a/lib/client-api/src/args.ts
+++ b/lib/client-api/src/args.ts
@@ -1,9 +1,12 @@
 import { Args, ArgTypes } from '@storybook/addons';
+import { isPlainObject } from 'lodash';
 
 type ValueType = { name: string; value?: ObjectValueType | ValueType };
 type ObjectValueType = Record<string, ValueType>;
 
-const map = (arg: any, type: ValueType): any => {
+export const INCOMPATIBLE = Symbol('incompatible');
+const map = (arg: unknown, type: ValueType): any => {
+  if (arg === undefined || arg === null) return arg;
   switch (type?.name) {
     case 'string':
       return String(arg);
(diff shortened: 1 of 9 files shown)


lib/core-client/src/preview/parseArgsParam.ts
```diff
@@ -1,20 +1,34 @@
...
-  if (key === null || value === null) return false;
+const validateArgs = (key = '', value: unknown): boolean => {
+  if (value === null || value === undefined) return true; // encoded as `!null` or `!undefined`
+  if (key === null) return false;
   if (key === '' || !VALIDATION_REGEXP.test(key)) return false;
+  if (typeof value === 'number') return true;
   if (typeof value === 'string') return VALIDATION_REGEXP.test(value);
-  if (Array.isArray(value)) return value.every((v) => validateArgs(key, v));
-  return Object.entries(value).every(([k, v]) => validateArgs(k, v));
+  if (Array.isArray(value)) return !value.some((v) => !validateArgs(key, v));
```
coderkevin, line 14: I'm guessing this line was changed for performance reasons, so it short-circuits out on the first failure?

lib/core-client/src/preview/parseArgsParam.ts
```diff
@@ -1,20 +1,34 @@
...
+  if (isPlainObject(value)) return !Object.entries(value).some(([k, v]) => !validateArgs(k, v));
+  return false;
```
coderkevin, line 16: I'm curious as to what conditions would get this far in the validation check.

lib/router/src/utils.ts
```diff
@@ -37,27 +38,48 @@ interface Args {
...
   if (typeof value !== typeof update) return update;
-  if (Array.isArray(value)) return update;
-  if (update && typeof update === 'object') {
-    return Object.keys(update).reduce((acc, key) => {
-      const diff = deepDiff(value[key], update[key]);
-      return diff === undefined ? acc : Object.assign(acc, { [key]: diff });
+  if (deepEqual(value, update)) return DEEPLY_EQUAL;
+  if (Array.isArray(value) && Array.isArray(update)) {
+    return value.length > update.length
+      ? [...new Array(value.length)].map((_, i) => update[i])
```
coderkevin, line 47: I think the spread into a new array is unnecessary here and creating two new arrays instead of one.

Also, I realize the implicit behavior of accessing an out-of-bounds index on an array returned `undefined`, but I think something like this could be a bit more explicit:
`[...update, ...Array(value.length - update.length).fill(undefined)]`

lib/router/src/utils.ts
```diff
@@ -37,27 +38,48 @@ interface Args {
...
+export const deepDiff = (value: unknown, update: unknown): any => {
   if (typeof value !== typeof update) return update;
-  if (Array.isArray(value)) return update;
-  if (update && typeof update === 'object') {
-    return Object.keys(update).reduce((acc, key) => {
-      const diff = deepDiff(value[key], update[key]);
-      return diff === undefined ? acc : Object.assign(acc, { [key]: diff });
+  if (deepEqual(value, update)) return DEEPLY_EQUAL;
+  if (Array.isArray(value) && Array.isArray(update)) {
+    return value.length > update.length
```
coderkevin, line 46: I wonder if an additional "early return if" might be more readable than a multi-line ternary:
`if (update.length >= value.length) return update;`

lib/client-api/src/args.ts
```diff
@@ -1,9 +1,12 @@
 import { Args, ArgTypes } from '@storybook/addons';
+import { isPlainObject } from 'lodash';
 
 type ValueType = { name: string; value?: ObjectValueType | ValueType };
 type ObjectValueType = Record<string, ValueType>;
 
-const map = (arg: any, type: ValueType): any => {
+export const INCOMPATIBLE = Symbol('incompatible');
```
coderkevin, line 7: This doesn't need to be exported as it's never returned from an exported function.

coderkevin (approved):
Just a few optional comments on this. In general, it looks good to me! 👍

lib/client-api/src/args.ts
```diff
@@ -12,24 +15,40 @@ const map = (arg: any, type: ValueType): any => {
...
 export const mapArgsToTypes = (args: Args, argTypes: ArgTypes): Args => {
   return Object.entries(args).reduce((acc, [key, value]) => {
-    return Object.assign(acc, { [key]: map(value, argTypes[key]?.type) });
+    const mapped = map(value, argTypes[key]?.type);
+    return mapped === INCOMPATIBLE ? acc : Object.assign(acc, { [key]: mapped });
   }, {});
 };
+
+export const combineArgs = (value: any, update: any): Args => {
+  if (isPlainObject(value) && isPlainObject(update)) {
```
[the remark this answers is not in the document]
ghengeveld: I agree, that's my preference too. In this case I anticipated we would have another conditional block to deal with arrays, but that turned out not to be necessary. At least not until we start merging arrays rather than overwriting them, which might happen when we start to use sparse arrays in the URL (which the next version of `qs` should support).

lib/client-api/src/args.ts
```diff
@@ -12,24 +15,40 @@ const map = (arg: any, type: ValueType): any => {
...
+    return Object.keys({ ...value, ...update }).reduce((acc, key) => {
+      if (key in update) {
+        const combined = combineArgs(value[key], update[key]);
+        if (combined !== undefined) acc[key] = combined;
```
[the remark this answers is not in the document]
ghengeveld: In this specific example, there is no `myProp in update`, so it falls through to the else clause below, where it sets `acc['myProp'] = undefined`.

lib/client-api/src/args.ts
```diff
@@ -12,24 +15,40 @@ const map = (arg: any, type: ValueType): any => {
...
+      } else {
+        acc[key] = value[key];
```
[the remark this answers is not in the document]
ghengeveld: We could do that, but then we'd have to `delete` keys from it (where the update is undefined). I'm not sure it would be simpler to understand. Note that I only used the object spread merge to get a union of the keys, not the values.

lib/core-client/src/preview/parseArgsParam.ts
ghengeveld: Non-POJOs (Date, Error, Promise, etc) and functions, I guess. We can't do anything with those. We have a separate control type for dates, which encodes to an ISO string.

lib/router/src/utils.ts
ghengeveld: Good call.

lib/router/src/utils.ts
ghengeveld: I like it. We don't even need the `fill` because undefined is what you get when spreading a sparse array.

lib/core-client/src/preview/parseArgsParam.ts
ghengeveld: I'm pretty sure the performance of `every(true)` is the same as `some(false)`, because both probably short circuit on the first failure

I honestly can't recall why I changed this. When I change it back the tests still pass, so I'll revert this.

PR #14054, opened 2021-02-25
Fix flaky color rendering

ghengeveld:
Issue: -

## What I did

This should fix the [flaky text color rendering](https://www.chromatic.com/test?appId=5a375b97f4b14f0020b0cda3&id=60148cecb5cbbe00215f0156) in Chromatic for the iframe zoom stories.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, should stop being flaky now and always render `mediumdark` rather than black text.


domyen:
LGTM! Thanks for the fix 🙇‍♂️

PR #14062, opened 2021-02-26
UI: Enable search for stories and fix `/` event listener

ghengeveld:
Issues: #13943 #13877 #13463

## What I did

This enables searching for story names. When both a story and component are found and they're parent/child, only the first one is displayed (whichever ranks higher), to avoid multiple results that lead to the same place.

Additionally, this fixes the event listener for `/`. Turns out we had two competing listeners. I removed the custom one and fixed the core one to also select the input text.

## How to test


shilman (approved):
Awesome 😍

PR #14067, opened 2021-02-26
Core: Namespace sidebar config options

ghengeveld:
Issue: -

## What I did

This deprecates `showRoots` in favor of `sidebar.showRoots`, moves the recently introduced `collapsedRoots` (beta) to be under `sidebar` as well and removes `sidebarAnimations` altogether.

Should we show a deprecation warning for `sidebarAnimations` as well? It's not going to break, but people may want to clean up their config.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes, nothing should change
- Does this need an update to the documentation? Yes, it's there

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -9,6 +9,7 @@
   - [6.2 Deprecations](#62-deprecations)
     - [Deprecated implicit PostCSS loader](#deprecated-implicit-postcss-loader)
     - [Deprecated default PostCSS plugins](#deprecated-default-postcss-plugins)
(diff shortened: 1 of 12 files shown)


shilman (approved):
This is great @ghengeveld -- thanks for cleaning up 🙏

nhoizey:
I lost a few minutes trying to find the documentation for `sidebarAnimations` I had in a project, so I would vote for a deprecation warning and at leat a mention in `MIGRATION.md`

PR #13733, opened 2021-01-26
Move to a fork of react-sizeme with updated React peer dependency

Hypnosphi:
One more step to eliminating peer dependency warnings

Diff with the original package: https://github.com/ctrlplusb/react-sizeme/pull/201/files

PR #14069, opened 2021-02-26
Move back to the original react-sizeme package

Hypnosphi:
See #13733 

They now have React 17 in peer deps


shilman (approved):
LGTM. thanks!

PR #14074, opened 2021-02-26
Lower z-index on "Show code" button

kylemh:
Issue:

## What I did
Removed set z-index from the div that wraps the "Show code" button in a story preview block. I don't wanna get into a fight with z-indeces anywhere - least of all within Storybook. My modals are rendered via portals which should be last in the DOM beating any z indeces; however, in Storybook, those portals are rendered within the preview block's iFrame, meaning they aren't rendered last.

### Before
Screen Shot 2021-02-26 at 12 55 44 PM

### After
Screen Shot 2021-02-26 at 12 55 32 PM

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200539748475788/1200567473293920) by [Unito](https://www.unito.io)

### DIFF
--- a/lib/components/src/ActionBar/ActionBar.tsx
+++ b/lib/components/src/ActionBar/ActionBar.tsx
@@ -9,7 +9,6 @@ const Container = styled.div<{}>(({ theme }) => ({
   maxWidth: '100%',
   display: 'flex',
   background: theme.background.content,
-  zIndex: 1,
 }));
 
 export const ActionButton = styled.button<{ disabled: boolean }>(
--- a/lib/components/src/ActionBar/ActionBar.tsx
+++ b/lib/components/src/ActionBar/ActionBar.tsx
@@ -9,6 +9,7 @@ const Container = styled.div<{}>(({ theme }) => ({
   maxWidth: '100%',
   display: 'flex',
   background: theme.background.content,
+  zIndex: 0,
 }));
 
 export const ActionButton = styled.button<{ disabled: boolean }>(


shilman (changes requested):
Looks like zIndex was added for this:

https://github.com/storybookjs/storybook/pull/11502#issuecomment-663774680

Can you investigate?

kylemh:
@domyen that video is 404ing from Loom. Do you remember the problem?

kylemh:
@ndelangen so I am using portals, but they render at the bottom of the body within the iFrame, meaning the "Show code" button appears after it in the page's DOM, so that's not a solution.

MichaelArestad:
Right. I did a few tests and I can confirm that the actions can't be used in some scenarios without the add `z-index: 1`. It looks like there is a `div` with the class`os-padding` that has a `z-index: 1` which makes those actions inaccessible when their `z-index` is removed.

I tried changing the `os-padding` `z-index` to `0` and that seemed to work, but may break something else. I'm not familiar with what is adding those styles, but I think it has something to do with scrollbars. @kylemh perhaps investigating that route further could prove fruitful?

kylemh:
@MichaelArestad zIndex 0 solved my issue. Committed here so you can see the result in terms of Storybook (rather than just my Storybook usage).

MichaelArestad:
@kylemh Right. I tried that yesterday. You can see [on the Docs/DocsPage story in the published storybook](https://5a375b97f4b14f0020b0cda3-tqlmlclwkk.chromatic.com/?path=/story/docs-docspage--with-subtitle) that the "Code" button is no longer clickable.

MichaelArestad:
Because this is breaking functionality and seems to be inactive, I'm going to close this PR. If you would like to try another iteration, I would be happy to reopen it. Thank you for contributing.

PR #14077, opened 2021-02-27
Add readonly option

Vadorequest:
Issue: https://github.com/storybookjs/storybook/issues/14048

## How to test

### DIFF
--- a/lib/components/src/blocks/ArgsTable/ArgControl.tsx
+++ b/lib/components/src/blocks/ArgsTable/ArgControl.tsx
@@ -22,7 +22,7 @@ const NoControl = () => <>-</>;
 
 export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
   const { key, control } = row;
-  console.log('row', row);
+
   const [isFocused, setFocused] = useState(false);
   // box because arg can be a fn (e.g. actions) and useState calls fn's
   const [boxedValue, setBoxedValue] = useState({ value: arg });
@@ -43,8 +43,10 @@ export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
   const onBlur = useCallback(() => setFocused(false), []);
   const onFocus = useCallback(() => setFocused(true), []);
 
-  if (control?.disable && control?.readonly) {
-    console.warn(`Both "disable" and "readonly" options were defined, applying "disable".`);
+  if (control?.disable && control?.readOnly) {
+    console.warn(
+      `Both "disable" and "readOnly" options were defined. The "disable" option takes precedence.`
+    );
   }
 
   if (!control || control.disable) return <NoControl />;
--- a/lib/components/src/blocks/ArgsTable/ArgControl.tsx
+++ b/lib/components/src/blocks/ArgsTable/ArgControl.tsx
@@ -22,7 +22,7 @@ const NoControl = () => <>-</>;
 
 export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
   const { key, control } = row;
-
+  console.log('row', row);
   const [isFocused, setFocused] = useState(false);
   // box because arg can be a fn (e.g. actions) and useState calls fn's
   const [boxedValue, setBoxedValue] = useState({ value: arg });
@@ -42,6 +42,10 @@ export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
 
   const onBlur = useCallback(() => setFocused(false), []);
   const onFocus = useCallback(() => setFocused(true), []);
+
+  if (control?.disable && control?.readonly) {
+    console.warn(`Both "disable" and "readonly" options were defined, applying "disable".`);
+  }
 
   if (!control || control.disable) return <NoControl />;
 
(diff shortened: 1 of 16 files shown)


lib/components/src/blocks/ArgsTable/ArgControl.tsx
```diff
@@ -22,7 +22,7 @@ const NoControl = () => <>-</>;
 
 export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
   const { key, control } = row;
-
+  console.log('row', row);
```
shilman, line 25:
```suggestion
```

shilman (changes requested):
Looking great. Can you please add some examples to `official-storybook`? And some documentation to 

https://github.com/storybookjs/storybook/blob/next/docs/essentials/controls.md

Vadorequest:
The current implementation doesn't work on `storybook-official` story.

image

When I add a console.log in `lib/components/src/controls/Text.tsx`, it shows the expected value.

```tsx
export const TextControl: FC<TextProps> = ({
  name,
  value,
  onChange,
  onFocus,
  onBlur,
  readOnly,
}) => {
  console.log('TextControl readOnly', readOnly) // Here, displays "true"
  const handleChange = (event: ChangeEvent<HTMLTextAreaElement>) => {
    onChange(event.target.value);
  };
```

But it only does so when clicking on the "Docs" tab, not on the "Canvas" tab.
I must be missing something.

Vadorequest:
@shilman Documentation is done, could you look at the implementation and let me know if you notice anything wrong? (it's still not working)

lib/components/src/controls/types.ts
```diff
@@ -9,6 +9,7 @@ export interface ControlProps<T> {
   onChange: (value: T) => T | void;
   onFocus?: (evt: any) => void;
   onBlur?: (evt: any) => void;
+  readOnly?: boolean;
```
Vadorequest, line 12: This doesn't look right to me, it fixes TS types in the various control option components, but I'm not sure it's correct.

Also, if it's correct, the `disable` property should be added there, too.

shilman:
Sure, I'll take a look today!

ghengeveld:
In light of https://github.com/storybookjs/storybook/pull/13890, shouldn't this property be one level up?

```js
argTypes: {
  foo: {
    table: { disable: true },    // deprecated
    disable: true,               // the new way to disable (hide) an arg
    control: { readonly: true }, // implemented in this PR
    readonly: true               // I suggest we do this instead
  }
}
```

I favor `readonly` over `readOnly` because of simplicity and being framework agnostic. `readOnly` seems like a React inspired prop name, but `readonly` (like the HTML attribute) is more common. In that regard I would've also preferred `disabled` over `disable`.

Vadorequest:
@ghengeveld I agree about the `readonly` naming, forgot storybook isn't react-specific.

Does the `readonly` option only apply to the `control`? Or could it be applied to other things? I don't really have an opinion whether `readonly` or `control.readonly` is better, but I wonder if it's an option that's specific to the `control`, or broader than that. Currently, my understanding is that it's specific to `control` and should be configured within it.

And, same goes for `disable`, I agree `control.table.disable` is too verbose, but I don't know if the `disable` option only disables it in the control, or if it's broader than that. And I think the final choice should be based on that.

ghengeveld:
I had a discussion with @shilman about this, and we concluded that my earlier suggestion is the right way to go. For this PR, that means:

- `readonly` should be a property on the arg itself, not on `control`.
- The `readonly` property should be respected by everything, not just Controls. That means it needs to be handled in the story store. Specifically, `mapArgsToTypes` needs to omit any `readonly` args (probably by flagging it `INCOMPATIBLE`): https://github.com/storybookjs/storybook/blob/4d50e9a8ee588e195e6582f7855755b1f524e445/lib/client-api/src/args.ts#L34-L39

We'll deal with `disable` separately.

Vadorequest:
I feel like I should probably wait for https://github.com/storybookjs/storybook/pull/13890 to be merged.

I'm glad TS types are being improved, it'll be easier to understand.

ndelangen:
@Vadorequest Thank you so soo much for this PR! We're really appreciative of any PR we get, because we know it takes a lot of time, effort and energy from contributors. This feature seems like something we want.

This has been in draft for over a year now, and the codebase has changed a lot since then, and with the upcoming changes for 7.0 that work would probably be even more complex.
Possibly not impossible, so please let me know if you're interested in re-opening this PR. But I'll close for now.

Again, we truly appreciate your time, energy and skill that you put into this PR! Thank you. I'm sorry for closing.

Vadorequest:
Thansk for the heads-up - I had totally forgotten about this. 

I won't work on it again, someone else should take the lead if you want this to be implemented :)

LeaveNhA:
I understand this PR has closed, but have you ever thinking anything about this requirement? 
It's a real-world requirement and I believe a lot of people might need it --like me, now--.

I will inspect the PR and if I wanna volunteer, I will inform you.

MaliJetal:
Hi,
I really need this change will you please repopen this issue or create one more. If you want i can create a PR.
@ndelangen @LeaveNhA @Vadorequest

PR #14080, opened 2021-02-28
Svelte: Fix async loaders in docs panel

j3rem1e:
Issue:

## What I did

Async loaders work in the preview panel, but not in the documentation panel. This PR fixes it.

However, this PR doesn' fix  tests with storyshot - it's another issue and there is already a PR for that.

## How to test


shilman (approved):
Looking good 🏅

PR #14098, opened 2021-03-02
Add Grafana UI to design systems list

tskarhed:
## What I did

Added Grafana UI to Design systems examples. I think Grafana fulfills the criteria to be on that list, but I'm also biased. Grafana UI is used by the Grafana core team and is a basis for all the plugin developers. It is a tool for the Grafana community, and the Storybook has been greatly appreciated by the teams working with plugins and UI development. I think it is a prime example of how Storybook can help with open source development and make it easier for people to contribute.


jonniebigodes:
@tskarhed thank you very much for the contribution and we're thrilled to see that the team over at Grafana uses Storybook as part of its development workflow. 

Self merging this

PR #14151, opened 2021-03-05
Chore: (Docs) Adjusts layout parameter documentation

jonniebigodes:
With this pull request the configure/story layout documentation is adjusted to mention the layout parameter is not exclusive to  a global approach within `.storybook/preview.js` but it can also be applied on a story basis.

What was done:
- Adjusted `configure/story-layout/`
- Included a couple of examples to demonstrate how to use within a story.

@shilman let me know if this is ok with you.

### DIFF
--- a/docs/configure/story-layout.md
+++ b/docs/configure/story-layout.md
@@ -2,7 +2,9 @@
 title: 'Story layout'
 ---
 
-The `layout` [global parameter](../writing-stories/parameters.md) allows you to configure how stories are positioned in Storybook's Canvas tab. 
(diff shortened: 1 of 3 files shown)


shilman (approved):
@jonniebigodes I would probably just point back to the central parameters documentation and mention that parameters can be set globally, at the component level, or at the story level, rather than duplicating the code/examples every time we want to emphasize that (much more to maintain).

However, this change gets the job done, so I'll let you decide whether to merge or make the change.

ndelangen (approved):

PR #14181, opened 2021-03-09
Chore: (Docs) Adjusts links to the addons gallery

jonniebigodes:
With this pull request, the documentation is updated to point readers to the addons gallery instead of the addon location in the monorepo and with that prevent 404 and issues with branches, such as mentioned in #13347 for instance.

@shilman one thing caught my eye when adjusting the links. It seems that for instance, the Docs addon in the gallery is pulling in Storybook readme as opposed to its own readme file. 
Also one other thing. It seems that searching for `storyshots` in the gallery is only pulling in the puppeteer version of it. Not the `@storybook/addon-storyshots` (searched for that specific and some permutations and didn't pan out).

Feel free to provide feedback

### DIFF
--- a/docs/writing-docs/docs-page.md
+++ b/docs/writing-docs/docs-page.md
@@ -9,7 +9,7 @@ title: 'DocsPage'
 </div>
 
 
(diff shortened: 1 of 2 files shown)


shilman (approved):
💯

PR #14228, opened 2021-03-13
Mention selectors requirement in Angular stories

literalpie:
Migration documentation update to communicate that some changes may be needed for Angular Stories. As discussed [here](https://github.com/storybookjs/storybook/pull/13939#issuecomment-796186037)

The asterisks on the `entryComponent` approach is kind of making me wonder if we should just give a better console warning if no component selector is provided. In our project, we will probably end up adding the selectors since it's the same amount of changes as `entryComponent` and will work more completely (with inputs).


gaetanmaisse:
@ThibaudAV does it look good to you?

ThibaudAV:
Yeah, LGTM 🚀

ThibaudAV (approved):

shilman (approved):
LGTM!!

PR #14258, opened 2021-03-17
WebComponents: Fix types and improve CLI detection

gaetanmaisse:
Related to https://github.com/storybookjs/storybook/pull/12395#issuecomment-800632234

## What I did

 - use types from `lit-html` instead of `lit-element` to avoid dependency issues for `lit-html`-only users
 - remove unneeded `lit-html` dep and keep only `lit-element` in config used to run WC E2E test
 - add detection of bare `lit-html` projects (and not only `lit-element` ones) in SB CLI

## How to test

- CI should be 🟢 
- Using this branch locally with yarn link should allow you to make SB runs in a bare `lit-html` project, and `sb init` should properly detect a WC setup.

### DIFF
--- a/lib/cli/src/detect.test.ts
+++ b/lib/cli/src/detect.test.ts
@@ -160,6 +160,16 @@ const MOCK_FRAMEWORK_FILES = [
     },
   },
   {
+    name: ProjectType.WEB_COMPONENTS,
+    files: {
+      'package.json': {
+        dependencies: {
+          'lit-html': '1.0.0',
+        },
+      },
+    },
+  },
+  {
     name: ProjectType.MITHRIL,
     files: {
       'package.json': {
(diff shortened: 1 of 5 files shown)


lib/cli/src/detect.test.ts
```diff
@@ -159,6 +159,16 @@ const MOCK_FRAMEWORK_FILES = [
       },
     },
   },
+  {
+    name: ProjectType.WEB_COMPONENTS,
+    files: {
+      'package.json': {
+        dependencies: {
+          'lit-html': '1.0.0',
```
ndelangen, line 167: should this be an exact dep?

gaetanmaisse: It's just for the test fixture ;) any version is ok

ndelangen: ok

ndelangen (approved):

ndelangen (approved):

PR #14255, opened 2021-03-17
Webpack5: Fix hash => fullhash

phaistonian:
Issue: #14253 

## What I did

Changed `[hash:..]` with `[fullhash:...]`.

PR #14261, opened 2021-03-17
Webpack5: Replace fullhash with contenthash

7rulnik:
According to https://webpack.js.org/configuration/output/#template-strings `fullhash` is hash of compilation (and `hash` was too). So it doesn't make sense to use it if we want some unique names for assets.

https://github.com/storybookjs/storybook/issues/14253#issuecomment-801056885
https://github.com/storybookjs/storybook/pull/14255

BTW we shouldn't use `contenthash` id dev mode https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling, but I believe it's another issue


shilman (approved):
LGTM! Thanks @7rulnik 🙏

khats:
@shilman should it be done for this line as well?
https://github.com/storybookjs/storybook/blob/908d3b001780274767a5b1c6c73c581d637b107a/lib/builder-webpack5/src/preview/iframe-webpack.config.ts#L124

PR #14279, opened 2021-03-19
Update web-component reload and updated kitchen-sink config 

brion-fuller:
Issue:

Web components cannot be hot reloaded as customElement.define can only be called once per page with an unique string.  This has been on the users to add to their config file.  Now this is available for free.

## What I did

I hooked into server-sent events for webpack HMR and listen for the built event in which I trigger a page reload.

## How to test


brion-fuller:
Going to look at some other solutions

brion-fuller:
@shilman @gaetanmaisse @yannbf This is the quick fix version of hot reloading the page for components

PR #14293, opened 2021-03-21
Core: Support some special values in URL args

ghengeveld:
Issue: -

## What I did

- [x] Support for sparse arrays
- [x] Support for Date objects
- [x] Support for hex colors
- [x] Support for rgb(a) colors
- [x] Support for hsl(a) colors
- [x] Update hex and date format according to feedback

## How to test


lib/router/src/utils.test.ts
```diff
@@ -178,6 +188,26 @@ describe('buildArgsParam', () => {
     expect(param).toEqual('foo.bar[0].key:!null;foo.baz:!null');
   });
 
+  it('encodes hex color values by replacing # with !', () => {
+    const param = buildArgsParam({}, { key: '#ff4785' });
+    expect(param).toEqual('key:!ff4785');
```
tmeasday, line 193: I feel like you might regret this format (to general, will limit future options). Do you think maybe it's better to do something more formal/deterministic like: `!hex(ff4785)`

lib/router/src/utils.test.ts
```diff
@@ -178,6 +188,26 @@ describe('buildArgsParam', () => {
...
+  });
+
+  it('encodes hsla color values by prefixing and compacting', () => {
+    const param = buildArgsParam({}, { hsl: 'hsl(45, 99%, 70%)', hsla: 'hsla(45, 99%, 70%, 0.5)' });
+    expect(param).toEqual('hsl:!hsl(45,99,70);hsla:!hsla(45,99,70,0.5)');
+  });
+
+  it('encodes Date objects as prefixed ISO string', () => {
+    const param = buildArgsParam({}, { key: new Date('2001-02-03T04:05:06.789Z') });
+    expect(param).toEqual('key:!2001-02-03T04:05:06.789Z');
```
tmeasday, line 208: Similarly, I feel like rather than matching on a regexp, why not just `!date(2001-02-03T04:05:06.789Z)` or something?

tmeasday (changes requested):
Code LGTM, I have some thoughts about the wire format and also I think we need one more set of tests.

lib/router/src/utils.test.ts
ghengeveld: Yeah sounds good!

lib/router/src/utils.test.ts
ghengeveld: Agreed.

shilman (approved):
Pretty nice!!!

PR #14296, opened 2021-03-21
Controls: Add automatic ids to all controls

YozhEzhi:
Issue: https://github.com/storybookjs/storybook/issues/13929

## What I did

Added `id` for components based on `name` prop.

## How to test

I've tested it with Chrome Dev Tools in running Storybook app.

## Questions

_- Is this testable with Jest or Chromatic screenshots?_

I think snapshots should be updated.

_- Does this need a new example in the kitchen sink apps?_

I don't know, because I don't familiar with kitchen, sorry.

_- Does this need an update to the documentation?_

No, it doesn't.


lib/components/src/controls/Date.tsx
```diff
@@ -99,15 +99,15 @@ export const DateControl: FC<DateProps> = ({ name, value, onChange, onFocus, onB
         type="date"
         max="9999-12-31" // I do this because of a rendering bug in chrome
         ref={dateRef as RefObject<HTMLInputElement>}
-        id={`${name}date`}
-        name={`${name}date`}
+        id={`${name}-date`}
+        name={`${name}-date`}
```
YozhEzhi, line 103: I've noticed that `Checkbox`'s and `Radio`'s `id` is formed as `${name}-${key}`, so I decided change `id` and `name` here for consistency. I'll revert it if you tell me.

shilman (changes requested):
Thanks for contributing this! We're in a feature freeze for 6.2, so we'll need to pick this up after 6.3 starts.

In the meantime, the primary issue I see is that AFAIK `name` can contain spaces and spaces are not allowed in ID attributes http://xahlee.info/js/html_allowed_chars_in_attribute.html

WDYT?

YozhEzhi:
> In the meantime, the primary issue I see is that AFAIK `name` can contain spaces and spaces are not allowed in ID attributes http://xahlee.info/js/html_allowed_chars_in_attribute.html
> 
> WDYT?

Yes. Looks like it's a good idea to improve `id` rendering.

What about to following `kebab-case` format for `id`?
Example: `<input id="my-cool-prop" />`, `<input id="my-cool-date-time" />`.

Also, would it be better to add some common prefix to all control `id`'s?

Example: `<input id="controls-my-cool-prop" />`, `<input id="controls-my-cool-date-time" />`.

`Id` becomes more unique. Or there is no need to increase `id`'s uniqueness?

shilman:
@YozhEzhi your proposal, including the prefix, sounds good to me!

examples/angular-cli/.storybook/preview.ts
```diff
@@ -4,7 +4,6 @@ import { prepareForInline } from '@storybook/addon-docs/angular/inline';
 import addCssWarning from '../src/cssWarning';
 
 // @ts-ignore
-// eslint-disable-next-line import/extensions, import/no-unresolved
```
YozhEzhi, line 7: Pre-commit hook fails on this O_o

examples/angular-cli/.storybook/preview.ts
```diff
@@ -1,9 +1,10 @@
 import { setCompodocJson } from '@storybook/addon-docs/angular';
 import addCssWarning from '../src/cssWarning';
 
+/* eslint-disable import/extensions, import/no-unresolved */
 // @ts-ignore
-// eslint-disable-next-line import/extensions, import/no-unresolved
 import docJson from '../documentation.json';
+/* eslint-enable import/extensions, import/no-unresolved */
```
YozhEzhi, line 7: @shilman What is wrong with this code or with me? :)

Git commit hook fails localy. There is no `documentation.json` in repo. Am I forgot to bootstrap something?

shilman: `documentation.json` gets generated dynamically when you run storybook in that example

lib/components/src/controls/helpers.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Adds `control` prefix to make ID attribute more specific.
+ * Removes spaces because spaces are not allowed in ID attributes
+ * @link http://xahlee.info/js/html_allowed_chars_in_attribute.html
+ * @example getControlId('my prop name') -> 'control-my-prop-name'
+ */
+export const getControlId = (value: string) => `control-${value.replace(/\s+/g, '-')}`;
```
shilman, line 7: Can you add some unit tests for this? I think you're missing some cases. Examples:

https://github.com/ComponentDriven/csf/blob/next/src/index.test.ts#L3

lib/components/src/controls/helpers.test.ts
```diff
@@ -0,0 +1,22 @@
...
+      'weird',
+      'some ,’–—―′¿`" weird <>()!.!!!{}[] id %^&$*#& 😀',
+      'control-some-,’–—―′¿`"-weird-<>()!.!!!{}[]-id-%^&$*#&-😀',
+    ],
+  ].forEach(([testName, input, output]) => {
+    it(`${testName}`, () => {
+      expect(getControlId(input)).toBe(output);
+    });
+  });
+});
```
YozhEzhi, line 22: @shilman Please, take a look.

lib/components/src/controls/helpers.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    ['upper case', 'SOME-ID', 'control-SOME-ID'],
+    ['handles basic substitution', 'a b$c?d😀e', 'control-a-b$c?d😀e'],
+    ['cyrillic', 'Кириллический идентификатор', 'control-Кириллический-идентификатор'],
+    ['korean', '바보', 'control-바보'],
+    [
+      'weird',
+      'some ,’–—―′¿`" weird <>()!.!!!{}[] id %^&$*#& 😀',
+      'control-some-,’–—―′¿`"-weird-<>()!.!!!{}[]-id-%^&$*#&-😀',
+    ],
+  ].forEach(([testName, input, output]) => {
```
shilman, line 17: Thanks @YozhEzhi ! A couple things:

1) can you convert this to jest `it.each` which is a more standard/idiomatic way of doing this: https://jestjs.io/docs/api#testeachtablename-fn-timeout

2) a bunch of these are not valid DOM id's:

> ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").

https://stackoverflow.com/questions/1077084/what-characters-are-allowed-in-dom-ids#:~:text=The%20W3C%20spec%20Basic%20HTML,%22).%22

shilman (changes requested):

lib/components/src/controls/helpers.test.ts
YozhEzhi: @shilman 
1. I'll do.
2. Is your advise to remove non valid cases? Ok. I'll do.

YozhEzhi: Done.

shilman: No, my point is that users can name their controls anything including non-valid characters. But the conversion function needs to turn those into IDs that only contain valid characters. So the conversion function needs to be a little smart to do that properly.

YozhEzhi: 🤔

YozhEzhi: This is not as simple as I thought :)

For example: What if user named control "Hello 😄" and "Hello 😄". Id should be unique. Now I have no idea how to deal that situation.

ndelangen: > What if user named control "Hello 😄" and "Hello 😄"

Then the users is going to discover that one of those controls does not show up?, If we use this Id as a react key, react will log a warning and render only 1 of them.

YozhEzhi: Guys, as I see in `next`[ branch code](https://github.com/storybookjs/storybook/blob/next/lib/components/src/controls/options/Checkbox.tsx#L69): some id's are generated like this:
```
const id = `${name}-${key}`;
```

And id becomes like `arg-Some value`. It's not valid.

In this PR I've made id generation to be like:
```
const id = `${name}-${index}`;
```

Example result:

YozhEzhi: @shilman @ndelangen
**IMHO**, my solution **does not degrade** current behaviour of controls. And it brings **some** improvements.
But maybe, I don't understand some usecases of controls.

So, please, feel free to close this PR if my solution will not work as it should to be.

ndelangen (approved):

shilman (approved):
Looking good! Thanks for your patience on this one 🙏

ndelangen (approved):

PR #14305, opened 2021-03-22
Components: Handle `null` when parsing input in color picker

gaetanmaisse:
Issue: The E2E are failing for the vast majority of setups (framework + config).

## What I did

It looks like the input can be null 🤷 it's the case for most of our E2E tests based on CLI `sb init`

## How to test

- CI should be back to 🟢 
- Locally, for instance : `yarn test:e2e-framework --clean --use-local-sb-cli sfcVue` then `yarn storybook` and try to navigate to `Button > Primary` story

### DIFF
--- a/lib/components/src/controls/Color.tsx
+++ b/lib/components/src/controls/Color.tsx
@@ -55,7 +55,9 @@ const expandShorthand = (value: string) => {
   return `#${r}${r}${g}${g}${b}${b}`;
 };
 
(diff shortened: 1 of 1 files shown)


gaetanmaisse:
@ghengeveld I just did a quick fix without any feature-oriented thinking, I let you take a look at it and improve things if needed.

ghengeveld (approved):

PR #14307, opened 2021-03-22
Core: Import `isPlainObject` directly from lodash

damianstasik:
## What I did
I changed how `isPlainObject` is imported from `lodash`, so that it directly imports the module instead of importing it from main module with many other exported modules.

## How to test


shilman (approved):
LGTM. good catch @visualfanatic !

cc @ghengeveld pls import directly from the lodash modules

PR #14308, opened 2021-03-22
UI: Fix toolbar text using theme color

rev3rsor:
Issue: #14664

Toolbar buttons don't use `barTextColor` from theme

## What I did

Change defaulted text color to theme variable

## How to test

- To test, set `barTextColor` in `theme.create()` e.g.

```
// examples/cra-kitchen-sink/.storybook/manager.js

addons.setConfig({
  ...
  theme: create({
    base: 'light',
    barTextColor: 'green', // <--
    brandTitle: 'CRA Kitchen Sink',
    brandUrl: 'https://github.com/storybookjs/storybook/tree/master/examples/cra-kitchen-sink',
    gridCellSize: 12,
  }),
});
```

Pre-fix: color only applies to icon buttons
Post-fix: unselected button text should be the same color as well


shilman (approved):
LGTM! Thanks for the fix @rev3rsor!

PR #14318, opened 2021-03-23
Docs configure addons

winkerVSbecks:
## What I did

Updated docs to refer to the addon kit instead of the old dev kits and added a section on various ways to configure an addon.

### DIFF
--- a/docs/addons/configure-addons.md
+++ b/docs/addons/configure-addons.md
@@ -6,7 +6,7 @@ Storybook is designed to be highly customizable. You can develop all kinds of wo
 
 ## Preset
 
-Preset options are global and are accessible from NodeJS. Great for things like Webpack loaders, Babel plugins, and other library or framework specific configurations.
+Preset options are global and are accessible from NodeJS. Great for things like Webpack loaders, Babel plugins, and other library or framework-specific configurations.
 
 Presets offload the burden of configuration from the user to the addon. They are convenient for integrating libraries with Storybook. Let's look at an example.
 
--- a/docs/addons/configure-addons.md
+++ b/docs/addons/configure-addons.md
@@ -1,0 +1,40 @@
+---
+title: 'Configure and communicate with an addon'
+---
+
+Storybook is designed to be highly customizable. You can develop all kinds of workflows around it. To support this, you can allow users to configure and communicate with addons in several different ways.
+
+## Preset
+
+Preset options are global and are accessible from NodeJS. Great for things like Webpack loaders, Babel plugins, and other library or framework specific configurations.
+
+Presets offload the burden of configuration from the user to the addon. They are convenient for integrating libraries with Storybook. Let's look at an example.
+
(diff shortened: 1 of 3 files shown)


docs/addons/configure-addons.md
```diff
@@ -0,0 +1,40 @@
+---
+title: 'Configure and communicate with an addon'
+---
+
+Storybook is designed to be highly customizable. You can develop all kinds of workflows around it. To support this, you can allow users to configure and communicate with addons in several different ways.
+
+## Preset
+
+Preset options are global and are accessible from NodeJS. Great for things like Webpack loaders, Babel plugins, and other library or framework specific configurations.
```
domyen, line 9:
```suggestion
Preset options are global and are accessible from NodeJS. Great for things like Webpack loaders, Babel plugins, and other library or framework-specific configurations.
```

docs/addons/configure-addons.md
```diff
@@ -0,0 +1,40 @@
...
+
+Presets offload the burden of configuration from the user to the addon. They are convenient for integrating libraries with Storybook. Let's look at an example.
+
+Many libraries require that the app be wrapped by a Provider. Which then provides data to components down the tree. For example, CSS-in-JS libraries use a ThemeProvider to access your theme. Presets can describe such behaviour. So, when the user installs the addon in their Storybook, all stories will automatically be wrapped by the ThemeProvider. Allowing folks to start using your library with Storybook, with just 1 line of config!
```
domyen, line 13:
```suggestion
Many libraries require that the app be wrapped by a `Provider`. Which then provides data to components down the tree. For example, CSS-in-JS libraries use a `ThemeProvider` to access the app's theme. Presets can describe behavior like adding wrappers automatically, without users having to do any manual configuration. If a user installs an addon that has Presets, the addon can instruct Storybook to wrap all stories in `ThemeProvider`. This allows folks to start using your library with Storybook, with just 1 line of config!
```

docs/addons/configure-addons.md
```diff
@@ -0,0 +1,40 @@
+---
+title: 'Configure and communicate with an addon'
+---
+
+Storybook is designed to be highly customizable. You can develop all kinds of workflows around it. To support this, you can allow users to configure and communicate with addons in several different ways.
```
domyen, line 5: Clarify who this article is for. 

Who are "users" in this case? Addon authors or end users? If the former the parameters section might need to be reframed. If the latter, the preset and channel sections might need to be reframed.

winkerVSbecks: @domyen the audience is meant to be Addon authors. I updated the intro to clarify this. The rest of the sections work as is.

shilman (approved):
LGTM!!

docs/addons/configure-addons.md
```diff
@@ -0,0 +1,38 @@
...
+Use the [`useParameter`](./addons-api#useparameter) hook to access the parameter values within your addon.
+
+```js
+export const Hover = () => <Button>Label</Button>;
+Hover.parameters = { pseudo: { hover: true } };
+```
+
+## Channels
+
+Channels enable two-way communication between the manager and the preview pane, using a NodeJS [EventEmitter](https://nodejs.org/api/events.html) compatible API. Your addons can plug into specific channels and respond to these events.
```
domyen, line 32: Is there more than one channel?

winkerVSbecks: Yea and you can define your own too.

domyen (approved):

shilman (approved):
LGTM

PR #14323, opened 2021-03-23
Refactor: Replace `lodash/range` with `Array.from`

damianstasik:
## What I did
I replaced [`lodash/range`](https://lodash.com/docs/4.17.15#range) with native [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) to create an array of N-elements.

```js
console.log(Array.from({ length: 3 })); // [undefined, undefined, undefined]
```

## How to test


ghengeveld (approved):
Nice. I didn't know Array.from accepted a map function.

PR #14325, opened 2021-03-23
Update note on cli-options.md

marksy:
Update the note regarding the use of `--no-manager-cache` to be more clear about using the flag

Issue:
The CLI docs for using the flag `--no-manager-cache` are a bit vague. 

## What I did
Changed the wording to clarify the usage.

## How to test


shilman (approved):
💯

PR #14333, opened 2021-03-25
Angular: Improve docs inline rendering setup

ThibaudAV:
Issue:

## What I did

Re adds the changes made in this [PR](https://github.com/storybookjs/storybook/pull/14270
) and revert in this [one](https://github.com/storybookjs/storybook/pull/14310) 

But this time adds a Webpack configuration to make `@storybook/angular/element-renderer` import optional if the sb client does not have the right dependencies

> Set up Angular prepareForInline inside the preset. Users can enable it on by simply setting docs.inlineStories: true and we don't need to create another entry point or import in the user's code.
> NOTE that this is a breaking change for prerelease users, since this PR removes the unnecessary entry point.

## How to test

> yes it is tested by e2e 
> N/A
> Done

### DIFF
--- a/addons/docs/angular/README.md
+++ b/addons/docs/angular/README.md
@@ -220,12 +220,10 @@ Then update `.storybook/preview.js`:
 
 ```js
 import { addParameters } from '@storybook/angular';
-import { prepareForInline } from '@storybook/addon-docs/angular/inline';
 
 addParameters({
   docs: {
     inlineStories: true,
-    prepareForInline,
   },
 });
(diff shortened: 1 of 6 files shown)


gaetanmaisse (approved):

shilman (approved):
WHOA this is awesome -- great job @ThibaudAV  💯

ThibaudAV:
Made wiith @gaetanmaisse ;)
Thanks @shilman

PR #14338, opened 2021-03-25
UI: Add aria-labels to buttons without discernible text

adagar:
Issue: https://github.com/storybookjs/storybook/issues/13797

## What I did
Added aria-label attributes with values to expandable text and toggle indicator
## How to test
Run axe devtools and observe that the error indicated in the issue above are no longer being flagged


shilman (approved):
💯

PR #14356, opened 2021-03-26
CSF: Optionally pass Args generic type from BaseAnnotations to ArgTypes

maxmezzomo:
Issue: #14301 

## What I did
Added the possibility got type Meta's generic Args to be passed to the ArgTypes property for the BaseAnnotation interface.

## How to test
I didn't see any _ts-expect_ or similar libraries so automatically test for types. Manually it can be tested.

## Notes
The Args generic could be propagated to ArgTypes in more situations, other than just the BaseAnnotations type. 
For example in StoryContext:
```
export declare type StoryContext = StoryIdentifier & {
    ...
    args: Args;
    argTypes: ArgTypes;
    ...
};
```
These kind of changes however may have required many more changes on related types. And I am not sure how used it would be. Meta seems to be the type most directly used by a user. 
The default generic for ArgTypes is therefore set to `Record<string, any>`

### DIFF
--- a/lib/addons/src/types.ts
+++ b/lib/addons/src/types.ts
@@ -43,9 +43,9 @@ export interface ArgType {
   [key: string]: any;
 }
 
-export interface ArgTypes {
-  [key: string]: ArgType;
-}
+export type ArgTypes<Args = Record<string, any>> = {
+  [key in keyof Partial<Args>]: ArgType;
+};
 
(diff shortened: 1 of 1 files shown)


shilman (approved):
Sorry for the slow response on this @maxmezzomo, this PR got lost in the shuffle and looks great to me. @yannbf WDYT?

PR #14383, opened 2021-03-29
Svelte: Fix Cannot read property '__docgen' of undefined

j3rem1e:
Issue: https://github.com/storybookjs/addon-svelte-csf/issues/15

When a svelte stories doesn't set the component in the default export (or with <Meta/> in addon-svelte-csf), then the stories doesn't show with "Cannot read property '__docgen' of undefined"

## What I did

The svelte source decorators was using the component without checking if it's set or not. I have added a null check here.

## How to test


shilman (approved):
LGTM!

PR #14384, opened 2021-03-29
CLI: Compatibility with sveltekit

j3rem1e:
Issue:

The current CLI doesn't work with sveltekit : 

1. sveltekit uses a `svelte.config.cjs` instead of `svelte.config.js`
2. it uses a project with `"type":"module"` : main and preview should be generated with a cjs extension

## What I did

Auto-detecting a svelte.config.cjs: in this case, it generated a `main.cjs` which imports this configuration.

## How to test


shilman (approved):
LGTM! 🚀

PR #14387, opened 2021-03-30
Controls: Fix defaultValue without PropType

tmeasday:
Issue: #14377

## What I did

Handle the case where `type` is `null`

## How to test

Chromatic - yes
Jest - couldn't get the `extractDocgenProps.test` to easily recreate this situation (you need the first prop to have a type and the second to not, which isn't possible in the current test harness).


shilman (approved):
LGTM! Thanks for the quick turnaround!

PR #14393, opened 2021-03-30
fix(ArgControl): unbox the value to simplify the code

stevensacks:
Issue:
The useState value was boxed unnecessarily.

## What I did
I unboxed the state.

## How to test
All the controls still work in all the controls in examples/official-storybook

I don't think so. This is a minor implementation change.

No.

No.


stevensacks:
Going to put this into a branch

PR #14394, opened 2021-03-30
Controls: Clean up arg unboxing and switch statements

stevensacks:
Issue:

## What I did
Cleaned up the ArgControl and Options by unboxing the value and using a dict instead of a large switch case.

## How to test
It's a minor implementation change and it works in all the examples/official-storybook actions stories.

No

No

No


shilman:
@ghengeveld can you take a look at the chromatic changes and let me know what you think?

ndelangen:
Those changes in chromatic look all right to me, this one seems a bit odd:

Screenshot 2021-04-22 at 17 25 41

Looks like the default value for this arg is a function that returns a string.
I'd think this can't have a control. (how would you edit a function?)

lib/components/src/blocks/ArgsTable/ArgControl.tsx
```diff
@@ -18,22 +18,39 @@ export interface ArgControlProps {
...
+
 const NoControl = () => <>-</>;
 
 export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
   const { key, control } = row;
 
   const [isFocused, setFocused] = useState(false);
-  // box because arg can be a fn (e.g. actions) and useState calls fn's
-  const [boxedValue, setBoxedValue] = useState({ value: arg });
+  const [value, setValue] = useState(() => arg);
```
ndelangen, line 45: what's the point of doing this:

```ts
useState(() => arg);
```

lib/components/src/controls/options/Options.tsx
```diff
@@ -28,28 +28,36 @@ const normalizeOptions = (options: Options, labels?: Record<any, string>) => {
   return options;
 };
 
+const Controls: Record<string, FC> = {
+  check: CheckboxControl,
+  'inline-check': CheckboxControl,
```
ndelangen, line 33: @shilman I think in the future we might want to rename these to `check-inline`, minor DX improvement maybe. Not 100% sure if it's worth it, but could be an addition so no breaking change, WDYT?

ndelangen (approved):
Code changes look good to me.

lib/components/src/blocks/ArgsTable/ArgControl.tsx
shilman: I "boxed" the arg because if the arg was a function it was getting executed, but the boxed version fixes that. However, the function is better than boxing because it doesn't allocate extra memory

ghengeveld: Please re-add a comment explaining why we're doing "lazy" state initialization here.

ghengeveld (approved):

lib/components/src/blocks/ArgsTable/ArgControl.tsx
shilman:
```suggestion
  // this is because the value can be a function
  const [value, setValue] = useState(() => arg);
```

PR #14422, opened 2021-03-31
Addon-docs/ArgsTable: Fix sort parameter propagation

tooppaaa:
Issue: #14414

## What I did
Pass down storySort


tooppaaa:
@shilman, I think this can be a quick fix but there's something weird here.
Docs is picking the storySort of the current story when I think this should be the one from <Primary />. 

However the cases where we have multiple sorts per story should be pretty rare right ?

shilman (approved):
LGTM!

PR #14437, opened 2021-04-01
UI: Add show toolbar T in menu

DallasGale:
Issue: #13182

## What I did
1. The  ```useMenu``` hook has been extended to include  ```const toolbarToogle = useMemo(...)```. 
2. ```isToolshown``` boolean has been used in ```useMenu``` and in  the ```Sidebar``` component to pass down the state change to ```useMenu```

## How to test

Test by running the app and clicking the ```...``` in the toolbar. Press ```T``` to see the state change inside the menu.


shilman (approved):
Looking great & tested ✅ . Thank you!

PR #14439, opened 2021-04-01
Addon-docs: Add transparency support to color swatch

bart-krakowski:
## What I did
I added a transparency checkerboard (https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRjwfs5goq3Bgox6EKk4LNxHLAyHKixVyN8pQ&usqp=CAU) to the ColorItem component.
I think the display of semi-transparent colors will be easier to read from now on:
Screenshot 2021-04-01 at 15 38 43

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200539748475788/1200567601702391) by [Unito](https://www.unito.io)


ndelangen:
I really like this! @MichaelArestad WDYT?

MichaelArestad:
This is a cool idea! Nice work @bart-krakowski 

1. Should this be optional?
2. I would apply it to the container wrapping all of the color swatches so the pattern is seamless between colors. There is a transparent border on this container, but I think it would be a good idea to make this border a solid color.
3. I find the checker pattern nice, but a bit heavy. Perhaps something with a less grey would be better to make it easier to see the unaltered color. I tried the following:

```css
background-image: repeating-linear-gradient(-45deg, #ccc, #ccc 1px, #fff 1px, #fff 16px)
```

Which looks like this:
image


What do you think?

ndelangen:
@bart-krakowski do you think you could make this adjustment?

bart-krakowski:
@ndelangen, yes, I will introduce it soon.

shilman:
@bart-krakowski this is awesome!! however, there is some weirdness in the chromatic build. can you please take a look? https://www.chromatic.com/build?appId=5a375b97f4b14f0020b0cda3&number=22752

bart-krakowski:
@shilman, I see it works better now :) 
https://www.chromatic.com/build?appId=5a375b97f4b14f0020b0cda3&number=22755

shilman (approved):
Great job @bart-krakowski @MichaelArestad 💯

shilman (changes requested):
Hmm @bart-krakowski can we add this without adding an extra prop? 🤔

bart-krakowski:
> Hmm @bart-krakowski can we add this without adding an extra prop? 🤔



> Hmm @bart-krakowski is there any way to achieve this without needing to add an extra prop? 🤔

bart-krakowski:
> Hmm @bart-krakowski is there any way to achieve this without needing to add an extra prop? 🤔

OK, I have removed the `isTransparent` prop.

lib/components/src/blocks/ColorPalette.tsx
```diff
@@ -64,6 +79,8 @@ const SwatchColors = styled.div(({ theme }) => ({
   height: 50,
   marginBottom: 5,
   overflow: 'hidden',
+  backgroundColor: 'white',
+  backgroundImage: `repeating-linear-gradient(-45deg, #ccc, #ccc 1px, #fff 1px, #fff 16px)`,
 }));
```
MichaelArestad, line 84: I would also add `background-clip: padding-box;` to prevent the background from showing under the transparent border.

MichaelArestad (changes requested):

shilman (approved):
This is so awesome. Thanks for all the fixes @bart-krakowski 💯

PR #14447, opened 2021-04-02
Addon-docs/Angular: Keep inlineStories to false by default

ThibaudAV:
Issue: 

🙈  multiple reverts and bounces of this subject have lost a small essential change

## What I did

keep a inlineStories = false as expected ... 

## How to test

image


shilman (approved):
LGTM

PR #14450, opened 2021-04-02
Document support for stories as function

avaly:
## What I did

Source: https://github.com/storybookjs/storybook/issues/11181#issuecomment-674844736


shilman (approved):
Nice one!!!

Issue #12850, opened 2020-10-21
Vue ArgsTable missing prop if slot with same name


fallemand:
**Describe the bug**
If I define a prop and a slot with the same name, `ArgsTable` only displays the `slot`

**To Reproduce**
Create a vue component with same `prop` and `slot`

**Screenshots**
**Using the same name**
![image](https://user-images.githubusercontent.com/16105726/96722815-14d9fe00-13ae-11eb-9f13-201c4a31ed96.png)

**Changing the prop name**
![image](https://user-images.githubusercontent.com/16105726/96722942-3b983480-13ae-11eb-80a3-d0683780d04f.png)

**Code snippets**
```vue
<template>
  <span>
    <!-- @slot Pass a custom title -->
    <slot name="title">
      {{ title }}
    </slot>
  </span>
</template>

<script lang="ts">
import Vue from 'vue';

export default Vue.extend({
  name: 'Title',
  props: {
    /** Title */
    title: {
      type: String,
      default: null,
    },
  },
});
</script>
```

**System**
```sh
Environment Info:

  System:
    OS: macOS 10.15.6
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  Binaries:
    Node: 12.18.0 - ~/.nvm/versions/node/v12.18.0/bin/node
    npm: 6.14.4 - ~/.nvm/versions/node/v12.18.0/bin/npm
  Browsers:
    Chrome: 86.0.4240.80
    Firefox: 79.0
    Safari: 13.1.2
  npmPackages:
    @storybook/addon-a11y: ^6.0.26 => 6.0.26 
    @storybook/addon-essentials: ^6.0.26 => 6.0.26 
    @storybook/vue: 6.0.26 => 6.0.26 
```

braddialpad:
I am getting this also. Slot and prop with the same name is a valid case in vue.

braddialpad:
Seems this is the same issue as this https://github.com/storybookjs/storybook/issues/12224

gaetansenn:
Any news about this bug ?

lee-chase:
I think I prefer grouping props and slots, perhaps in the second parameter passed to the template.

fallemand:
It will be awesome to see it happening in `6.3`.
This has big implications for most vue design systems, since usually slots and props have the same name.

atflick:
Hacky workaround:
[...]

PR #14465, opened 2021-04-03
Vue 3: rename slot args if prop args with same name already exist

TimonLukas:
Issue: https://github.com/storybookjs/storybook/issues/12850

## What I did

If a slot argument is generated and a prop argument with the same name already exists, the slot argument will have a `slot` prefix (e.g. `default` -> `slotDefault`).

## How to test

- **Is this testable with Jest or Chromatic screenshots?** Probably, but I wasn't able to find an existing test suite for Vue 3, and I'm not experienced enough to build one up from the ground.
- **Does this need a new example in the kitchen sink apps?** I think it would make sense. I created a small working example in this branch: https://github.com/TimonLukas/storybook/tree/vue3-prop-slot-args-example I'm not sure what the best practice for merging it is - should it be integrated into an already existing story?
- **Does this need an update to the documentation?** It should be documented, but I wasn't able to find Vue 3-specific documentation beyond the blog posts.

### DIFF
--- a/addons/docs/src/frameworks/vue3/extractArgTypes.ts
+++ b/addons/docs/src/frameworks/vue3/extractArgTypes.ts
@@ -1,3 +1,4 @@
+import upperFirst from 'lodash/upperFirst';
 import { ArgTypes } from '@storybook/api';
 import { ArgTypesExtractor, hasDocgen, extractComponentProps } from '../../lib/docgen';
 import { convert } from '../../lib/convert';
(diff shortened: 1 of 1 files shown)


lee-chase:
I think I would prefer always prefixing with `slot` for consistency.

That said it does leave the user still with the headache of excluding slots from being passed from args as props.
Object.keys(args).reduce((a, k) => { if (!k.startsWith('slot')) a[k] = args[k]; return a; }, {})
I wonder if we could split props and slots, perhaps in the 2nd parameter passed to the template.

TimonLukas:
I agree it's better to have consistent behavior. For now I made it dynamic so it's not a breaking change, but since it's still early in the Vue3 support's life I guess that wouldn't be too much of a problem.

I wonder if it's worth it to split the arguments. As long as you don't explicitly have a prop `slotXyz` it doesn't matter if you bind them. But on the other hand, splitting them into a separate argument would remove any need for "automagic" names and make the whole process a lot more explicit...

Maybe it's also a good idea to take a look how this is handled in other Storybook implementations? Does e.g. Angular have a similar situation?

TimonLukas:
Let's move discussion to #14470.

PR #14472, opened 2021-04-04
Core: Don't shadow the window global variable

eirslett:
Issue: Vite's hot module reloading doesn't work properly with the 'global' package

This snippet:

```
import { window } from 'global';
```

will let you mock window variables in Jest tests.
However, it breaks hot module reloading in Vite,
because Vite injects code snippets in top of every
file:

```
window.STUFF = 'Vite injected some HMR code here';
import { window } from 'global';
// this fails: Uncaught ReferenceError: Cannot access 'window' before initialization
```

## What I did
The simple solution is to rename `window` when importing it:

```
import { window as globalWindow } from 'global';
globalWindow.alert('hello world');
```

The code works like before, but Vite will be able to
inject its code snippet without interfering with the
rest of the code.

## How to test

Run Storybook CI, and check that nothing is broken.
I verified locally that this fixes the Vite HMR bug (but the Vite stuff is going on in branches in forks of Storybook, so it's hard for others to test).


shilman (approved):
LGTM 👍 

@ndelangen @tmeasday any objections?

shilman (approved):
Merging to unblock the vite work

PR #14492, opened 2021-04-06
Upgrade `dotenv-webpack` to v6

tessro:
Issue: #14403

## What I did

This PR bumps `dotenv-webpack` to v6 in places that were still using v1.8, resolving #14403.

## How to test

All tests pass.

I also reviewed `dotenv-webpack`'s changelog to ensure there weren't side effects of this upgrade. Some parts of Storybook are already using v6 and there do not appear to be any breaking changes from this upgrade, so I feel pretty confident it will work fine.

### DIFF
--- a/lib/builder-webpack4/package.json
+++ b/lib/builder-webpack4/package.json
@@ -80,7 +80,7 @@
     "case-sensitive-paths-webpack-plugin": "^2.3.0",
     "core-js": "^3.8.2",
     "css-loader": "^3.6.0",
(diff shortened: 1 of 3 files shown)


shilman (approved):
Thanks @paulrosania! Changes LGTM. Usually I'd release this in an alpha, but since the change is small and can be easily reverted if needed, I'm going to merge as is and cross my fingers instead.

paulrosania:
🤞🏼 Thank you!!

PR #14493, opened 2021-04-06
Controls: Reset ArgsTable state when switching stories

ghengeveld:
Issue: #14474

## What I did

Added the `key` prop to force React to create a fresh component rather than reuse a previous instance.
Added a missing dependency to `useCallback` in two places which could lead to the wrong onChange handler getting called.

This also fixes a similar issue with the JSON raw editor using stale data when switching stories.

## How to test

### DIFF
--- a/addons/controls/src/ControlsPanel.tsx
+++ b/addons/controls/src/ControlsPanel.tsx
@@ -1,5 +1,5 @@
 import React, { FC } from 'react';
-import { ArgTypes, useArgs, useArgTypes, useParameter } from '@storybook/api';
+import { ArgTypes, useArgs, useArgTypes, useParameter, useStorybookState } from '@storybook/api';
 import { ArgsTable, NoControlsWarning, PresetColor, SortType } from '@storybook/components';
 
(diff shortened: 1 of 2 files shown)


addons/controls/src/ControlsPanel.tsx
```diff
@@ -36,6 +37,7 @@ export const ControlsPanel: FC = () => {
       {showWarning && <NoControlsWarning />}
       <ArgsTable
         {...{
+          key: path, // resets state when switching stories
```
tmeasday, line 40: Would `storyId` make more sense here?

shilman (approved):
LGTM!

addons/controls/src/ControlsPanel.tsx
ghengeveld: Yeah I did that at first but it would then have to be:

```
key: refId ? `${refId}_${storyId}` : storyId
```

`path` achieves the same thing but without the added complexity of dealing with `refId` (it's in the path already).

PR #14494, opened 2021-04-06
Core: Don't include args param in docs mode URL

ghengeveld:
Issue: -

## What I did

- Omit args param on the docs tab, to eliminate some weird behavior when changing args on the docs page while the 2nd or later story is selected.
- Restore the args param when switching between stories, or navigating from docs to canvas. This means any arg changes made on the docs page *will* be reflected in the URL when switching back to the canvas.

## How to test


shilman (commented):
Code looks good. I'm not sure I agree with the rationale. Currently what this will do is apply the args to the currently selected story. In the case of DocsPage stories running in `--docs` mode, this means it will set the controls in ArgsTable and also update the primary (i.e. first) story's args, which makes perfect sense from a UX standpoint. There are plenty of other, more confusing, cases that this fix eliminates. Can we do a quick call to discuss before merging?

ghengeveld:
Updated as discussed.

shilman (approved):
💯

PR #14511, opened 2021-04-07
Core: Allow string in object arg and support fractional numbers in URL args

ghengeveld:
Issue: #14508

## What I did

- Add support for fractional numbers in URL args
- Allow raw strings and numbers when parsing object arg param (e.g. for ReactNode)

## How to test


shilman (approved):
LGTM!

PR #14528, opened 2021-04-08
Addon-a11y: Fix broken highlight result styles

damianstasik:
## What I did

I fixed broken styles that are added by A11Y addon when result highlighting is enabled. The issue was caused by the duplicated closing bracket. I also removed the unnecessary comma at the end of the object format which caused the `box-shadow` to be missing in the addon's highlight story.

## How to test

Screen Shot 2021-04-09 at 00 24 59

### DIFF
--- a/addons/a11y/src/highlight.ts
+++ b/addons/a11y/src/highlight.ts
@@ -2,11 +2,10 @@ export const highlightStyle = (color: string) => `
   outline: 2px dashed ${color};
   outline-offset: 2px;
   box-shadow: 0 0 0 6px rgba(255,255,255,0.6);
-}
 `;
 
 export const highlightObject = (color: string) => ({
   outline: `2px dashed ${color}`,
   outlineOffset: 2,
-  boxShadow: '0 0 0 6px rgba(255,255,255,0.6),',
+  boxShadow: '0 0 0 6px rgba(255,255,255,0.6)',
 });


tooppaaa (approved):

shilman:
@domyen @tooppaaa can you check chromatic for me?

tooppaaa:
Indeed there's something weird here. @visualfanatic do you want me to look at the issue on dark background ?
https://www.chromatic.com/test?appId=5a375b97f4b14f0020b0cda3&id=606f86dc15d25e00215c5857

domyen:
Thanks @visualfanatic!

This brings up a good question, why is the box-shadow there in the first place? I'm not sure it helps the outline standout more and would suggest removing the box-shadow altogether.

e.g., For example, in this test the box-shadow is quite confusing because it respects element dimensions/radius but the outline does not.
image

tooppaaa:
@visualfanatic Do you want me to take over the chromatic issues ? I'd like to see this fix moving into the addon :)

shilman:
@winkerVSbecks we found this unmerged PR and know that it overlaps with the refactor work you've recently done for the highlight addon #17800 . mind reviewing this and seeing if there are any lessons/fixes that might help make highlight better? you can close this when you're done. @visualfanatic thanks so much for the contribution, and sorry for letting it sit for so long 🙈

winkerVSbecks:
@shilman I'll include that fix in my PR. But afaik that bit of code doesn't actually get used anywhere except a couple of stories which I had already updated to use the new API.

ndelangen:
@damianstasik Thank you for your efforts, time and energy you put into making this PR. It's very much appreciated, it really is. @winkerVSbecks worked on a new core-ish API for storybook which allows for an API to be called which then promps Storybook's manager UI to overlay something over the Canvas.

He initially built this for an addon to showcase spacings between components etc, but it's obvious we should employ this for more uses, such as the A11Y addon which also needs to highlight things to the storybook users (where are those darn links without text?!).

So with that said, I'm going to close this PR, because ultimately we need to do better then injecting CSS into the canvas to highlight things... this will always be prone to breaking the user's components, or simply to overridden and thus breaking the storybook feature.

I'm going to pass this to @winkerVSbecks who -I think- is already planning to make significant changes to the A11Y addon to have it make use of these newer APIs.

Again: I'd like to thank you for the work you did on storybook, It's appreciated! I'm sorry I'm closing your PR.

PR #14504, opened 2021-04-07
ADD nx tooling

ndelangen:

PR #14535, opened 2021-04-09
Build: Add NX bootstrap optimization

FrozenPandaz:
## What I did

This PR adds Nx to run the prepare stage. Nx has remote caching which would allow developers to reuse builds done on the CI server. This will make getting started with the repo as well as CI faster.

For more information about Nx, see the [official documentation](https://nx.dev/)

## How to test

Running `nx run-many --target prepare --parallel --all` should successfully build all projects.

Closes https://github.com/storybookjs/storybook/pull/14504


ndelangen:
@FrozenPandaz do you want to pair again to try and get this across the line & merged?

ndelangen:
Super excited about this!

app/angular/package.json
```diff
@@ -81,7 +81,7 @@
     "@angular/forms": "^11.2.0",
     "@angular/platform-browser": "^11.2.0",
     "@angular/platform-browser-dynamic": "^11.2.0",
-    "@nrwl/workspace": "^11.1.5",
+    "@nrwl/workspace": "^11.6.3",
```
FrozenPandaz, line 84: This is an update to the devDependency.

This new version has some allowances that allow Nx to be tested _within_ an Nx repo.

scripts/bootstrap.js
```diff
@@ -95,8 +95,8 @@ function run() {
       command: () => {
         log.info(prefix, 'prepare');
         spawn(
-          `lerna run prepare ${
-            process.env.CI ? `--concurrency ${maxConcurrentTasks} --stream` : ''
+          `nx run-many --target=prepare --all --parallel ${
```
FrozenPandaz, line 98: This changes the command to use Nx instead of `lerna` which has remote caching!

lib/core-server/src/core-presets.test.ts
```diff
@@ -9,9 +9,13 @@ import { executor as managerExecutor } from './manager/builder';
 import { buildDevStandalone } from './build-dev';
 import { buildStaticStandalone } from './build-static';
 
+// nx-ignore-next-line
```
FrozenPandaz, line 12: These were some imports that created circular dependencies. These comments ignore the dependencies generated by these lines.

app/angular/src/server/angular-read-workspace.ts
```diff
@@ -24,6 +24,7 @@ export const readAngularWorkspaceConfig = async (
 
     const nxWorkspace = require('@nrwl/workspace').readWorkspaceConfig({
       format: 'angularCli',
+      path: dirToSearch,
```
FrozenPandaz, line 27: Now that this method is used within a Nx workspace, it began to resolve the `workspace.json` at the root of _this repo_ in unit tests. 

This should not affect end users of Storybook. This is a fix for some tests that use this method.

ndelangen:
@FrozenPandaz I think there's a bug in the NX integration.. When I click on "All runs for this branch" I see this url:
https://nx.app/branch?workspaceId=606dc2e0c2e8d5671d813305&branchName=14535

I think the branch-name is wrong. That looks like the PR-number to me.

lib/core-server/src/core-presets.test.ts
```diff
@@ -9,9 +9,13 @@ import { executor as managerExecutor } from './manager/builder';
...
 import reactOptions from '../../../app/react/src/server/options';
+// nx-ignore-next-line
 import vue3Options from '../../../app/vue3/src/server/options';
+// nx-ignore-next-line
 import htmlOptions from '../../../app/html/src/server/options';
+// nx-ignore-next-line
```
ndelangen, line 18: @shilman If we wanted to remove these circular dependencies we could extract this into it's own util, or duplicate the code (wouldn't recommend that)

ndelangen (approved):

FrozenPandaz:
> I think the branch-name is wrong. That looks like the PR-number to me.

Yes, you are right, that is the PR number.

The PR number is automatically picked up from circle.

We will make it less confusing. :+1:

ndelangen (approved):

shilman (approved):
This is amazing 🤯  Thank you @FrozenPandaz !!!

PR #14556, opened 2021-04-12
Core: Fix URL handling in Firefox

raon0211:
Issue:

image

When opening a storybook build in a local filesystem with Firefox, an error occurs.

The error indicates that some string like `null/Users/raon0211/dev/repo-name/storybook-build/iframe.html` was passed to the constructor of the native URL class.

This is because when Firefox parses URLs with protocol `file:` (e.g. `file:///Users/raon0211/dev/repo-name/storybook-build/index.html`), the resulting url object has origin `"null"`. (See this [bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=878297).)

image

As documented in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL/origin), the `origin` of URL objects is browser dependent when the url has the `file:` protocol.

## What I did

I used the browser-independent `protocol` and `host` instead of `origin` to fix the problem.

## How to test

This issue only occurs in Firefox.


lib/channel-postmessage/src/index.ts
```diff
@@ -231,8 +231,8 @@ const getEventSourceUrl = (event: MessageEvent) => {
 
   if (frame && remainder.length === 0) {
     const src = frame.getAttribute('src');
-    const { origin, pathname } = new URL(src, document.location);
-    return origin + pathname;
+    const { protocol, host, pathname } = new URL(src, document.location);
+    return `${protocol}//${host}${pathname}`;
```
ndelangen, line 235: I think this will drop the port, if there's 1 present.

ndelangen (changes requested):

lib/channel-postmessage/src/index.ts
raon0211: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL/host), It seems that `host` includes the port.

ndelangen: I stand corrected 🙇

ndelangen (approved):

PR #14563, opened 2021-04-12
Storyshots: Add `beforeAxeTest` hook

coreyjv:
Issue: #14536

## What I did

Per #14536 I added a method named `beforeAxeTest` that is executed before the axe tests are run. 

## How to test

I did not see any test suites for `imageSnapshot` or `axeTest`. If I missed them please let me know and I'll add the necessary tests.

### DIFF
--- a/addons/storyshots/storyshots-puppeteer/src/axeTest.ts
+++ b/addons/storyshots/storyshots-puppeteer/src/axeTest.ts
@@ -1,5 +1,5 @@
 import '@wordpress/jest-puppeteer-axe';
-import { defaultCommonConfig, CommonConfig } from './config';
+import { defaultAxeConfig, AxeConfig } from './config';
 import { puppeteerTest } from './puppeteerTest';
 
 declare global {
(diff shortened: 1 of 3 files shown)


coreyjv:
Would it be possible to get feedback on this PR? Thanks!

shilman (approved):
LGTM!!!

addons/storyshots/storyshots-puppeteer/src/axeTest.ts
```diff
@@ -11,13 +11,17 @@ declare global {
   }
 }
 
-export const axeTest = (customConfig: Partial<CommonConfig> = {}) =>
+export const axeTest = (customConfig: Partial<AxeConfig> = {}) => {
+  const config = { ...defaultAxeConfig, ...customConfig };
+  const { beforeAxeTest } = config;
+
   puppeteerTest({
```
Hypnosphi, line 18: Did it actually work for you @coreyjv? I'm asking because the result of `puppeteerTest` is no longer returned, so instead of AXE it runs the default snapshot tests.

Ideally, we would catch that on CI, but unfortunately the corresponding GitHub Action was eventually removed: 
https://github.com/storybookjs/storybook/commit/9be0706518861b851b645b84a6bd0a52f7e78363

@ndelangen do you remember what exactly wasn't functioning back then?

coreyjv: I'm sorry about that mistake @Hypnosphi! I didn't realize that this wouldn't be caught by CI. I can open another PR to address this if you'd like.

Hypnosphi: No problem, I'm working on another PR for the same package, so I'll just include the fix there

coreyjv: @Hypnosphi -- PR created to address the regression here: https://github.com/storybookjs/storybook/pull/15005

Again my apologies for not catching this the first time.

PR #14567, opened 2021-04-12
Bump vue-docgen-api to 4.38.0

### LINKED ISSUE #14524: vue-docgen-api: failed to parsed when exported component use "as" with typescript
**Current behavior**
We have exported components this way (for better typing extended components with composition api)
```ts
import { defineComponent } from '@vue/composition-api'
import CeHeader from './CeHeader'

const CeHeaderTheme = defineComponent({
  extends: CeHeader
})

export default CeHeaderTheme as typeof CeHeader
```

But this way of exporting causes warnings in storybook: 
```
[vue-docgen-loader] failed to parse the component file with docgen-api: No suitable component definition found on...
```

**To reproduce**

Just export some component as default and add some as typeof operator:

```ts
export default {
  name: 'Test'
} as {
  name: string
}
```
**Expected behavior**

I would like to use "as" operator without any warnings.

mercs600:
Issue: #14524

## What I did
Just bump of  vue-docgen-api to 4.38.0 in vue and vue3 packages. It resolves #14524 
## How to test

just deps bump - any crucial changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Features
 
-* Core: Enable community builders ([#14545](https://github.com/storybookjs/storybook/pull/14545))
+- Core: Enable community builders ([#14545](https://github.com/storybookjs/storybook/pull/14545))
 
 ## 6.3.0-alpha.4 (April 10, 2021)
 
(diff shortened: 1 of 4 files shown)


shilman (approved):
Thanks @mercs600 looks great! 🙏

PR #14522, opened 2021-04-08
CI: Remove Travis, fix TeamCity, rework E2E on CircleCI

gaetanmaisse:
## What I did

 - Remove TravisCI integration as it was no-op for 2 years
 - Fix TeamCity builds by using Node 12 and update Yarn commands (to Yarn 2 version)
 - Rework the E2E tests:
   - `e2e-tests-core` for "officially" supported frameworks : Vue + Angular (+ React + Web Components in e2e-tests-pnp)
   - `e2e-tests-extended` for other frameworks: Preact, Svelte, Riot, Rax etc
   - `e2e-tests-pnp` for some frameworks/setups compatible with PnP: React + WC
  - Upgrade to Cypress 6.8 and use Cypress docker image in CircleCI: Previously Cypress was installed manually to be sure to run with Node 10. We can now use the docker images provided by Cypress team (at least for the 6.8) because they are based on Node 12, the one we are using across the CI.
  - Merge multiple GitHub Actions related to DangerJS into one and fix the behaviour i.e. trigger a new run each time a label is set/unset on a PR.

 

## How to test

- All CircleCI + TeamCity builds should be 🟢

PR #14568, opened 2021-04-12
Build: Disable yarn immutable install by default during E2E tests

gaetanmaisse:
## What I did

When running the E2E tests we are doing a fresh install of Yarn using the latest version; as of Yarn 3.0.0 rc1 a behaviour on dep install has been updated:
 > The enableImmutableInstalls will now default to true on CI (we still recommend to explicitly use --immutable on the CLI).
You can re-allow mutations by adding YARN_ENABLE_IMMUTABLE_INSTALLS=false in your environment variables.

So I updated the Yarn config used in the E2E to be able to properly bootstrap the different projects.

For details, see https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md#300-rc1

## How to test

- CI should be back to 🟢

### DIFF
--- a/scripts/run-e2e.ts
+++ b/scripts/run-e2e.ts
@@ -107,6 +107,8 @@ const configureYarn2 = async ({ cwd }: Options) => {
     // Disable fallback mode to make sure everything is required correctly
     `yarn config set pnpFallbackMode none`,
     `yarn config set enableGlobalCache true`,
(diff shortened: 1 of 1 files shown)


ndelangen (approved):

gaetanmaisse:
@shilman, self-merging to fix CircleCI checks on `next`, other CI improvements and fixes will be done with https://github.com/storybookjs/storybook/pull/14522

PR #14573, opened 2021-04-12
globalTypes with toolbar requires addon-toolbars

SamKirkland:
Adding a note that @storybook/addon-toolbars is required to use globalTypes to define toolbars.

Had a really tough time finding why the documentation wasn't working properly. It was because I updated from a old version that didn't have @storybook/addon-essentials.

Issue:
After updating the the latest storybook version I was underable to add custom toolbars. This was due to not having the toolbar plugin. It's unclear this is a requirement in the existing documentation.

## What I did
Updated the documentation on preview.js to denote that the toolbars addon is needed in order for the toolbars to be shown.

## How to test
Verify documentation is updated

- Is this testable with Jest or Chromatic screenshots? Nope
- Does this need an update to the documentation? This is a documentation update

### DIFF
--- a/docs/essentials/toolbars-and-globals.md
+++ b/docs/essentials/toolbars-and-globals.md
@@ -72,6 +72,10 @@ The <code>icon</code> element used in the examples loads the icons from the <cod
 
 </div>
 
+<div class="aside">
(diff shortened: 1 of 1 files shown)


shilman (approved):
Nice one! Yeah, the docs are "optimized" for fresh installs 🙈  cc @jonniebigodes

PR #14582, opened 2021-04-13
Storyshots: Preserve authentication information in Storybook URL

avendiart:
Issue: #14107

## What I did

Refactored the `constructUrl` function to only make adjustments to the provided URL instead of re-creating it from scratch. Also added the tests to make sure authentication information is preserved within the URL.


shilman (approved):
great fix @avendiart 👍

PR #14586, opened 2021-04-13
Angular: set the @ViewChild with a non-empty value in StorybookWrapperComponent

ThibaudAV:
Issue: https://github.com/storybookjs/storybook/issues/14575. maybe other ??? 🤔 

## What I did

Fix the error :
> Unhandled Promise rejection: Can't construct a query for the property "storyComponentElementRef" of "StorybookWrapperComponent" since the query selector wasn't defined. ; Zone: <root> ; Task: Promise.then ; Value: Error: Can't construct a query for the property "storyComponentElementRef" of "StorybookWrapperComponent" since the query selector wasn't defined.
which appears on angular v11 once the ngcc command is executed

## How to test

> hardly :/ I don't know how to do it 
> nope
> nope 


shilman (approved):
LGTM! thanks @ThibaudAV 🙏

PR #14636, opened 2021-04-16
Modified Swatches keys to avoid duplicates

Dschungelabenteuer:
Issue: #14580

## What I did

Both `renderSwatch` and `renderSwatchLabel` were using the color value as the key by whether mapping the `colors` array or the `colors` object values and keys. Using the same color twice would then result in duplicate keys, which should be avoided for obvious reasons. I concatenated each color value with their respective index from the `Array.prototype.map` function to ensure there is no duplicate keys.

## How to test

There are two ways of feeding the `colors` prop, I have tested both ways with duplicates values:

**Array**
```jsx
  <ColorItem
    title="theme.color.primary"
    subtitle="Coral"
    colors={['#FF4785', '#FF4785']}
  />
```

**Object**
```jsx
  <ColorItem
    title="theme.color.positive"
    subtitle="Green"
    colors={{
      Apple: 'rgba(102,191,60,1)',
      Apple80: 'rgba(102,191,60,.8)',
      Apple80: 'rgba(102,191,60,.8)',
      Apple80: 'rgba(102,191,60,.7)',
      Apple60: 'rgba(102,191,60,.6)',
      Apple50: 'rgba(102,191,60,.6)',
      Apple30: 'rgba(102,191,60,.3)',
    }}
  />
```

Please note how I deliberately used the same `Apple80` key several times. It was done on purpose to check the resulting behaviour: in this case, only the last value will be considered, which is how it is supposed to be.

Before the fix, the duplicate value between `Apple60` and `Apple50` would raise a duplicate key error. This should work fine now!

## Note

I think I could have only relied on the index for the key instead of concatenating both value and index, but it felt safer to me. Feel free to provide some feedback if I'm doing the wrong way!

### DIFF
--- a/lib/components/src/blocks/ColorPalette.tsx
+++ b/lib/components/src/blocks/ColorPalette.tsx
@@ -121,10 +121,10 @@ interface ColorProps {
   colors: Colors;
 }
 
-function renderSwatch(color: string) {
+function renderSwatch(color: string, index: number) {
   return (
     <Swatch
-      key={color}
+      key={`${color}-${index}`}
       title={color}
       style={{
         background: color,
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM @Dschungelabenteuer !! thanks!

PR #14647, opened 2021-04-19
Add `funding` to manifests

tessro:
Issue: #14514

## What I did

I added a `funding` property to all `package.json` files other than the example projects.

I was intimidated by the number of projects, so I used the following script:

```bash
#!/bin/bash

# Find all `package.json` files other than examples
manifests=`git ls-files | grep 'package.json' | grep -v '^examples'`

function add_funding {
  file=$1

  # Insert `funding` block after `repository` block
  sed -i .bak -e '/^  "repository"/i\
  "funding": {\
    "type": "opencollective",\
    "url": "https://opencollective.com/storybook"\
  },' $file
 }

for manifest in $manifests
do
  add_funding $manifest
done

# Remove `sed` backup files (and this script, as a side effect)
git clean -f
```

## How to test

Run `npm fund` from the monorepo root. The top entry should be all the subprojects listed under Storybook's Open Collective URL.

### DIFF
--- a/addons/a11y/package.json
+++ b/addons/a11y/package.json
@@ -20,6 +20,10 @@
     "url": "git+https://github.com/storybookjs/storybook.git",
     "directory": "addons/a11y"
   },
+  "funding": {
+    "type": "opencollective",
+    "url": "https://opencollective.com/storybook"
+  },
   "license": "MIT",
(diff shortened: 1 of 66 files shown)


shilman (approved):
This is great! Thanks so much @paulrosania ❤️

Issue #13768, opened 2021-01-29
Support Ivy renderer in @storybook/angular


kroeder:
**Is your feature request related to a problem? Please describe**
Currently, `@storybook/angular/client` produces ViewEngine / JiT Code. ViewEngine will be dropped in the future by the Angular Team and replaced by Ivy. This will most likely cause `@storybook/angular` to not work anymore in the future.

On top of that, producing ViewEngine code is not what a real-world application would produce. (Angular >= 9)

Ivy reference: https://angular.io/guide/ivy

**Starter branch for testing**
@ThibaudAV provided a branch with a minimum Storybook example setup that can be used as a proof of concept
https://github.com/storybookjs/storybook/tree/new-angular/empty-app/app

**Describe the solution you'd like**
We need to add `@ngtools/webpack` to our webpack configuration. That alone won't do the trick, though. The `AngularCompilerPlugin` has several requirements such as an entry point. 

The first step should be to produce Ivy / JiT code and ignore AoT for now (https://angular.io/guide/aot-compiler).

**Describe alternatives you've considered**
I'm not sure there are alternatives to `@ngtools/webpack`

**Additional context**

`@storybook/angular`: https://github.com/storybookjs/storybook/tree/next/app/angular/src

Storybook Angular handles things dynamically which means there is no unique `AppModule` or a `bootstrapModule`. Both are part of the `render` function: https://github.com/storybookjs/storybook/blob/next/app/angular/src/client/preview/angular-beta/RendererService.ts#L58

There's already a PR that is more or less a playground PR including changes that do not make sense (yet 🙃 ): https://github.com/storybookjs/storybook/pull/11157

# Collection of information I've gathered 

## AngularCompilerPlugin setup
**Some more details about the `AngularCompilerPlugin**

The `AngularCompilerPlugin` does toggle JIT-Mode by checking the `skipCodeGeneration` option value 

https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/angular_compiler_plugin.ts#L223

**Documentation https://www.npmjs.com/package/@ngtools/webpack**

> skipCodeGeneration. Optional, defaults to false. Disable code generation and do not refactor the code to bootstrap. This replaces templateUrl: "string" with template: require("string") (and similar for styles) to allow for webpack to properly link the resources.

I assume setting `skipCodeGeneration` to true is mandatory for using JIT instead of AOT (which is not yet our priority in this issue)

**directTemplateLoading**

[...]

PR #14649, opened 2021-04-19
Angular: Support Ivy addon

kroeder:
Issue: #13768

## What I did

This is an intermediate solution until the Ivy support can be integrated directly to storybook/angular

## How to test

See https://github.com/storybookjs/addon-angular-ivy


shilman (approved):
💯💯💯 Victory!!! 💯💯💯

Issue #11542, opened 2020-07-14
Story source code shows "MDXCreateElement" instead of actual source code


NicolaSansom:
**Describe the bug**
The output of Show Code tabs mismatches the source 

**To Reproduce**

1. Create a new `.mdx` story:
```
# Alert 

<Preview>
  <Story name="Alert with one option">
    {() => {
      const [visible, toggleVisible] = useState(false);
      return (
        <div>
          <button onClick={() => toggleVisible(!visible)}>Open Alert with one action</button>
          <Alert
            visible={visible}
            toggleVisible={() => toggleVisible(false)}
            title="Title"
            description="Description"
          />
        </div>
      );
    }}
  </Story>
</Preview>
```
2. Click on 'Show Code'
3. Scroll down to screenshot of output 

**Expected behavior**
Output match contents of the `<Story>`

**Screenshots**
![image](https://user-images.githubusercontent.com/17367104/87429042-22c3fa00-c5db-11ea-8be7-08e771d84db1.png)

**Code snippets**
Full Story
```
import { Story, Preview, Props } from '@storybook/addon-docs/blocks';
import { withA11y } from '@storybook/addon-a11y';
import { useState } from 'react';
import OverlayAlert from './OverlayAlert';

<Meta title="Components/Overlay - Alert" component={OverlayAlert} decorators={[withA11y]} />

# Overlay - Alert

<Preview>
  <Story name="Alert with one option">
    {() => {
      const [visible, toggleVisible] = useState(false);
      return (
        <div>
          <button onClick={() => toggleVisible(!visible)}>Open Alert with one action</button>
          <Alert
            visible={visible}
            toggleVisible={() => toggleVisible(false)}
            title="Title"
            description="Description"
          />
        </div>
      );
    }}
  </Story>
</Preview>
```

Main.js
```
module.exports = {
  stories: ['../src/docs/*.mdx', '../src/**/*.stories.(js|mdx)'],
  addons: [
    '@storybook/preset-create-react-app',
    '@storybook/addon-controls',
    '@storybook/addon-actions',
    '@storybook/addon-links',
    '@storybook/addon-a11y/register',
    '@storybook/addon-viewport/register',
    {
      name: '@storybook/addon-docs',
      options: { configureJSX: true },
    },
  ],
};
```
Show code output
```
<MDXCreateElement
  mdxType="div"
  originalType="div"
>
  <MDXCreateElement
    mdxType="button"
    onClick={() => {}}
    originalType="button"
[...]

PR #14652, opened 2021-04-19
Addon-docs: Exclude decorators in dynamic source snippets

lauracarballo:
Issues: #12022 #11542

## What I did

The dynamic code snippet is generated by a decorator called [`jsxDecorator`](https://github.com/storybookjs/storybook/blob/next/addons/docs/src/frameworks/react/jsxDecorator.tsx) which is included in `kindMetadata.decorators`.

This decorator runs after the story decorators provided by the developer, which normally include i18n providers, data fetching or visual styling that shouldn't be shown in the code snippet.

The order of the decorators can be seen here:
https://github.com/storybookjs/storybook/blob/next/lib/client-api/src/story_store.ts#L379

Re-ordering the decorators would be the easiest solution, but that seems like it would be breaking change to library due to some addons possibly expecting the output of story decorators.

So, to bypass the story decorators, the solution myself and @jamesb3ll came up with is to use the `component` from the story context. If the developer has provided a component we can construct the JSX element using `React.createElement` together with the `args` as the props. That component is then converted to readable code string using the `react-element-to-jsx-string` library.

We've updated the tests to reflect this, any feedback on this solution would be great.

### DIFF
--- a/addons/docs/src/frameworks/react/jsxDecorator.test.tsx
+++ b/addons/docs/src/frameworks/react/jsxDecorator.test.tsx
@@ -155,12 +155,13 @@ describe('renderJsx', () => {
 });
 
 // @ts-ignore
-const makeContext = (name: string, parameters: any, args: any): StoryContext => ({
+const makeContext = (name: string, parameters: any, args: any, extra?: object): StoryContext => ({
   id: `jsx-test--${name}`,
   kind: 'js-text',
   name,
   parameters,
   args,
+  ...extra,
 });
 
 describe('jsxDecorator', () => {
@@ -183,21 +184,29 @@ describe('jsxDecorator', () => {
     );
   });
 
-  it('should render dynamically when provided a component', () => {
-    const Component = ({ className }: { className: string }) => (
-      <div className={className}>component</div>
+  it('should not render decorators when provided excludeDecorators parameter', () => {
+    const storyFn = (args: any) => <div>args story</div>;
+    const decoratedStoryFn = (args: any) => (
+      <div style={{ padding: 25, border: '3px solid red' }}>{storyFn(args)}</div>
     );
-    const storyFn = (args: any) => <Component {...args} />;
     const context = makeContext(
       'args',
-      { __isArgsStory: true, component: Component },
-      { className: 'test' }
+      {
+        __isArgsStory: true,
+        docs: {
+          source: {
+            excludeDecorators: true,
+          },
+        },
+      },
+      {},
+      { originalStoryFn: storyFn }
     );
-    jsxDecorator(storyFn, context);
+    jsxDecorator(decoratedStoryFn, context);
     expect(mockChannel.emit).toHaveBeenCalledWith(
       SNIPPET_RENDERED,
       'jsx-test--args',
-      '<Component className="test" />'
+      '<div>\n  args story\n</div>'
     );
   });
 
--- a/addons/docs/src/frameworks/react/jsxDecorator.test.tsx
+++ b/addons/docs/src/frameworks/react/jsxDecorator.test.tsx
@@ -183,6 +183,24 @@ describe('jsxDecorator', () => {
     );
   });
 
+  it('should render dynamically when provided a component', () => {
+    const Component = ({ className }: { className: string }) => (
+      <div className={className}>component</div>
+    );
+    const storyFn = (args: any) => <Component {...args} />;
+    const context = makeContext(
+      'args',
+      { __isArgsStory: true, component: Component },
+      { className: 'test' }
+    );
+    jsxDecorator(storyFn, context);
+    expect(mockChannel.emit).toHaveBeenCalledWith(
+      SNIPPET_RENDERED,
+      'jsx-test--args',
+      '<Component className="test" />'
+    );
+  });
+
   it('should skip dynamic rendering for no-args stories', () => {
     const storyFn = () => <div>classic story</div>;
     const context = makeContext('classic', {}, {});
--- a/addons/docs/src/frameworks/react/jsxDecorator.tsx
+++ b/addons/docs/src/frameworks/react/jsxDecorator.tsx
@@ -166,8 +166,14 @@ export const jsxDecorator = (storyFn: any, context: StoryContext) => {
     ...(context?.parameters.jsx || {}),
   } as Required<JSXOptions>;
 
+  let storyOrComponent = story;
+  const { parameters, args } = context;
+  if (parameters.component) {
+    storyOrComponent = React.createElement(parameters.component, { ...args });
+  }
+
   let jsx = '';
-  const rendered = renderJsx(story, options);
+  const rendered = renderJsx(storyOrComponent, options);
   if (rendered) {
     jsx = applyTransformSource(rendered, options, context);
   }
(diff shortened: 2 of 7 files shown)


shilman (commented):
Hi @lauracarballo, thanks for the fix! However, I'm a little unclear on how it will work when the component needs contexts provided by the decorators, such as i18n or theme.

shilman:
One more wrinkle. A story function can contain arbitrary JSX, but this only shows the `default.component`. If this solution works for components that depend on contexts from decorators, we can probably come up a similar solution that can show the story's JSX.

lauracarballo:
Hi @shilman, thanks for reviewing the PR. I've updated the cra-kitchen-sink with an example including decorators to showcase what's being solved in this PR in a better way. 

Currently decorators are being included inside the code snippet. To my understanding decorators should not be appearing inside the source code snippet since they are something "external" to the original component.

For example:

Current source snippet using decorators

With the fix provided, the `jsxDecorator` still returns the original story including decorators but before doing so, we grab the `component` from the `context` and use that to generate the code snippet. 

This way we end up rendering a story that includes decorators in the Canvas and Doc Blocks but does not include them in the source code snippet. 

Fixed sourced code snippet using decorators

ndelangen:
There a difference between these 2 stories? 1 shows a function returning JSX, the other shows just the JSX.

I don't have a strong preference which we do, But I'd prefer it if it was consistent.

Screenshot 2021-04-22 at 14 59 13

lauracarballo:
Hi @ndelangen, thanks for your reply and for having a look at my PR.

The 2 stories added in `decorators.stories.js` are to demonstrate how [source code snippets](https://storybook.js.org/docs/react/writing-docs/doc-blocks#source) render for 2 different styles of stories.
**Basic** story:
```js
export const Basic = () => <Button onClick={action('clicked', { depth: 1 })}>Basic</Button>;
```
**Args** story:
```js
export const WithArgs = (args) => <Button {...args} />;
WithArgs.args = { onClick: action('clicked', { depth: 1 }), children: 'With args' };
```
According to [DocsPage documentation](https://storybook.js.org/docs/react/writing-docs/doc-blocks#docspage-1) when `docs.source.type` is the default value of `auto` the code snippet output will:

> Use `dynamic` snippets if the story is written using Args and the framework supports it.

And otherwise, if not an Args story it will render using `code` which is the raw story code parsed at build time.

This is why the `decorators.stories.js` file shows two different outputs: JSX (`dynamic`) for Args Story and a function returning JSX (`code`) for a Basic story.

What I'm attempting to fix in this PR is that currently the Args stories code snippets are broken when adding decorators as mentioned in this issue (https://github.com/storybookjs/storybook/issues/12022).

Thanks again for reviewing

shilman (changes requested):
Following up on my previous comment, I did a little testing on this branch. This is a fantastic start, but the solution will be a little more involved.

From what I can tell, if a StoryFn relies some context, such as themes provided by a decorator:

- `ReactDOM.render(<StoryFn />)` needs decorators/context
- `<StoryFn />` doesn't need them

This means that we can render JSX for the component without decorators even when we can't render the component, which is great news!

However, this PR as it stands is only accurate when the user's story function is:

```js
const Template = (args) => <Component {...args} />
```

This is the most common case, but there are plenty of times when this is not the case. For example, it's also common to show compositions of components in a story:

```js
export const Empty = (args) => <List {...args} />;

export const ManyItems = (args) => (
  <List {...args}>
    <ListItem />
    <ListItem />
    <ListItem />
  </List>
);
```

The first story would render correctly with the changes in this PR but the second story wouldn't.

I think the solution is to render the undecorated story function, which we happen to keep around in the story store. Unfortunately, it's currently not available in the StoryContext at render time.

So I'd like to see a few changes to this:
- [ ] Make the undecorated story function available in the story context
- [ ] Render JSX from the undecorated story function instead of from the synthetic `<Component {...args />` function
- [ ] Make this feature either opt-in or opt-out using a story parameter, because I think there are cases where people WANT to see the decorated story. For example, if my story needs a React.Context to work, I might want to show that in the code sample.

I'm happy to help make these changes. What do you think?

lauracarballo:
Thanks a lot for reviewing. Your solution seems great, I was hesitant to add extra data to the context but having your feedback it seems like the best place to access the original story.

I'm looking into this, I'll get back to you if I need any extra help.

Thanks for the detailed write up and tasks.

shilman:
Awesome @lauracarballo 💯 -- please ask any q's here or feel free to grab me or any of the maintainers (@ndelangen @tmeasday etc.) on Discord https://discord.gg/storybook. This is going to be really great.

gabiseabra:
> Re-ordering the decorators would be the easiest solution, but that seems like it would be breaking change to library due to some addons possibly expecting the output of story decorators.

Hey, I’m also looking into issue #12022 and from what I gather `jsxDecorator` is the first entry in the global decorators list (not kind decorators), so it runs immediately after user-defined decorators (kind & story) and before library-defined decorators.

So I don’t understand why re-ordering `jsxDecorator` would be an issue as long as it keeps running before all other global decorators, only moving "down" the stack by skipping user-defined ones. I may be missing something, can you share any more insights? Which addon would be affected by this?

lauracarballo:
Hi @shilman, sorry for the delay in the response. I've made updated on the PR based on your previous feedback :)

I have added the `originalStoryFn` to the `context` so we can access the original story in the `jsxDecorator`. Following your advice I have added an extra parameter `docs.source.includeDecorators` to opt-in or opt-out of showing decorators in the source code snippet. 
Note that I have excluded the decorators by default and this could be a breaking change, consequently a lot of the tests are breaking as the `context` object has been changed. I would like to receive some feedback on this matter before attempting to fix the tests.

Thanks for your patience and guidance

shilman:
Hi @lauracarballo thanks so much for following up, this is looking great!

How about making it opt-out just to be on the safe side, with `excludeDecorators` instead of `includeDecorators`? That way:
1. It won't be a breaking change
2. It won't crash when the stories require decorators to render properly

Since parameters cascade, you could opt-in to this feature globally with one line in `.storybook/preview.js`. I think after that, it's just a matter of cleaning up in various places to make TS types line up.

What do you think? Storybook 6.3 is going into beta tomorrow, so it would be great if we could wrap this up this week. Happy to do whatever it takes to help get this over the line!

lauracarballo:
Hi @shilman, thanks for the quick response :)

I have updated the PR to make it opt-in and opt-out using `excludeDecorators` and I have also set this in `.storybook/preview.js` in the `cra-kitchen-sink` example. The test has also been updated for this change.

shilman (approved):
Awesome!!! 💯 x 💯

jonniebigodes:
@lauracarballo one additional thing if you're ok with it. If you could jump into our [Discord server](https://discord.com/invite/UUt2PJb) and message me (same username), I wanted to follow up with you on this pull request.

lauracarballo:
Hi @jonniebigodes, I messaged you on Discord already. Thanks :)

raduliviu:
> I have updated the PR to make it opt-in and opt-out using `excludeDecorators` and I have also set this in `.storybook/preview.js` in the `cra-kitchen-sink` example. The test has also been updated for this change.

Hi @shilman sorry to revive this thread, but I was curious if this `parameters.docs.source.excludeDecorators` parameter can be found anywhere in the official Storybook docs?

PR #14685, opened 2021-04-21
Addon-docs: Add classnames for Preview block

frassinier:
Issue: #14673 

## What I did

I've added className for `Show/Hide code` toggle and for `Copy` button of the `Preview` block

## How to test


shilman (approved):
LGTM! thanks @frassinier !!! 🙏

Ayushsunny (approved):
Awesome idea looks good to me!

You can say it's working from this message in the build: https://github.com/storybookjs/storybook/pull/14685/checks

But make sure that finishes with ✔️. After that, it should be good to merge

lib/components/src/blocks/Preview.tsx
```diff
@@ -121,13 +122,21 @@ const getSource = (
     case expanded: {
       return {
         source: <StyledSource {...withSource} dark />,
-        actionItem: { title: 'Hide code', onClick: () => setExpanded(false) },
+        actionItem: {
+          title: 'Hide code',
+          className: 'code-toggle code-toggle--expanded',
```
ndelangen, line 127: Perhaps these classnames are a bit too generic? Should they possible be more specifically named for storybook?

We have this:
https://github.com/storybookjs/storybook/blob/08ed3928329c069b45dffa908955c21ab8bcdac1/lib/components/src/blocks/ArgsTable/ArgsTable.tsx#L380-L381

Prefixing with `docblock-argstable-`.. 

https://github.com/storybookjs/storybook/blob/04acc44510fc189003e643c3262e1db46275f859/lib/components/src/blocks/ArgsTable/ArgValue.tsx#L180

Prexiging with `sbdocs-`

https://github.com/storybookjs/storybook/blob/f74a04a411c438880b2f08e4e7a5e8cc8b2359f9/lib/ui/src/components/sidebar/Tree.tsx#L127

Not really prefixing at all.. 😱 

https://github.com/storybookjs/storybook/blob/50e7385d84972366b0ee2ae6a9e50bda82929cc4/lib/components/src/blocks/Preview.tsx#L214

Prefixing with `docs-`

https://github.com/storybookjs/storybook/blob/3377bb9861efccb8c9bb4dfc1c5515b59da11efe/lib/ui/src/settings/shortcuts.tsx#L272

No prefix!

https://github.com/storybookjs/storybook/blob/99d0e7a343dd0e978530054c8b43f80003121634/lib/components/src/tabs/tabs.tsx#L168

No prefix!

https://github.com/storybookjs/storybook/blob/f74a04a411c438880b2f08e4e7a5e8cc8b2359f9/lib/ui/src/components/sidebar/Sidebar.tsx#L111

A prefix-less `container` and a second class on the same element prefixed with `sidebar-`

---

I propose to rename these classes to be prefixed with  `docblock-`.

ndelangen (changes requested):

lib/components/src/blocks/Preview.tsx
frassinier: Totally agreed!
The "hard" part of the PR was to choose a class name TBH.
I can rename all of them but keep in mind that some people rely on these class names to override the style.
Can I suggest adding prefixed class names instead of a whole renaming?

ndelangen: Whatever we have now/already, we can never ever change. that 🚢 has sailed..
( we could change these classes in 7.0, if we deem it important enough )

But what we can do, is not make the same 'mistake' again, and use a proper css-"namespace" / prefix for anything we add starting now.

ndelangen: > I propose to rename these classes to be prefixed with docblock-.

I should have been more clear. I meant "these classes we're adding now"

frassinier: Let me fix that then!

lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
```diff
@@ -155,15 +155,7 @@ export const SyntaxHighlighter: FunctionComponent<Props> = ({
...
       {copyable ? (
-        <ActionBar
-          actionItems={[
-            {
-              title: copied ? 'Copied' : 'Copy',
-              className: `copy-button ${copied ? 'copy-button--copied' : ''}`,
-              onClick,
-            },
-          ]}
-        />
```
ndelangen: You decided these classnames were not really needed?

ndelangen (approved):

lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
frassinier: Yes because it was more about the code part itself which is not directly related to an SB block! It will not solve #14673

ndelangen:
ODD CI failure, asking @shilman to take a look.

PR #14689, opened 2021-04-22
Docs: Remove `babel-loader` and `@babel/core` peer deps

gaetanmaisse:
Issue: https://github.com/storybookjs/storybook/issues/13183 

## What I did

Docs:
 - Remove `babel-loader` peer dep as `babel-loader` is required in the context of `builder-webpack4` so there is no need to have it as a dep.
 - Remove `@babel/core` peer dep as it is already there as regular dep

## How to test

- As these changes are quite critical they are somehow covered by all the examples + chromatic + E2E tests

### DIFF
--- a/addons/docs/README.md
+++ b/addons/docs/README.md
@@ -103,10 +103,10 @@ First add the package. Make sure that the versions for your `@storybook/*` packa
 yarn add -D @storybook/addon-docs
 ```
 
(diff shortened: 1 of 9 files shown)


shilman (approved):
LGTM. should we patch this into 6.2?

gaetanmaisse:
The real issue with `babel-loader` was fixed a while ago and this PR is just some maintenance/cosmetic improvements so it's not mandatory to backport them to 6.2 I guess.

PR #14690, opened 2021-04-22
Build: Create webpack stats target directory if needed and accept boolean flag

ghengeveld:
Issue: -

## What I did

Instead of `writeFile` this uses `outputFile` so that it will create directories if needed.
With this change, `--webpack-stats-json` won't require a value anymore. If the flag is present with no value (`true`), it will use the outputDir as its value.
Also I updated the args definitions to correctly use `<>` (required) and `[]` (optional)

## How to test


shilman (approved):
LGTM 💯

PR #14692, opened 2021-04-22
Core: Don't recreate a bound story function each time we call a decorated story

tmeasday:
Issue:#12255

## What I did

Rather than using a closure and rebinding the story function with the context as it is decorated, instead use a getStoryContext() function which is defined once at decoration time to supply the context.

## How to test

We have a Chromatic test. Will think about a Jest one

There is one

No

### DIFF
--- a/lib/addons/src/types.ts
+++ b/lib/addons/src/types.ts
@@ -53,6 +53,7 @@ export interface StoryIdentifier {
   name: StoryName;
 }
 
+export type StoryContextUpdate = Partial<StoryContext>;
 export type StoryContext = StoryIdentifier & {
   [key: string]: any;
   parameters: Parameters;
@@ -93,8 +94,13 @@ export interface OptionsParameter extends Object {
 
 export type StoryGetter = (context: StoryContext) => any;
 
+// This is the type of story function passed to a decorator -- does not rely on being passed any context
+export type PartialStoryFn<ReturnType = unknown> = (p?: StoryContextUpdate) => ReturnType;
+// This is a passArgsFirst: false user story function
 export type LegacyStoryFn<ReturnType = unknown> = (p?: StoryContext) => ReturnType;
+// This is a passArgsFirst: true user story function
 export type ArgsStoryFn<ReturnType = unknown> = (a?: Args, p?: StoryContext) => ReturnType;
+// This is either type of user story function
 export type StoryFn<ReturnType = unknown> = LegacyStoryFn<ReturnType> | ArgsStoryFn<ReturnType>;
 
 export type StoryWrapper = (
@@ -136,16 +142,16 @@ export interface StoryApi<StoryFnReturnType = unknown> {
 }
 
 export type DecoratorFunction<StoryFnReturnType = unknown> = (
-  fn: StoryFn<StoryFnReturnType>,
+  fn: PartialStoryFn<StoryFnReturnType>,
   c: StoryContext
-) => ReturnType<StoryFn<StoryFnReturnType>>;
+) => ReturnType<LegacyStoryFn<StoryFnReturnType>>;
 
 export type LoaderFunction = (c: StoryContext) => Promise<Record<string, any>>;
 
 export type DecorateStoryFunction<StoryFnReturnType = unknown> = (
-  storyFn: StoryFn<StoryFnReturnType>,
+  storyFn: LegacyStoryFn<StoryFnReturnType>,
   decorators: DecoratorFunction<StoryFnReturnType>[]
-) => StoryFn<StoryFnReturnType>;
+) => LegacyStoryFn<StoryFnReturnType>;
 
 export interface ClientStoryApi<StoryFnReturnType = unknown> {
   storiesOf(kind: StoryKind, module: NodeModule): StoryApi<StoryFnReturnType>;
--- a/lib/client-api/src/decorators.test.ts
+++ b/lib/client-api/src/decorators.test.ts
@@ -2,7 +2,7 @@ import { StoryContext } from '@storybook/addons';
 
 import { defaultDecorateStory } from './decorators';
 
-function makeContext(input: Record<string, any>): StoryContext {
+function makeContext(input: Record<string, any> = {}): StoryContext {
   return {
     id: 'id',
     kind: 'kind',
@@ -24,7 +24,7 @@ describe('client-api.decorators', () => {
     const decorated = defaultDecorateStory(() => order.push(4), decorators);
 
     expect(order).toEqual([]);
-    decorated();
+    decorated(makeContext());
     expect(order).toEqual([3, 2, 1, 4]);
   });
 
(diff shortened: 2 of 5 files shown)


lib/addons/src/types.ts
```diff
@@ -93,8 +94,13 @@ export interface OptionsParameter extends Object {
...
 export type StoryGetter = (context: StoryContext) => any;
 
-export type LegacyStoryFn<ReturnType = unknown> = (p?: StoryContext) => ReturnType;
-export type ArgsStoryFn<ReturnType = unknown> = (a?: Args, p?: StoryContext) => ReturnType;
+// This is the type of story function passed to a decorator -- does not rely on being passed any context
+export type PartialStoryFn<ReturnType = unknown> = (p?: StoryContextUpdate) => ReturnType;
+// This is a passArgsFirst: false user story function
+export type LegacyStoryFn<ReturnType = unknown> = (p: StoryContext) => ReturnType;
+// This is a passArgsFirst: true user story function
+export type ArgsStoryFn<ReturnType = unknown> = (a: Args, p: StoryContext) => ReturnType;
```
tmeasday, line 102: I tightened some of these types up, by introducing a `PartialStoryFn` type (the thing a decorator expects in its first argument). 

I'm not sure if that's going to cause other problems but it seems like "A good thing".

ndelangen: LGTM

lib/addons/src/types.ts
```diff
@@ -136,16 +142,16 @@ export interface StoryApi<StoryFnReturnType = unknown> {
 }
 
 export type DecoratorFunction<StoryFnReturnType = unknown> = (
-  fn: StoryFn<StoryFnReturnType>,
+  fn: PartialStoryFn<StoryFnReturnType>,
   c: StoryContext
-) => ReturnType<StoryFn<StoryFnReturnType>>;
+) => ReturnType<LegacyStoryFn<StoryFnReturnType>>;
```
ndelangen, line 147: Why is this changed to `LegacyStoryFn`?

lib/client-api/src/decorators.ts
```diff
@@ -25,15 +11,43 @@ const defaultContext: StoryContext = {
...
-        storyFn({ ...context, ...innerContext }),
-      context
-    );
+const bindWithContext = (
+  storyFn: LegacyStoryFn,
+  getStoryContext: () => StoryContext
+): PartialStoryFn =>
+  // (NOTE: You cannot override the parameters key, it is fixed)
+  ({ parameters, ...innerContext }: StoryContextUpdate = {}) =>
+    storyFn({ ...getStoryContext(), ...innerContext });
```
ndelangen, line 20: should innerContext be called `contextUpdate`?

lib/client-api/src/decorators.ts
```diff
@@ -25,15 +11,43 @@ const defaultContext: StoryContext = {
...
+  let contextStore: StoryContext;
+  const decoratedWithContextStore = decorators.reduce(
+    (story, decorator) => decorateStory(story, decorator, () => contextStore),
+    storyFn
+  );
+  return (context) => {
+    contextStore = context;
+    return decoratedWithContextStore(context);
+  };
+};
```
ndelangen, line 53: Let's add a bunch of unit-tests asserting:
```
  // We use a trick to avoid recreating the bound story function inside `decorateStory`.
  // Instead we pass it a context "getter", which is defined once (at "decoration time")
  // The getter reads a variable which is scoped to this call of `decorateStory`
  // (ie to this story), so there is no possibility of overlap.
  ```

holds true, now and in the future.

File: lib/client-api/src/decorators.ts (hunk 3 above)
Comment (tmeasday) [reply lib/client-api/src/decorators.ts:20]: That would make sense

File: lib/client-api/src/decorators.ts (hunk 4 above)
Comment (tmeasday) [reply lib/client-api/src/decorators.ts:53]: I agree, I am just scratching my head a bit on how to actually do that!

File: lib/addons/src/types.ts (hunk 1 above)
Comment (tmeasday) [reply lib/addons/src/types.ts:102]: I am currently having to update a *lot* of things to get bootstrap to pass. I wonder if I should split it into a different PR.

File: lib/addons/src/types.ts (hunk 2 above)
Comment (tmeasday) [reply lib/addons/src/types.ts:147]: Decorators deal in `LegacyStoryFn`s always (they never expect to pass args as the first argument, or receive it).

In the store, we transform `original` (an `ArgsStoryFn`) into `finalStoryFn` (a `LegacyStoryFn`) before decorating it. When we actually render the decorated story, we pass `context` in as first argument.

File: lib/addons/src/types.ts (hunk 1 above)
Comment (ndelangen) [reply lib/addons/src/types.ts:102]: Yeah, we should prioritize making the runtime work correctly first, considering the code is very dynamic and complex, meaning the types are too.

Also this code is used in many places. So making the types strict is going to be a lot of work.

Deferring the "stricter types" work sounds reasonable to me.

File: lib/addons/src/types.ts (hunk 2 above)
Comment (ndelangen) [reply lib/addons/src/types.ts:147]: I see ✅

**Comment by tmeasday** [comment]
@ndelangen I figured out some tests and it's passing 🎉 

Here's the typing change: https://github.com/storybookjs/storybook/pull/14701

File: lib/client-api/src/decorators.test.ts [hunk 5]
```diff
@@ -42,6 +42,50 @@ describe('client-api.decorators', () => {
...
+    const decoratedTwo = defaultDecorateStory((c) => contexts.push(c), decorators);
+
+    decoratedOne(makeContext({ story: 1 }));
+    decoratedTwo(makeContext({ story: 2 }));
+
+    resolve();
+    await fence;
+
+    expect(contexts[0].story).toBe(1);
+    expect(contexts[1].story).toBe(2);
```
ndelangen, line 86: I'd really like to get rid of the `.story` property at some point. Can we instead test for `.id`? AFAIK the `.story` property we don't actually use, and is just there to not break backwards compatibility.

lib/client-api/src/decorators.ts
```diff
@@ -25,15 +21,44 @@ const defaultContext: StoryContext = {
...
+  let contextStore: StoryContext;
+  const decoratedWithContextStore = decorators.reduce(
+    (story, decorator) => decorateStory(story, decorator, () => contextStore),
+    storyFn
+  );
+  return (context = defaultContext) => {
+    contextStore = context;
+    return decoratedWithContextStore(context); // Pass the context directly into the first decorator
+  };
+};
```
ndelangen, line 64: Do we need to make changes to the other decorateStory functions (vue has one, and I think angular has one too)

ndelangen: Or is that simply not needed since the problem we're solving is react-specific?

ndelangen: Not creating a new function is a general performance optimization though I think.. Should use less memory and fewer garbage collections needed.

Unless I'm wrong about that?

lib/client-api/src/decorators.test.ts
tmeasday: Oh, this `.story` property is just one I made for this test. Let's call it something else.

lib/client-api/src/decorators.ts
tmeasday: I guess you are right. It does make the `decorateStory` function a little bit more complex and hard to reason about (perhaps that's debatable?) and the angular/vue/etc versions are sort of complicated already.

I'm open to changing it if you want, I'm not sure the perf difference is really going to be a big deal (stories don't get rendered all that often).

ndelangen: I'd prefer making all decorateStory function be act similar, so that when you understand 1, you understand them all.

tmeasday: Discussed this with @shilman.

We aren't super comfortable changing the already quite complex decorator "applicators" for other view layers. We don't understand the view layers well enough to know if it is a good idea or will change the behaviour in unexpected ways.

I think this will serve as a template for future view layers and if maintainers of other view layers want to bring it across as an optimization/simplification they can.

lib/client-api/src/decorators.test.ts
tmeasday, line 86:
```suggestion
    decoratedOne(makeContext({ value: 1 }));
    decoratedTwo(makeContext({ value: 2 }));

    resolve();
    await fence;

    expect(contexts[0].value).toBe(1);
    expect(contexts[1].value).toBe(2);
```

shilman (approved):
LGTM! 💯

PR #14701, opened 2021-04-23
Tighten up `StoryFn` types

tmeasday:
In https://github.com/storybookjs/storybook/pull/14692 I added a `PartialStoryFn` type, which is the thing that a decorator accepts as an argument:

```
export type PartialStoryFn<ReturnType = unknown> = (p?: StoryContextUpdate) => ReturnType;
```

A decorator doesn't *have* to pass the context into the story it is passed, it's automatically bound.

This fact led to our other story types (`LegacyStoryFn`, `ArgsStoryFn` and the union `StoryFn`) having optional arguments, which was actually really wrong, you would expect them to break if passed no context.

This PR attempts to tighten that up. It's kind of a big change, not sure if it needs to be a major version change.

### DIFF
--- a/app/angular/src/client/preview/types.ts
+++ b/app/angular/src/client/preview/types.ts
@@ -11,7 +11,7 @@ export interface ICollection {
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
 }
 
 export interface IStorybookSection {
--- a/app/aurelia/src/client/preview/types.ts
+++ b/app/aurelia/src/client/preview/types.ts
@@ -1,9 +1,9 @@
-import { StoryFn } from '@storybook/addons';
+import { PartialStoryFn } from '@storybook/addons';
 import { IRegistry, IContainer, Constructable } from 'aurelia';
 import { Component } from './decorators';
 
 export interface RenderMainArgs {
-  storyFn: StoryFn<Partial<StoryFnAureliaReturnType>>;
+  storyFn: PartialStoryFn<Partial<StoryFnAureliaReturnType>>;
   selectedKind: string;
   selectedStory: string;
   showMain: () => void;
@@ -26,7 +26,7 @@ export interface ShowErrorArgs {
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
 }
 
 export interface IStorybookSection {
--- a/app/html/src/client/preview/types.ts
+++ b/app/html/src/client/preview/types.ts
@@ -4,7 +4,7 @@ export type StoryFnHtmlReturnType = string | Node;
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
 }
 
 export interface IStorybookSection {
--- a/app/mithril/src/client/preview/types.ts
+++ b/app/mithril/src/client/preview/types.ts
@@ -4,7 +4,7 @@ export type { RenderContext } from '@storybook/core';
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
 }
 
 export interface IStorybookSection {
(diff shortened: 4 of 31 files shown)


tmeasday:
One thing that makes this PR kind of wrong right now is that actually decorated stories aren't `LegacyStoryFn`s because they do make the first argument optional, thanks to this kind of crazy thing:

image

I was going to get rid of that in https://github.com/storybookjs/storybook/pull/14692 but I worry it is a breaking change.

In this PR we either need to get rid of that (and make `decorateStory` return a `LegacyStoryFn`), or else add **another** type `OptionalLegacyStoryFn = (p: StoryContextUpdate) => ReturnType;` (note this is different to the `PartialStoryFn` above).

app/angular/src/client/preview/types.ts
```diff
@@ -11,7 +11,7 @@ export interface ICollection {
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
```
ndelangen, line 14: You cannot create types like this in TypeScript, what you need to do is create overloading:

official docs:
https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

tutorial that explains how to, pretty well:
https://www.tutorialsteacher.com/typescript/function-overloading

app/aurelia/src/client/preview/types.ts
```diff
@@ -26,7 +26,7 @@ export interface ShowErrorArgs {
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
```
ndelangen, line 29: user overloading instead

app/html/src/client/preview/types.ts
```diff
@@ -4,7 +4,7 @@ export type StoryFnHtmlReturnType = string | Node;
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
```
ndelangen, line 7: user overloading instead

app/mithril/src/client/preview/types.ts
```diff
@@ -4,7 +4,7 @@ export type { RenderContext } from '@storybook/core';
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
```
ndelangen, line 7: user overloading instead

app/preact/src/client/preview/types.ts
```diff
@@ -11,7 +11,7 @@ export interface ShowErrorArgs {
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
```
ndelangen, line 14: user overloading instead

app/react/src/client/preview/types.ts
```diff
@@ -13,7 +13,7 @@ export type StoryFnReactReturnType = ReactElement<unknown>;
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
```
ndelangen, line 16: use overloading instead

app/server/src/client/preview/types.ts
```diff
@@ -13,7 +13,7 @@ export type FetchStoryHtmlType = (
 
 export interface IStorybookStory {
   name: string;
-  render: () => any;
+  render: ((context: any) => any) | ((args: any, context: any) => any);
```
ndelangen, line 16: use overloading instead

tmeasday:
@ndelangen do you know why we don't share types between the different view layers (or that we don't use the types from `@storybook/addons` in the first place?)

ndelangen:
Reason are
- avoid circular dependencies
- unfinished cleanup

A large scale cleanup of all storybook types would be very nice some day.

tmeasday:
Also discussed this with @shilman and this feels like a breaking/7.0 change. Probably easiest to do with the large scale cleanup you mention @ndelangen.

Going to leave this PR to bitrot for now as a suggestion for later changes. Should we close it?

MichaelArestad:
>  Should we close it?

Yep!

PR #14736, opened 2021-04-27
Addon-a11y: Document best practices for handling failing rules

kaelig:
Issue: when folks disable a11y checks in their stories, it doesn’t communicate what rule was failing and why. On a large codebase we quickly ended up with a lot of disabled a11y checks, because people didn't know there were better ways of handling these than fully disabling checks.

## What I did

I documented various ways of overriding rules to make checks go green without fully disabling a11y tests.

I'd appreciate feedback on voice & tone, the examples given, and how to write code examples that will look familiar to the Storybook audience.

## How to test

- Does this need an update to the documentation? Probably not?

### DIFF
--- a/addons/a11y/README.md
+++ b/addons/a11y/README.md
@@ -36,11 +36,82 @@ export const inaccessible = () => (
 );
 ```
 
+## Handling failing rules
+
+When a Story has failing rules (accessibility violations), there are mutliple ways to handle these failures.
+
+### Global overrides
+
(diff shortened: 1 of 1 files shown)


shilman (approved):
❤️ ❤️ ❤️

addons/a11y/README.md
```diff
@@ -36,11 +36,82 @@ export const inaccessible = () => (
 );
 ```
 
+## Handling failing rules
+
+When a Story has failing rules (accessibility violations), there are mutliple ways to handle these failures.
```
atesgoral, line 41: typo: mutliple

PR #14744, opened 2021-04-28
Build: Change nx cloud access token to read-only token

FrozenPandaz:
Issue:

## What I did

The token checked into the repo is a `read-write` token which is not a huge deal but should be a `read-only` token.

## How to test

You should still be able to get cache hits locally and CI should populate the cache.

If you pull down this branch, the following should be instant after installing dependencies:

`yarn prepare` > Core & Examples (core)

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -22,7 +22,7 @@
           "package",
           "prepare"
         ],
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM! 💯

PR #14745, opened 2021-04-28
Core: Fix `features`, `core`, `logLevel` in main.js config types

valerybugakov:
Issue: `features` setting [was added recently](https://github.com/storybookjs/storybook/pull/14478) to the `main.js` config. It's not reflected in the `StorybookConfig` type.

## What I did

Added `features`, `core` and `loglevel` to the `StorybookConfig` type.

## How to test

Updated usage of `StorybookConfig` in examples should throw an error if the config type doesn't have some field.

### DIFF
--- a/examples/official-storybook/main.ts
+++ b/examples/official-storybook/main.ts
@@ -2,7 +2,7 @@
 
 import type { StorybookConfig } from '@storybook/react/types';
 
-module.exports = {
+const config: StorybookConfig = {
   stories: [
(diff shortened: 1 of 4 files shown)


shilman (commented):
@valerybugakov great addition! however I only see `features` added. Did you forget to check something in?

valerybugakov:
@shilman, right, I've pushed the last commit 👍

shilman (approved):
💯 !!!

PR #14749, opened 2021-04-29
Addon-a11y/Storyshots: Upgrade axe-core to 4.2.0 and related dependencies

kaelig:
## What I did

Axe-core v4.2.0 was just released ([release notes](https://github.com/dequelabs/axe-core/releases/tag/v4.2.0)), let's update to this new version across the board.

**[Possible breaking change] (?)** This may result in failing builds for projects that automate a11y tests in CI, as new accessibility violations from rules introduced in 4.2.0 may be reported.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes (as in: tests may break)

### DIFF
--- a/addons/a11y/package.json
+++ b/addons/a11y/package.json
@@ -53,7 +53,7 @@
     "@storybook/components": "6.3.0-alpha.15",
     "@storybook/core-events": "6.3.0-alpha.15",
     "@storybook/theming": "6.3.0-alpha.15",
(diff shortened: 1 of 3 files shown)


shilman (approved):
Thanks @kaelig 🙏 

This is in the semver grey zone, and I'm ok with the change. I'll put it on alpha and roll back if anybody complains too loudly.

PR #14765, opened 2021-04-29
Chore: (Docs): Minor tweaks for Vue 3 and FAQ updates

jonniebigodes:
With this pull request the FAQ, and install instructions for Vue 3 based projects were updated to drop the usage of the pre-release version (6.2). 

Also, some work was done on the FAQ section to reflect this change as we were incorrectly mentioning to the users that they need to specifically set the `vue3` for Storyshots, when that's not the case. It automatically detected based on a recent test.

Feel free to provide feedback

### DIFF
--- a/app/vue/README.md
+++ b/app/vue/README.md
@@ -28,11 +28,11 @@ You can also build a [static version](https://storybook.js.org/docs/vue/workflow
 
 ## Vue Notes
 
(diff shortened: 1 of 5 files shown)


shilman (approved):
Great stuff @jonniebigodes 🔥

PR #14777, opened 2021-04-30
fix: mdx compiler switch to new path

saranrapjs:
This is a proposed patch to an existing draft branch on the main storybook repo here: https://github.com/storybookjs/storybook/pull/14769


shilman (approved):
Nice @saranrapjs -- makes sense. Thanks for putting this together!

PR #14783, opened 2021-05-02
Add documentation section about linking to other stories and pages from MDX

Dschungelabenteuer:
## What I did

As asked [on the Discord server](https://discord.com/channels/486522875931656193/501692020226654208/837504203169071135) by @shilman, I've tried to explain quite briefly how to provide links (and also include anchors) to other stories and content. With a small notice about the anchors being ignored when targeting the "story" view mode (Canvas tab).

Not sure this is clear enough, and whether that video is relevant/useful.

## How to test

- Does this need a new example in the kitchen sink apps? Maybe
- Does this need an update to the documentation? This is the update to the documentation


ndelangen (approved):

docs/writing-docs/mdx.md
```diff
@@ -160,6 +160,29 @@ To get a "documentation-only story", in your UI, define a `<Meta>` as you normal
...
+```md
+[Go to specific story canvas](?path=/story/some--id)
+```
+
+You can also use anchors to target a specific section of a page:
+```md
+[Go to the conclusion of the documentation page](?path=/docs/some--id#conclusion)
+```
+
+Note that when using the Controls addon, anchors are ignored by the Canvas tab since the URL gets rewritten to track the args values.
```
jonniebigodes, line 182: @Dschungelabenteuer one thing about this. If you could adjust this to  :

```html
<div class="aside">
💡 <strong>Note:</> By applying this pattern with the Controls addon, all anchors will be ignored in Canvas based on how Storybook handles URLs to track the args values.
</div
```

Something like this. Let me know what you thing?

jonniebigodes (changes requested):
@Dschungelabenteuer thank you for the time and effort you put into this pull request. I left one note regarding it. Let me know and we'll go from there.

Stay safe

docs/writing-docs/mdx.md
Dschungelabenteuer: Much better 👍

Dschungelabenteuer: Much better 👍

shilman (approved):
Thanks so much for documenting @Dschungelabenteuer !! We should fix this @ghengeveld ...

PR #14806, opened 2021-05-04
Update MIGRATION.md with ESM jest transform tip

kylegach:
Issue: #13478 

## What I did

Add tip about transforming newly-available ESModule packages in jest

## How to test

- Does this need an update to the documentation? This _is_ an update to documentation


shilman (approved):
Awesome @kylegach. Thank you so much!

PR #14808, opened 2021-05-04
Accessibility: Adds title to close button on settings page

thawkin3:
Issue: N/A

## What I did

I've added a `title` attribute to the Close button on the Settings page.

This is needed to provide a helpful title when using a screen reader, otherwise the screen reader only sees a button element without telling the user what the button does. This also provides a helpful selector since previously there wasn't a good way to identify the generic button in something like a Cypress test.

## How to test

1. Start any Storybook app locally
2. Navigate to the `?path=/settings/about` page
3. In the top-right, hover over the X button with your mouse and verify that you see the title text "Close settings page"

## Screenshots

Screen Shot 2021-05-04 at 3 06 48 PM
Screen Shot 2021-05-04 at 3 07 12 PM
Screen Shot 2021-05-04 at 3 07 45 PM

### DIFF
--- a/lib/ui/src/settings/index.tsx
+++ b/lib/ui/src/settings/index.tsx
@@ -82,6 +82,7 @@ const Pages: FunctionComponent<{
             e.preventDefault();
             return onClose();
           }}
+          title="Close settings page"
         >
           <Icons icon="close" />
         </IconButton>


thawkin3:
I checked the review app, and it's behaving nicely!

shilman (approved):
Thank you @thawkin3! 👍

thawkin3:
@shilman No prob! Thanks for the speedy review!

PR #14819, opened 2021-05-05
refactor(core-server): rename options prop from 'staticDir' to 'staticDirs'

oscard0m:
## 📝 Summary

Rename option property `staticDir` to `staticDirs`

## ⛱ Motivation and Context

The variable is `string[]` type so it has sense to be in plural.

Context: https://github.com/storybookjs/storybook/pull/14068#pullrequestreview-615204622
@shilman

### DIFF
--- a/lib/core-common/src/types.ts
+++ b/lib/core-common/src/types.ts
@@ -124,7 +124,7 @@ export interface CLIOptions {
   ignorePreview?: boolean;
   previewUrl?: string;
   host?: string;
-  staticDir?: string[];
+  staticDirs?: string[];
   configDir?: string;
   https?: boolean;
   sslCa?: string[];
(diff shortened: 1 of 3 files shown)


shilman (commented):
Did you test? I think the problem is here:

https://github.com/storybookjs/storybook/blob/next/lib/core-server/src/cli/dev.ts#L17
https://github.com/storybookjs/storybook/blob/next/lib/core-server/src/cli/prod.ts#L29

If we update the name of the CLI flag, then we need to do a bunch of extra code to make it a non-breaking change. Since this will be replaced by https://github.com/storybookjs/storybook/issues/13250, I think it's probably not worth the effort.

oscard0m:
> Did you test? I think the problem is here:
> 
> https://github.com/storybookjs/storybook/blob/next/lib/core-server/src/cli/dev.ts#L17
> https://github.com/storybookjs/storybook/blob/next/lib/core-server/src/cli/prod.ts#L29

I just run `yarn test` in my local. It's a bit confusing the feedback in PR checks, I will ask some questions in my next contribution about it :)

> If we update the name of the CLI flag, then we need to do a bunch of extra code to make it a non-breaking change. Since this will be replaced by #13250, I think it's probably not worth the effort.

Yep, with this #13250 I would not go fior this refactor. Closing this PR.

PR #14408, opened 2021-03-31
UI: Boolean control change for UI Fix toggle button

cjmcintyre:
Issue: [UI: Fix toggle button for custom theming #13924](https://github.com/storybookjs/storybook/issues/13924)

**Boolean control change for UI Fix toggle button for custom themeing**

_Apologies in advance if anything here is out of your norm.. First time contributing to an O/S project + first time using storybook... 😟_

## What I did
**Reproduced by:** git cloning, added custom theming [ .storybook/YourTheme.js (complex example) ](https://storybook.js.org/docs/react/configure/theming) and a control button as per examples on playbook website.

When adding the custom theme it seems Control takes the property form **barBg** as indicated beforehand, the propose change will make it take from **appBg** instead.

Example of my custom theme 

```
    // UI
    appBg: 'white',
    appContentBg: 'silver',
    appBorderColor: 'grey',
    appBorderRadius: 4,  

    // Toolbar default and active colors
    barTextColor: 'silver',
    barSelectedColor: 'black',
    barBg: 'black',
  
```

## How to test
**Tested by:**  Reproducing above and then applying fix to Boolean.tsx and building locally with **yarn build-storybook** to test (Sorry no automation this time...)

playbook

- Is this testable with Jest or Chromatic screenshots?
Not sure, sorry.
- Does this need a new example in the kitchen sink apps?
Not that I'm aware of.
- Does this need an update to the documentation?
Not that I'm aware of.

If your answer is yes to any of these, please make sure to include it in your PR.

PR #14820, opened 2021-05-05
UI: Styling updates

domyen:
Issue: #14737 and other UI improvements

## What I did
- ArgsTable when in controls panel: Fix ugly black border in Safari
- Boolean control: Switch position of `True | False` to `False | True` per #14737
- Sidebar: Buttons for root heading and expand/collapse now only get the outline-y focus state if they're keyboarded to
- Toolbar icons: Update copy icon to a link (which is what it does) and share icon to the alternative share icon
- Icons: Update icon set to have parity with the design system
- a11y addon: minor layout alignment fixes

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes. Go to the PR check.


shilman (approved):
Looks great @domyen! :100:

shilman:
@domyen I think this PR introduces a slight regression. In the case where there's a boolean arg with an `undefined` value, in the old version the control would display `false`. After this PR the control displays `true`. While neither option is technically correct, the old behavior is more correct than the new one. I think we are dealing with this in the "undefined values" work, so I'm not sure if it's critical to fix, but I wanted to get it on your radar.

domyen:
@shilman That's strange, this was a styling change it doesn't affect whether the boolean was `checked`. When I go to the Storybook itself I see that it still displays "false" [here](https://5a375b97f4b14f0020b0cda3-gbgvhyavvb.chromatic.com/?path=/story/controls-boolean--undefined).

image

ndelangen:
🤔

shilman:
@domyen I'll keep an eye out for it after the undefined values changes come in and provide a repro if it's still broken then

domyen:
Thanks! I handed the undefined state project back to @tmeasday yesterday so he'll probably know if anything is wrong.

shilman:
Ok @tmeasday @domyen you can see the changes in every new PR that has unaccepted changes related to this bug. I'm not sure why it didn't show up in the original PR.

shilman:
False alarm. It was introduced here: #14408, reverting

PR #14822, opened 2021-05-06
Added if on viewMode to avoid incorrect Settings Router load

dChancellor:
Issue: #14718

## What I did

Added an if statement that checks the viewMode before loading the Settings Route. This fixes the bug mentioned in issue #14718.

### DIFF
--- a/lib/ui/src/app.tsx
+++ b/lib/ui/src/app.tsx
@@ -47,11 +47,16 @@ const App = React.memo<AppProps>(
           {
             key: 'settings',
             render: () => <SettingsPages />,
-            route: (({ children }) => (
-              <Route path="/settings" startsWith>
-                {children}
-              </Route>
-            )) as FunctionComponent,
+            route: (({ children }) => {
+              if (viewMode === 'settings') {
+                return (
+                  <Route path="/settings" startsWith>
+                    {children}
+                  </Route>
+                );
+              }
+              return null;
+            }) as FunctionComponent,
           },
         ],
       }),


lib/ui/src/app.tsx
```diff
@@ -47,11 +47,16 @@ const App = React.memo<AppProps>(
           {
             key: 'settings',
             render: () => <SettingsPages />,
-            route: (({ children }) => (
-              <Route path="/settings" startsWith>
-                {children}
-              </Route>
-            )) as FunctionComponent,
+            route: (({ children }) => {
+              if (viewMode === 'settings') {
```
ndelangen, line 51: viewMode changes over time, but this will never re-evaluate because of useMemo without dependencies..

This will mean the setting page will never, ever render anything...

ndelangen (changes requested):

lib/ui/src/app.tsx
dChancellor: Ah, okay. My apologies, I must have only tested it by loading the settings page by going directly to /settings/about rather than navigating there.

I will go back and take a look at the code base and try and wrap my head around a different solution. I appreciate your patience, as it's my first OSS contribution.

ndelangen: >  as it's my first OSS contribution.

Really really appreciated!

> Would simply adding viewMode as a dependency cause unwanted performance issues?

OI'm pretty sure the router gives it as an argument too!

darleendenno:
Hey @dChancellor 👋 Thanks for taking a stab at this! Are you still working on a solution?

darleendenno:
Hey @dChancellor I'm going to close this PR since it's a bit stale -- if you want to work on this again, let us know. 

We appreciate your contribution  ❤️

PR #14835, opened 2021-05-06
Storyshots: Make `vue-jest` and `svelte` optional peer dependencies

merceyz:
Issue: #14758 

## What I did

Made `vue-jest` and `svelte` optional peer dependencies

## How to test

Run an install and see that there is no complaint about missing peer dependencies for a framework you're not using


shilman (approved):
Thanks @merceyz !!! 🚀

PR #14880, opened 2021-05-10
Build: Remove outdated `@types/cpy` dependency

gaetanmaisse:
## What I did

Remove outdated @types/cpy dependency

> ➤ YN0061: │ @types/cpy@npm:7.1.3 is deprecated: This is a stub types definition for cpy. cpy provides its own type definitions, so you don't need @types/cpy installed!

## How to test

- Build step in CI should be 🟢


shilman (commented):
LGTM

PR #14900, opened 2021-05-12
Addon-docs: Remove all defaultValue eval-ing

tmeasday:
Telescoping on https://github.com/storybookjs/storybook/pull/14899

Issue: #14579 

## What I did

 - Simply do not eval and set `argType.defaultValue` in any case.

## What I did *not* do

 - If `argType.defaultValue` is set, we continue to use it as the initial arg value (if the arg is otherwise unset).
    - This means we do not break backwards compatibility
    - Also it means we don't need to change the implementation of actions' argType enhancer

 - I think maybe we can clean up some test cases, @shilman we should probably walk through it together

## How to test

Yes, lots of stories will be affected.


shilman (approved):
LGTM!

PR #14893, opened 2021-05-11
Chore: (Docs) Adjusts snippets in decorators for React

jonniebigodes:
With this pull request, the documentation is updated to document how to use the Story() within decorators in React.

Addresses and closes #13356

What was done:
- Added some snippets to document how to use the Story() within decorators.
- Adjusted the imports in TS stories for clarity.
- Renamed one snippet to avoid repetition.

Following the issue above I saw that it goes to #12255. And the issue was closed with the release of Storybook 6.3.x alpha version. Until Storybook 6.3 is released, at least we have both documented. 

Feel free to provide feedback.

PR #14907, opened 2021-05-12
Chore: (Docs) Minor polish to the decorators docs

### LINKED ISSUE #13927: Custom styles for components based on view mode

I want to add custom styles if a component is shown in the 'canvas' tab that are not applied in the 'docs' tab (or the other way around).

An example is stories about full page containers: in de 'docs' tab these should have a fixed height to show smaller examples, however on the 'canvas' tab I want them to fill the whole page.

Some way to access some storybook-context variables, such as the viewMode, inside the stories (or decorators) itself. However I'm not sure about the best way to add this.

Right now we have a decorator like this (example with React):

```tsx
(Story: any) => {
  const isFullscreen = window.location.search.includes("&viewMode=story");
  if (isFullscreen) {
    return <Story />;
  }
  return (
    <div style={{ height: 600 }}>
      <Story />
    </div>
  );
}
```

I have not found any other solution to get the same results.

If I could get pointed to the right direction, I can contribute to add this feature.

**Additional context**

/

jonniebigodes:
With this small pull request, the writing-stories/decorators function was updated and polished a bit to address and close #13927.

What was done:
- Polished the documentation.
- Included the missing items available to the decorators context.

Merge this after #14893 to prevent issues with the build.

Feel free to provide feedback

### DIFF
--- a/docs/writing-stories/decorators.md
+++ b/docs/writing-stories/decorators.md
@@ -2,13 +2,13 @@
 title: 'Decorators'
 ---
 
-A decorator is a way to wrap a story in extra “rendering” functionality. Many addons define decorators in order to augment your stories with extra rendering or gather details about how your story is rendered.
+A decorator is a way to wrap a story in extra “rendering” functionality. Many addons define decorators to augment your stories with extra rendering or gather details about how your story is rendered.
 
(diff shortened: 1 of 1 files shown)


shilman (commented):
LGTM!

PR #14917, opened 2021-05-13
Build: Put E2E tests back on track

gaetanmaisse:
## What I did

 - Web components: Run `web_components_typescript` e2e test with node_modules linker instead of PnP. It looks like there is an issue with PnP, it sometimes fails to parse TypeScript stories, I didn't understand why yet 🤷🏻  
 - Angular: Add a config for Angular 11 and use it in the e2e until Angular 12 is supported.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,7 +195,7 @@ jobs:
           command: yarn wait-on http://localhost:6000
       - run:
           name: Run E2E tests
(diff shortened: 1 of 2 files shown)


gaetanmaisse:
@shilman @ndelangen Self-merging to get `next` back to 🟢 . We will update the Angular part as soon as Angular 12 will be supported, and I'm still working on the WC TypeScript + PnP issues.

Issue #14926, opened 2021-05-14
Controls do not handle `null` or incorrect typed args properly


tmeasday:
**Describe the bug**
Arg type inference can be incorrect; users can make mistakes.

An arg can be "wrong" in various ways:

 1. It can be `null` for non-object arg types
 2. It can be `undefined` for required arg types
 3. It can be the wrong type (e.g. a string when the arg type is number).

Most of our controls either don't show any error, show the value wrong, or even error out when you do 1&3. For 2, they work but they should probably tell you that's an issue.

**To Reproduce**

Will add stories in a branch.

penx:
This may be partially fixed by #17102

tmeasday:
@penx this issue is more for the reverse of what #17102 -- when the controls receive a value they do not expect.

valentinpalkovic:
Thanks for filing this tracking issue back in 2021! Controls were merged into core in v9 and we've progressively added parse-error UI (red borders + tooltips) for numbers and objects. The original null/undefined/wrong-type sub-cases are largely handled in v10. Closing as resolved; please open a focused issue if a specific control still misbehaves.

PR #14927, opened 2021-05-14
Handle incorrect arg values in controls

tmeasday:
Issue: #14926

## What I did

Added stories highlighting the problem.


lib/components/src/controls/Boolean.stories.tsx
```diff
@@ -11,11 +11,18 @@ const Template = (initialValue?: boolean) => {
   return (
     <>
       <BooleanControl name="boolean" value={value} onChange={(newVal) => setValue(newVal)} />
-      <p>value: {typeof value === 'boolean' ? value.toString() : value}</p>
+      <pre>{JSON.stringify(value) || 'undefined'}</pre>
```
ghengeveld, line 14: `String(value)` would've done the trick. Unless you want additional quotes around strings.

lib/components/src/controls/Boolean.tsx
```diff
@@ -80,16 +81,23 @@ const format = (value: BooleanValue): string | null => (value ? String(value) :
...
-      checked={value || false}
-      {...{ name, onBlur, onFocus }}
-    />
-    <span>False</span>
-    <span>True</span>
-  </Label>
-);
+export const BooleanControl: FC<BooleanProps> = ({ name, value, onChange, onBlur, onFocus }) => {
+  const onSetFalse = useCallback(() => onChange(false), [onChange]);
+  if (value === undefined) {
```
ghengeveld, line 86: What about `null`?

lib/components/src/controls/Number.stories.tsx
```diff
@@ -6,22 +6,22 @@ export default {
...
-  return (
-    <>
-      <NumberControl name="number" value={value} onChange={(newVal) => setValue(newVal)} />
-      <p>{value}</p>
-    </>
-  );
-};
+export const Basic = () => Template(10);
+
+export const Zero = () => Template(0);
```
ghengeveld, line 21: What about decimals?

lib/components/src/controls/Range.tsx
```diff
@@ -27,6 +27,13 @@ const RangeInput = styled.input(({ theme }) => ({
     cursor: 'pointer',
   },
 
+  // FIXME
```
ghengeveld, line 30: Would be nice to explain what needs to be fixed here.

lib/components/src/controls/Boolean.stories.tsx
tmeasday: This is just common to all the stories for the different controls

ghengeveld (approved):

lib/components/src/controls/Boolean.tsx
tmeasday: Yes, sorry this PR was supposed to be a draft, and was based on other PRs that handled `undefined.

It doesn't actually handle incorrect arg values yet -- e.g. `null` or the wrong type.

lib/components/src/controls/Number.stories.tsx
tmeasday: They work, but sure, we should add a story.

lib/components/src/controls/Range.tsx
tmeasday: This isn't really part of this PR. There is a separate open ticket about this exact thing.

tmeasday (commented):
Sorry this was a bit confusing @ghengeveld

tmeasday:
Not sure why this PR still includes a bunch of changes that were already merged to `next`. Probably it needs to be rebased.

ndelangen (approved):

tmeasday:
I guess we can't really merge this until we get the controls to handle the errors at least a little better.

domyen:
@tmeasday Is there anything I need to do here?

tmeasday:
@domyen, yeah, let's figure out a design for what controls look like when they have invalid content.

A simple example is something like

```js
export const MyStory = Template.bind({})
MyStory.argTypes = { foo: { type: 'string' } }
MyStory.args = { foo: 7 }
```

We've said `foo` is a string, but supplied the value `7`. What should we show in the controls UI in this case? (Chances are the actually story will error out).

Another example is when an arg is required but the story does not supply a value.

tmeasday:
@ndelangen -- have discussed this with @shilman and we were more of the opinion that we just let the user do what they want here -- keep in mind that `argTypes` can be inferred from TS/propTypes and TBH I don't entirely trust docgen to always get them right.

So we were thinking the issue is more just how do we deal with trying to render a value with a control that it is not equipped to handle.

ndelangen:
> could choose to render an error rather then the story
OK so that's out of the question then.

How about rendering the warning where the control would be?

tmeasday:
> How about rendering the warning where the control would be?

Seems like an OK solution. Perhaps we want to let the override the "broken" value with the control though?

ndelangen:
Like a button to set it to the default value?

tmeasday:
Sure, I guess the PR isn't helping much unmerged. But FTR I don't think we currently handle all possibilities of weird incorrect things you could pass into controls.

PR #14932, opened 2021-05-14
Storyshots: Make preact-render-to-string optional peer dependency

darekkay:
Issue: #14758

Continuation of #14835

## What I did

Made `preact-render-to-string` optional peer dependency

## How to test

Run an install and see that there is no complaint about missing peer dependencies for a framework you're not using

### DIFF
--- a/addons/storyshots/storyshots-core/package.json
+++ b/addons/storyshots/storyshots-core/package.json
@@ -126,6 +126,9 @@
     "preact": {
       "optional": true
     },
+    "preact-render-to-string": {
+      "optional": true
+    },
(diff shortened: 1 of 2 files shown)


merceyz:
This isn't necessary, see https://github.com/storybookjs/storybook/issues/14758#issuecomment-841137060 and https://github.com/storybookjs/storybook/issues/14758#issuecomment-841139481

darekkay:
I think this can be indeed closed, as stated by @merceyz 

> preact-render-to-string isn't a peer dependency so it wouldn't help

PR #14935, opened 2021-05-14
Addon-a11y: Highlight all elements correctly

yannbf:
Issue: Couldn't find an open issue for this

## What I did

There was an extra curly bracket in the css for highlighting elements in the a11y addon, that resulted in a broken stylesheet.

Before:
image

After:
image

## How to test

Build the addon and run any storybook, then toggle all elements to see the difference

```
yarn build addon-a11y
yarn start
```

Or check the chromatic build and test it there as well!

### DIFF
--- a/addons/a11y/src/a11yHighlight.ts
+++ b/addons/a11y/src/a11yHighlight.ts
@@ -21,9 +21,12 @@ const highlight = (infos: HighlightInfo) => {
   const id = HIGHLIGHT_STYLE_ID;
   resetHighlight();
 
+  // Make sure there are no duplicated selectors
+  const elements = Array.from(new Set(infos.elements));
+
   const sheet = document.createElement('style');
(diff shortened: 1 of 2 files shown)


addons/a11y/src/a11yHighlight.ts
```diff
@@ -21,9 +21,12 @@ const highlight = (infos: HighlightInfo) => {
   const id = HIGHLIGHT_STYLE_ID;
   resetHighlight();
 
+  // Make sure there are no duplicated selectors
+  const elements = Array.from(new Set(infos.elements));
```
yannbf, line 25: I noticed that sometimes the elements array would be of size 20, but actually 10 of these selectors would be just duplicated, resulting in a longer stylesheet than necessary

shilman (commented):
LGTM!

PR #14937, opened 2021-05-14
docs: update framework support

yannbf:
Angular and Svelte have been revamped in 6.2 and now support new features, we should update the support table in our docs.

## What I did

- Add angular to Inline rendering table
- Add svelte to args table
- Add vue, angular and svelte to dynamic source


shilman (approved):
LGTM!! Thanks @yannbf !!

PR #14942, opened 2021-05-14
CLI: Keep Webpack 4 builder for Angular lower than 12

gaetanmaisse:
## What I did

 - I fixed the computation of Angular version used in the users' projects
 - I enabled Angular 12 e2e tests as part of our CI checks. Angular 11 is kept for now as the release is out only for a few days and SB + Angular 12 use Webpack4 builder and SB + Angular 11 is using Webpack 5 builder

## How to test

- CI should be 🟢

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -218,8 +218,8 @@ jobs:
           command: yarn wait-on http://localhost:6000
       - run:
           name: Run E2E tests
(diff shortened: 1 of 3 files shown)


yannbf (approved):
LGTM! As soon as this is good to go I can update my other PRs and they'll be green as well :D

shilman (approved):
Great job @gaetanmaisse! Thanks for the fix!

PR #14949, opened 2021-05-16
Core: Remove updateGlobals warning message

yannbf:
## What I did

Remove the log when updating globals with same values. I thought the log would be helpful, but it just ends up adding clutter. It could happen that globals are updated based on different scenarios, like HMR etc, and a fresh installed Storybook is logging that already, so I think we should not log it.

The fact that a fresh install already logs that might tell us that there's some code running multiple times where it shouldn't, though 🤔

## How to test


shilman (approved):
LGTM!

Issue #13578, opened 2021-01-08
Provide a new theme variable for toolbar icon hover color


thesobercoder:
**Is your feature request related to a problem? Please describe**
Currently, the `ThemeVars` interface accepts a `colorSecondary` parameter which is used to color the selected item on the left sidebar and the toolbar icons hover state. However, in my custom theme I need the toolbar icon hover color to be something diffrerent because I'm using the same color for `barBg` and `colorSecondary`. 

```
export interface ThemeVars { 
   colorSecondary?: string;
   ...
}
```

**Describe the solution you'd like**
Create a new property in `ThemeVars` interface to accept the hover color.

```
export interface ThemeVars { 
   barHover?: string;
   ...
}
```

**Describe alternatives you've considered**
For the time being, I'm using a different color `colorSecondary` as a workaround, but that is not making the theme look nice and what I exactly want.

**Are you able to assist to bring the feature to reality?**
Yes

Tomastomaslol:
Hi @thesobercoder! 

Does it make sense to you to also introduce a variable to control the `barIconActiveColor`? Is this implementation suitable for your use case? 🙂  

**Suggested implementation:**

```js
const EXAMPLE_THEME = create({
  base: 'light',
  barIconHoverColor: 'red',
  barIconActiveColor: 'green',
});

```

**Would result in:**

<img width="283" alt="Screen Shot 2021-05-18 at 1 22 22 pm" src="https://user-images.githubusercontent.com/1872246/118585472-4e57d800-b7dc-11eb-89d7-3337d159442f.png">

thesobercoder:
@Tomastomaslol Yeah this should work.

Tomastomaslol:
Great! I just updated the documentation and requested a PR review. Hopefully, this should be available for you fairly soon! 🙂

ndelangen:
I don't think we should add specific hover colors for specific elements/components to our ThemeVars.

I respect you want to theme storybook and customize it to suit your wishes, but every bit of customization we allow adds complexity for us. In this case, I don't think the costs outweighs the benefits.

PR #14971, opened 2021-05-18
UI: introduce theme colors for bar icon buttons

Tomastomaslol:
Issue: #13578

## What I did
* Added new optional theme variables `barIconHoverColor` & `barIconActiveColor`
* Use new varabiles if defined for active and hover state for bar  `IconButton`

## How to test

**Example theme**

```js
create({
  base: 'light',
  barIconHoverColor: 'red',
  barIconActiveColor: 'green',
});

```

**Results in:**

Screen Shot 2021-05-18 at 1 22 22 pm


lib/theming/src/types.ts
```diff
@@ -27,6 +27,8 @@ export interface ThemeVars {
   barTextColor?: string;
   barSelectedColor?: string;
   barBg?: string;
+  barIconHoverColor?: string;
+  barIconActiveColor?: string;
```
ndelangen, line 31: Can you show me an example on why the icon should be different in color from the text?

Tomastomaslol: Personally I don't have use-case for this but I there was scenario raised in the issue where this would be useful: 

> Currently, the ThemeVars interface accepts a colorSecondary parameter which is used to color the selected item on the left sidebar and the toolbar icons hover state. However, in my custom theme I need the toolbar icon hover color to be something diffrerent because I'm using the same color for barBg and colorSecondary.

https://github.com/storybookjs/storybook/issues/13578 

It seemed like a reasonable feature to me but please let me know if I don't think it's worth the extra complexity and can close this PR.

ndelangen: Yeah I personally don't think adding this to our ThemeVars is worth the extra complexity indeed.
We keep the ThemeVars object small on purpose, we know this means theming is limited.

Tomastomaslol: No worries! Have a great day! 🤔

PR #14973, opened 2021-05-18
Controls: Style `undefined` range slider different to filled one

tmeasday:
Issue: #14972


shilman (changes requested):
The `core-server/prebuilt` files should not be checked in

tmeasday:
@shilman done

shilman (approved):
Awesome!

PR #14980, opened 2021-05-18
Build: Improve e2e script

gaetanmaisse:
## What I did

 - Improve CLI with prompts
 - Clean some duplicated code (use function from SB CLI directly)
 - Reduce log and improve e2e output with a sum-up of the passing/failing tests

## How to test

- Locally: ` yarn local-registry --port 6000 --open --publish` and `yarn test:e2e-framework --clean --use-local-sb-cli cra`
- On CI: the e2e steps should be way more understandable


scripts/run-e2e.ts
```diff
@@ -123,120 +114,202 @@ const runTests = async ({ name, ...rest }: Parameters) => {
...
+
+  return e2eResult;
 };
 
-perform().then(() => {
+perform().then((e2eResult) => {
+  logger.info(`🧮 E2E Results`);
+
+  Object.entries(e2eResult).forEach(([configName, result]) => {
+    logger.info(`${configName}: ${result ? 'OK' : 'KO'}`);
```
yannbf, line 311: This is nice! But KO reads very similar to OK, maybe another word would be better?

yannbf (approved):

ndelangen (approved):

PR #14999, opened 2021-05-21
Update code snippets to use `argTypes[key].options` instead of `control.options`

ghengeveld:
Issue: -

## What I did

Checked all code snippets for references to `control.options` and hoisted them all. Also removed instances of `controls: { type: 'select' }` since that is the default when `options` is specified.

## How to test


shilman (commented):
LGTM!

PR #15006, opened 2021-05-21
Storyshots-puppeteer: Remove the usage of GPL-licensed `@wordpress/jest-puppeteer-axe` package

Hypnosphi:
It turned out that `@wordpress/jest-puppeteer-axe` is GPL-licensed which means that it can't be used in projects with other licenses.

This also fixes `yarn test-puppeteer` command, so I think it makes sense to bring back the related CI job


shilman (approved):
Thanks @Hypnosphi !!! Great catch 💯

PR #15029, opened 2021-05-25
Add info about Lit in migration guide

gaetanmaisse:
## What I did

Related to https://github.com/storybookjs/storybook/issues/13489

I added a small section about Lit support in Storybook 6.3 as some users may face issues with wrong `lit-html`/`lit-element` versions. SB 6.3 require `lit-html` 1.4.x or `lit-element` 2.5.x and according to the package manager used in their projects users may need to bump this dep manually (in the vast majority of cases it should/will be handled automatically)


shilman (commented):
LGTM!

PR #15043, opened 2021-05-25
Do not try to named import from global

IanVS:
Issue: #14913 

## What I did

Instead of trying to import named exports from the `global` package which don't exist, this imports a default `global` from the package, and then destructures off the pieces that are desired.  This prevents the error experienced in the referenced issue.

I got agreement from @ndelangen for this approach in [discord](https://discord.com/channels/486522875931656193/501692020226654208/846779578332872704) prior to tackling this one.

## How to test

I had a hard time linking my changes into my reproduction.  It seems that snowpack does not play well with the yarn symlinks that are created via `npx sb@next link --local`.  I did, however, copy the built `dist` folder from `client-logger` out of my changed repo and into my reproduction, and the error changed from one usage to another (there are many, and it's hard to copy them all over).  So I'm pretty certain this change does fix the issue.

- Is this testable with Jest or Chromatic screenshots? I don't think so

### DIFF
--- a/addons/links/src/preview.test.ts
+++ b/addons/links/src/preview.test.ts
@@ -1,8 +1,10 @@
 import addons from '@storybook/addons';
 import { SELECT_STORY } from '@storybook/core-events';
 
-import { __STORYBOOK_STORY_STORE__ } from 'global';
+import globalPkg from 'global';
 import { linkTo, hrefTo } from './preview';
+
+const { __STORYBOOK_STORY_STORE__ } = globalPkg;
 
 jest.mock('@storybook/addons');
 jest.mock('global', () => ({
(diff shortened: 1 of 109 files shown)


addons/links/src/preview.test.ts
```diff
@@ -1,9 +1,11 @@
 import addons from '@storybook/addons';
 import { SELECT_STORY } from '@storybook/core-events';
 
-import { __STORYBOOK_STORY_STORE__ } from 'global';
+import globalPkg from 'global';
```
IanVS, line 4: I had to name this one slightly differently due to jest wanting to mock the actual `global` below.

ndelangen (approved):

PR #15049, opened 2021-05-26
Server: Add support for stories written in YAML

titouanmathis:
## What I did

- Add support for stories written in YAML to the server mode

## Context

As writing JSON by hand can be tedious, I added support for stories written in YAML for the server app.


jonspalmer (commented):
This is an awesome addition. Two requests: 
1. Let's add a 'test' of the YML. We have `json-to-csf-compiler.test.ts`. It's barely testing the JSON part but let's add a smaller YAML version so there is some yml examples in the addon vs the example.
2. Can we create a new example for yml in the kitchen sink vs repurposing the ally example. I'd like to keep the addon examples consistent to avoid confusion.

titouanmathis:
Hi @jonspalmer, I pushed some commits addressing your requests, let me know if this seems good for you!

app/server/src/lib/compiler/json-to-csf-compiler.test.ts
```diff
@@ -1,13 +1,15 @@
 import 'jest-specific-snapshot';
 import path from 'path';
 import fs from 'fs-extra';
+import YAML from 'yaml';
 import { compileCsfModule } from '.';
 
-const inputRegExp = /\.json$/;
+const inputRegExp = /\.(json|ya?ml)$/;
```
jonspalmer, line 7: Rather than this can we organize it as two `describe` blocks so something like:

```
['json', 'yaml'].each (fileType) -> {
  const inputRegExp = new RegExp(`\.(${fileType}\$`);

  describe(`${fileType}-to-csf-compiler`, () => {
    .....
```

titouanmathis: It is fixed in d1be58d 🙂

jonspalmer (approved):
Looks awesome. Thank you for the contribution

shilman (approved):
Oooh man I'm a sucker for YAML ❤️

PR #15055, opened 2021-05-26
Core: Set `loose: true` in babel/preset-env config

IanVS:
Issue: #14805

## What I did

The warnings being thrown seem to be coming from storybook's default babel configs mixing `loose: true` in some of the babel plugins but not on `@babel/preset-env`, which defaults to `loose: false`.  Those settings all technically need to match up, to avoid the warnings.  Changing loose to `false` would be a possibly-breaking change, but setting `loose: true` on the preset-env silences the warnings while keeping the actual current behavior.  (The warnings were saying basically, "We can't use false like you want").  

## How to test

- Is this testable with Jest or Chromatic screenshots? - No

I made these changes in the `node_modules/@storybook` files of my project (which has no custom babel config), and verified that the warnings are no longer shown.

### DIFF
--- a/lib/core-common/src/utils/babel.ts
+++ b/lib/core-common/src/utils/babel.ts
@@ -43,7 +43,7 @@ const plugins = [
 ];
 
 const presets = [
-  [require.resolve('@babel/preset-env'), { shippedProposals: true }],
+  [require.resolve('@babel/preset-env'), { shippedProposals: true, loose: true }],
   require.resolve('@babel/preset-typescript'),
 ];
 
--- a/lib/core-common/src/utils/es6Transpiler.ts
+++ b/lib/core-common/src/utils/es6Transpiler.ts
@@ -34,6 +34,7 @@ export const es6Transpiler: () => RuleSetRule = () => {
               {
                 shippedProposals: true,
                 modules: false,
+                loose: true,
                 targets: 'defaults',
               },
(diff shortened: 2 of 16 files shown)


IanVS:
I ran yarn test locally and it passed.  I'll need to troubleshoot it looks like.

IanVS:
It seems like the CI jobs that are failing here are also failing on other PRs as well, so I don't think it's related to my changes here. 

With that, note that the output in the e2e CI (https://app.circleci.com/pipelines/github/storybookjs/storybook/18801/workflows/86961a94-b6c5-4b52-bf91-f7b7b0355a59/jobs/320757) does not contain the "loose" warnings as another recent example PR does: (https://app.circleci.com/pipelines/github/storybookjs/storybook/18800/workflows/a78f192d-6762-488f-96c8-814001f9d7b7/jobs/320741)

shilman (approved):
Working great @IanVS -- thanks so much for this 🙏 

I'd like a few more eyes on this before I merge. Discussion on loose mode here: https://2ality.com/2015/12/babel6-loose-mode.html

cc @ndelangen @tmeasday @gaetanmaisse

gaetanmaisse (approved):
LGTM. Failing e2e tests are related to https://github.com/yarnpkg/berry/issues/2938

PR #15059, opened 2021-05-27
Controls: Initialize color control on reset

88ta:
Issue: https://github.com/storybookjs/storybook/issues/14978

## What I did
Fixed a bug where the color control was not initialized when the reset button was pressed.

## How to test


shilman (commented):
Thank you @88ta! Mind reviewing @ghengeveld @tmeasday ?

lib/components/src/controls/Color.tsx
```diff
@@ -213,6 +221,13 @@ const useColorInput = (initialValue: string, onChange: (value: string) => string
   const [color, setColor] = useState(() => parseValue(value));
   const [colorSpace, setColorSpace] = useState(color?.colorSpace || ColorSpace.HEX);
 
+  useEffect(() => {
+    if (initialValue !== undefined) return;
+    const parsed = parseValue('');
+    setValue('');
+    setColor(parsed);
+  }, [initialValue]);
```
tmeasday, line 229: Should this call `updateValue`?

tmeasday (changes requested):
@ghengeveld what do you think about this?

I feel like the "correct" solution to this kind of problem is to lift the state up to the level of the reset button and make the state get cleared when the reset button is pressed. That's probably a bit tricky to implement however for controls.

How is this solved (or is it not solved) for the JSON editor?

lib/components/src/controls/Color.tsx
amartinez62: @ghengeveld

ghengeveld (approved):
I made some adjustments and I think this is good to go now.

shilman (approved):
LGTM! 🙏

ghengeveld:
Build is failing in IE11 due to Acorn not being transpiled. This was fixed before afaik. Merging in `next` which will hopefully fix that.

PR #15066, opened 2021-05-27
react-dev-utils dependencies vulnerabilities

RomanZabila:
react-dev-utils has browserslist in dependencies which cause vulnerabilities:
https://www.npmjs.com/advisories/1747

Issue:

## How to test

npm run audit

### DIFF
--- a/app/react/package.json
+++ b/app/react/package.json
@@ -58,7 +58,7 @@
     "global": "^4.4.0",
     "lodash": "^4.17.20",
     "prop-types": "^15.7.2",
-    "react-dev-utils": "^11.0.3",
+    "react-dev-utils": "^11.0.4",
     "react-docgen-typescript-plugin": "^0.6.2",
     "react-refresh": "^0.8.3",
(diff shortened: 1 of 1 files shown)


app/react/package.json
```diff
@@ -58,7 +58,7 @@
     "global": "^4.4.0",
     "lodash": "^4.17.20",
     "prop-types": "^15.7.2",
-    "react-dev-utils": "^11.0.3",
+    "react-dev-utils": "^11.0.4",
```
sgerrand, line 61: 💭 The `react-dev-utils` package doesn't have a version available yet which has a "safe" version of `browserslist` available. Does this change help?

sgerrand (commented):
💭 I'm unsure how this change will help resolve the stated transitive vulnerability, as the `react-dev-utils` package doesn't have a version available yet which has a "safe" version of `browserslist` available.

app/react/package.json
RomanZabila: yes, you are right, I'm closing this PR.

PR #15083, opened 2021-05-28
Update markdown-to-jsx version to fix url links in comments

TheMightyPenguin:
Issue: https://github.com/storybookjs/storybook/issues/14176

## What I did

There was an issue with `markdown-to-jsx`, in that it was generating a link twice, when `compodocs` already generated an anchor tag for a given URL string. So when `markdown-to-jsx` would see this:

```html
<a href="https://example.com">
  https://example.com
</a>
```

It would then generate the following:

```html
<a href="https://example.com">
  <a href="https://example.com">
    https://example.com
  </a>
</a>
```

As it adds an anchor tag automatically when it sees a URL link in text.

I issued a PR to `markdown-to-jsx` fixing the duplicated nested link issue (https://github.com/probablyup/markdown-to-jsx/pull/381), it was merged a few days ago. This PR updates the version of the library to apply the fix on this side.

## How to test

1. Add a JSDoc comment with links to the ChipsGroup component on the angular-cli example: https://github.com/storybookjs/storybook/blob/abc74c3f93d48734a7b759e2bec10d234fa81b7e/examples/angular-cli/src/stories/basics/ng-module/angular-src/chips-group.component.ts#L41

It could be something like:
```
  /**
   * This is a test
   * https://example.com
   */
```

2. Run the `angular-cli` example using `yarn storybook`, and go to the ChipsGroup docs page: http://localhost:9008/?path=/docs/basics-ngmodule-module-with-multiple-component--chips-group
3. Validate that the link is generated, and that upon inspecting the DOM elements, there's a single `a` element present.

### DIFF
--- a/lib/components/package.json
+++ b/lib/components/package.json
@@ -52,7 +52,7 @@
     "fast-deep-equal": "^3.1.3",
     "global": "^4.4.0",
     "lodash": "^4.17.20",
-    "markdown-to-jsx": "^7.1.0",
+    "markdown-to-jsx": "^7.1.3",
     "memoizerific": "^1.11.3",
     "overlayscrollbars": "^1.13.1",
     "polished": "^4.0.5",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6548,7 +6548,7 @@ __metadata:
     global: ^4.4.0
     jest: ^26.6.3
     lodash: ^4.17.20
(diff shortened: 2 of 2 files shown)


shilman (approved):
Thank you @TheMightyPenguin !!! 🙏  Great work 💪

PR #15091, opened 2021-05-29
Loosen puppeteer dependencies to allow last major version

vmalay:
Issue: Pupeteer version used by storyshots is very old. It is 6 major updates late. Following https://github.com/storybookjs/storybook/issues/15079

## What I did
I updated the puppeteer version to the last major version in the package.json for @storybook/storyshots-puppeteer. 

## How to test


shilman (commented):
Since puppeteer is a peer dependency, can we accept the range of compatible versions? This change forces any project that uses it to upgrade, but I think that's more disruptive than it needs to be. WDYT?

vmalay:
Yes you're right @shilman , I've updated the PR !

shilman (approved):
LGTM. Also upgraded the monorepo and tested with:

```
yarn storyshots-puppeteer
```

vmalay:
PR has been reverted. New PR: https://github.com/storybookjs/storybook/pull/15163

PR #15107, opened 2021-05-31
Essentials: Add measure and outline addons

winkerVSbecks:
Issue:

## What I did
Added the Measure and Outline addons to `@storybook/addon-essentials`

## How to test
Build `@storybook/addon-essentials` and load up any of the example Storybooks

No

No

Yes, should add Measure and Outline to the list of essential addons

### DIFF
--- a/addons/essentials/package.json
+++ b/addons/essentials/package.json
@@ -50,7 +50,7 @@
     "@storybook/node-logger": "6.3.0-beta.8",
     "core-js": "^3.8.2",
     "regenerator-runtime": "^0.13.7",
-    "storybook-addon-outline": "^1.2.0",
+    "storybook-addon-outline": "^1.3.0",
     "ts-dedent": "^2.0.0"
   },
   "devDependencies": {
--- a/addons/essentials/package.json
+++ b/addons/essentials/package.json
@@ -50,7 +50,7 @@
     "@storybook/node-logger": "6.3.0-beta.8",
     "core-js": "^3.8.2",
     "regenerator-runtime": "^0.13.7",
-    "storybook-addon-outline": "^1.3.0",
+    "storybook-addon-outline": "^1.3.1",
     "ts-dedent": "^2.0.0"
   },
   "devDependencies": {
--- a/addons/essentials/package.json
+++ b/addons/essentials/package.json
@@ -43,6 +43,7 @@
     "@storybook/addon-backgrounds": "6.3.0-beta.8",
     "@storybook/addon-controls": "6.3.0-beta.8",
     "@storybook/addon-docs": "6.3.0-beta.8",
+    "@storybook/addon-measure": "^1.1.0",
     "@storybook/addon-toolbars": "6.3.0-beta.8",
     "@storybook/addon-viewport": "6.3.0-beta.8",
     "@storybook/addons": "6.3.0-beta.8",
@@ -50,7 +51,7 @@
     "@storybook/node-logger": "6.3.0-beta.8",
     "core-js": "^3.8.2",
     "regenerator-runtime": "^0.13.7",
-    "storybook-addon-outline": "^1.3.1",
+    "storybook-addon-outline": "^1.3.2",
     "ts-dedent": "^2.0.0"
   },
   "devDependencies": {
--- a/addons/essentials/package.json
+++ b/addons/essentials/package.json
@@ -50,6 +50,7 @@
     "@storybook/node-logger": "6.3.0-beta.8",
     "core-js": "^3.8.2",
     "regenerator-runtime": "^0.13.7",
+    "storybook-addon-outline": "^1.2.0",
     "ts-dedent": "^2.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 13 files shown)


yarn.lock
```diff
@@ -5798,7 +5798,50 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@storybook/addon-essentials@6.3.0-beta.5, @storybook/addon-essentials@workspace:*, @storybook/addon-essentials@workspace:addons/essentials":
+"@storybook/addon-essentials@portal:/Users/varunvachhar/Documents/chromatic/storybook/addons/essentials::locator=%40storybook%2Froot%40workspace%3A.":
```
yannbf, line 5801: Did you link something local to the monorepo while testing?

winkerVSbecks: I did, and I thought I removed it. Good catch.

yannbf:
This is super awesome! The only concern I have, which I don't really know if it's a valid one, is that now with these addons, out of the box, Storybook already contains 5 toolbar icons. As soon as people start adding a couple more addons, it gets really crowded. Example of a crowded toolbar (3 from essentials, 3 custom):

image

I'm unsure whether having too many items there becomes a challenging UI for the user. I guess it is, when people access via mobile. What are your thoughts?

winkerVSbecks:
@yannbf there is always the user always has the option to disable essential addons that they are not using. But, I think in the long term we need a tweak to the toolbar to support more addons.

addons/essentials/package.json
```diff
@@ -43,13 +43,15 @@
     "@storybook/addon-backgrounds": "6.3.0-beta.5",
     "@storybook/addon-controls": "6.3.0-beta.5",
     "@storybook/addon-docs": "6.3.0-beta.5",
+    "@storybook/addon-measure": "latest",
```
shilman, line 46: This should be a `^x.x.x` version so you can release breaking changes to the addon without breaking `addon-essentials` users

yarn.lock
```diff
@@ -6115,6 +6137,26 @@ __metadata:
   languageName: unknown
   linkType: soft
 
+"@storybook/addons@npm:^6.0.0":
```
shilman, line 6140: Hmm we probably don't want this in here...

shilman (changes requested):
A couple minor issues

winkerVSbecks:
@shilman updated, could you take another look please

shilman (changes requested):
@winkerVSbecks There are a couple issues here.
- the PnP tests are failing because `addon-outline` uses `ts-dedent` but doesn't declare it as a dependency in `package.json`. Adding it & republishing the addon should fix it.
- Chromatic is failing on some measure-related error

Home stretch!

shilman (approved):
Looking good! I checked the perf and there is a slight uptick in install size & manager bundle size, which is to be expected. Merging & will release later today.

IanVS:
It might be good to update the docs (or at least readme) too, it seems there are a few more essentials than are indicated on https://storybook.js.org/addons/tag/essentials/ and https://github.com/storybookjs/storybook/tree/next/addons/essentials#contents.

shilman:
Thanks @IanVS. We'll be sure to update everything as part of the 6.3 release cc @winkerVSbecks @jonniebigodes

hckhanh:
Why this cool feature is disabled in `rc.0` version?

shilman:
@hckhanh it's enabled in `rc.0`. screenshot from a fresh install:

Example___Button_-_Primary_⋅_Storybook

hckhanh:
Hi @shilman, I don't know why but when I upgrade from `beta.11` to `rc.0`. The Measure and Outline tools are disappeared. I leave my configurations bellow:

image

## main.js

```js
module.exports = {
  stories: ["../src/**/*.stories.mdx", "../src/**/*.stories.@(js|jsx|ts|tsx)"],
  addons: [
    "@storybook/addon-links",
    "@storybook/addon-essentials",
    "@storybook/preset-create-react-app",
  ],
  core: {
    // builder: 'webpack5',
  },
};
```

## manager.js

```js
import { addons } from "@storybook/addons";
import { themes } from "@storybook/theming";

addons.setConfig({
  theme: themes.dark,
});
```

## preview.js

```js asasdas
import { themes as storybookThemes } from "@storybook/theming";
import { ThemeProvider } from "styled-components";
import themes, { GlobalStyles } from "../src/themes";

export const parameters = {
  actions: { argTypesRegex: "^on[A-Z].*" },
  controls: {
    matchers: {
      color: /(background|color)$/i,
      date: /Date$/,
    },
  },
  docs: {
    theme: storybookThemes.dark,
  },
};

export const decorators = [
  (Story) => (
    <ThemeProvider theme={themes}>
      <GlobalStyles />
      <Story />
    </ThemeProvider>
  ),
];
```

shilman:
@hckhanh Did you try removing `node_modules` and/or lockfiles and reinstalling?

hckhanh:
Óp I am using Yarn 2 with plug'nplay

PR #15111, opened 2021-05-31
Components: Accordion and AccordionList components

IgorSzyporyn:
## What I did

Created the new components ```<Accordion />```, ```<AccordionItem />```, ```<AccordionHeader />```, ```<AccordionBody />``` & ```<AccordionList />```

The aim came from a wish to make the list feature used in a11y add-on commonly accessible with a uniform visual expression.

More documentation can be found in the ```official-storybook```

## How to test

Open the ```official-storybook``` example after you spin up with a ```yarn build components```

- Does this need a new example in the kitchen sink apps? Yes, I made stories and documents
- Does this need an update to the documentation? Do not think so


IgorSzyporyn:
About to add fixes per Deepscan

lib/components/src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+type StateChange = {
+  id: string;
+  index: number;
+};
+
+export type AccordionProps = {
+  allowMultipleOpen?: boolean;
+  defaultOpen?: number | number[];
+  bordered?: boolean;
```
yannbf, line 17: The border at the last element stands out in both versions:
image

image

Is that intentional?

lib/components/src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,200 @@
...
+import { AccordionContext } from './AccordionContext';
+
+import type { AddToMapFn, OpenMap } from './AccordionContext';
+
+type AccordionMap = Record<string, { index: number; id: string }>;
+
+type StateChange = {
+  id: string;
+  index: number;
+};
```
ndelangen, line 12:
```suggestion
interface StateChange {
  id: string;
  index: number;
};
```

lib/components/src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,200 @@
...
+} & React.HTMLAttributes<HTMLUListElement>;
+
+/**
+ *
+ * Controlling wrapper component that will control the open/close functionality to
+ * child ```<AccordionItem />```'s and style the other support components
+ * ```<AccordionHeader />``` and ```<AccordionBody />```
+ *
+ * Although each component can work independently and be controlled from the outside
+ * via an **open** prop, it is not recommended to use ```<AccordionItem />```,
```
ndelangen, line 31: I think it's a good idea to use propNames like `isOpen` rather then `open`.

When passing props into `styled` props like `open` can make their way into the actual html, which can be unexpected.

lib/components/src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,200 @@
...
+      }
+
+      if (oldOpen) {
+        onClose(oldOpen);
+      }
+
+      setOpen({ ...open, ...newOpen });
+      onOpen({ id, index: itemMap.current[id].index });
+    },
+    [open, setOpen, itemMap]
```
ndelangen, line 103: You use onOpen here, but it's not added to the useCallback dependency?

I suggest not having a default value for onOpen, but instead do something like:

```
onOpen && onOpen({ id, index: itemMap.current[id].index });
```

lib/components/src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,200 @@
...
+      }
+
+      &[data-sb-state-prevent-expander='true'] {
+        & > [data-sb-accordion-header] {
+          cursor: default;
+        }
+      }
+    }
+  `}
+`;
```
ndelangen, line 200: We use the object style notation:

```suggestion
const Wrapper = styled.ul<WrapperProps>(
  ({ theme }) => ({
    overflow: 'hidden',
    listStyle: 'none',
    margin: 0,
    padding: 0,
    backgroundColor: theme.background.content,
  }),
  ({ theme, bordered }) =>
    bordered
      ? {
          border: `1px solid ${theme.appBorderColor}`,
        }
      : {},
  ({ theme, rounded }) =>
    rounded
      ? {
          borderRadius: theme.appBorderRadius,
        }
      : {},
  ({ theme, lined }) =>
    lined
      ? {
          '& > [data-sb-accordion-item]': {
            borderBottom: `1px solid ${theme.appBorderColor}`,
            '&:last-child': {
              borderBottom: '0 none',
            },
            '&[aria-expanded="true"]': {
              '& > [data-sb-accordion-header]': {
                borderBottom: `1px solid ${theme.appBorderColor}`,
              },
            },
          },
        }
      : {},

  ({ theme }) => ({
    '& > [data-sb-accordion-item]': {
      '& > [data-sb-accordion-header]': {
        padding: 16,
        cursor: 'pointer',
        border: '1px solid transparent',
        '&:hover': {
          backgroundColor: theme.background.hoverable,
        },
      },
      '& > [data-sb-accordion-body]': {
        fontSize: 14,
      },
      '&[aria-expanded="true"]': {
        '& > [data-sb-accordion-body] > [data-sb-accordion-body-inner]': {
          padding: '16px 16px 16px 42px',
          backgroundColor: theme.background.app,
        },
      },
      '&[data-sb-state-prevent-expander="true"]': {
        '& > [data-sb-accordion-header]': {
          cursor: 'default',
        },
      },
    },
  })
);
```

I'm not a huge fan of the super specific selectors. can you explain why:
```
& [data-sb-accordion-body-inner]
```
Is not good enough, and it has to be:
```
& > [data-sb-accordion-item] > [data-sb-accordion-body] > [data-sb-accordion-body-inner]
```

Do we expect accordions within accordions?

IgorSzyporyn:
I went it over once again, and delegated some styling responsibilities and also exposed some used properties for scope control so you can set defaults for each item on Accordion level, but override per AccordionItem

The indentation of body content is now optional and can be set via the prop `indentBody`

The DOM and styling is better for the rotating icon as well

I have changed all types into interfaces where possible, and used object notation for styling

yannbf:
Hey @domyen I think you're going to enjoy checking this work!

Super awesome and intended to be used in the addons a11y and jest, in order to achieve better consistency. WDTY?

[Storybook build](https://5a375b97f4b14f0020b0cda3-yijaktpwiy.chromatic.com/?path=/story/basics-accordion--controllable)

IgorSzyporyn:
Use case for an a11y update proposal I will have when this is merged


https://user-images.githubusercontent.com/6172338/120811028-dca5bb00-c54b-11eb-812f-e6a99e90aab9.mov

lib/components/src/Accordion/AccordionContext.tsx
```diff
@@ -0,0 +1,19 @@
...
+export type OpenMap = Record<string, boolean>;
+
+interface AccordionContextProps {
+  openState: OpenMap;
+  addToMap: AddToMapFn;
+  onOpen: (id: string) => void;
+  onClose: (id: string) => void;
+  lined?: boolean | undefined;
+  bordered?: boolean | undefined;
+  indentBody?: boolean | undefined;
```
yannbf, line 14: isn't `indentBody?: boolean | undefined;` same as `indentBody?: boolean;`?

lib/components/src/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,191 @@
+import React, { useCallback, useContext, useEffect, useRef, useState, Children } from 'react';
+import { uniqueId } from 'lodash';
+import { styled } from '@storybook/theming';
+import { AccordionContext } from './AccordionContext';
+import { AccordionItemContext } from './AccordionItemContext';
+
+// Props are also available from Accordion context provider, but local props
+// takes predecense for scope control
```
yannbf, line 8:
```suggestion
// takes precedence for scope control
```

lib/components/src/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,191 @@
...
+  ...rest
+}: AccordionItemProps) => {
+  const [open, setOpen] = useState(_open);
+  const context = useContext(AccordionContext);
+  const id = useRef(uniqueId('AccordionItem-'));
+  const preventOpen = Children.count(children) < 2;
+  const initialOpen = useRef(_open);
+
+  const onExpand = useCallback(() => {
+    if (!preventOpen && _open !== true) {
```
yannbf, line 33: what about assigning this to a variable with a meaningful name? this is used in multiple places

lib/components/src/Accordion/stories/Accordion.stories.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import { Accordion } from '../Accordion';
+import { AccordionItem } from '../AccordionItem';
+import { AccordionHeader } from '../AccordionHeader';
+import { AccordionBody } from '../AccordionBody';
+import { Icons } from '../../icon/icon';
+
+// eslint-disable-next-line import/order
```
yannbf, line 8: What's the reason for this?

lib/components/src/Accordion/stories/Accordion.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+import type { Story, Meta } from '@storybook/react';
+import type { AccordionProps } from '../Accordion';
+
+export default {
+  title: 'Basics/Accordion',
+  component: Accordion,
+  argTypes: { onOpen: { action: 'open' }, onClose: { action: 'close' } },
+  parameters: {
+    layout: 'padded',
```
yannbf, line 17: This is the default value, you can remove it
```suggestion
```

lib/components/src/Accordion/stories/AccordionBody.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default {
+  title: 'Basics/Accordion/AccordionBody',
+  component: AccordionBody,
+  parameters: {
+    layout: 'padded',
+  },
+} as Meta;
+
+const Template: Story<AccordionBodyProps> = (args) => (
```
yannbf, line 16: There's a new type ComponentMeta which is quite convenient, you don't need to export the props anymore:

```suggestion
import type { ComponentStory, Meta } from '@storybook/react';

import { AccordionBody } from '../AccordionBody';

export default {
  title: 'Basics/Accordion/AccordionBody',
  component: AccordionBody,
} as Meta;

const Template: ComponentStory<typeof AccordionBody> = (args) => (
```

lib/components/src/Accordion/stories/AccordionHeader.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+export default {
+  title: 'Basics/Accordion/AccordionHeader',
+  component: AccordionHeader,
+  argTypes: { onOpen: { action: 'open' }, onClose: { action: 'close' } },
+  parameters: {
+    layout: 'padded',
+  },
+} as Meta;
+
+const Template: Story<AccordionHeaderProps> = (args) => (
```
yannbf, line 17:
```suggestion
import type { ComponentStory, Meta } from '@storybook/react';

import { AccordionHeader } from '../AccordionHeader';

export default {
  title: 'Basics/Accordion/AccordionHeader',
  component: AccordionHeader,
  argTypes: { onOpen: { action: 'open' }, onClose: { action: 'close' } },
} as Meta;

const Template: ComponentStory<typeof AccordionHeader> = (args) => (
```

lib/components/src/Accordion/stories/AccordionItem.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export default {
+  title: 'Basics/Accordion/AccordionItem',
+  component: AccordionItem,
+  parameters: {
+    layout: 'padded',
+  },
+} as Meta;
+
+const Template: Story<AccordionItemProps> = (args) => (
```
yannbf, line 18:
```suggestion
import React from 'react';
import type { Story, Meta } from '@storybook/react';

import { AccordionItem } from '../AccordionItem';
import { AccordionHeader } from '../AccordionHeader';
import { AccordionBody } from '../AccordionBody';

export default {
  title: 'Basics/Accordion/AccordionItem',
  component: AccordionItem,
} as Meta;

const Template: ComponentStory<typeof AccordionItem> = (args) => (
```

domyen:
Amazing contribution 🙏. 

 @winkerVSbecks @MichaelArestad can you take a look.

IgorSzyporyn:
I do not understand why it say that unit testing is failing

All tests are working as you can see below @shilman @ndelangen 

Screenshot 2021-06-07 at 10 51 06

lib/components/src/Accordion/Accordion.tsx
MichaelArestad: I agree. I might also change `narrow` to either `isCompact` or maybe even `size` with the values of `medium` and `small`. The latter is what is more common in the Storybook Design System.

MichaelArestad:
I love it.

I have only some minor suggestions:
* As we discussed briefly, I would recommend changing the text size to `13px` in both the `AccordionHeader` and `AccordionBody` just to be consistent with the rest of the Storybook interface.
* Though I dig the padding, I also would change the padding to be consistent with other elements in Storybook. I'm seeing `10p` used regularly as padding. Perhaps `10px 15px`?

Have you tried out adding some buttons or other controls to the right side of the `AccordionHeader`?

IgorSzyporyn:
> I love it.
> 
> I have only some minor suggestions:
> 
> * As we discussed briefly, I would recommend changing the text size to `13px` in both the `AccordionHeader` and `AccordionBody` just to be consistent with the rest of the Storybook interface.
> * Though I dig the padding, I also would change the padding to be consistent with other elements in Storybook. I'm seeing `10p` used regularly as padding. Perhaps `10px 15px`?
> 
> Have you tried out adding some buttons or other controls to the right side of the `AccordionHeader`?

Will make these changes

Font size is based on theme values - I have been trying to avoid hardcoding any CSS where possible to avoid it

Padding changes coming up also

MichaelArestad:
Awesome. Thank you, @IgorSzyporyn

ndelangen:
I do not understand why the test is exiting with non-0 either. @shilman any ideas?

shilman:
@ndelangen @gaetanmaisse I see this error in the logs:

```
ReferenceError: You are trying to access a property or method of the Jest environment after it has been torn down.

      at runWithJestRealTimers (node_modules/@testing-library/dom/dist/helpers.js:38:8)
      at jestFakeTimersAreEnabled (node_modules/@testing-library/dom/dist/helpers.js:57:12)
      at MutationObserverImpl.checkRealTimersCallback [as _callback] (node_modules/@testing-library/dom/dist/wait-for.js:114:49)
```

So perhaps the assertions are passing but there's some teardown problem in one of the tests?

IgorSzyporyn:
> Awesome. Thank you, @IgorSzyporyn

I have been thinking

I do not believe that changing the font and padding to hardcoded values for theme consistency is a good or proper way.

Should it not be the theme values that changes then - the values provided from theme should be the ones providing theme consistency IMO

IgorSzyporyn:
> @ndelangen @gaetanmaisse I see this error in the logs:
> 
> ```
> ReferenceError: You are trying to access a property or method of the Jest environment after it has been torn down.
> 
>       at runWithJestRealTimers (node_modules/@testing-library/dom/dist/helpers.js:38:8)
>       at jestFakeTimersAreEnabled (node_modules/@testing-library/dom/dist/helpers.js:57:12)
>       at MutationObserverImpl.checkRealTimersCallback [as _callback] (node_modules/@testing-library/dom/dist/wait-for.js:114:49)
> ```
> 
> So perhaps the assertions are passing but there's some teardown problem in one of the tests?

Will look at that - thanks

lib/components/src/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,205 @@
+import React, { useCallback, useContext, useEffect, useRef, useState, Children } from 'react';
+// @TODO get rid of lodash usage
+// shilman wish to get rid of lodash for bundle size, so maybe "nanoid" can be
+// added for unique id requirements?
+import { uniqueId } from 'lodash';
```
shilman, line 5: If you're going to use lodash, I think you need to import like:

```
import uniqueId from 'lodash/uniqueId';
```

shilman (changes requested):

domyen:
**font-size and layout**
Agree with @MichaelArestad, we should set it in the font-size and layout for how we expect the UI to render in Storybook. For example, the existing Placeholder.tsx component uses the theme to set the font-size like [so](https://github.com/storybookjs/storybook/blob/next/lib/components/src/placeholder/placeholder.tsx#L14).

That reduces the work for contributors and maintainers who want to consume the component.

IgorSzyporyn:
> **font-size and layout**
> Agree with @MichaelArestad, we should set it in the font-size and layout for how we expect the UI to render in Storybook. For example, the existing Placeholder.tsx component uses the theme to set the font-size like [so](https://github.com/storybookjs/storybook/blob/next/lib/components/src/placeholder/placeholder.tsx#L14).
> 
> That reduces the work for contributors and maintainers who want to consume the component.
> 
> **stories**
> There are more variations of the accordion component than we intend to include in the design. For example, no-border, header-only, custom icon. We typically don't include these other variations in the official storybook unless they're actually used in the app. Can we reduce the scope?

I was not aware of an accordion component was in the works - so I just made one with enough properties to cover most of the use cases I could quickly imagine on the top of my head.

No border etc... are there because there are actual use cases for it pending (I have one myself, and @shilman has one as well) - for instance if we do not allow to not have a border, then the component will not be able to be used in an absolute setting without either overriding the styles or having double border

Yes I can, and will, do `theme.typography.size.s2 - 1` - but for me then a theme variable has no meaning the second you mutate it's values, and we should still look at why the official theme font size used is a mutated theme value

That to me is more confusing than anything else - it is very little apparent that you are to subtract 1 pixel

IgorSzyporyn:
All requested changes should be in now

@shilman @domyen @MichaelArestad @yannbf

MichaelArestad:
Moved to https://github.com/storybookjs/storybook/pull/15412

IgorSzyporyn:
> Moved to #15412

Forgot that - thanks man

MichaelArestad:
Teamwork 🙌

PR #15121, opened 2021-06-02
Core: Fix prebuilt manager usage on first run

ghengeveld:
Issue: #15115

## What I did

When the prebuilt manager can be used, ensure that we actually do so and not continue compiling the manager anyway.
I've also made the webpack version part of the cache key for manager caching, so you won't end up using a cached webpack4 manager when switching to webpack5.

## How to test

- Does this need a new example in the kitchen sink apps? We should have one that uses the prebuilt manager


shilman (approved):
Looking good!! 💯

PR #15146, opened 2021-06-05
Build: Fix WC example and e2e tests

gaetanmaisse:
## What I did

 - WC Kitchen Sink: quick hack to make the CI happy. To remove as soon as we are using `workspace:*` in our deps instead of pin versions. For details see: https://yarnpkg.com/features/workspaces#publishing-workspaces. @shilman and I will do a pair programming session on Monday to move the release from Lerna to Yarn and use `workspace:*` in the monorepo
 -  CLI: Use a curated list of templates for interaction with users. All the templates are still available when defining the `--template` option directly but are not displayed in CLI prompt.

## How to test

- CI should be 🟢


shilman (approved):
LGTM!

PR #15160, opened 2021-06-07
Angular: Clear root / docs-root when navigating from one tab to the other

ThibaudAV:
Issue:

when I switch from canvas to docs the canvas story is not displayed
exactly the same problem :
https://discord.com/channels/486522875931656193/490770949910691862/850254520281595914
And the problem is that it continues to update the component / story in the root and not root-docs html element

## What I did

clear root / docs-root when navigating from one to the other 
otherwise angular will select the 1st component that matches the ng selector (which is the story id) and will not work for the 2nd

Usually angular's destroy PlatformBrowserDynamic does the cleaning. but in some applications it doesn't. this solution makes sure it's done in all cases 

## How to test


kroeder (approved):
LGTM

PR #15191, opened 2021-06-10
Update examples.md

yceballost:
Multibrand and multiplatform component library built for Telefónica company


jonniebigodes:
@yceballost thanks for the time and effort you put into this pull request. We're thrilled to see that Telefonica is using Storybook in your component library and we're more than glad to add it to our examples.  

Self merging this @shilman

PR #15194, opened 2021-06-10
Angular: Use docsMode to set docs options

ThibaudAV:
Issue: 

## What I did

use docsMode to set docs options

Setting the docs property to false seems to break things. probably used for other things.
docsMode seems to be the right option for the standalone run

## How to test


shilman (commented):
LGTM!

PR #15213, opened 2021-06-11
Core: Preserve other query params when changing args/globals

ghengeveld:
Issue: #15193

## What I did

When changing the `args` or `globals` query params, make sure we also keep other existing query params unchanged.

## How to test


shilman (changes requested):
this would be a good one to add tests for!

ghengeveld:
> this would be a good one to add tests for!

Done!

shilman (approved):
💥

PR #15231, opened 2021-06-14
Feature: Added toggle prop to toolbar addon for a button without menu to cycle through values

IgorSzyporyn:
## What I did
Added a new parameter to the toolbar addon config named "toggle"

This will invoke a button that can cycle (since it will accept any number of items and not just two to toggle between) between the item values.

## How to test
You can just go to the `examples/official-storybook` and spin it up - then edit `preview.js` and change the localization config object to include a `toggle: true` and it should just cycle between all the locale values and keybinds should still work.

If a default value is not chosen from start you will have to add an icon value for the button to show initially

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200539748475788/1200567474272563) by [Unito](https://www.unito.io)

### DIFF
--- a/addons/toolbars/src/components/ToolbarMenuToggle.tsx
+++ b/addons/toolbars/src/components/ToolbarMenuToggle.tsx
@@ -1,0 +1,45 @@
+import React, { useCallback, FC } from 'react';
+import { useGlobals } from '@storybook/api';
+import { ToolbarMenuButton } from './ToolbarMenuButton';
+import { withKeyboardCycle, WithKeyboardCycleProps } from '../hoc/withKeyboardCycle';
+import { getSelectedIcon } from '../utils/get-selected-icon';
+import { ToolbarMenuProps } from '../types';
+
+type ToolbarMenuToggleProps = ToolbarMenuProps & WithKeyboardCycleProps;
+
+export const ToolbarMenuToggle: FC<ToolbarMenuToggleProps> = withKeyboardCycle(
+  ({ id, name, description, cycleValues = [], toolbar: { title: _title, items, showName } }) => {
+    const [globals, updateGlobals] = useGlobals();
+
+    const currentValue = globals[id];
+    const hasGlobalValue = !!currentValue;
+    const icon = getSelectedIcon({ currentValue, items });
+
+    let title = _title;
+
+    // Deprecation support for old "name of global arg used as title"
+    if (showName && !title) {
+      title = name;
+    }
+
+    const setNext = useCallback(() => {
+      const currentIndex = cycleValues.indexOf(currentValue);
+      const currentIsLast = currentIndex === cycleValues.length - 1;
+
+      const newCurrentIndex = currentIsLast ? 0 : currentIndex + 1;
+      const newCurrent = cycleValues[newCurrentIndex];
+
+      updateGlobals({ [id]: newCurrent });
+    }, [currentValue, updateGlobals]);
+
+    return (
+      <ToolbarMenuButton
+        active={hasGlobalValue}
+        description={description}
+        onClick={setNext}
+        icon={icon}
+        title={title}
+      />
+    );
+  }
+);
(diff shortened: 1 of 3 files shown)


MichaelArestad:
@IgorSzyporyn Can you add some screenshots or a gif to this? I'm having a hard time testing.

addons/toolbars/src/types.ts
```diff
@@ -31,6 +31,7 @@ export interface NormalizedToolbarConfig {
   preventDynamicIcon?: boolean;
   items: ToolbarItem[];
   shortcuts?: ToolbarShortcuts;
+  toggle?: boolean;
```
darleendenno, line 34: Would a more descriptive prop be `cycle`? `isCycling` feels more accurate than `isToggle` to me. What do you think?

addons/toolbars/src/components/ToolbarMenuToggle.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+      updateGlobals({ [id]: newCurrent });
+    }, [currentValue, updateGlobals]);
+
+    return (
+      <ToolbarMenuButton
+        active={hasGlobalValue}
+        description={description}
+        onClick={setNext}
+        icon={icon}
```
darleendenno, line 40: This assumes a default value is set, which is not always the case. I would like to see if there's a way to use the default toolbar icon, when no default value is present, rather than not displaying the button at all. Does this seem possible?

darleendenno (commented):
Hey @IgorSzyporyn 👋 I've been testing this out and it's looking great! I can see the value of being able to configure the toolbar to quickly cycle through values rather than always navigating to the dropdown menu.

cycle

My concern regards your note about requiring defaultValues to display an icon. I think we should default to the toolbar icon if no defaultValue is present. Does that seem possible?

addons/toolbars/src/types.ts
IgorSzyporyn: > Would a more descriptive prop be `cycle`? `isCycling` feels more accurate than `isToggle` to me. What do you think?

It was cycle, then changed to toggle - I'll let others decide, I was worried about "cycle" not be too apparent in what it does, where "toggle" is easier to grasp what it does - but is technically not right

addons/toolbars/src/components/ToolbarMenuToggle.tsx
IgorSzyporyn: Offset is - no icon, no button (well button, but 0px wide)

I do think it is not an unfair agreement to make with the users, that if you do not provide something to show, then we show nothing.

I can put in a default icon for when there is nothing though

IgorSzyporyn: Misunderstood in previous comment - yes will look at that

MichaelArestad:
This is an interesting design pattern. It's pretty cool, but I'm not sure I would expect this behavior as a user. It also raises some accessibility concerns for me. In a list longer than 2 items, a user would need to click more times to get to their desired option and, for some folks, clicking can be difficult or painful.

Because this interaction seems optimized for advanced users, I would suggest you making this into an addon. What do you think?

IgorSzyporyn:
> This is an interesting design pattern. It's pretty cool, but I'm not sure I would expect this behavior as a user. It also raises some accessibility concerns for me. In a list longer than 2 items, a user would need to click more times to get to their desired option and, for some folks, clicking can be difficult or painful.
> 
> Because this interaction seems optimized for advanced users, I would suggest you making this into an addon. What do you think?

Then you dont use the toggle option I would say but use the dropdown instead - or just use the keybinds which will toggle through

It is an option, not a requirement - the other way to go is to code the possibility of endless cycling away - which is weird to me.

It's a toggle that can (if you want) cycle

MichaelArestad:
> Then you dont use the toggle option I would say but use the dropdown instead - or just use the keybinds which will toggle through

Right. I like the idea of a toggle, but in order to avoid unexpected interactions it should always be a boolean toggle treated like an on/off switch. Could you adjust this to restrict toggle to only off/on rather than cycling through longer lists?

PR #15256, opened 2021-06-16
Feature: new ScrollArea component and removed react-sizeme + overlayscrollbars dependencies

IgorSzyporyn:
## What I did

Created a new ScrollArea component (for now named ScrollBar till takeover is approved and a task for this is made) that will get rid of the dependency [overlayscrollbars](https://www.npmjs.com/package/overlayscrollbars).

It differs in that it

- only adds 2 DOM elements per scroller
- can have customized size
- used observable API for dynamic rescaling
- will fully remove and add a scrollers DOM upon requirements
- can have customized padding from edge
- can have customized color - both hex/rgba/hsl/etc... and theme colors (primary, secondary,  etc..)
- can have customized opacity
- can have scrollers placed on any of the 4 sides

## How to test

- Spin up `examples/official-storybook` and kick it under Basic/ScrollBar


lib/components/scripts/writeCssScript.js
```diff
@@ -8,7 +8,7 @@ const fs = require('fs');
 const { parse } = require('css');
 const { isNaN } = require('global');
 
-const INPUT = require.resolve('overlayscrollbars/css/OverlayScrollbars.min.css');
```
IgorSzyporyn, line 11: This is a part of the whole nx thing, but I can not find anywhere where this file is referenced other than inside node_modules - which is weird....

IgorSzyporyn:
Closing this PR as I am making the right telescoped PR's

PR #15279, opened 2021-06-17
Angular: Fix some Ivy rendering glitches

ThibaudAV:
Issue: @yngvebn https://discord.com/channels/486522875931656193/490770949910691862/854687468321767444

## What I did
- fix(angular): call AbstractRenderer.resetCompiledComponents after full render
to avoid having `ngIf/ngFor` ivy errors

- fix(angular): destroy component when doc re render
to correctly destroy the components after an HMR

## How to test

- Is this testable with Jest or Chromatic screenshots? 🤷‍♂️

### DIFF
--- a/app/angular/src/client/preview/angular-beta/DocsRenderer.ts
+++ b/app/angular/src/client/preview/angular-beta/DocsRenderer.ts
@@ -27,9 +27,11 @@ export class DocsRenderer extends AbstractRenderer {
     });
 
     await super.render({ ...options, forced: false });
-
-    await AbstractRenderer.resetCompiledComponents();
   }
 
   async beforeFullRender(): Promise<void> {}
+
+  async afterFullRender(): Promise<void> {
+    await AbstractRenderer.resetCompiledComponents();
+  }
 }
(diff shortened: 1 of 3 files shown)


app/angular/src/client/preview/angular-beta/DocsRenderer.ts
```diff
@@ -26,6 +26,15 @@ export class DocsRenderer extends AbstractRenderer {
       await DocsRenderer.resetPlatformBrowserDynamic();
     });
 
+    /**
+     * Destroy and recreate the PlatformBrowserDynamic of angular
+     * when doc re render. Allows to call ngOnDestroy of angular
+     * for previous component
+     */
+    channel.once(Events.DOCS_RENDERED, async () => {
```
kroeder, line 34: Does `DOCS_RENDERED` mean "after docs have rendered"? 
Why do we need to reset everything after the rendering? Or is the Event name just misleading?

ThibaudAV: when we are on the page of docs and that we make a change in the html for example the doc is rendered again. and it is thus necessary to destroy the previous component 

tested with `/story/basics-component-with-ngondestroy--simple-component` 

we see clearly the problem after several HMR in console

shilman (approved):
LGTM, i guess? 🤷

kroeder (approved):

yngvebn:
A bit late to the party, but I can confirm that the errors are gone in my repro. 🎉

PR #15292, opened 2021-06-18
CLI: Update types in React typescript templates

yannbf:
Issue: N/A

## What I did

1 - Update types to use `ComponentStory` and `ComponentMeta`

2 - Remove `React.FC` as the components don't receive children so they shouldn't be FC

3 - Fix a small thing in the CLI for e2e where users were able to press enter without selecting a framework, and the script would do nothing

## How to test

- Does this need an update to the documentation? Not this, but it's good to think about updating the React typescript docs if we want more people to use this type.


shilman (approved):
LGTM! 💯

PR #15304, opened 2021-06-20
feature: resize-observer dependency added and useDOMRect hook created

IgorSzyporyn:
**Telescopes INTO #15308**

## What I did

Added a new dependency [resize-observer](https://www.npmjs.com/package/resize-observer) which is to take over our usage of [react-sizeme](https://www.npmjs.com/package/react-sizeme) in a ready telescoped PR that will also make the dependency [overlayscrollbars](https://www.npmjs.com/package/overlayscrollbars) obsolete

Created 2 hooks
- useCallbackRef - a hook to attach refs to your other hooks
- useDOMRect - a hook that describes the size and position of a ref via [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect)


IgorSzyporyn:
@shilman I had to manually revert some changes in package.json that confuses me - this PR wanted to overwrite rc11 with rc5 and I also had to manually revert value of gitHead

lib/addons/src/hooks/useDOMRect.ts
```diff
@@ -0,0 +1,61 @@
+import { useState, useEffect, RefObject } from 'react';
+import ResizeObserver from 'resize-observer-polyfill';
+import { useCallbackRef } from './useCallbackRef';
+
+const isStateDirty = (a: UseDOMRectBounds, b: UseDOMRectBounds) => {
```
IgorSzyporyn, line 5: Would like some help to get this function memoized

lib/addons/src/hooks/useDOMRect.ts
```diff
@@ -0,0 +1,61 @@
...
+  width: 0,
+  left: 0,
+  top: 0,
+  bottom: 0,
+  right: 0,
+  x: 0,
+  y: 0,
+};
+
+export const useDOMRect = <T extends HTMLElement = HTMLDivElement>(ref: RefObject<T>) => {
```
ndelangen, line 31: Please add a description of what this hook is useful for, it's not immediately obvious

lib/addons/src/hooks/useDOMRect.ts
```diff
@@ -0,0 +1,61 @@
+import { useState, useEffect, RefObject } from 'react';
+import ResizeObserver from 'resize-observer-polyfill';
+import { useCallbackRef } from './useCallbackRef';
+
+const isStateDirty = (a: UseDOMRectBounds, b: UseDOMRectBounds) => {
+  return JSON.stringify(a) !== JSON.stringify(b);
+};
+
+export interface UseDOMRectBounds {
```
ndelangen, line 9: These's existing interfaces for this, I'm sure

lib/addons/src/hooks/useDOMRect.ts
```diff
@@ -0,0 +1,61 @@
+import { useState, useEffect, RefObject } from 'react';
+import ResizeObserver from 'resize-observer-polyfill';
```
ndelangen, line 2: Is the polyfill ALWAYS included?

IgorSzyporyn: Do not understand, as in is it always needed?

This is what caniuse.com say - if we can live with this then I will remove the polyfill

Screenshot 2021-06-21 174221

lib/addons/src/hooks/useDOMRect.ts
IgorSzyporyn: Will do

lib/addons/src/hooks/useDOMRect.ts
```diff
@@ -0,0 +1,61 @@
...
+  y: 0,
+};
+
+export const useDOMRect = <T extends HTMLElement = HTMLDivElement>(ref: RefObject<T>) => {
+  const { ref: element, fn: attachRef } = useCallbackRef<T>();
+  const [bounds, setBounds] = useState<UseDOMRectBounds>({ ...initialState });
+
+  useEffect(() => {
+    const onResize = ([entry]: ResizeObserverEntry[]) => {
+      const newBounds = entry.target.getBoundingClientRect().toJSON();
```
ndelangen, line 37:
```suggestion
      const newBounds = entry.target.getBoundingClientRect();
```

lib/addons/src/hooks/useCallbackRef.ts
```diff
@@ -0,0 +1,14 @@
...
+  const [ref, setRef] = useState<RefObject<T | null>>(null);
+  const fn = useCallback(
+    (node: RefObject<T | null>) => {
+      setRef(node);
+    },
+    [setRef]
+  );
+
+  return { ref, fn };
+}
```
ndelangen, line 14: Can you explain to me what the point of this is?

lib/addons/src/hooks/useDOMRect.ts
```diff
@@ -0,0 +1,61 @@
...
+  left: 0,
+  top: 0,
+  bottom: 0,
+  right: 0,
+  x: 0,
+  y: 0,
+};
+
+export const useDOMRect = <T extends HTMLElement = HTMLDivElement>(ref: RefObject<T>) => {
+  const { ref: element, fn: attachRef } = useCallbackRef<T>();
```
ndelangen, line 32: I think you're looking to just use `useRef` ?

lib/addons/src/hooks/useDOMRect.ts
IgorSzyporyn: done

lib/addons/src/hooks/useCallbackRef.ts
IgorSzyporyn: Not exactly as in "is this absolutely necessary" - it seems it just makes it easier to have a cleaner output from the hook useDOMRect as the useEffect that uses the setBounds gets called one time less

This is as I remember it, this is code I copied some time ago from somewhere and tweaked :-D

If you can see we can avoid it and not have extra render due to useDOMRect having setBounds called once too many in the start then yes - lets get rid of it!

IgorSzyporyn: I will use `useLayoutEffect` and now this is not required anymore

IgorSzyporyn: Might have to be required, will have to start coding against a use case again

React has the type RefCallback made for this situation it seems, and other hooks I have been looking at that does the same also employes a callback for ref in one variant or another ([like here](https://github.com/ZeeCoder/use-resize-observer/blob/master/src/index.ts))

lib/addons/src/hooks/useDOMRect.ts
IgorSzyporyn: This might be a solution instead

https://www.npmjs.com/package/resize-observer

It also sports an optional install feature out of the box to overwrite window.ResizeObserver on demand

lib/addons/src/hooks/useCallbackRef.ts
IgorSzyporyn: > I will use `useLayoutEffect` and now this is not required anymore

won't work for delayed mounts - will need the callback

lib/addons/src/hooks/useDOMRect.ts
IgorSzyporyn: Been reading up and I do need it for delayed mounts

I can however create the ref in the hook and return it to make outside code cleaner

yarn.lock
```diff
@@ -6531,6 +6532,7 @@ __metadata:
   resolution: "@storybook/components@workspace:lib/components"
   dependencies:
     "@popperjs/core": ^2.6.0
+    "@storybook/addons": 6.3.0-rc.11
```
yannbf, line 6535: does this make sense? I think this is the reason the install step in the pipeline is failing

lib/addons/src/hooks/useCallbackRef.ts
```diff
@@ -0,0 +1,14 @@
...
+  const [callbackRef, setRef] = useState<MutableRefObject<T | null>>(null);
+  const setCallbackRef = useCallback(
+    (ref: MutableRefObject<T | null>) => {
+      setRef(ref);
+    },
+    [setRef]
+  );
+
+  return { callbackRef, setCallbackRef };
+}
```
ndelangen, line 14: I think this hook is pointless

ndelangen (changes requested):

yarn.lock
IgorSzyporyn: Seems outdated now I merged in next - lets hope it fixes it, thanks for finding that man!

lib/addons/src/hooks/useCallbackRef.ts
IgorSzyporyn: It is required for more complex use cases when it comes to DOM element positions

Every hook I have been looking at does it as well (not that bandwagon argumentation is a thing) - but tons of posts about it as well

I do not claim to be able to fully convince you with argumentation as I lack that kind of detailed in-depth knowledge though.

Can we have a chat about it and go in-depth @ndelangen?

IgorSzyporyn: Hook is gone

yarn.lock
IgorSzyporyn: Problem seems to be a hard link to @storybook/addons 6.3.0 in the yarn.lock file - only place I can find a reference to it anywhere, and that is what e2e test is complaining about

lib/addons/src/hooks/use-dom-rect/useDOMRect.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export const useDOMRect = <T extends HTMLElement = HTMLDivElement>({
+  rounded,
+  live,
+  ref: customRef,
+}: UseDOMRectProps<T> = {}) => {
+  const hookRef = useRef<T>(null);
+  const ref: MutableRefObject<T> = customRef || hookRef;
+  const [_rect, setRect] = useState<UseDOMRectBounds>({ ...initialState });
+  const rect = _rect as DOMRect;
```
ndelangen, line 74: Thisshould not be needed, right?

lib/addons/src/hooks/use-dom-rect/useDOMRect.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    return () => {
+      if (observer) {
+        observer.disconnect();
+      }
+    };
+  }, [ref]);
+
+  return { ref, rect: rect as DOMRect };
```
ndelangen, line 118: I'm not sure, why this hook would return something that was passed into it?
Seems to me, like it could return just the `rect`

ndelangen (changes requested):

lib/addons/src/hooks/use-dom-rect/useDOMRect.ts
IgorSzyporyn: Often when using hooks that requires a ref I find the file to be be a little ref cluttered, so it's so you can either be given the ref yourself - or be courtesy handed one back so you don't have to create it yourself.

```ts
import React, { useRef } from 'react'
import { useDomRect } from '@storybook/addons'

const MyComp = () => {
  const ref = useRef<HTMLButtonElement>(null);
  const { rect: { width } } = useDomRect<HTMLButtonElement>({ref});
  
  return <button ref={ref}>{width}</button>
}
```

vs

```ts
import React from 'react'
import { useDomRect } from '@storybook/addons'

const MyComp = () => {
  const { ref, rect: { width } } = useDomRect<HTMLButtonElement>({ref});
  
  return <button ref={ref}>{width}</button>
}
```

lib/addons/src/hooks/use-dom-rect/useDOMRect.ts
IgorSzyporyn: It should not be, but it is - been trying to figure it out since `as` declarations annoys me 😄

PR #15308, opened 2021-06-21
Components: New ScrollArea component

IgorSzyporyn:
**Telescoped FROM #15304**
**Telescopes INTO #15323**

## What I did
Created a new ScrollArea component using the new hook `useDOMRect` which makes the dependency [overlayscrollbars](https://www.npmjs.com/package/overlayscrollbars) obsolete in  #15323 

## How to test
Fire up the `official-storybook` storybook and go to `Basics/ScrollBars`

I have only named the folder "ScrollBars" to make the transition to take over ScrollArea easier in next PR


lib/components/src/ScrollBar/ScrollArea.tsx
```diff
@@ -0,0 +1,344 @@
...
+    (event) => {
+      if (showOn === 'hover') {
+        setState({
+          ...state,
+          horizontal: { ...state.horizontal, show: true },
+          vertical: { ...state.vertical, show: true },
+        });
+      }
+
+      if (rest.onMouseEnter) {
```
yannbf, line 215: shouldn't this be just handled as `onMouseEnter` prop?

lib/components/src/ScrollBar/ScrollArea.tsx
```diff
@@ -0,0 +1,344 @@
...
+    (event) => {
+      if (showOn === 'hover') {
+        setState({
+          ...state,
+          horizontal: { ...state.horizontal, show: false },
+          vertical: { ...state.vertical, show: false },
+        });
+      }
+
+      if (rest.onMouseLeave) {
```
yannbf, line 233: same question as `onMouseEnter`

lib/components/src/ScrollBar/ScrollArea.tsx
```diff
@@ -0,0 +1,344 @@
...
+  sliderColor,
+  sliderFadeout = SLIDER_DEFAULT_FADEOUT,
+  sliderOpacity = SLIDER_DEFAULT_OPACITY,
+  sliderPadding = SLIDER_DEFAULT_PADDING,
+  sliderSize = SLIDER_DEFAULT_SIZE,
+  sliderType,
+  vertical: enableVertical = true,
+  verticalPosition = 'right',
+  onScroll,
+  ContentProps = {},
```
yannbf, line 31: Why pascal case?

lib/components/src/ScrollBar/components/TrackHorizontal.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { FC, useCallback, useRef } from 'react';
+import { document as _document } from 'global';
+import * as Styled from '../styled';
+import { ScrollAreaTrackProps } from '../types';
+
+const document = _document as Document;
```
yannbf, line 6: why? just for the types?

lib/components/src/ScrollBar/styled.tsx
```diff
@@ -0,0 +1,113 @@
...
+  sliderSize: number;
+}
+
+export const Slider = styled.div<SliderProps>(({ theme, sliderColor, sliderType, sliderSize }) => ({
+  backgroundColor: sliderType ? theme.color[sliderType] : sliderColor || theme.color.dark,
+  borderTopLeftRadius: sliderSize / 2,
+  borderTopRightRadius: sliderSize / 2,
+  borderBottomLeftRadius: sliderSize / 2,
+  borderBottomRightRadius: sliderSize / 2,
+  cursor: 'grabbing',
```
yannbf, line 104: shouldn't the scrollbar function like the default ones from the system? just decrease opacity on hover and keep the normal cursor style?

yannbf:
I noticed this when dynamically changing the dimensions of the element. Is it expected?

scroll-bug

lib/components/src/ScrollBar/ScrollArea.tsx
IgorSzyporyn: I am not interested in the onMouseEnter prop from the outside world myself

And since the  `<Styled.Wrapper />` is the one to get the spreaded props (which would include the onMouseEnter) but the `<Styled.ScrollableContainer />` is the one to actually hold the event I just give the option of having a "native" feel to the event handling in this case, without breaking anything.

Use still gets his onMouseEnter - but due to some fancy DOM setup it has to be derived from another actual DOM element

IgorSzyporyn: Hmmmm - you got me thinking... will get back to you here 😄

lib/components/src/ScrollBar/components/TrackHorizontal.tsx
IgorSzyporyn: Yes - the global we are using is however pretty bad, and I am proposing in another PR that we use [window-or-global](https://www.npmjs.com/package/window-or-global) which has nice types, 0 dependencies and is maintained

lib/components/src/ScrollBar/styled.tsx
IgorSzyporyn: Let's vote 😃 

I don't feel there is a standard for this across systems, and some leverage is available - I quite like it because it is visually declarative in terms of what my options are.

IgorSzyporyn:
> I noticed this when dynamically changing the dimensions of the element. Is it expected?
> 
> scroll-bug

Which part?

It should adjust itself according to ratio and end up filling up to 100% and then go away when there is no longer a need for the scroller because container is as high/higher

lib/components/src/ScrollBar/ScrollArea.tsx
IgorSzyporyn: I like to differentiate betweeen props that is for the wrapper and props that are for support components

IgorSzyporyn: If this is an approach not to the teams liking I will change off course

shilman: definitely prefer camelCase!

lib/components/src/ScrollBar/styled.tsx
IgorSzyporyn: I will change back to default - been looking around what is out there and you are right 😃

IgorSzyporyn: fixed

lib/components/src/ScrollBar/ScrollArea.tsx
IgorSzyporyn: fixed

lib/components/src/ScrollBar/ScrollArea.tsx
IgorSzyporyn: You are right - fixed

lib/components/src/ScrollBar/ScrollArea.tsx
IgorSzyporyn: fixed

IgorSzyporyn:
Thanks @yannbf @shilman @ndelangen for the good PR - changes as requested are up 😃

yannbf:
> > I noticed this when dynamically changing the dimensions of the element. Is it expected?
> > scroll-bug
> 
> Which part?
> 
> It should adjust itself according to ratio and end up filling up to 100% and then go away when there is no longer a need for the scroller because container is as high/higher
> 
> The ratio between the track and the slider should be the same as the one between the outer and inner scroll area

If you notice, it seems like 

> > I noticed this when dynamically changing the dimensions of the element. Is it expected?
> > scroll-bug
> 
> Which part?
> 
> It should adjust itself according to ratio and end up filling up to 100% and then go away when there is no longer a need for the scroller because container is as high/higher
> 
> The ratio between the track and the slider should be the same as the one between the outer and inner scroll area

Sorry my message was very generic hehe

I'm on my phone so can't find the exact code easily, but it seems like if the width is 100 and then it's set to 500, the element will recalculate its width pixel by pixel, rather than just adjusting itself directly to 500 or animating via css. If you see the video, for every recalculation, the styled element recreates a css class.. That doesn't seem well performant

IgorSzyporyn:
> > > I noticed this when dynamically changing the dimensions of the element. Is it expected?
> > > scroll-bug
> > 
> > 
> > Which part?
> > It should adjust itself according to ratio and end up filling up to 100% and then go away when there is no longer a need for the scroller because container is as high/higher
> > The ratio between the track and the slider should be the same as the one between the outer and inner scroll area
> 
> If you notice, it seems like
> 
> > > I noticed this when dynamically changing the dimensions of the element. Is it expected?
> > > scroll-bug
> > 
> > 
> > Which part?
> > It should adjust itself according to ratio and end up filling up to 100% and then go away when there is no longer a need for the scroller because container is as high/higher
> > The ratio between the track and the slider should be the same as the one between the outer and inner scroll area
> 
> Sorry my message was very generic hehe
> 
> I'm on my phone so can't find the exact code easily, but it seems like if the width is 100 and then it's set to 500, the element will recalculate its width pixel by pixel, rather than just adjusting itself directly to 500 or animating via css. If you see the video, for every recalculation, the styled element recreates a css class.. That doesn't seem well performant

yeah that is a slip-up, thanks man!!

All changes that are related to the scrolling and calculations should only have been using inline styles

lib/components/src/ScrollBar/styled.tsx
```diff
@@ -0,0 +1,112 @@
...
+  borderBottomRightRadius: sliderSize / 2,
+}));
+
+export const VerticalSlider = styled(Slider)(({ sliderSize }) => ({
+  width: sliderSize,
+}));
+
+export const HorizontalSlider = styled(Slider)(({ sliderSize }) => ({
+  height: sliderSize,
+}));
```
ndelangen, line 112: This is bad, should use inline-styles here

lib/components/src/ScrollBar/styled.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+interface SliderProps {
+  sliderColor?: string;
+  sliderType?: keyof Theme['color'];
+  sliderSize: number;
+}
+
+export const Slider = styled.div<SliderProps>(({ theme, sliderColor, sliderType, sliderSize }) => ({
+  backgroundColor: sliderType ? theme.color[sliderType] : sliderColor || theme.color.dark,
```
ndelangen, line 99: This should switch based on theme.base

lib/components/src/ScrollBar/components/TrackHorizontal.tsx
```diff
@@ -0,0 +1,92 @@
...
+          data-sb-scrollarea-slider=""
+          data-sb-scrollarea-slider-horizontal=""
+          sliderColor={sliderColor}
+          sliderSize={sliderSize}
+          sliderType={sliderType}
+          onMouseDown={dragStart}
+          style={{
+            transform: `translateX(${state.slider.position}px)`,
+            width: state.slider.size,
+          }}
```
ndelangen, line 87: You could apply these inline styles in `handleScroll` on the ref directly, instead of managing it in state and having react apply them.

ndelangen: I think setting state for every scroll event is a really bad idea.
preferably, we change the dom directly in the side effect, and not have react do reconciliation on every scroll frame

ndelangen (changes requested):

lib/components/src/ScrollBar/components/TrackHorizontal.tsx
IgorSzyporyn: Agree - Yann pointed some out that this does

PR #15312, opened 2021-06-21
Toolbars: Add missing `regenerator-runtime` dependency

gaetanmaisse:
Issue: https://github.com/storybookjs/storybook/issues/13830#issuecomment-864242992

## What I did

I just added `regenerator-runtime` as dependency in `@storybook/addon-toolbars`


shilman (approved):
lgtm

PR #15331, opened 2021-06-23
Core: Support `argType.type.summary` when mapping url args

ghengeveld:
Issue: #15278

## What I did

Sometimes we get a `type.summary` rather than a `type.name`, which causes the url args mapping to fail. This fixes that.

## How to test


ghengeveld:
Closing in favor of #15332

We should just make sure we always have a `type.name`.

PR #15332, opened 2021-06-23
Svelte: Fix argType.type.name extraction

ghengeveld:
Issue: #15278

## What I did

ArgTypes for Svelte were incorrectly setting `type.summary` which should be `type.name`, as `summary` only exists in the context of `table`.

## How to test


shilman (approved):
LGTM! 🙏

PR #15354, opened 2021-06-24
Essentials: Update addon measure and outline

winkerVSbecks:
Issue:

## What I did
Updated the measure and outline addons. This fixes the dependency warnings.


shilman (commented):
Thanks @winkerVSbecks :pray:

shilman (approved):

PR #15363, opened 2021-06-25
feature: new tabs component

IgorSzyporyn:
**NB: Should be telescoped off #15323 - but for some reason it say no changes between these branches**

## What I did
Created a new TabsBar component

Created two components for export without TabsBar capabilities `TabsButton` & `TabsTool` for usage when the look and feel of a Tab is required

Created the component `ContentTab` for tabs that has content to show
Created the component `ButtonTab` for tabs that has no content to show
Created the component `MenuTab` for tabs that has menu content to show
Created the component `ToolTab` for tabs that are just tools to be clicked
Created the component `SeperatorTab` for when a separator of tabs is required

TabsBar will accept any children - but for functionality the components ContentTab, ButtonTab, MenuTab, ToolTab and SeparatorTab are to be used.

The new TabsBar will accept the property `staticTools` to lock the tools right side in place

Also added the hook `useWindowSize`

## How to test

Spin up `official-storybook` and go to `Basics/TabsBar`


lib/addons/src/hooks/useWindowSize.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface WindowSize {
+  width: number | undefined;
+  height: number | undefined;
+}
+
+// Hook
+export const useWindowSize = () => {
+  // Initialize state with undefined width/height so server and client renders match
+  // Learn more here: https://joshwcomeau.com/react/the-perils-of-rehydration/
```
ndelangen, line 12: We don't SSR?

lib/components/src/TabsBar/TabsButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+);
+
+interface WrapperProps {
+  active: boolean;
+  textColor: string;
+  activeColor: string;
+  narrow: boolean;
+}
+
+const Wrapper = styled.button<WrapperProps>(
```
ndelangen, line 73: I don't think `Wrapper` is the best name of this, this is actually the Button, and what's called `TabsButton` is the Wrapper?

lib/components/src/TabsBar/components/ToolTab.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const { addToMap, removeFromMap } = useContext(TabsBarContext);
+  const tabProps = getTabProps(props);
+
+  useEffect(() => {
+    addToMap(tabProps);
+
+    return () => {
+      removeFromMap(tabProps);
+    };
+  }, []);
```
ndelangen, line 18: Can you explain this pattern to me?

It seems odd to me that this component adds itself to a context like this?

ndelangen (changes requested):
Can you explain the `addToMap(tabProps);` pattern in detail, and why it's needed and why it's the best pattern to use here?

I feel as if this Tabs component could be vastly simplified.

lib/components/src/TabsBar/TabsButton.tsx
IgorSzyporyn: I just consistently call my components outside styled wrappers for `Wrapper` to avoid having to think and having the name I come up with have variant patterns.

lib/components/src/TabsBar/components/ToolTab.tsx
IgorSzyporyn: The offset was the component `tabs` and it's usage - and especially `tabs` and the modified and borrowed and simulated usage across storybook.

This component (as it was) wanted to be able to just accept children and then pipe them into the whole logic train of the TabsBar, by not actually returning or using the given component/node, but creating a new, and altering props.

This pattern provides a hook-in to register to the context - via the context - and kind of yell "yes I want to be a part of this setup", and the TabsBar knows nothing of this and just accepts whatever wants to register with it.

The original usage is preserved and the syntax is for me more declarative and precise

This PR aims at TabsBar now being able to handle all of our tab bar requirements as we currently semi-mock and bypass many many places, as tabs with the older pattern can not handle what is required - where before it could only handle

- Tab with content

It can now handle

- Tab with content
- Tab as a button
- Tab as a button with a menu
- Tab that is just a separator
- And we can easily add to this range without having to do more than write that new tab component type.

It will then follow the pattern and syntax from `Accordion`

```ts
<TabsBar>
  <ContentTab id="canvas" label="Canvas">Show the canvas</ContentTab>
  <ContentTab id="docs" label="Docs">Show the docs</ContentTab>
  <SeparatorTab />
  <ToolTab id="zoom" icon="zoom" />
  <ToolTab id="zoom-out" icon="zoom-out" />
  <ToolTab id="zoom-reset" icon="zoom-reset" />
  <SeparatorTab />
  <ButtonTab id="grid" icon="grid" />
  <MenuTab id="viewport" icon="grow" menu={[{ id: 'small-mobile' label: 'Small Mobile'}]} />
</TabsBar>
```

Or you can do it as an config object

```ts
<TabsBar
  tabs={[
    { id: "canvas", label: "Canvas", content: "Show the canvas" },
    { id: "docs", label: "Docs", content: "Show the docs" },
    { type: "separator" },
    { id: 'zoom-in', type: 'tool', icon: 'zoom' },
    { id: 'zoom-out', type: 'tool', icon: 'zoom-out' },
    { id: 'zoom-reset', type: 'tool', icon: 'zoom-reset' },
    { type: "separator" },
    { id: 'grid', icon: 'grid', type: 'button' },
    {
      id: 'viewport',
      icon: 'grow',
      type: 'menu',
      menu: [
        { id: 'small-mobile', label: 'Small Mobile' },
        { id: 'mobile', label: 'Mobile' },
        { id: 'large-mobile', label: 'Large Mobile' },
        { id: 'tablet', label: 'Tablet' },
      ],
    },
  ]}
/>
```

lib/components/src/TabsBar/TabsButton.tsx
IgorSzyporyn: I can change that and make my rule component name + wrapper so it would be

`TabsButtonWrapper`

IgorSzyporyn: As you also mentioned you would like styled components in a file of it's own this would be sorted as it could then easily become

`Styled.Button`

lib/addons/src/hooks/useWindowSize.ts
IgorSzyporyn: I don't know if we do or not

Just though I would add this as it would be exposed to the outside world

PR #15377, opened 2021-06-26
Documentation: Change isToolshown type from String to Boolean

apapadopoulou:
Issue: While going through the documentation, I noticed that the `isToolshown` api value is categorized as `String`, but I think that it should be `boolean`.

## What I did
I changed the type of `isToolshown` from `String` to `Boolean`.
## How to test
You can view the file preview.


jonniebigodes:
@apapadopoulou thanks for catching this and help us with the documentation improvements. Indeed you're correct and it should be a Boolean as opposed to a String value.

Merging this.

PR #15408, opened 2021-06-29
Core: Fix decorator context update

tmeasday:
Issue: #15400 

Previously we always passed the combination of the original context and the previous decorator's passed arguments into the next decorator.

That meant if an earlier decorator set something on the context, it would get lost by the time it reached the story.

## What I did

Updated the "decorator with context store" so that the context store keeps getting updated after each decorator along the chain.

## How to test

PR includes a jest test and a story that demonstrate the behaviour


shilman (approved):
LGTM -- tricky!

PR #15410, opened 2021-06-29
Angular: Fix circular reference not being handled in moduleMetadata

Marklb:
Issue: #15333

## What I did

Fixed error when `moduleMetadata` contains a circular reference.

This change adds `telejson` as a dependency, when the only thing it is being used for is stringifying an object for comparing with previous. It may be better to just add a method that can handle the circular reference, instead of adding a dependency, but haven't had time to try that. I just wanted to go on an submit my quick solution and can try to avoid the added dependency when I get a chance, unless someone else wants to do it.

## How to test

### DIFF
--- a/app/angular/package.json
+++ b/app/angular/package.json
@@ -65,6 +65,7 @@
     "regenerator-runtime": "^0.13.7",
     "sass-loader": "^10.1.0",
     "strip-json-comments": "3.1.1",
+    "telejson": "^5.3.2",
     "ts-dedent": "^2.0.0",
     "ts-loader": "^8.0.14",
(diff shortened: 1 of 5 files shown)


Marklb:
The more I think about it, maybe adding the additional `telejson` dependency is worth it, since it is tested and small. I am not really sure what a reasonable depth would be though.

ThibaudAV (approved):
i don't know `telejson`. 
I don't know if storybook already has a similar dependency to avoid multiplying them 🤷‍♂️ 

but it looks good to me 👍

shilman (approved):
looking good!! 💯

PR #15415, opened 2021-06-29
Replace uuid and uuid-browser with nanoid

IgorSzyporyn:
## What I did

Removed the two packages [uuid](https://www.npmjs.com/package/uuid) and [uuid-browser](https://www.npmjs.com/package/uuid-browser) for [nanoid](https://www.npmjs.com/package/nanoid)

## Why I did it

We do have a unique id requirement in places, and nanoid provides an easier to use interface and has less bells and whistles and comes in at half the cost in unpacked size (52.7 kB vs 116 kB) - and is only 108 bytes minified/gzipped and they claim it is 60% faster than UUID

## How to test

Pipeline tests should be enough

### DIFF
--- a/addons/actions/package.json
+++ b/addons/actions/package.json
@@ -51,13 +51,13 @@
     "fast-deep-equal": "^3.1.3",
     "global": "^4.4.0",
     "lodash": "^4.17.20",
+    "nanoid": "^3.1.23",
     "polished": "^4.0.5",
     "prop-types": "^15.7.2",
     "react-inspector": "^5.1.0",
     "regenerator-runtime": "^0.13.7",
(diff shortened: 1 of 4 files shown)


ndelangen (approved):

shilman (approved):
LGTM

shilman (changes requested):
Looks like this PR breaks IE11 support:

https://www.chromatic.com/build?appId=5def62c97bd45f0020ef370e&number=8752

IgorSzyporyn:
> Looks like this PR breaks IE11 support:
> 
> https://www.chromatic.com/build?appId=5def62c97bd45f0020ef370e&number=8752

looking

PR #15418, opened 2021-06-29
Angular: Fix duplicate stories in docs view

dexster:
Issue: #15170 
 
## What I did

Allow stories to be added to mdx docs more than once  #15170 

I used nanoid so it's the same version that @IgorSzyporyn has just committed 

## How to test

Duplicate story has been added to kitchen sink

snapshots have also been updated. I don't think they've been updated for a while


dexster:
@ThibaudAV @IgorSzyporyn  Do you think we could get this merged. In "Files changed" tab it's just the package.json and prepareForInline that would need to be reviewed. All other files are just snapshot updates.

ThibaudAV:
This is strange this snapshot updates 🤔  i don't think theses change should cause them

dexster:
I don't think my changes caused them. I think updated snapshots have not been committed in the last 5 months so many other changes might have caused this.

Should I remove the snapshot updates from the commit?

ThibaudAV:
I don't think so, otherwise the test would have failed.

you have to start by fixing the circleci test of your pr [here](https://github.com/storybookjs/storybook/pull/15418/checks?check_run_id=2943198828)

apparently :
`The lockfile would have been modified by this install, which is explicitly forbidden`

Could you not add yarn.lock update ?

dexster:
Too many issues. Unsure why snapshot updates were required. I will submit a new PR

PR #15428, opened 2021-06-29
Addon-docs: Cache DocsContext on window to prevent duplication

eirslett:
This is intended to solve https://github.com/eirslett/storybook-builder-vite/issues/40
and related issues.

## How to test

Look at the linked issue.
Hopefully this shouldn't break any existing functionality in addon-docs, but looking at the code I don't think it will.

It would be nice to have this backported to the 6.3 release as well!

Arguably, the actual problem here is that Vite is making untrue assumptions, so ideally it should be fixed there. I think it's

### DIFF
--- a/addons/docs/src/blocks/DocsContext.ts
+++ b/addons/docs/src/blocks/DocsContext.ts
@@ -1,4 +1,5 @@
 import { Context, createContext } from 'react';
+import { window as globalWindow } from 'global';
 
 export interface DocsContextProps {
   id?: string;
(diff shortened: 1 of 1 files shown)


shilman:
This is pretty ugly but I suppose we can use it if there’s no proper fix in Vite-land?

eirslett:
I had a little look in Vite and couldn't find a simple solution. So if it's acceptable we could put this in Storybook until we figure out a better way.

shilman (approved):
Works for me... 🤷

Pewtro:
I believe this PR broke some functionality in so that I get the following error when trying to utilise addon-docs and addon-storyshot at the same time, but maybe I'm missing something in my setup, but when going through the documentation I've done everything that's required
image

shilman:
@eirslett what do you think about wrapping your addition in `if(globalWindow) { ... }`?

eirslett:
> @eirslett what do you think about wrapping your addition in if(globalWindow) { ... }?


That should work, I suppose. If `globalWindow` is not defined, we could just create the context directly.

Pewtro:
I wrote a patch-package script to handle it but I basically did this: 
```js
-if (_global.window.__DOCS_CONTEXT__ === undefined) {
+if (_global.window && _global.window.__DOCS_CONTEXT__ === undefined) {
   _global.window.__DOCS_CONTEXT__ = /*#__PURE__*/(0, _react.createContext)({});
   _global.window.__DOCS_CONTEXT__.displayName = 'DocsContext';
 }
 
-var DocsContext = _global.window.__DOCS_CONTEXT__;
+var DocsContext = _global.window ? _global.window.__DOCS_CONTEXT__ : null;
```

My storyshots aren't picking up the mdx files for now, and I'm unsure if that's because of what I did here or not - but atleast storyshots is picking up the regular stories and snapshotting them now whilst keeping functionality the same in my build/development environments.

eirslett:
Looks good! You could maybe just run `createContext()` instead of `null` in the last ternary expression. But I think you have the right fix!

PR #15456, opened 2021-07-01
typo(cli): fix error message (yarn → npm)

PiDelport:
Fix the error message in `NPMProxy.runGetVersions` to refer to "npm" instead of "yarn".

I ran into this while troubleshooting problems with `sb init`.


shilman (commented):
thank you!! 🙏

PR #15459, opened 2021-07-01
UI: Fix sidebar toggle in fullscreen mode

yannbf:
Issue: related to #15369

The toggle functionality didn't work when in fullscreen.

## What I did

Fixed so that when clicking on the toggle icon when in fullscreen mode, it will show the sidebar correctly.

## How to test

- run storybook
- Press A then S (to hide both addons panel and sidebar, effectively going to fullscreen)
- Click in the hamburger icon on top left. It should open the sidebar

- Is this testable with Jest or Chromatic screenshots? I guess better if tested in Cypress


darleendenno (approved):
Great catch! Thank you! This is much better in fullscreen mode.

PR #15472, opened 2021-07-03
Fix: Recalculate height of ZoomElement when child element updates

Tomastomaslol:
Issue: #15368 

## What I did
- Removed duplicated never called function `browserSupportsCssZoom`
- Added an observer to "watch" for changes in child elements
-  Update calculated height when child elements are updated

In browsers without support for CSS `zoom` the height of the element needs to be known to be able to simulate zoom using `transform`. When the height of the child `ZoomElement` updates the height needs to be re-calculated.

### New behavior
https://user-images.githubusercontent.com/1872246/146660086-8e103823-4cc7-4929-8999-6b17718d9c85.mp4

### Current behavior
https://user-images.githubusercontent.com/1872246/146660138-3f25e815-eb82-41ba-8304-c5e2b84b6631.mp4


darleendenno:
Hey @Tomastomaslol is this ready for review?

Tomastomaslol:
Hi @darleendenno! Sorry, I took a long break from this PR to focus on other things.

I think I have figured out a decent solution to the issue described in #15368 and I'm planning to spend some more time improving the code and "productionize" it before it's ready for review. 

I will let you know when I'm happy with the code and update the status of the PR. 🙂

lib/components/src/Zoom/ZoomElement.tsx
```diff
@@ -20,16 +21,52 @@ type ZoomProps = {
...
+  options = {},
+  callback,
+}: {
+  element: RefObject<Element>;
+  options: MutationObserverInit;
+  callback: MutationCallback;
+}): void => {
+  const observer = useMemo(
+    () =>
+      new MutationObserver((mutationRecord, mutationObserver) => {
```
Tomastomaslol, line 35: Handle `MutationObserver` not defined

lib/components/src/Zoom/ZoomElement.tsx
```diff
@@ -20,16 +21,52 @@ type ZoomProps = {
   children: ReactElement | ReactElement[];
 };
 
+const useMutationObserver = ({
```
Tomastomaslol, line 24: Can this live in this file or should this be moved somewhere else?

lib/components/src/Zoom/Zoom.tsx
```diff
@@ -1,19 +1,6 @@
-import global from 'global';
 import { ZoomElement as Element } from './ZoomElement';
 import { ZoomIFrame as IFrame } from './ZoomIFrame';
 
-const { window: globalWindow } = global;
-
-export const browserSupportsCssZoom = (): boolean => {
```
Tomastomaslol, line 7: This function is never called and it looks like it moved here:

https://github.com/storybookjs/storybook/blob/a8db9585f07d8cfc7354e68f6b780a6ff6536b38/lib/components/src/Zoom/browserSupportsCssZoom.ts#L5

tmeasday:
> Does this change still really do anything when rendering in modernInlineRendering (which will be the default in 7.0 anyway @shilman @tmeasday ?

I'm not sure modern inline rendering really changes the behaviour of this component @ndelangen? Can you explain what you were thinking here, I'm missing something I think.

I think this PR should be considered as part of the doc blocks 2.0 work. I'm not quite sure what the exact plan is there @shilman.

ndelangen:
@tmeasday with inline rendering, the size of the container will just auto-update. so there's no need to keep it in sync, because the layout engine is doing that the natural way.
That's what I was referring to.

lib/components/src/Zoom/ZoomElement.tsx
```diff
@@ -15,6 +24,35 @@ const ZoomElementWrapper = styled.div<{ scale: number; height: number }>(({ scal
...
+  options = {},
+  callback,
+}: {
+  element: RefObject<Element>;
+  options: MutationObserverInit;
+  callback: MutationCallback;
+}): void => {
+  const observer = useMemo(
+    () =>
+      typeof MutationObserver === 'function'
```
ghengeveld, line 39: I don't think we need this check, `MutationObserver` has been available in all browsers since 2014. Save for Opera Mini which frankly nobody is going to be using to view their Storybook in anyway.

lib/components/src/Zoom/ZoomElement.tsx
```diff
@@ -24,15 +62,28 @@ export function ZoomElement({ scale, children }: ZoomProps) {
...
+
   useEffect(() => {
     if (componentWrapperRef.current) {
       setHeight(componentWrapperRef.current.getBoundingClientRect().height);
     }
   }, [scale, componentWrapperRef.current]);
 
+  useMutationObserver({
+    element: componentWrapperRef,
+    options: { subtree: true, childList: true },
```
ghengeveld, line 77: `options` should be memoized else we'll have a new object with each render, which will cause the `useEffect` above to disconnect and re-observe.

ghengeveld (approved):

lib/components/src/Zoom/ZoomElement.tsx
ghengeveld: If this is the only place we use a MutationObserver, I'm fine with it being in this file.

lib/components/src/Zoom/ZoomElement.tsx
```diff
@@ -24,15 +62,28 @@ export function ZoomElement({ scale, children }: ZoomProps) {
...
+    callback: handleMutations,
+  });
+
   return (
     <ZoomElementWrapper scale={scale} height={height}>
-      <div ref={componentWrapperRef} className="innerZoomElementWrapper">
+      <div
+        ref={browserSupportsCssZoom() ? null : componentWrapperRef}
```
ghengeveld, line 84: I think we can just call `browserSupportsCssZoom` once, outside the component render function, put it in a variable and use that here. Otherwise we'll be running this check over and over again.

lib/components/src/Zoom/ZoomElement.tsx
ndelangen: I strongly **agree** with this, Storybook 7.0 will target `chrome100`. If it's supported in there, don't bother downgrading, transpiling, of feature detecting!

tmeasday:
> @tmeasday with inline rendering

@ndelangen well if you are talking about *inline* rendering, then that hasn't changed in 7.0. The default was to inline render already. What's changed is the mechanism to do it, which doesn't affect how the browser lays things out AFAIK. I guess this is useful for non-inline mode? Although TBH I wasn't aware that a mutation observer would work there..

PR #15483, opened 2021-07-05
Webpack5: Quit process after finishing a static build

bebraw:
Issue: https://github.com/storybookjs/storybook/issues/15227

## What I did

I added a `try`/`catch` around Angular's build process and made sure webpack compiler `close` method gets called. See https://github.com/angular/angular-cli/issues/21261#issuecomment-875153057 for reference.

## How to test

It's best to test the change against https://github.com/hoffination/storybook-timeout-repro . Ideally a subset of that would be pulled here as a test somehow.

### DIFF
--- a/app/angular/src/server/build.ts
+++ b/app/angular/src/server/build.ts
@@ -5,9 +5,6 @@ import options from './options';
 async function build() {
   try {
     await buildStatic(options);
-
-    // #15227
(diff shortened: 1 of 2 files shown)


shilman (approved):
Looking good! Thanks for tracking this down @bebraw !!

PR #15486, opened 2021-07-05
Addon-docs: Fix absolute path links in docs

stefan-schweiger:
Issue: #9005 

## What I did
Removed the check for `isStorybookPath` in documentation links. Since it only checks if the URL starts with a `/` this does not make too much sense to begin with and breaks absolute links.

## How to test

- **Is this testable with Jest or Chromatic screenshots?**
I've added Jest tests
- **Does this need a new example in the kitchen sink apps?**
No
- **Does this need an update to the documentation?**
Probably not?


ndelangen:
This is a feature exists to allow users to link to other stories... Which I'm sure people use.

I do agree that matching on `/` and considering all absolute paths as "storybook" urls is terrible, and makes not sense.

Perhaps we should match on `/(story|docs)\/.` instead?

stefan-schweiger:
The documentation tells us to use `?path=/story/my-super-story` if you need to link to other stories, so I don't see much benefit of matching `/` at all, except for people already using it incorrectly.

stefan-schweiger:
Another solution we might use is to allow something like `?external=/absolut/url` and handle this correctly. But I still think handling everything that isn't `?path=` as real links would be the correct way to do it.

ndelangen:
I agree, but changing that would technically be a breaking change.

stefan-schweiger:
@ndelangen so how do you want to proceed with this issue?

ndelangen:
@shilman how would you feel about putting this in 6.4, and adding a section to the migration guide

ndelangen:
ok, 6.5 it is

stefan-schweiger:
@shilman @ndelangen is this still on the 6.5 roadmap?

ndelangen:
@stefan-schweiger I tried to get the tests moved to the proper (new) place, but it seems there's a failure now that the base branch has been merged in... I understand it's been a while, but could you maybe see if you could make the CI happy?

PR #15491, opened 2021-07-06
Addon-docs/Angular: Fix numeric args default value handling

stefan-schweiger:
Issue: #15490

## What I did
Added a check that if the type is numeric (boolean and string were already checked) then return the valueType.


shilman (approved):
Looking good! 🙈

PR #15496, opened 2021-07-06
Chore: (Docs) Minor adjustments to the theming docs

jonniebigodes:
With this pull request, the theming docs are updated to remove the reference to the usage of `--no-manager-cache` flag as it's no longer required as Storybook will apply the changes consistently.

What was done:
- Removed the flag mention in the said docs


jonniebigodes:
@shilman self merging this!

PR #15523, opened 2021-07-08
Build: Fix cache setup in GitHub Actions workflow

### LINKED ISSUE #15522: Considering it was added in 2.2.0 you would need to update this to v2
Considering it was added in 2.2.0 you would need to update this to v2, you also need to checkout the code before setting up node so it can actually get the cache folder

_Originally posted by @merceyz in https://github.com/storybookjs/storybook/pull/15480#r665372047_

oscard0m:
## 📝 Summary

- [x] Bump `actions/setup-node` to `v2` in `tests-unit.yml`
- [x] Move `actions/checkout` as first step in `tests-unit.yml`

## ⛱ Motivation and Context
Considering it was added in 2.2.0 you would need to update this to v2, you also need to checkout the code before setting up node so it can actually get the cache folder

Fix #15522

### DIFF
--- a/.github/workflows/tests-unit.yml
+++ b/.github/workflows/tests-unit.yml
@@ -7,11 +7,11 @@ jobs:
     name: Core Unit Tests
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
           node-version: "12.x"
           cache: yarn
-      - uses: actions/checkout@v2
       - name: install, bootstrap
         run: |
           yarn install --immutable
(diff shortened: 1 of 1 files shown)


oscard0m:
@merceyz Just in case you want to follow up on this.

gaetanmaisse (approved):
Totaly missed that in the first review 🤦🏻 

It looks better now: https://github.com/oscard0m/storybook/runs/3021295094?check_suite_focus=true

oscard0m:
> Totaly missed that in the first review 🤦🏻
> 
> It looks better now: https://github.com/oscard0m/storybook/runs/3021295094?check_suite_focus=true

I missed it too. Let's see how execution time goes with these changes! 💪🏽

PR #15526, opened 2021-07-08
Essentials: Add outline addon to monorepo

winkerVSbecks:
Issue: #15411 

## What I did

* Added storybook-addon-outline to the monorepo 
* Renames the package to @storybook/addon-outline and use same versioning strategy as other addons
* Updated the reference in addon essentials

@shilman I based the `package.json` on other addons. Not sure if I missed anything. Also, will switching the package name to `@storybook/addon-outline` cause any issue?

## How to test

Added a story to the official-storybook


shilman (approved):
Cool, thanks for getting this over! Updated some yarn/nx bureaucracy and snapshots to get CI unstuck ☝️

PR #15527, opened 2021-07-08
Composition: Fix refs ordering

### LINKED ISSUE #12188: Composite storybooks display order is undefined
**Describe the bug**
When main.js has refs to other storybooks, the order that they show up in the list seems to be based on which one loads first rather than their definition order.  There does not seem to be documentation for how to control their order.

**To Reproduce**

1. Add multiple refs to create a composite storybook
2. Add stories to the storybook in the last ref, but don't add stories to the storybook in the first ref
3. When running the composite storybook, the last storybook in the ref list will show up first

**Expected behavior**
The order should not change or be based on documented behavior that can be controlled via configuration

dexster:
Issue: #12188

## What I did
Fix #12188. Order is now set according to how the refs are entered in main.js

## How to test
I added a composition project to the examples


shilman (changes requested):
The fix looks great! However, can you please add the test to an existing example rather than creating a new one? We're in the process of fixing this, but currently every new example bloats the monorepo and slows down installation etc.

dexster:
Done. I moved the refs to the angular example project as that's the one I'm familiar with

shilman (approved):
Thanks so much! Awesome PR -- will merge & release today!!! 🙏

PR #15545, opened 2021-07-09
Essentials: Add measure addon to monorepo

### LINKED ISSUE #15411: Essentials: Bring measure/outline addons into monorepo

winkerVSbecks:
## What I did
* Moved @storybook/addon-measure to the monorepo
* Update the version to match versioning strategy used by other addons

closes #15411

## How to test

Added a stories to the official-storybook

### DIFF
--- a/addons/essentials/package.json
+++ b/addons/essentials/package.json
@@ -43,7 +43,7 @@
     "@storybook/addon-backgrounds": "6.4.0-alpha.13",
     "@storybook/addon-controls": "6.4.0-alpha.13",
     "@storybook/addon-docs": "6.4.0-alpha.13",
-    "@storybook/addon-measure": "^2.0.0",
+    "@storybook/addon-measure": "6.4.0-alpha.13",
     "@storybook/addon-outline": "6.4.0-alpha.13",
     "@storybook/addon-toolbars": "6.4.0-alpha.13",
     "@storybook/addon-viewport": "6.4.0-alpha.13",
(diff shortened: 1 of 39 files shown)


shilman (approved):
Thanks @winkerVSbecks ! `cra-bench` CI job is green again 💚

PR #15549, opened 2021-07-11
Controls: Fix color matching behavior for non-string types

yannbf:
Issue: #15443

## What I did

Fixed the assignment for color control based on the type of the passed arg

## How to test


shilman (approved):
LGTM! Do we need to worry about null/undefined values here?

PR #15556, opened 2021-07-12
Build: Fix `publish` step on CircleCI

gaetanmaisse:
## What I did

 - Fix Verdaccio config for `addon-measure`: as it is now on the monorepo there is no need to add a custom verdaccio rule for it 
 - Also improve the way Cypress records are stored as CircleCI artefacts by using specific folder for each framework, for instance: https://app.circleci.com/pipelines/github/storybookjs/storybook/19873/workflows/b7b2686a-52c6-4c80-a99a-576afd9f61a1/jobs/330350/artifacts
image

## How to test

- `publish` CI step should be 🟢

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,7 +198,7 @@ jobs:
           command: yarn test:e2e-framework --clean --all --skip angular11 --skip angular --skip vue3 --skip web_components_typescript --skip cra
           no_output_timeout: 5m
       - store_artifacts:
-          path: /tmp/storybook/cypress
+          path: /tmp/cypress-record
           destination: cypress
   e2e-tests-core:
     executor:
@@ -224,7 +224,7 @@ jobs:
           command: yarn test:e2e-framework vue3 angular angular11 web_components_typescript web_components_lit2
           no_output_timeout: 5m
       - store_artifacts:
-          path: /tmp/storybook/cypress
+          path: /tmp/cypress-record
           destination: cypress
   cra-bench:
     executor:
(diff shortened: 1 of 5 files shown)


yannbf:
You are my hero <3

gaetanmaisse:
@yannbf and I investigate a bit the issue with Angular e2e tests and we narrowed the issue to compodoc 1.1.12 as everything is working with 1.1.11. We need to check in details what could be the cause; changelog https://github.com/compodoc/compodoc/blob/master/CHANGELOG.md#1112-2021-06-07

yannbf:
Quick summary of the angular issue:

I think the issue was introduced after [this change](https://github.com/compodoc/compodoc/pull/1035/commits/031f4821d19a5a8b18326cd2c2804d14a8404d9c) in compodoc 1.1.12. [Here's a diff](https://www.diffchecker.com/SXq3ruVy) with changes that came in the componentMetadata from `1.1.11`(working) to `1.1.13`(failing). 

That change affected a check for non-supported properties and simply returned `{ type: { name: "void" } }` and then most of the argTypes passed to controls were incorrect.

I added a small fix, however as I don't have a lot of context there, I need someone else to help me check if the change makes sense.

After the change, controls is able to read/write internal properties. @shilman bug or feature? 😂 

image


On the positive side, in compodoc there's now support for the `deprecated` field and possibly other interesting things we can take advantage of!

shilman (approved):
LGTM. thanks so much @gaetanmaisse 🙏

PR #15557, opened 2021-07-12
Update MIGRATION.md to clarify default values inference

mayank99:
Issue: https://github.com/storybookjs/storybook/issues/15378#issuecomment-878481356

## What I did

Moved the part about `argType.defaultValue` above the deprecations section and added some wording to clarify that users now need to use top-level `args` instead of relying on sb to infer default values from component.

### Rationale

The inference change was breaking apps which don't even use `defaultValue`. "Deprecation" also implies that the previous behavior is preserved, not removed.


shilman (approved):
thanks @mayank99 💯

PR #15559, opened 2021-07-12
UI: Allow keyboard shortcut to copy code in preview blocks

melindali255:
## What I did
The keyboard shortcut for copying can be used to copy the source code of preview blocks. The source code does not need to be visible/expanded for the shortcut to work. After successfully copying, the action bar briefly displays a 'Copied' confirmation message.


darleendenno (commented):
Hey @melindali255 👋 Thank you for taking the time to improve Storybook!!

I love this! ❤️  It's very clean, and works as expected (this is me pressing CTRL + C a _lot_)
copy

Is there a use-case where people may want to copy partial code? 

In the current version, if I highlight a partial code block, I can copy that text. In your solution, I am only able to copy the full block. I'm eager to hear your thoughts! 

If this doesn't seem like a big deal from a ux perspective, then I think this code is good-to-go.

melindali255:
Hi @darleendenno! Thank you for your feedback!

I could see a use case where someone may want to only copy a certain section of the code block. Let me add some changes to allow that to happen.

lib/components/src/syntaxhighlighter/syntaxhighlighter.tsx
```diff
@@ -152,10 +160,15 @@ export const SyntaxHighlighter: FunctionComponent<Props> = ({
...
 
-  const onClick = (e: MouseEvent<HTMLButtonElement>) => {
+  const onClick = (e: MouseEvent<HTMLButtonElement> | ClipboardEvent<HTMLDivElement>) => {
     e.preventDefault();
 
-    copyToClipboard(highlightableCode)
+    const textToCopy =
+      e.type !== 'click' && globalWindow.getSelection().toString() !== ''
+        ? globalWindow.getSelection().toString()
+        : highlightableCode;
```
darleendenno, line 169: This works great!

darleendenno (approved):
❤️❤️❤️❤️ I LOVE THIS! ❤️❤️❤️❤️

Thank you for this improvement. Regression passes for me locally, and now users can copy full code blocks without highlighting. Great work!!

darleendenno:
@shilman uncertain why chromatic is displaying a different baseline for the sidebar -- will let you sort that out. This is good to go on my end 👍

shilman (approved):
Works great!!! 🏅 Thanks @melindali255!!

jonniebigodes:
@melindali255 I've kept an eye on this pull request and one final thing we'd like from you if you don't mind. Could you jump in our [Discord Server](https://discord.gg/UUt2PJb) and message me (same username), as there's one item to address.  Sounds good?

Stay safe

PR #15561, opened 2021-07-13
[Documentation : Configure Addons] fix broken link to withRoundTrip example

neilbryson:
## What I did
Change link from .js to .ts for the `withRoundTrip` example to avoid dead links


shilman (commented):
LGTM! thanks @neilbryson!!!

PR #15563, opened 2021-07-13
Angular: Fix actions argType auto generation

yannbf:
Issue: #15537

## What I did
- The order in the sections matters because properties are overridden, and actions were being lost now that there's a new property coming from compodoc

## How to test
- yarn boostrap --core
- yarn workspace angular-cli start-storybook
- Test the actions addon. Should work!


shilman (approved):
LGTM!

PR #15564, opened 2021-07-13
Preact typescript pragma fix

### LINKED ISSUE #15531: storybook cli doesn't add @babel/typescript when scaffolding preact project
**Describe the bug**
Babel config for typescrit should come included when scaffolding a preact project

**To Reproduce**
Add Storybook to an existing project of tsx preact. With writing your first story you'll get an error
`h is not defined`
Adding babel config to main.js will correct error
```
  babel: async options => ({
    ...options,
    presets: [["@babel/typescript", { jsxPragma: "h" }]],
  }),
```

**System**
Environment Info:

  System:
    OS: Windows 10 10.0.19041
    CPU: (32) x64 AMD Ryzen 9 3950X 16-Core Processor
  Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 2.4.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 91.0.4472.124
    Edge: Spartan (44.19041.1023.0), Chromium (91.0.864.67)
  npmPackages:
    @storybook/addon-actions: ^6.3.4 => 6.3.4
    @storybook/addon-docs: ^6.3.4 => 6.3.4
    @storybook/addon-essentials: ^6.3.4 => 6.3.4
    @storybook/addon-links: ^6.3.4 => 6.3.4
    @storybook/addon-postcss: ^2.0.0 => 2.0.0
    @storybook/preact: ^6.3.4 => 6.3.4

**Additional context**
I found the solution from this dev.to article: https://dev.to/sadnessojisan/storybook-with-preact-ts-goober-4ghk
Otherwise I would have been totally lost

KrofDrakula:
Issue:

## What I did

- Updated Babel configuration to use `automatic` runtime and set `importSource` to `preact` by default. This does not require importing  `h` from `preact`, and importing `React` works for compatibility.
- Updated `tsconfig` to take account of default JSX factories in `preact-kitchen-sink`.
- Pragmas should not be required in story files.
- Fixed stories to reflect new rules.

## How to test

- Is this testable with Jest or Chromatic screenshots? ✅ Yessir.
- Does this need a new example in the kitchen sink apps? ✅ Already contains one.
- Does this need an update to the documentation? ❌ There are no current Preact snippets available.


KrofDrakula:
@shilman Storyshots seem to fail after updating the Webpack plugin configuration to make Storybook run, but I'm unsure on how to solve the storyshots addon problems. Any pointers?

ndelangen (approved):

ndelangen:
Hey @KrofDrakula This looks really good! Thank you for this PR. I'm real sorry it's been left without any attention for so long..

I merged in next (4000+ commits behind, wow!), and I notice that the storyshots test is failing.. Could you look at this? I assume there's some babel config trickery that need the be updated to make it work.

Happy to merge once that's fixed!

ndelangen:
Yes I understand, mind that a dirty hack to get things working is totally acceptable here, because we're replacing storyshots with the storybook test-runner.

Storyshots will not survive much longer, but we also don't want it to break by accident.

KrofDrakula:
@ndelangen Finally caught some time to fix this. The underlying issue was that Jest wasn't transpiling JSX correctly (using `React.createElement` instead of `h` as is customary for Preact) and TypeScript wasn't setup correctly for Jest transpilation. I've renamed the files to JS and all the Jest tests pass now since introducing a Babel overridefor the Preact kitchen sink example. This means that TSX won't work currently compile correctly without more work.

Also, I've had to reintroduce the `h` import, but that's idiomatic within the Preact community anyways, so I felt that was OK.

Let me know if there's anything that may need a closer look at.

KrofDrakula:
@ndelangen Actually I jumped the gun on that one, the issues is with TypeScript, so will charm up a fix soon.

KrofDrakula:
@ndelangen OK, tests pass for storyshots and `examples/preact-kitchen-sink` now. There are some unrelated failures in snapshots for Angular since I merged the latest `next` branch, but I assume that's currently expected? I may need some help on that if we decide we should address those too in this PR.

ndelangen (approved):

code/examples/preact-kitchen-sink/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "jsx": "react-jsx",
+    "jsxFactory": "h",
+    "jsxFragmentFactory": "Fragment",
```
ndelangen, line 5: This seems to give me error/warnings when I open the file:
Screen Shot 2022-07-27 at 11 31 24

ndelangen:
@KrofDrakula I updated the branch with `next` which had some pretty drastic changes.
Could you have a look at the PR changed file, to see if I made the right calls in resolving the merge conflicts? 🙏 Thanks!

To me this PR looks good, if the CI checks are 🟢 we can merge!

KrofDrakula:
@ndelangen Thanks, will have a look. Yeah, I wasn't too sure about the `tsconfig` change, the editor error might be because of mismatching TS runtime versions (I'm using the latest one vs. the version specified in `package.json` for the Storybook repo, which is pretty old).

I'll have to look into that one; I'm always confused about that compile option, as it doesn't quite align with the Babel JSX plugin config.

PR #15581, opened 2021-07-14
Avoid slow regex.match call in renderJsx

maerten:
Issue:

## What I did

Added: a faster `indexOf` check to see if a slower regex `match` can be avoided.

The regex can cause slow renders if images are passed in base64 encoded format. The regex will then attempt to find the string 'quot&' in the entire base64 string.

## How to test

This should not impact behavior, other than performance.
Storybook > Docs  > Show Code should now be a little faster when passing long base64 strings as props.


shilman (approved):
LGTM!! Thanks @maerten 💯

PR #15592, opened 2021-07-14
Types: Export BaseStoryFn and  BaseStoryObject

yannbf:
Issue: N/A

## What I did

With CSF3, the Story type is now a union of two base types: `BaseStoryFn` and `BaseStoryObject`. This makes it so that `Story` is not directly callable, as it can be just an object.
Some libraries (such as @storybook/testing-react) might need to have a return type that is not `Story` anymore, but rather `BaseStoryFn`, in order to return something that can be called like `<Story/>`. 

This PR export these types, in order to help other libraries do proper inferring or typecasting.

### DIFF
--- a/lib/addons/src/types.ts
+++ b/lib/addons/src/types.ts
@@ -273,14 +273,14 @@ export interface BaseMeta<ComponentType> {
   subcomponents?: Record<string, ComponentType>;
 }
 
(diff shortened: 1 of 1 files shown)


shilman (approved):
LGTM

PR #15611, opened 2021-07-18
UI: Only render settings page when viewMode is settings

vivek12345:
Issue: https://github.com/storybookjs/storybook/issues/14718

## What I did
Taking reference from https://github.com/storybookjs/storybook/pull/14822
I re-added the if check and added viewMode as a dependency for `useMemo`

## How to test

### DIFF
--- a/lib/ui/src/app.tsx
+++ b/lib/ui/src/app.tsx
@@ -47,15 +47,20 @@ const App = React.memo<AppProps>(
           {
             key: 'settings',
             render: () => <SettingsPages />,
-            route: (({ children }) => (
-              <Route path="/settings" startsWith>
-                {children}
-              </Route>
-            )) as FunctionComponent,
+            route: (({ children }) => {
+              if (viewMode === 'settings') {
+                return (
+                  <Route path="/settings" startsWith>
+                    {children}
+                  </Route>
+                );
+              }
+              return null;
+            }) as FunctionComponent,
           },
         ],
       }),
-      []
+      [viewMode]
     );
 
     if (!width || !height) {


vivek12345:
I am still new to this repo so not entirely sure how can I write a test case for this functionality. If someone can point me in the right direction then may be I can add a few tests.

darleendenno (commented):
Hi @vivek12345 👋 Thank you for building off of @dChancellor's PR to get this done! 🚀 🚀 

I've pulled this branch down locally and all appears well and working for me. If you wanted to write a test for this, you could write a story 🎉  Perhaps in `app.stories.tsx` we could add a lil story like:

```
const settingsHistory = createHistory(createMemorySource('/?path=/story/settings'));

export const WithSettingsId = () => (
  <App provider={(new FakeProvider() as unknown) as Provider} history={settingsHistory} />
);
```

I'm going to add in @ndelangen for a second opinion as I am a bit new to this codebase as well, but it looks good to me 😄

vivek12345:
Thank you for reviewing @darleendenno . I will start working on the test case now

lib/ui/src/app.tsx
```diff
@@ -47,15 +47,20 @@ const App = React.memo<AppProps>(
...
-              <Route path="/settings" startsWith>
-                {children}
-              </Route>
-            )) as FunctionComponent,
+            route: (({ children }) => {
+              if (viewMode === 'settings') {
+                return (
+                  <Route path="/settings" startsWith>
+                    {children}
+                  </Route>
```
ndelangen, line 55: Isn't this redundant?
https://github.com/storybookjs/storybook/blob/d0b533db1178abf32df0796d6da17b73f82a35e7/lib/router/src/router.tsx#L104-L113

The Route component won't render it's children if the route doesn't match..

darleendenno: @ndelangen I don't think so? If you pull down and test the branch locally, are you able to set a story's id to `settings`, view the story, and the controls panel?

ndelangen: Hmmm, it's supposed to only match on `/settings` not on `/story/settings-...`.

The true bug here is that `<Route path="/settings" startsWith>` doesn't actually behave correctly.

ndelangen:
related: https://github.com/storybookjs/storybook/pull/16963

PR #15616, opened 2021-07-18
Upgrade react-refresh plugin to fix fast refresh on Webpack5

benbender:
Per default installation, even on `@next`, you'll end up with the following error if you enable `react-refresh` in conjunction with `webpack5`:

````
Error: NormalModuleFactory.afterResolve (CaseSensitivePathsPlugin, ReactRefreshPlugin) is no longer a waterfall hook, but a bailing hook instead. Do not return the passed object, but modify it instead. Returning false will ignore the request and results in no module created.
````

The error can be fixed by manually override the package resolution for `react-refresh` and `@pmmmwh/react-refresh-webpack-plugin` to their latest versions. To render the override obsolete, I would like to kindly request to uplift both dependencies for @next.


shilman (approved):
LGTM! Thanks @benbender

PR #15621, opened 2021-07-19
Examples: Fix react-ts to be runnable standalone

graup:
Issue: the react-ts example is not runnable standalone

## What I did

- Added babelrc (this fixes the issue)
- Cleaned up main.ts (removed addons already included in essentials)
- Reordered package.json dependencies

## How to test

Building the example is part of the E2E tests


shilman (approved):
LGTM!

PR #15628, opened 2021-07-19
Fix/12745 babel config loading

ndelangen:
Issue: https://github.com/storybookjs/storybook/issues/14425 https://github.com/storybookjs/storybook/issues/2758 https://github.com/storybookjs/storybook/issues/12745 https://github.com/storybookjs/storybook/issues/15502

## What I did
* removed the docs telling users babelrc in project root would be loaded (it isn't)
* added support for loading babel.config.ts files in storybook config directory
* added support for the babel config file exporting a function
* layer the config resolution from main.js and the babel config files

## How to test
1. create a storybook
1. expect the babelconfig to be the default
1. add a `.storybook/babel.config.ts` which exports a function (returning valid babel config)
1. run storybook and expect the babel config to be used which was specified in the above config file; the default config is not used
1. add a babel export to `main.js` with valid babel config ( a function that takes config and returns changed config)
1. run storybook and expect a the babel config to be merged with `main.js` overriding `babel.config.ts`; the default config is not used

### DIFF
--- a/docs/configure/babel.md
+++ b/docs/configure/babel.md
@@ -18,9 +18,7 @@ Check out our [source](https://github.com/storybookjs/storybook/blob/next/lib/co
 
 ### Custom config file
 
-If your project has a `.babelrc` file, we'll use that instead of the default config file.
-
-You can also place a `.storybook/.babelrc` file to use a special configuration for Storybook only.
+If your project has a `.storybook/.babelrc` file, we'll use that instead of the default config file.
 
 ### Custom configuration
 
--- a/lib/core-common/src/utils/load-custom-babel-config.ts
+++ b/lib/core-common/src/utils/load-custom-babel-config.ts
@@ -87,6 +87,7 @@ export const loadCustomBabelConfig = async (
     loadFromPath(path.resolve(configDir, '.babelrc.json')) ||
     loadFromPath(path.resolve(configDir, '.babelrc.js')) ||
     loadFromPath(path.resolve(configDir, 'babel.config.json')) ||
+    loadFromPath(path.resolve(configDir, 'babel.config.ts')) ||
     loadFromPath(path.resolve(configDir, 'babel.config.js'));
 
   if (babelConfig) {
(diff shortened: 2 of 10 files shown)


docs/configure/babel.md
```diff
@@ -18,9 +18,7 @@ Check out our [source](https://github.com/storybookjs/storybook/blob/next/lib/co
 
 ### Custom config file
 
-If your project has a `.babelrc` file, we'll use that instead of the default config file.
-
-You can also place a `.storybook/.babelrc` file to use a special configuration for Storybook only.
+If your project has a `.storybook/.babelrc` file, we'll use that instead of the default config file.
```
benediktarnold, line 21: According to the code https://github.com/storybookjs/storybook/blob/next/lib/core-common/src/utils/load-custom-babel-config.ts#L79 there are more filenames considered for loading in the `.storybook` folder

- .babelrc
- .babelrc.json
- .babelrc.js
- babel.config.json
- babel.config.js

yannbf: Important to note as well that the recommended setting for Babel 7 is to use `babel.config.json`, lots of people are renaming their configuration file so it would be nice indeed to see that extension in the docs

Ref: https://babeljs.io/docs/en/config-files#6x-vs-7x-babelrc-loading

lib/core-common/src/utils/load-custom-babel-config.ts
```diff
@@ -48,10 +49,16 @@ function loadFromPath(babelConfigPath: string): TransformOptions {
       throw error.js;
     }
 
-    config = { ...config, babelrc: false };
+    if (config instanceof Function) {
+      config = config();
```
benediktarnold, line 53: Calling the config function without an argument could be problematic. Babel calls the function with an `api` object https://babeljs.io/docs/en/config-files#config-function-api
In my opinion it is ok to stub that api and make sure that calls to the methods don't fail. The `env` method should be easy to reimplement.

PR #15637, opened 2021-07-20
Chore: (Docs) Adds measure and outline main readme

jonniebigodes:
With this small pull request, the measure and outline addons are introduced into the main Readme file.


jonniebigodes:
Self merging this.

PR #15654, opened 2021-07-22
Controls: Don't set arg in validateOptions if it would be `undefined`

### LINKED ISSUE #15630: Storybook controls shouldn't pass 'undefined' by default
Now that storybook doesn't infer defaults for controls since 6.3+, there's an unexpected behavior where stories will get `undefined` explicitly as control values. This can mess up some components (in this case, React), because `{ someProp: undefined }` is different than `{}` and it will result in strange behavior. For example, if you're handling defaults by doing something like `{ ...DEFAULT_PROPS, props }`, then a value of `undefined` will overwrite the default.

See discussion in issue #15378 for more details. Seems like this was introduced in #14579.

Mentioning people involved / with the same issue: tmeasday shilman penx mboettcher

tmeasday:
Fixes #15630

Issue:

Code set `args[X] = undefined` if it wasn't already defined. This was made much more obvious by the removal of default values from args.

## What I did

Ensure the `args` has the key.

## How to test

Yes, Jest test.

### DIFF
--- a/lib/client-api/src/args.ts
+++ b/lib/client-api/src/args.ts
@@ -72,7 +72,9 @@ export const combineArgs = (value: any, update: any): Args => {
 export const validateOptions = (args: Args, argTypes: ArgTypes): Args => {
   return Object.entries(argTypes).reduce((acc, [key, { options }]) => {
     if (!options) {
-      acc[key] = args[key];
+      if (key in args) {
+        acc[key] = args[key];
+      }
       return acc;
     }
 
--- a/lib/client-api/src/args.test.ts
+++ b/lib/client-api/src/args.test.ts
@@ -185,6 +185,11 @@ describe('combineArgs', () => {
 });
 
 describe('validateOptions', () => {
+  // https://github.com/storybookjs/storybook/issues/15630
+  it('does not set args to `undefined` if they are unset', () => {
+    expect(validateOptions({}, { a: {} })).toStrictEqual({});
+  });
+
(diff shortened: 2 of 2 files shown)


lib/client-api/src/args.ts
```diff
@@ -72,7 +72,9 @@ export const combineArgs = (value: any, update: any): Args => {
 export const validateOptions = (args: Args, argTypes: ArgTypes): Args => {
   return Object.entries(argTypes).reduce((acc, [key, { options }]) => {
     if (!options) {
-      acc[key] = args[key];
+      if (key in args) {
```
tmeasday, line 75: We could also use `args[key] !== undefined`, or various other techniques. I honestly don't know what's the right way.

ghengeveld: Good catch. I think this is fine and better than checking against undefined. The alternative would be hasOwnProperty, but since we're dealing with plain objects, it doesn't matter.

ghengeveld (approved):

shilman (approved):
💯

kylegach:
Thank you for fixing this (and back-porting to 6.3)! I was able to remove a bunch of decorators working around this issue. 💥

lib/client-api/src/args.ts
tmeasday: Yeah, I guess technically someone could have an arg called something like `hasOwnProperty` but I suspect they'll be in a world of hurt if they do that anyway.

PR #15669, opened 2021-07-23
feat: allow aliasing require

mjuniper:
Issue: We have a somewhat unique situation in our ember app where `require` is aliased so it is not available on the `window` when the storybook preview attempts to load. This causes the preview to fail to load.

## What I did

I check for an environment variable, `STORYBOOK_REQUIRE_ALIAS`, and, if present, I use that as the name of the `require` function. If the environment variable is not present, I just use `require`. I am certainly open to other suggestions.

## How to test

### DIFF
--- a/app/ember/src/client/preview/globals.ts
+++ b/app/ember/src/client/preview/globals.ts
@@ -4,3 +4,4 @@ const { window: globalWindow } = global;
 
 globalWindow.STORYBOOK_NAME = process.env.STORYBOOK_NAME;
 globalWindow.STORYBOOK_ENV = 'ember';
+globalWindow.STORYBOOK_REQUIRE_ALIAS = process.env.STORYBOOK_REQUIRE_ALIAS;
--- a/app/ember/src/client/preview/render.ts
+++ b/app/ember/src/client/preview/render.ts
@@ -8,8 +8,11 @@ declare let Ember: any;
 
 const rootEl = document.getElementById('root');
 
-const config = globalWindow.require(`${globalWindow.STORYBOOK_NAME}/config/environment`);
-const app = globalWindow.require(`${globalWindow.STORYBOOK_NAME}/app`).default.create({
+const requireFuncName = globalWindow.STORYBOOK_REQUIRE_ALIAS || 'require';
+const requireFunc = globalWindow[requireFuncName];
+
+const config = requireFunc(`${globalWindow.STORYBOOK_NAME}/config/environment`);
+const app = requireFunc(`${globalWindow.STORYBOOK_NAME}/app`).default.create({
   autoboot: false,
   rootElement: rootEl,
   ...config.APP,


shilman (commented):
Is this an ember thing to override require? I'm a little skeptical, but I see that the code that it's replacing is already quite dynamic so ... I'll let the experts weigh in. @gossi @gabrielcsapo any thoughts?

gossi:
Aaaah, took me a while to understand your problem 😵😅

- I was wondering, if this is detectable rather than configurable?
- Also, I think this is likely _special_ to your use-case. Would you be able in your ember project, to alias it back to `window.require` - as this is kinda the "interface" being used here. Would it be possible for you to do this in either `index.js` or `ember-cli-build.js` of your project?

gabrielcsapo:
@mjuniper can you provide a situation in which you are using require at runtime? 

https://github.com/ef4/ember-auto-import should be used for lazy importing of non-ember modules into your application. 

As ember continues to remove a lot of the custom magic, with https://github.com/embroider-build/embroider landing soon 🤞 , we should stay away from letting the require scoping leak into other projects.

tomwayson:
I'll weigh in as the author of the Ember addon that @mjuniper is using, and one of the addon authors who have been dealing with this issue w/in the Ember ecosystem since 2015 when [it was first proposed that Ember's loader provide a no-conflict mode](https://github.com/ember-cli/ember-cli/issues/3420). Unfortunately that never happened so ever since addon developers have had to resort to aliasing `window.require()` in certain scenarios.

The specific reason that happens w/ my addon is that the application developer wants to use a pre-built AMD library hosted on a CDN w/in an Ember application. In these scenarios, the application the developer (in this case @mjuniper) is not able to re-name `require()` in the AMD build (as has been suggested above) on the CDN, so addons like mine that wrap such a library resort to aliasing the `window.requre()` at build time that is introduced by Ember's loader.

So in this case, there's nothing else @mjuniper can do if he wants to use that AMD library in his Ember application. It's not likely that he's using `require()` in his own code, it's that he's using a library that does. So he can't just `ember-auto-import` his way out of this.

I've [already asked](https://github.com/embroider-build/embroider/issues/358), and unfortunately, Ember's magical embroider future isn't going to _just fix this_ w/o someone like myself or one of the other addon authors stepping up and making some changes to how embroider will work.

I don't see a better way to handle this than what @mjuniper is proposing.

gabrielcsapo:
@tomwayson Okay this is great context.

Just a quick question on the actual failure point that we are trying to fix so that we can have tests for this feature and make sure that we can keep pushing ember away from having expectations of the module loading infra to work as expected.

Is this https://github.com/Esri/esri-loader/blob/master/src/modules.ts#L12 the failure point that we are trying to provide a fix for?

tomwayson:
Well, as I understand @mjuniper's issue, the actual point of failure he's experiencing is that `globalWindow.require()` is undefined for him when attempting to execute this line:

https://github.com/storybookjs/storybook/pull/15669/files#diff-df14a7166bc151dfb966d1884c99516fc9bf304c6dc4c320b7da4f301b4e5abdL11

And that is b/c during the Ember build, the ember-esri-loader addon has already renamed it to `window.equireray()` during the Ember build. I suspect that above line must happen before esri-loader is useed to inject the script containing the Dojo AMD loader, which will set `window.require()`.

gabrielcsapo:
Ahh, okay so the default require functionality is being overridden by a `window.equireray()`. 

Would we instead be able to inject a script via the options that could be added as a result of https://github.com/storybookjs/ember-cli-storybook/issues/38? 

If we can inject head tags as a result of your config your application could opt into this functionality by overriding `window.require` to be actually window.equireray()`. This would allow us to add more functionality for all users of ember-cli-storybook.

mjuniper:
> Would we instead be able to inject a script via the options that could be added as a result of [storybookjs/ember-cli-storybook#38](https://github.com/storybookjs/ember-cli-storybook/issues/38)?

I am pretty sure that would work. Unfortunately, in my attempts to verify that, I ran into _another_ issue on our end. But bottom line is that I think your proposed solution will work. What is the time frame for getting that implemented?

gabrielcsapo:
I'll put a PR up tonight to fix that issue.

gabrielcsapo:
https://github.com/storybookjs/ember-cli-storybook/pull/80 this should enable the functionality you are looking for.

PR #15689, opened 2021-07-27
CLI: Fix Svelte CLI template markup

benbender:
Issue:
Svelte examples using JSX-notation for the SVGs. Svelte uses HTML and not JSX. So this fixes those small errors.


shilman (approved):
LGTM. Thanks @benbender !

PR #15697, opened 2021-07-28
CSF3: Add auto-titles from standard glob patterns

jonspalmer:
## What I did

Breaking out the glob processing from #15612 

glob patterns like `dir/**/*.stories.js` now support story auto titles in CSF3.


shilman (approved):
Great stuff @jonspalmer 💪

PR #15708, opened 2021-07-29
fix: a bug of require react

ron0115:
sometimes this  context judgement cause bug.

Issue:

## How to test

### DIFF
--- a/app/react/src/server/framework-preset-react.ts
+++ b/app/react/src/server/framework-preset-react.ts
@@ -38,7 +38,7 @@ const context = storybookReactDirName.includes('node_modules')
 
 const hasJsxRuntime = () => {
   try {
-    require.resolve('react/jsx-runtime', { paths: [context] });
+    require.resolve('react/jsx-runtime', { paths: [path.join(storybookReactDirName, '../../node_modules', path.join(storybookReactDirName, '../../'))] });
     return true;
   } catch (e) {
     return false;


merceyz (changes requested):
`hasJsxRuntime` should ideally be passed the path to the file, no assumptions should be made about the layout on disk

Ref https://github.com/storybookjs/storybook/pull/12899#discussion_r511676973
Ref https://github.com/storybookjs/storybook/pull/13899

shilman:
Do you a have a reproduction repo you can share that demonstrates this bug? If not, can you create one? See [how to create a repro](https://storybook.js.org/docs/react/contribute/how-to-reproduce).

ron0115:
I think the judge about monorepo or normal repo may be a hard code. So why the command path including all paths as much as possible ? It could cover more situations.




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "storybookjs/storybook"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2021年8月2日(星期一) 中午1:01
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [storybookjs/storybook] fix: a bug of require react (#15708)





 
Do you a have a reproduction repo you can share that demonstrates this bug? If not, can you create one? See how to create a repro.

shilman:
Thanks for submitting this @ron0115 but I don' think we can merge unless we know exactly what cases it fixes. Happy to re-open if you can provide that! 🙏

PR #15709, opened 2021-07-29
fix: react build-storybook hang process

JayBee007:
Issue: build-storybook output static files but the process never finishes

## What I did

Copied the fix for angular's unfinished build-storybook process from the **PR:** https://github.com/storybookjs/storybook/pull/15483

## How to test

Try to build storybook with a fully controlled custom webpack config on CI/CI

### DIFF
--- a/app/react/src/server/build.ts
+++ b/app/react/src/server/build.ts
@@ -1,4 +1,16 @@
 import { buildStatic } from '@storybook/core/server';
+import { logger } from '@storybook/node-logger';
 import options from './options';
 
-buildStatic(options);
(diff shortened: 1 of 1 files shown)


shilman (commented):
Thanks for the PR @JayBee007! However, I think this change is redundant with the code that you're wrapping:

https://github.com/storybookjs/storybook/blob/next/lib/core-server/src/build-static.ts#L147-L149

Closing this for now. Please let me know if I'm missing something! 🙏

JayBee007:
Got it! Thanks for the correction.

PR #15740, opened 2021-08-03
UI: Add skip to canvas/sidebar links

darleendenno:
Issue: #4667 

> It is currently quite difficult to test the screen reader accessibility of a component, as you need to navigate past all of the navigation links and then step into the iframe to interact with the component.

## What I did

Added three separate links to improve keyboard navigation around Storybook. All links are only visible on keyboard focus.
1. sidebar/Heading - this link should be the first tabbed content after the address bar; takes the user to the canvas
   * now accepts a `skipLinkHref` prop to determine when a skip link should render
2. sidebar/Tree - this link is visible when a story is selected and focused; takes the user to the canvas
3. preview/FramesRenderer - this link takes the user back to the sidebar

I also added the option to override theme on a per-story basis via global/parameters

## Screenshots
Only using keyboard (TAB and then ENTER) to navigate between the sidebar and iframe
skip-to-content-link

## What I did not do
* Improve the entirety of keyboard navigation for Storybook. This is just a small improvement and should stay as such.
* Make this available for mobile navigation -- that's an entirely different mountain to climb

## How to test

* Open the Storybook available on this PR (see the status check links at the bottom)
* Use a screen reader and attempt to navigate to the iframe; is this easier than the baseline?
* Use a keyboard to navigate between stories, the sidebar, and the canvas; is this easier than the baseline?
* Check out Tree & Heading stories (additional play functionality added 🧪 )

---
- Is this testable with Jest or Chromatic screenshots? **Yes, additional stories added**

#### Questions from the authors (seeking a11y community members for comment)
* How can this be improved upon?
* Is there any desire to improve the navigation relationship between sidebar/canvas and the addon panel?

#### Shoutouts
Thank you to @smithambera, @yannbf, @MichaelArestad, and @ndelangen for your wise collaboration ❤️


shilman (commented):
Looking great @darleendenno -- there are a couple rendering errors in Chromatic, can you take a look?

darleendenno:
> Looking great @darleendenno -- there are a couple rendering errors in Chromatic, can you take a look?

@shilman both of these errors are in components using the new `play` functionality -- does chromatic support those interactions?

shilman:
@darleendenno Chromatic does not officially support `play`, but in practice it should work. Something funny is going on here and there are two instances of `data-testid="heading--skip"` on the page so TestingLib is throwing an error. I'm not sure why yet, but will see if I can repro on my local machine

darleendenno:
> @darleendenno Chromatic does not officially support `play`, but in practice it should work. Something funny is going on here and there are two instances of `data-testid="heading--skip"` on the page so TestingLib is throwing an error. I'm not sure why yet, but will see if I can repro on my local machine

I see now -- I'll fix it up today. May need a delay or some syntax refactoring. Thanks for catching!

lib/ui/src/components/sidebar/Tree.stories.tsx
```diff
@@ -95,3 +96,21 @@ export const SingleStoryComponents = () => {
...
+    isMain: true,
+    refId,
+    data: stories,
+    highlightedRef: { current: { itemId: 'tooltip-tooltipbuildlist--default', refId } },
+    setHighlightedItemId: log,
+    selectedStoryId: 'tooltip-tooltipbuildlist--default',
+    onSelectStoryId: () => {},
+  },
+  play: () => {
+    screen.getByTestId('node--skip').focus();
```
shilman, line 114: @darleendenno try returning the promise so Chromatic can `await` it properly

MichaelArestad:
Looking good! Thank you for making this happen!

Almost all the buttons worked as expected. The exception is the one in the canvas:
image

I can't get it to work. I tried `return`, `space`, and clicking with no luck.

I also noticed that I can trigger that button twice. Try `shift + tab` and `tab` around it.

skip-link-test-1


FF latest. Mac latest.

lib/ui/src/components/sidebar/Tree.stories.tsx
darleendenno: Feels weird to me -- but I can give it a whirl

image

lib/ui/src/components/preview/FramesRenderer.tsx
```diff
@@ -12,6 +13,23 @@ const getActive = (refId: FramesRendererProps['refId']) => {
...
 
+const SkipToSidebarLink = styled(Button)(({ theme }) => ({
+  display: 'none',
+  '@media (min-width: 600px)': {
+    display: 'block',
+    position: 'absolute',
+    top: '10px',
+    right: '15px',
+    padding: '10px 15px',
+    fontSize: theme.typography.size.s1,
```
ghengeveld, line 24: Be aware in styled-components (e.g. in Chromatic), numbers will use the standard unit for the property, which is `em` for font-size. In Emotion it will always use `px` except for unitless properties (e.g. z-index), so in this case it works out.

lib/ui/src/components/preview/FramesRenderer.tsx
```diff
@@ -72,6 +90,9 @@ export const FramesRenderer: FunctionComponent<FramesRendererProps> = ({
   return (
     <Fragment>
       <Global styles={styles} />
+      <SkipToSidebarLink secondary isLink tabIndex={0} href={`#${storyId}`}>
+        Skip to sidebar
+      </SkipToSidebarLink>
```
ghengeveld, line 95: This button should not be rendered when in `singleStory` mode. In this case there is no sidebar: `showNav` is always `false`. You can enable this mode by passing `singleStory=true` as a query param.

It's also possible the sidebar is hidden (e.g. in fullscreen mode, or by pressing `S`). We should then either enable the sidebar when pressing the button, or not render the button at all (which would also cover singleStory mode). I'm fine with either approach. @domyen which you do you think makes sense here?

See [`lib/api/src/modules/layout.ts`](https://github.com/storybookjs/storybook/blob/next/lib/api/src/modules/layout.ts).

ghengeveld (changes requested):

lib/ui/src/components/preview/FramesRenderer.tsx
darleendenno: Excellent catch, @ghengeveld 💯

darleendenno: I believe I've fixed this, but feel free to unresolve if the issue persists 👍

examples/official-storybook/preview.js
```diff
@@ -87,7 +87,8 @@ const ThemedSetRoot = () => {
 };
 
 export const decorators = [
-  (StoryFn, { globals: { theme = 'light' } }) => {
+  (StoryFn, { globals, parameters }) => {
+    const theme = parameters.theme || globals.theme || 'light';
```
darleendenno, line 91: This was added to assist with Chromatic and the `play` functionality -- The decorator in `examples/official-storybook` was causing chromatic to run each story 'stacked' with both light & dark themes. This made my play function invalid as it was only expecting a single element, but chromatic was rendering the story twice. This allows stories to override theme via parameters/global on a case-by-case basis. Hope this change is okay!

shilman (approved):
Great job--looking fantastic! 💯

examples/official-storybook/preview.js
shilman: Cool, I like it 👍

darleendenno:
> Great job--looking fantastic! 💯

Thank you!! I have one more thing to fix up with theming per @domyen but I am hoping to have it ready early next week 🙏

yannbf:
Super nice @darleendenno! Works great both in normal keyboard navigation as well as with Voice Over 👌

examples/official-storybook/preview.js
darleendenno: Thank you! I ended up reverting this as it caused issues with the theming addon -- but something to keep in mind for future features 😄

darleendenno:
Ok this is ready for re-review & merging. Let me know what you think @ghengeveld ❤️

MichaelArestad:
I'm having trouble getting the skip to sidebar button to work with keyboard or by clicking on FF Mac latest. Any ideas?

darleendenno:
> I'm having trouble getting the skip to sidebar button to work with keyboard or by clicking on FF Mac latest. Any ideas?

I’ve been having issues on initial render as well. Refresh and try again. May have to file an issue or get some help figuring out why that’s happening.

lib/ui/src/components/sidebar/Heading.tsx
```diff
@@ -34,13 +36,36 @@ const HeadingWrapper = styled.div({
...
+  '@media (min-width: 600px)': {
+    display: 'block',
+    position: 'absolute',
+    width: '100%',
+    padding: '10px 15px',
+    fontSize: theme.typography.size.s1,
+    zIndex: 1,
+    transform: 'translate(0,-100px)',
+    '&:focus': {
+      transform: 'translate(0,-10px)',
```
MichaelArestad, line 50: I might change the focused positioning to `translate(0, 0)`  or  `translateY(0)` so it aligns  over the logo:

image

MichaelArestad (approved):
IT WORKS! :shipit: Nice work. I had only one minor suggestion. Thanks for making this happen.

callamwilliams:
How can I disable this? For me, it just appears as a 'skip to canvas' button that when clicked moves down 200px and does nothing

MichaelArestad:
@callamwilliams 
Can you share an example of this happening? I would love to get it fixed.

adamnel:
Here's an example of it happening for us https://covalent.teradata.lol

MichaelArestad:
@adamnel It looks like one of your custom styles made this button visible all the time:


By default, it will not appear unless using keyboard navigation:


And both actually do send focus to the canvas. There are not always interactive elements in the canvas, but they do work as expected for folks using keyboard nav.

kharrop:
+1 I'm seeing the above issue where the "Skip to canvas" button is displaying by default after upgrading to the latest version.

yairEO:
I have seen this for months also, now I'm on Storybook version `6.5.13` with a custom brand logo set in `manager.js`

```
const theme = create({
    base: 'light',
    brandTitle: 'CatalogONE',
    brandImage: 'https://www.amdocs.com/sites/default/files/2022-03/Amdocs-Catalog.png',
})

addons.setConfig({
  isFullscreen: false,
  showNav: true,
  showPanel: true,
  panelPosition: 'bottom',
  enableShortcuts: false,
  isToolshown: true,
  theme,
  selectedPanel: undefined,
  initialActive: 'sidebar',
  sidebar: {
    showRoots: false,
    collapsedRoots: ['other'],
  },
  toolbar: {
    title: { hidden: false, },
    zoom: { hidden: false, },
    eject: { hidden: false, },
    copy: { hidden: false, },
    fullscreen: { hidden: false, },
  },
});
```

It's especially visible when the sidebar is expanded:

https://user-images.githubusercontent.com/845031/205020607-dc879161-61c8-45c7-bf40-e3c298faa9f4.mp4

image

PR #15742, opened 2021-08-03
Angular: Add global CSF3 renderer

yngvebn:
Issue: https://github.com/storybookjs/storybook/issues/15498

## What I did

Added a basic global renderer for Angular and CSF3
Enabled previewCsfV3 for the angular-cli example storybook

## How to test

Added a couple of stories to examples/angular-cli


yngvebn:
@shilman / @ThibaudAV I'm struggling with getting the tests to run green locally, even though I haven't touched any of the files or tests it's complaining about... Would greatly appreciate if someone would help out.. The same goes for this PR: https://github.com/storybookjs/storybook/pull/15743

examples/angular-cli/src/stories/csf3/input-with-play.stories.ts
```diff
@@ -0,0 +1,28 @@
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { InputComponent } from './sb-input.component';
+
+export default {
+  title: 'CSF3/Input',
```
ThibaudAV, line 6: maybe `Core / CSF3 / Input` ?

yngvebn: Would "Preview" be a fitting prefix perhaps?

ThibaudAV: - In core I tried to place the stories showing and testing the native functionality of sotrybook
- basics it's more for examples of storybook use 
- addons for addons
..
so if CSF3 is to become something native in storybook main set, I can see it in `core/...` 

but it's up to you, knowing that ;)

shilman (approved):
Awesome! Thanks @yngvebn 🚀

yngvebn:
> Awesome! Thanks @yngvebn 🚀 

You bet!

PR #15774, opened 2021-08-06
Addon-docs/Angular: Use compodoc rawdescription where available

### LINKED ISSUE #15591: Angular: Compodoc character issues
**Describe the bug**
Certain characters are not translated correctly. See image below of a Controls description.
Screenshot 2021-07-14 at 23 10 00

This issue above is caused by an apostrophe in the jsdoc comment 

This is because Compodoc does some parsing of the jsdoc when generating the `description` property. It also adds HTML tags to this `description` property

Compodoc 1.1.14 has a new property called `rawdescription` which does not encode characters or add HTML tags. The difference looks like this

Screenshot 2021-07-14 at 23 21 00

 I propose we move from using the `description` property to the `rawdescription` property.  I can make the change if you're happy to go ahead with this?

dexster:
Issue: #15591 #14116

## What I did

Use rawdescription over description for generated comments when available. rawdescription has no HTML tags and escaped characters so the Storybook comments. This will also fix the nested <p> error in console logs 

This fixes #15591 & #14116

Compodoc also updated to 1.1.14 to allow this change.

## How to test

I updated a Story property comment to include an apostrophe. Snapshots also had to be updated as HTML formatting is now removed when using rawdescription

### DIFF
--- a/addons/docs/src/frameworks/angular/__testfixtures__/doc-button/argtypes.snapshot
+++ b/addons/docs/src/frameworks/angular/__testfixtures__/doc-button/argtypes.snapshot
@@ -19,8 +19,8 @@ Object {
   },
   "_value": Object {
     "defaultValue": "Private hello",
-    "description": "<p>Private value. </p>
-",
+    "description": "
+Private value.",
     "name": "_value",
     "table": Object {
       "category": "properties",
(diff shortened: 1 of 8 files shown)


shilman (approved):
LGTM! thanks @dexster 🙏

PR #15775, opened 2021-08-06
Core: Fix main.js glob resolution for direct paths in stories

lkuechler:
Issue: #15746

## What I did

I extended the fallback variant to exclude the slash directly after the base path as recommended by @dcwarwick in the linked issue.
There is now also a test for this case.


shilman (approved):
Thanks for the fast followup @lkuechler, much appreciated! 🙏

PR #15789, opened 2021-08-08
Server: Update example to use options and labels for options controls

jonspalmer:
## What I did
Options controls take options and labels. This updates the example


shilman (approved):
💯

PR #15798, opened 2021-08-09
CSF: Infer defaultValue of argtype based on arg

yannbf:
An extension on top of #14356

## What I did

ArgTypes is now able to infer the values based on the Args being passed.
If a custom argtype is being passed (that's not a key of passed down Args), it will still infer the correct properties, but `defaultValue` will be of type `unknown`.

## How to test

- Does this need a new example in the kitchen sink apps? not necessarily
- Does this need an update to the documentation? not sure [?]

### DIFF
--- a/lib/addons/src/types.ts
+++ b/lib/addons/src/types.ts
@@ -36,16 +36,20 @@ export interface Args {
   [key: string]: any;
 }
 
-export interface ArgType {
+export interface ArgType<Arg = unknown> {
   name?: string;
   description?: string;
-  defaultValue?: any;
+  defaultValue?: Arg;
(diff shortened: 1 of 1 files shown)


yannbf:
Hey @maxmezzomo! Thanks a lot for the previous contribution you did in #14356! I noticed that the autocompletion was not happening from keys that were not defined in the Args, so I added a fix to that. I also thought it would be nice to get type inference from the property (as in `Args<key>`) and added that as well.

What do you think?

shilman (approved):
LGTM 💪

yannbf:
@shilman this seems good to go!

PR #15799, opened 2021-08-09
Add checkboxes to pull request template

yannbf:
## What I did

Here's an idea:
Rethink a bit on the template so it actually has a checklist rather than a yes/no questionnaire, serving as a good reminder that has to be always checked.

**Idea 1 - Something like this example:**

### Checklist
- [ ] Added label to PR
- [ ] Tested with Cypress, Jest or Chromatic screenshots (if applicable)
- [ ] Added a new example in the kitchen sink apps (if applicable)
- [ ] Updated documentation (if applicable)

**Idea 2:**

Just update the template so there's no need to type yes/no, but click the checks instead.

----

Do you think this will add more value, or make it more confusing?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,9 +4,9 @@ Issue:
 
 ## How to test
 
-- Is this testable with Jest or Chromatic screenshots?
-- Does this need a new example in the kitchen sink apps?
-- Does this need an update to the documentation?
+- [ ] Is this testable with Jest or Chromatic screenshots?
+- [ ] Does this need a new example in the kitchen sink apps?
+- [ ] Does this need an update to the documentation?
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,9 +4,9 @@ Issue:
 
 ## How to test
 
-- Is this testable with Jest or Chromatic screenshots?
-- Does this need a new example in the kitchen sink apps?
-- Does this need an update to the documentation?
+- [ ] Is this testable with Jest or Chromatic screenshots?
+- [ ] Does this need a new example in the kitchen sink apps?
+- [ ] Does this need an update to the documentation?
```
darleendenno, line 9: I like option 1, personally. Thanks for improving the PR template ❤️

shilman (approved):
LGTM - good idea @yannbf we can probably do more with this

PR #15812, opened 2021-08-10
Controls: Fix `esm is not defined` error with built Storybook

gaetanmaisse:
Issue: #15773

## What I did

I reverted https://github.com/storybookjs/storybook/pull/14394 because it looks like it's the commits that introduced the `esm is not defined` error. It will need some deeper investigation.

## How to test

- e2e ci steps should be back to 🟢

### DIFF
--- a/lib/components/src/blocks/ArgsTable/ArgControl.tsx
+++ b/lib/components/src/blocks/ArgsTable/ArgControl.tsx
@@ -42,15 +42,16 @@ export const ArgControl: FC<ArgControlProps> = ({ row, arg, updateArgs }) => {
   const { key, control } = row;
 
   const [isFocused, setFocused] = useState(false);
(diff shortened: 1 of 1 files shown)


gaetanmaisse:
Angular related E2E tests are still failing but for another reason, addon-controls doesn't consider `user` as an arg 🤔  (with a def value)


image

shilman:
@gaetanmaisse the old boxing/unboxing code was a hack because when you try to use React's `useState` with a function value it breaks unless you either (1) box it, or (2) do what @stevensacks proposed which is simpler and slightly more efficient. However, until we have a fix for the problem at hand (which maybe has something to do with how the fix interacts with our webpack settings? or maybe just isn't quite right?) we should revert since the bug is a blocker and getting in the way of people using the alphas.

ndelangen (approved):

Issue #15327, opened 2021-06-22
Issue with "Controls - Dealing with complex values" guide for Vue


santiagoarizti:
**Describe the bug**
Following [this guide](https://storybook.js.org/docs/vue/essentials/controls#dealing-with-complex-values) for Storybook+vue doesn't produce output expected.

In this guide "dealing with complex values" we find that we can map the value of a control in a story through a complex function to produce some complex prop for a component.

Following the guide, although very closely, doesn't work: the modified "arg"'s properties don't do anything to the rendered component.

**To Reproduce**
https://bitbucket.org/cpsanti/storybook-bug-report1/src/storybook-bug-report1/

**System**
note: I tried to use npx sb@next but it didn't work. I copied the output of system from another issue and altered it to match my team's setup:

    System:
      OS: Windows 10 and MacOS (team uses both and this fails for all of us)
    Binaries:
      Node: 16.3.0
      npm: 7.15.1
    Browsers:
      Chrome: Version 91.0.4472.114 (Official Build) (x86_64)
    npmPackages:
        "@vue/cli": "^4.5.13",
        "core-js": "^3.6.5",
        "vue": "^2.6.11"
        "@babel/core": "^7.14.3",
        "@storybook/addon-actions": "^6.2.9",
        "@storybook/addon-essentials": "^6.2.9",
        "@storybook/addon-links": "^6.2.9",
        "@storybook/vue": "^6.2.9",
        "@vue/cli-plugin-babel": "~4.5.0",
        "@vue/cli-plugin-eslint": "~4.5.0",
        "@vue/cli-service": "~4.5.0",
        "babel-eslint": "^10.1.0",
        "babel-loader": "^8.2.2",
        "eslint": "^6.7.2",
        "eslint-plugin-vue": "^6.2.2",
        "vue-loader": "^15.9.7",
        "vue-template-compiler": "^2.6.11"

**Additional context**
The repo I built was stripped down of any complexity, it contains exactly 1 compoent and 1 story file.
Launching storybook in my repo shows the same explanation of the but I provided here.

jonniebigodes:
@santiagoarizti thanks for bringing this to our attention, let me take a look and see what might be the issue and if need be I'll update the documentation accordingly.

santiagoarizti:
Ok, I upgraded to storybook 6.3 and to builder webpack5 and still have the issue. It would be nice to learn if the documentation works for someone or is this use case broken for everybody

willwill96:
[...]

PR #15839, opened 2021-08-12
Chore: (Docs) Removes snippets from vue docs

jonniebigodes:
With this pull request,  #15327 is addressed, to avoid adding confusion and informing readers to lower their Storybook version to an outdated one in which the snippet works.

Once and when this issue is addressed, I'll add the snippet back for Vue 2.

What was done:
- Updated the controls and doc blocks documentation to remove the reference to the snippet.
- Deleted the snippets mentioned above to avoid cluttering.

Feel free to provide feedback


jonniebigodes:
Self merging @shilman

PR #15845, opened 2021-08-13
fix warning when default exports exists

sudosubin:
Issue: None

## What I did

- When using `.storybook/main.js`, it does not change anything.
- When using `.storybook/main.ts`, it removes unnecessary warnings.

## How to test

## Description

```text
WARN Unable to find main.js addons: /path/to/.storybook/main
```

I found an unnecessary warning above during `start-storybook`. I surely had `addons` field, but the warning was not removed.

I debugged for minutes, and I found out `.storybook/main.ts` with below made the warning:

```typescript
// .storybook/main.ts
export default {
  stories: ['../src/**/*.stories.tsx'],
  addons: ['@storybook/addon-essentials'],
};
```

Since it is es module, the properties (ex `stories`, `addons`) cannot be access directly with `require(mainFile)`. It is nested.

```javascript
// ideally `require(mainFile)`
{
  stories: [/* storybook stories */],
  addons: [/* storybook addons */]
}

// actually `require(mainFile)`
{
  default: {
    stories: [/* storybook stories */],
    addons: [/* storybook addons */]
  }
}
```

This PR adds simple guards for this case.

### DIFF
--- a/addons/controls/src/preset/ensureDocsBeforeControls.ts
+++ b/addons/controls/src/preset/ensureDocsBeforeControls.ts
@@ -30,11 +30,12 @@ export const ensureDocsBeforeControls = (configDir: string) => {
   try {
     // eslint-disable-next-line global-require,import/no-dynamic-require
     const main = require(mainFile);
-    if (!main?.addons) {
+    const addons = main?.addons || main?.default?.addons;
+    if (!addons) {
       logger.warn(`Unable to find main.js addons: ${mainFile}`);
       return;
     }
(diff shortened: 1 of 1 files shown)


shilman (commented):
Hi @sudosubin, thanks for taking the time to make a PR. However, we don't actually want users to use the default export. The configuration fields should be named exports. In Typescript this could be:

```ts
export const stories = [...];
export const addons = [...];
```

OR

```ts
const stories = ...;
const addons = ...;
export { stories, addons };
```

Closing this PR for now.

PR #15860, opened 2021-08-17
Addon-measure: Update z-index to fit with libraries with also high z-index

Aarbel:
Libraries like radix-ui use high z-indez for modals, so to measure on them, we need a higher z-index measure tool, cf #15859


winkerVSbecks:
As @MichaelArestad pointed out in the issue, browsers limit the maximum value for z-index to `2147483647`. That's what radix uses too. Could you update the PR to use that value.

Aarbel:
@winkerVSbecks done

winkerVSbecks (approved):

MichaelArestad (approved):

MichaelArestad:
@shilman I'm not sure why the checks are in a waiting state. Any idea?

PR #15863, opened 2021-08-17
Build: Update `caniuse-lite` dependency

gaetanmaisse:
## What I did

Browserslist was throwing a `caniuse-lite is outdated` warning causing some unit tests to fail.
The provided command `npx browserslist@latest --update-db` wasn't working with Yarn 2 so I updated it with: `yarn up -R caniuse-lite`

## How to test

- CI should be back to 🟢

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -15968,17 +15968,10 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


shilman (approved):
You rock @gaetanmaisse 🎸

PR #15868, opened 2021-08-17
Chore: (Docs) Adds an asset to be used throughout the third party docs sites

jonniebigodes:
With this pull request, I'm introducing a converted gif from the video currently in place as a demonstration of Storybook running. Said gif can be used by any third-party documentation site. With the added bonus that should it change we can always update said gif without the need of updating any third-party docs, (i.e., CRA).


jonniebigodes:
Self merging this

PR #15871, opened 2021-08-18
Core: On demand store

tmeasday:
### From the migration notes:

Introduces an opt-in feature flag, `features.storyStoreV7`, which loads stories in an "on demand" way (that is when rendered), rather than up front when the Storybook is booted. This way of operating will become the default in 7.0 and will likely be switched to opt-out in that version.

The key benefit of the on demand store is that stories are code-split automatically (in `builder-webpack4` and `builder-webpack5`), which allows for much smaller bundle sizes, faster rendering, and improved general performance via various opt-in Webpack features.

The on-demand store relies on the "story index" data structure which is generated in the server (node) via static code analysis. As such, it has the following limitations:

- Does not work with `storiesOf()`
- Does not work if you used dynamic story names or component titles.

However, the `autoTitle` feature is supported.

#### Behavioral differences

The key behavioral differences of the v7 store are:

- `SET_STORIES` is not emitted on boot up. Instead the manager loads the story index independenly.
- A new event `STORY_PREPARED` is emitted when a story is rendered for the first time, which contains metadata about the story, such as `parameters`.
- All "entire" store APIs such as `extract()` need to be proceeded by an async call to `loadAllCSFFiles()` which fetches all CSF files and processes them.

#### Using the v7 store

To activate the v7 mode set the feature flag in your `.storybook/main.js` config:

```js
module.exports = {
  // ... your existing config
  features: {
    storyStoreV7: true,
  },
};
```

NOTE: `features.storyStoreV7` implies `features.buildStoriesJson` and has the same limitations.

---

This PR is too large to be sensibly reviewed, instead it has been split into several smaller PRs that can be examined separately and changes will be cherry picked back here:

- [ ] https://github.com/storybookjs/storybook/pull/15999
- [ ] https://github.com/storybookjs/storybook/pull/16004
- [ ] https://github.com/storybookjs/storybook/pull/16005
- [ ] https://github.com/storybookjs/storybook/pull/16006
- [ ] https://github.com/storybookjs/storybook/pull/16009
- [ ] https://github.com/storybookjs/storybook/pull/16010
- [ ] https://github.com/storybookjs/storybook/pull/16011
- [ ] https://github.com/storybookjs/storybook/pull/16012
- [ ] https://github.com/storybookjs/storybook/pull/16034

--- 

Current limitations:

 - [ ] HMR is not fully supported in the v7 store, adding/removing stories will not be picked up, you need to reload the page manually.
 - [ ] v7 mode currently does not work with Storyshots
 - [ ] v7 mode currently does not work with Composition
 - [ ] v7 mode currently does not work with Links
 - [ ] No UI is currently show for async story rendering (not necessarily a big problem but could be improved).

These limitations will be addressed in follow up PRs.


addons/a11y/package.json
```diff
@@ -48,10 +48,10 @@
     "@storybook/addons": "6.4.0-alpha.19",
     "@storybook/api": "6.4.0-alpha.19",
     "@storybook/channels": "6.4.0-alpha.19",
-    "@storybook/client-api": "6.4.0-alpha.19",
     "@storybook/client-logger": "6.4.0-alpha.19",
     "@storybook/components": "6.4.0-alpha.19",
     "@storybook/core-events": "6.4.0-alpha.19",
+    "@storybook/csf": "0.0.2--canary.a1972cb.0",
```
ghengeveld, line 54: Shouldn't this be `@componentdriven/csf`?

addons/controls/package.json
```diff
@@ -47,11 +47,14 @@
...
     "@storybook/api": "6.4.0-alpha.19",
-    "@storybook/client-api": "6.4.0-alpha.19",
+    "@storybook/client-logger": "6.4.0-alpha.19",
     "@storybook/components": "6.4.0-alpha.19",
+    "@storybook/csf": "0.0.2--canary.a1972cb.0",
     "@storybook/node-logger": "6.4.0-alpha.19",
+    "@storybook/store": "6.4.0-alpha.19",
     "@storybook/theming": "6.4.0-alpha.19",
     "core-js": "^3.8.2",
+    "lodash": "^4.17.20",
```
ghengeveld, line 57: I don't see Lodash being used anywhere.

PR #15874, opened 2021-08-18
Angular: Fix Cannot read property 'selector' of undefined

saulodias:
Issue: #14828

## What I did

This fixes the [issue 14828](https://github.com/storybookjs/storybook/issues/14828). 
```
 Cannot read property 'selector' of undefined
```
This issue happens when trying to import an Angular component which has been already compiled. The Issue is explained [here](https://github.com/storybookjs/storybook/issues/14828#issuecomment-882118035) by @KondakovArtem. 

Here is a [reproduction of the issue](https://github.com/saulodias/storybook-i14828). All the code except for the last commit, where I created the `my-lib.stories.ts` file has been automatically generated with the following commands, using the `@angular/cli@11.2.14`.

```
ng new storybook-i14828 --create-application=false
cd storybook-i14828
ng generate library my-lib
npx sb init
```

## How to test

  - The previous Jest test already safisfies the condition. I made sure the method will still return undefined, which was already a valid tested return type, if it doesn't find any decorators.

### DIFF
--- a/app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.ts
+++ b/app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.ts
@@ -136,5 +136,14 @@ export const getComponentDecoratorMetadata = (component: any): Component | undef
       ? Reflect.getOwnPropertyDescriptor(component, decoratorKey).value
       : component[decoratorKey];
 
-  return (decorators || []).find((d) => d instanceof Component);
+  if (!decorators) {
+    return (
+      component.decorators &&
+      component.decorators[0] &&
+      component.decorators[0].args &&
+      component.decorators[0].args[0]
+    );
+  }
+
+  return decorators.find((d) => d instanceof Component);
 };
--- a/app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.test.ts
+++ b/app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.test.ts
@@ -13,7 +13,12 @@ import {
 import { TestBed } from '@angular/core/testing';
 import { BrowserDynamicTestingModule } from '@angular/platform-browser-dynamic/testing';
 
-import { getComponentInputsOutputs, isComponent, isDeclarable } from './NgComponentAnalyzer';
+import {
+  getComponentInputsOutputs,
+  isComponent,
+  isDeclarable,
+  getComponentDecoratorMetadata,
+} from './NgComponentAnalyzer';
 
 describe('getComponentInputsOutputs', () => {
   it('should return empty if no I/O found', () => {
@@ -156,6 +161,15 @@ describe('isComponent', () => {
     class FooDirective {}
 
     expect(isComponent(FooDirective)).toEqual(false);
+  });
+});
+
+describe('getComponentDecoratorMetadata', () => {
+  it('should return Component with a Component', () => {
+    @Component({})
+    class FooComponent {}
+
+    expect(getComponentDecoratorMetadata(FooComponent)).toBeInstanceOf(Component);
   });
 });
 


yarn.lock
```diff
@@ -15968,17 +15968,10 @@ __metadata:
...
-  linkType: hard
-
-"caniuse-lite@npm:^1.0.30001219":
-  version: 1.0.30001228
-  resolution: "caniuse-lite@npm:1.0.30001228"
-  checksum: d6ab115abd93789fe0919773f108a2fbd2efb4b6abe802d29d68655756ec82e6b4dd9a2728a629ae39060eac3e3b094c7cb4899fc3454a274bc04e547d770c34
+"caniuse-lite@npm:^1.0.0, caniuse-lite@npm:^1.0.30000981, caniuse-lite@npm:^1.0.30001032, caniuse-lite@npm:^1.0.30001035, caniuse-lite@npm:^1.0.30001109, caniuse-lite@npm:^1.0.30001125, caniuse-lite@npm:^1.0.30001181, caniuse-lite@npm:^1.0.30001219":
+  version: 1.0.30001251
+  resolution: "caniuse-lite@npm:1.0.30001251"
+  checksum: 027c0ee6533854f416ff8595acec24f985481572a7371212395fb106ff00504a145c6e45fe80e5524b85f62662d7ed54d899e5c9fd97c52e6fc4e1ca7805fc8c
```
ThibaudAV, line 15974: these changes don't seem necessary for this PR, no ?

app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.ts
```diff
@@ -136,5 +136,14 @@ export const getComponentDecoratorMetadata = (component: any): Component | undef
       ? Reflect.getOwnPropertyDescriptor(component, decoratorKey).value
       : component[decoratorKey];
 
-  return (decorators || []).find((d) => d instanceof Component);
+  if (!decorators) {
```
ThibaudAV, line 139: Could you add some test on this file ? https://github.com/storybookjs/storybook/blob/5bcfdae1b6f5d7849a0a48b58a306410977e0cf5/app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.test.ts

To run simple test `yarn jest NgComponentAnalyzer.test.ts` or with watch mode `yarn jest NgComponentAnalyzer.test.ts --watch`
don't hesitate to ask me for help if needed ;)

ThibaudAV:
❤️  thanks for contribution :) LGTM (Just the test, if possible)

yarn.lock
saulodias: I have rebased my commits and now those changes are gone.

app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.ts
saulodias: Hey, I tried my best to create a test at least for the "regular" case. That is, when you are trying to use a component which is not from an Angular library bundle. That test didn't exist indeed. However, once you compile the components into a library, everything becomes very "plain" and we cannot take advantage of types that much to create clean tests, or at least I could not come up with a way to do that in a clean way. So what here is some evidence, if you will, just for the sake of traceability in regards to the expected behavior:

image

Instead of:
image

ThibaudAV: 👍 
Top these evidence if you want you can add them here :

https://github.com/storybookjs/storybook/tree/1508807079ecb35a34f8f05c7d60a179b770504a/examples/angular-cli/src/stories/basics
As you can see there are others.  It's also a way to test the application when unit testing is not efficient / possible 
The examples try to be autonomous and represent a basic use case

ThibaudAV (approved):

app/angular/src/client/preview/angular-beta/utils/NgComponentAnalyzer.ts
saulodias: I'm working on a minimal Angular Library setup for that `angular-cli` workspace. I'll let you know when I'm done and we can discuss some minor details. But we'll cross that bridge when we get to it.

PR #15881, opened 2021-08-19
Addon-docs/Angular: Fix default values in ArgsTable

dexster:
Issue:

## What I did
Added defaultValue to table.defaultValue.summary as per @shilman comments in #12644.

Also, added support for JSDoc @default and @defaultvalue to allow scenarios like accessor defaults where Compodoc will not find default values

## How to test

Kitchen sink app has been updated to show defaultValues and also new accessor defaultValues


shilman (changes requested):
Thanks for the fix @dexster! 🙏  What do you think about this?

addons/docs/src/frameworks/angular/compodoc.ts
```diff
@@ -157,7 +158,17 @@ export const extractType = (property: Property, defaultValue: any) => {
 const extractDefaultValue = (property: Property) => {
   try {
     // eslint-disable-next-line no-eval
-    const value = eval(property.defaultValue);
+    let value = eval(property.defaultValue);
```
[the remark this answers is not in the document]
dexster: Definitely. I only left the eval in as I was not sure about backwards compatibility and what might break if removed.

dexster:
@shilman, eval removed. Note that this change also adds defaults to Outputs, something like this - `new EventEmitter<Event>()`. Are there any concerns regarding this?

addons/docs/src/frameworks/angular/compodoc.ts
```diff
@@ -156,8 +157,17 @@ export const extractType = (property: Property, defaultValue: any) => {
 
 const extractDefaultValue = (property: Property) => {
   try {
-    // eslint-disable-next-line no-eval
-    const value = eval(property.defaultValue);
+    let value = property.defaultValue?.replace(/^'(.*)'$/, '$1');
+    if (value == null && property.jsdoctags.length > 0) {
+      property.jsdoctags.forEach((tag: JsDocTag) => {
+        if (['default', 'defaultvalue'].includes(tag.tagName.escapedText)) {
+          const tmp = window.document.createElement('DIV');
```
shilman, line 164: Is this just to unescape HTML? If so, is there a library we can use for this?

dexster: I originally used DOMParser code below but husky pre-commit throws this error - `'DOMParser' is not defined`. 
``` 
const dom = new DOMParser().parseFromString(tag.comment, 'text/html');
value = dom.body.textContent;
```
If I use window.DOMParser then Typescript has errors. If I add `// @ts-ignore` then all is well. WDYT?
```
 // @ts-ignore
const dom = new window.DOMParser().parseFromString(tag.comment, 'text/html');
value = dom.body.textContent;
```

shilman: i like this better if it works 😁

shilman:
@dexster sorry for the slow reply on this. CI is failing --- are you able to see the failures? any idea what's going on?

addons/docs/src/frameworks/angular/compodoc.ts
```diff
@@ -156,10 +157,23 @@ export const extractType = (property: Property, defaultValue: any) => {
...
+          // @ts-ignore
+          const dom = new window.DOMParser().parseFromString(tag.comment, 'text/html');
+          value = dom.body.textContent;
+        }
+      });
+    }
+
     return value;
   } catch (err) {
+    console.log('UNABLE TO EVAL: ', err);
```
shilman, line 176:
```suggestion
```

shilman:
Looks like the user arg is being incorrectly set in the Angular11 e2e test:

Mouse_Highlight_Overlay

addons/docs/src/frameworks/angular/compodoc.ts
dexster: Oops 😮 Removed

shilman (approved):
Looking great! Thanks for your persistence on this! 🙏

addons/docs/src/frameworks/angular/compodoc.ts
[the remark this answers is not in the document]
amaino: Hi @shilman
Removing the eval has made all default values into strings for the component when using docs as these are set on the component when the story is loaded. I have created a PR to readd this but not sure if there is a better solution?

https://github.com/storybookjs/storybook/pull/17353

PR #15885, opened 2021-08-20
fix(build): avoid logging an object on compilation errors

BlackHole1:
When building with webpack, all information from stats will be printed when an error occurs.

图片

In my project, stats will print out over 5,000 lines, which is hard to swallow..


ndelangen:
Hello @BlackHole1 really appreciate the assist here, could you show us the error occurring somewhere perhaps?
Does it happen in one of the examples in this monorepo?

BlackHole1:
> Hello @BlackHole1 really appreciate the assist here, could you show us the error occurring somewhere perhaps? Does it happen in one of the examples in this monorepo?

yes, the project uses monorepo.

BlackHole1:
see: https://github.com/netless-io/flat/runs/4975672133?check_suite_focus=true#step:7:380

ndelangen:
I think the true reason we're logging this stat is because of this: (we are throwing here with the stats
https://github.com/storybookjs/storybook/blob/89830ad76384faeaeb0c19df3cb44232cdde261b/lib/builder-webpack4/src/index.ts#L149

and then this: (we are awaiting these 2 promises (the preview is throwing)
https://github.com/storybookjs/storybook/blob/89830ad76384faeaeb0c19df3cb44232cdde261b/lib/core-server/src/build-static.ts#L146

and then finally the thrown stats object is caught here, and logged as an error:
https://github.com/storybookjs/storybook/blob/89830ad76384faeaeb0c19df3cb44232cdde261b/lib/core-server/src/build-static.ts#L173

I think we should throw with a generic error instead here:
https://github.com/storybookjs/storybook/blob/89830ad76384faeaeb0c19df3cb44232cdde261b/lib/builder-webpack4/src/index.ts#L149

```ts
          return fail(new Error('compiler encountered errors builder the preview'));
```

@BlackHole1 could you adjust the PR?
mind we have a similar code path in the builder-webpack5:

https://github.com/storybookjs/storybook/blob/89830ad76384faeaeb0c19df3cb44232cdde261b/lib/builder-webpack5/src/index.ts#L137-L140

and also for the manager builders:
https://github.com/storybookjs/storybook/blob/89830ad76384faeaeb0c19df3cb44232cdde261b/lib/manager-webpack5/src/index.ts#L164
https://github.com/storybookjs/storybook/blob/89830ad76384faeaeb0c19df3cb44232cdde261b/lib/manager-webpack4/src/index.ts#L166

BlackHole1:
@ndelangen Great! I will be making adjustments later today.

ndelangen:
@yannbf this seems relevant for our changes we were making today

ndelangen (approved):

PR #15899, opened 2021-08-23
Fix resetArgs example snippet

agriffis:
The example for `resetArgs` was cut-n-paste from typings without fixing
it to be valid code.


shilman (approved):
Thanks @agriffis !! LGTM 🙏

PR #15915, opened 2021-08-25
Addon-docs: Fix embedding selected story in canvas block

quickgiant:
## What I did

Fixed an issue where embedding the currently selected story in a canvas block resulted in a preview crash.

Previously the `CURRENT_SELECTION` token was resolved to a real story ID only if the Source component was passed as a single `id` prop, but since the Canvas component passes an `ids` array to Source regardless of how many children are present, the handling was skipped.

To fix, the Source component was modified to handle `CURRENT_SELECTION` if it appears anywhere in `id` or `ids`. This occurs before IDs are passed to `useStories` since the story store expects resolved story IDs instead of the `"."` token.

## How to test

Open the `Addons/Docs/mdx` story and verify that the new example for a selected story in a canvas renders correctly with source.


quickgiant:
As a heads up, the behavior with this bug has changed after recent updates to `next`. Before, the canvas block would just show "No code available", but now, due to refactors to the story store, it now errors with the following message:
```
Uncaught (in promise) Error: Didn't find '.' in story index (`stories.json`)
    at StoryIndexStore.storyIdToEntry (StoryIndexStore.js:156)
    at StoryStore.loadCSFFileByStoryId (StoryStore.js:243)
    at StoryStore._callee3$ (StoryStore.js:371)
    at tryCatch (runtime.js:63)
    at Generator.invoke [as _invoke] (runtime.js:293)
    at Generator.next (runtime.js:118)
    at asyncGeneratorStep (StoryStore.js:13)
    at _next (StoryStore.js:15)
    at StoryStore.js:15
    at new Promise (<anonymous>)
```

I've moved the `CURRENT_SELECTION` handling to before the `useStories` call to fix this issue as well.

shilman (approved):
Thanks for fixing this @quickgiant! 🙏

PR #15941, opened 2021-08-30
feat: Add parameter to enable grid by default

mbicknese:
Issue: #7367

## What I did

Adds the `isActive` parameter to the `grid` section of the backgrounds
addon. It is `false` by default and may be set globally or per story.

Refs #7367

## How to test

One can add the following configuration to the `preview.js` or story parameters
```js
backgrounds: {
  grid: {
    isActive: true,
  },
},
```

This parameter has been set to true for the `GridCellProperties` story in the official-storybook example. As well as it has been added to the common snippets in the documentation.

### DIFF
--- a/addons/backgrounds/src/decorators/withGrid.ts
+++ b/addons/backgrounds/src/decorators/withGrid.ts
@@ -18,7 +18,9 @@ const deprecatedCellSizeWarning = deprecate(
 export const withGrid = (StoryFn: StoryFunction, context: StoryContext) => {
   const { globals, parameters } = context;
   const gridParameters = parameters[BACKGROUNDS_PARAM_KEY].grid;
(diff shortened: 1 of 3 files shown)


darleendenno (changes requested):
Hey @mbicknese 👋 Thanks for taking the time to improve this addon!

I checked out your update and noticed that the toggle button for the background addon no longer works (see gif). If you're able to fix that to where there are no regressions, I think we can safely merge.

addon-toggle


There has been discussion around the best way to handle default addon settings in a nice and unified manner; it will most likely require a very large change (coming in maybe 8.0?). Merging this would cause some technical debt down the road in that case, but I'm open to it if this is highly desired right now.

mbicknese:
Hi @darleendenno, thanks for taking the time to look into this! I've briefly looked into this and can't seem to find an easy solution. So my conclusion is to wait for the unified addon settings method and implement it right in one go.

PR #15960, opened 2021-09-01
Chore: (Docs)Fixes the addon snippet for the toolbars

### LINKED ISSUE #15947: Toolbar & Globals error in the getTheme function
**Describe the bug**

I am trying to create an addon to update the theme name selected in the toolbar for example if I select English the text in the toolbar will be "English" instead of "Language". 

See the following screenshot for the result I want to achieve:

image

I followed the documentation for how to [Write an Addons](https://storybook.js.org/docs/react/addons/writing-addons/) along with the [Toolbar & Globals](https://storybook.js.org/docs/react/essentials/toolbars-and-globals) but when using the `getTheme` function I have the following error `Uncaught ReferenceError: getTheme is not defined`

jonniebigodes:
With this small pull request the snippet used in this [section](https://storybook.js.org/docs/react/essentials/toolbars-and-globals#consuming-globals-from-within-an-addon) of the documentation is updated to prevent issues when using it.

Closes #15947

What was done:
- The snippet was updated to actually work. It was missing an import and updated the destructuring to avoid issues.

### DIFF
--- a/docs/snippets/common/addon-consume-globaltype.js.mdx
+++ b/docs/snippets/common/addon-consume-globaltype.js.mdx
@@ -1,20 +1,37 @@
 ```js
 // your-addon-register-file.js
 
+import React from 'react';
+
 import { useGlobals } from '@storybook/api';
+
 import { AddonPanel, Placeholder, Separator, Source, Spaced, Title } from '@storybook/components';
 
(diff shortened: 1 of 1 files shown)


jonniebigodes:
Self merging this.

PR #15969, opened 2021-09-02
Core: Add 'staticDirs' config option

oscard0m:
Issue: #13250

## ☑️ What I did
- [x] Added `staticDirs` property to  `StorybookConfig` type
- [x] Added missing type to `useStatics(...)` second argument `options` as `Options` from `@storybook/core-common`
- [x] `useStatics` reads `staticDirs` property from Storybook's configuration and, if it's defined, it adds a route for each directory path passed to staticDirs:
   - The property supports a `string` or an object `{from: string, to: string}`
   - if it's a string, the route of the server for those statics should be `/` and serve the files from `${staticDir}`
   - if it's an object, the route of the server for those statics should be `/${to}/` and serve the files from `${from}`
- [x] Deprecate `--staticDir` | `-s` option from Storybook CLI
   - [x] Mark properties defined in types as `/* deprecated */` (`prod` and `dev`)
   - [x] Add `--static-dir` flag to the collection of flags which should trigger an warning when running the CLI

### Manual testing
1. Choose one storybook framework example you want to run the manual test with _i.e. `examples/react-ts`_
2. Create a new folder with the name you want under the storybook example you chose to run the manual test. Create a file inside the folder you just created with the filename you want and the content you want. i.e. `examples/react-ts/public/public.txt`
3. Change the `main.ts` of the example you selected to run the test adding the `staticDirs` property and set the value of the folder you created in the previous step: i.e. 
```js
...
staticDirs: ['./public']
...
```
4. Run Storybook for the example you chose (make sure you do it in its corresponding folder). i.e.
```bash
> pwd
~/storybook/examples/react-ts

> yarn storybook
```
5. Once Storybook is running in your `localhost`, try to access your static file. _i.e. `http://localhost:<YOUR_PORT>/public.txt`_
image

6. Repeat all the process using the second approach of `{from, to}` and check your static files are served under its mapped path. i.e. `{ from: './public', to: './assets' }` -> `http://localhost:<PORT>/assets/test.txt`
image

### [🚧  WIP 🚧 ] Non-manual Tests
- [ ] Add test coverage for this new Storybook configuration option
    - [x] Evaluate [edge cases covered by previous CLI's flag: `--staticDir` tests](https://github.com/storybookjs/storybook/blob/next/lib/core-server/src/utils/__tests__/server-statics.test.ts)
    - [x] Add `staticDirs` option to `examples/official-storybook/main.ts`

The modified method has no test coverage but I would like to use this PR to add some coverage into it. I would like to test the express app using some library like [`supertest`](url). Is there any chosen library to test this kind of thins in storybook repo I should be aware of?

## Documentation
- [x] Add documentation about this new Storybook configuration option
- [x] Update documentation with deprecated CLI flag


lib/core-server/src/cli/utils.ts
```diff
@@ -18,21 +18,35 @@ export function getEnvConfig(program: Record<string, any>, configEnv: Record<str
...
+
 export function checkDeprecatedFlags(options: {
   dll?: boolean;
   uiDll?: boolean;
   docsDll?: boolean;
+  staticDirs?: string[];
 }) {
   if (!options.dll || options.uiDll || options.docsDll) {
     warnDLLsDeprecated();
+  } else if (options.staticDirs) {
```
ndelangen, line 49: should be it's own `if`, not `else if`

ndelangen:
Code LGTM, have you tested it; does it work?

oscard0m:
> Code LGTM, have you tested it; does it work?

I manually tested and it works to me:

With this file-tree:
image

And this configuration:
```js
...
staticDirs: [
    'statics/public',
    { from: 'statics/examples/example1', to: 'example1' },
    { from: 'statics/examples/example2', to: 'example2' },
],
...
```
I get these 2 links working:
drawing
drawing
drawing

---

Trying to see the deprecation warning message I spotted a bug. Fixed and improved types to catch this at compilation time 💅🏽 

image

---

🛠️  I just noticed I was not using `parseStaticDir()` utility which covers most of the use case between relative, non-relative path as values from Storybook configuration.

---
🤔 🔬 

I'm trying to create tests to cover `useStatics()` (it would help to make me feel safer landing this). I will keep you posted on the progress. If this is kind of urgent I can move it to a different PR if you prefer.

lib/core-server/src/cli/utils.ts
oscard0m: Good catch!

shilman (changes requested):
Looking great, thanks for putting this together @oscard0m ! Couple minor changes..

lib/core-server/src/cli/prod.ts
```diff
@@ -1,24 +1,27 @@
 import program, { CommanderStatic } from 'commander';
 import chalk from 'chalk';
 import { logger } from '@storybook/node-logger';
+import { CLIOptions } from '@storybook/core-common';
```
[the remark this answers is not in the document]
shilman: cc @ndelangen

lib/core-common/src/types.ts
```diff
@@ -249,6 +257,12 @@ export interface StorybookConfig {
       }
   >;
   core?: CoreConfig;
+  /**
+   * Sets a list of directories of static files to be loaded by Storybook server
+   *
+   * @example `['./public']` or `[{from: './public', 'to': '/assets'}]`
+   */
+  staticDirs: (DirectoryMapping | string)[];
```
shilman, line 265:
```suggestion
  staticDirs?: (DirectoryMapping | string)[];
```

oscard0m: Good catch!

shilman (changes requested):
This is awesome @oscard0m !!!

My only issue is that the other paths in `.storybook/main.js` are relative to the config directory, whereas currently the paths appear to be relative to the current working directory.

To reproduce, edit `cra-kitchen-sink/main.js` and set:

```js
  staticDirs: ['./public'],
```

I would expect this to be:

```js
  staticDirs: ['../public'],
```

WDYT?

oscard0m:
> To reproduce, edit `cra-kitchen-sink/main.js` and set:
> 
> ```js
>   staticDirs: ['./public'],
> ```
> 
> I would expect this to be:
> 
> ```js
>   staticDirs: ['../public'],
> ```
> 
> WDYT?

Sounds reasonable to me. Applied changes and also migrated all examples to use `staticDirs` option in configuration files instead of cli's `--static-dir | -s`

shilman (changes requested):
@oscard0m changes are looking good! there is a problem with the `ember-cli` example in CI -- it's also reproducible if you run it on your dev machine.

I think the static directory works a bit different in Ember but I don't know the details. @gossi @gabrielcsapo @TheMightyPenguin can you provide any help here?

Worst case, I think we can revert this change for now, but of course it would be better to find the proper resolution if possible. Thanks! 🙏

oscard0m:
> @oscard0m you can see the CI errors here at the end of the logs https://app.circleci.com/pipelines/github/storybookjs/storybook/21659/workflows/504e6875-0cde-4c62-8cc2-0659e9e56b73/jobs/348198

Hey @shilman ! After taking a look into this, I would like to get some guidance. What I think it's happening:

- When building `examples/*`, [`buildStaticStandalone()`](https://github.com/storybookjs/storybook/blob/f521a74479f5b989242190081b1b8c448809b5cd/lib/core-server/src/build-static.ts#L28) is executed, which is not taking in consideration Storybook's config file apparently.
- TS Types are not ready to simply extend the type to support [`StorybookConfig`](https://github.com/storybookjs/storybook/blob/f521a74479f5b989242190081b1b8c448809b5cd/lib/core-common/src/types.ts#L275)
- Also, I'm not sure if it would be correct merging Storybook's config (`.storybook/main.js`) when running `standalone`.

---

@shilman I would need some guidance (or maybe pair-program this together?) on how to touch this. Is it possible to fix it in an elegant and at the same time to not fall in a rabbit hole of file changes?

shilman:
@oscard0m I'd be happy to pair on this together. Can you jump on our discord and chat me?

https://discord.gg/storybook

shilman (changes requested):
@oscard0m Thanks so much for pairing with me on this! One thought I had afterwards: Storybook should only support one staticDirs specification (command line or main.js) and should throw an error if the user specifies both. Any chance you can add that for both the dev & production use cases?

oscard0m:
> @oscard0m Thanks so much for pairing with me on this! One thought I had afterwards: Storybook should only support one staticDirs specification (command line or main.js) and should throw an error if the user specifies both. Any chance you can add that for both the dev & production use cases?

Sure. Let me take a look into it and I'll ping you if I find any questions.

docs/configure/images-and-assets.md
```diff
@@ -76,23 +76,33 @@ Here `./public` is your static directory. Now use it in a component or story lik
...
+    { from: './public', to: '/assets' },
+    { from: './statics', to: '/resources' }
+  ];
+  ...
 }
 ```
 
+### **[⚠️ Deprecated]** Serving static files via Storybook CLI
+
+Using `--static-dir` or `-s` option with Storybook CLI is deprecated. It is recommended to use [Storybook static directoy Configuration option](#serving-static-files-via-storybook-configuration) instead.
```
yannbf, line 104:
```suggestion
Using `--static-dir` or `-s` option with Storybook CLI is deprecated. It is recommended to use [Storybook static directory configuration option](#serving-static-files-via-storybook-configuration) instead.
```

examples/ember-cli/package.json
```diff
@@ -45,6 +45,7 @@
     "ember-resolver": "^7.0.0",
     "ember-source": "~3.24.0",
     "loader.js": "^4.7.0",
+    "shx": "^0.3.2",
```
yannbf, line 48: was this missing before?

lib/core-common/src/utils/normalize-stories.ts
```diff
@@ -38,6 +38,22 @@ const isDirectory = (configDir: string, entry: string) => {
...
+  const directoryFromConfig = path.resolve(configDir, directory);
+  let directoryFromWorking = path.relative(workingDir, directoryFromConfig);
+
+  // relative('/foo', '/foo/src') => 'src'
+  // but we want `./src` to match importPaths
+  if (!directoryFromWorking.startsWith('.')) {
+    directoryFromWorking = `.${path.sep}${directoryFromWorking}`;
+  }
+  return directoryFromWorking;
+};
```
yannbf, line 55:
```suggestion
export const getDirectoryFromWorkingDir = ({
  configDir,
  workingDir,
  directory,
}: NormalizeOptions & { directory: string }) => {
  const directoryFromConfig = path.resolve(configDir, directory);
  let directoryFromWorking = path.relative(workingDir, directoryFromConfig);

  // relative('/foo', '/foo/src') => 'src'
  // but we want `./src` to match importPaths
  if (!directoryFromWorkingDir.startsWith('.')) {
    directoryFromWorkingDir = `.${path.sep}${directoryFromWorkingDir}`;
  }
  return directoryFromWorkingDir;
};
```

lib/core-server/src/build-static.ts
```diff
@@ -68,6 +70,24 @@ export async function buildStaticStandalone(options: CLIOptions & LoadOptions &
...
 
+  const staticDirs = await presets.apply<StorybookConfig['staticDirs']>('staticDirs');
+
+  if (staticDirs && options.staticDir) {
+    throw new Error(`Conflict when trying to read staticDirs:
+    * Storybook's configuration option: 'staticDirs'
+    * Storybook's CLI flag: '--staticDir' or '-s'
+     
+    Choose one of them, but not both.
+    `);
```
yannbf, line 81: Can you use `ts-dedent` here? This way the output in the CLI is properly aligned!
```suggestion
    throw new Error(dedent(`Conflict when trying to read staticDirs:
    * Storybook's configuration option: 'staticDirs'
    * Storybook's CLI flag: '--staticDir' or '-s'
     
    Choose one of them, but not both.
    `));
```

shilman (approved):
Great job @oscard0m -- awesome feature!!!

lib/core-server/src/build-static.ts
oscard0m: I didn't know this library. Thanks for sharing :)

oscard0m:
Thanks @shilman and @yannbf for bringing this to the finish line. I did not have time during this week to check this, sorry for that. I really appreciate you took your time to review, re-review, pair with me, apply some fixes... Thank you :)

IvanovRoman:
https://github.com/storybookjs/storybook/issues/16817

IvanovRoman:
https://github.com/storybookjs/storybook/issues/16817

PR #15973, opened 2021-09-03
webpack-manager: add ability to resolve js extensions in ts files

### LINKED ISSUE #15962: storybook cannot import .js paths in .tsx? files.
**Describe the bug**
I'm creating a UI Library that is distributed as a ES Module.

in typescript, to create an ES Module compliant package you must import/export all paths as `.js` as outlined here: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

storybook does not currently infer these paths and will fail to find the .ts files during development.

in order to provide a great developer experience, storybook should support this functionality.

**To Reproduce**

https://github.com/lifeiscontent/storybook-repro

**System**

```
Environment Info:

  System:
    OS: Linux 5.11 Zorin OS 16
    CPU: (4) x64 Intel(R) Core(TM) i3-10110U CPU @ 2.10GHz
  Binaries:
    Node: 14.17.4 - ~/.asdf/installs/nodejs/14.17.4/bin/node
    Yarn: 3.0.1 - ~/.asdf/installs/nodejs/14.17.4/.npm/bin/yarn
    npm: 6.14.14 - ~/.asdf/installs/nodejs/14.17.4/bin/npm
  Browsers:
    Chrome: 93.0.4577.63
    Firefox: 91.0.2
  npmPackages:
    @storybook/addon-actions: ^6.4.0-alpha.33 => 6.4.0-alpha.33 
    @storybook/addon-docs: ^6.4.0-alpha.33 => 6.4.0-alpha.33 
    @storybook/addon-essentials: ^6.4.0-alpha.33 => 6.4.0-alpha.33 
    @storybook/addon-links: ^6.4.0-alpha.33 => 6.4.0-alpha.33 
    @storybook/react: ^6.4.0-alpha.33 => 6.4.0-alpha.33 
```

**Additional context**

thanks for your help ❤️

lifeiscontent:
closes #15962

## What I did

add a Webpack plugin to resolve js extensions in ts files.

## How to test

- [x] Does this need a new example in the kitchen sink apps?

### DIFF
--- a/examples/cra-ts-kitchen-sink/src/components/List.stories.tsx
+++ b/examples/cra-ts-kitchen-sink/src/components/List.stories.tsx
@@ -1,0 +1,29 @@
+/* eslint-disable import/no-unresolved */
+import React from 'react';
+import { Meta } from '@storybook/react';
+import List from './List.js';
+import ListItem from './ListItem.js';
+
+export default {
+  title: 'Docgen/List',
+  component: List,
+  subcomponents: {
+    ListItem,
+  },
+} as Meta;
+
+export const SimpleList = () => {
+  return (
+    <List>
+      <ListItem>Hello World</ListItem>
+    </List>
+  );
+};
+
+export const WithMultiple = () => (
+  <List>
+    <ListItem>One</ListItem>
+    <ListItem>Two</ListItem>
+    <ListItem>Three</ListItem>
+  </List>
+);
--- a/examples/cra-ts-kitchen-sink/src/components/List.tsx
+++ b/examples/cra-ts-kitchen-sink/src/components/List.tsx
@@ -1,0 +1,7 @@
+import React, { FunctionComponent } from 'react';
+
+const List: FunctionComponent<unknown> = ({ children }) => {
+  return <ul>{children}</ul>;
+};
+
+export default List;
--- a/examples/cra-ts-kitchen-sink/src/components/ListItem.tsx
+++ b/examples/cra-ts-kitchen-sink/src/components/ListItem.tsx
@@ -1,0 +1,7 @@
+import React, { FunctionComponent } from 'react';
+
+const ListItem: FunctionComponent<unknown> = ({ children }) => {
+  return <li>{children}</li>;
+};
+
+export default ListItem;
--- a/lib/manager-webpack4/package.json
+++ b/lib/manager-webpack4/package.json
@@ -70,6 +70,7 @@
     "read-pkg-up": "^7.0.1",
     "regenerator-runtime": "^0.13.7",
     "resolve-from": "^5.0.0",
+    "resolve-typescript-plugin": "^1.1.0",
     "style-loader": "^1.3.0",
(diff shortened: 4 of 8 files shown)


lifeiscontent:
@shilman not sure what I'm doing, do I need to upgrade the version of storybook manager or is that a legit failure in the code I've updated? Maybe I'm adding this plugin in the wrong spot?

djcsdy:
Hi. Author of resolve-typescript-plugin here. 

It looks like this configuration should work although there are some differences in your setup that are quite different than environments I've tested in.

The way I would approach this is to step through resolve-typescript-plugin's [tapAsync callback](https://github.com/softwareventures/resolve-typescript-plugin/blob/04e58b42833e0c712af63103c4f9bc7499194341/index.ts#L28-L53) in a debugger and check that it is actually getting called and that it behaves as expected for the `*.js` imports.

Hopefully the callback is getting called and it's just behaving incorrectly, in which case the solution should become apparent. If it's not getting called at all then the problem is most likely somewhere in your webpack configuration.

Unfortunately webpack's file resolution system is very complicated and not very documented, and I'm not certain that I've hooked into it in the best possible way. So this problem might be evidence that I've got it wrong.

If you can't figure it out and you haven't heard from me in a few days then feel free to ping me and I'll try to find time to have a closer look, but I can't promise anything.

If you do find what's going on and it's something I need to change in resolve-typescript-plugin then of course [please raise an issue](https://github.com/softwareventures/resolve-typescript-plugin/issues).

Some things to consider:
 * It's possible that the Yarn PNP resolver is interfering somehow. I don't think it should but I haven't tested against it.
 * The `resolve` configuration I test against is: `resolve: {fullySpecified: true, plugins: [new ResolveTypeScriptPlugin()]}`. [Yours is quite a bit different](https://github.com/storybookjs/storybook/blob/3e352ae86e131517b63f96b55b143dbf90a43afc/lib/manager-webpack4/src/presets/manager-preset.ts#L160-L173) and it's possible that some other setting is interfering with the plugin somehow.
 * I haven't tested against webpack 4. I see no reason that it shouldn't work, as AFAIK there have been no significant changes to the resolver itself between webpack 4 and 5. However IIRC the default `resolve` configuration changed a bit in webpack 5, so that might be the problem.
 * Your [`tsconfig.json`](https://github.com/storybookjs/storybook/blob/3e352ae86e131517b63f96b55b143dbf90a43afc/examples/cra-ts-kitchen-sink/tsconfig.json#L9) has `module: "commonjs"`. I've only tested with `module: "es2020"`. It's possible that compiling to CommonJS causes the TypeScript compiler to get confused although I don't see why it should and in any case the logs show a problem during resolution, not during compilation. So that probably isn't the problem and AFAIK it should work but it's something I haven't tested.

lifeiscontent:
@djcsdy I'm pretty sure its an issue with webpack 4, I'm using these changes locally in a webpack 5 project and it works fine.

